PR #24, opened 2020-04-30
Fix jobs docs: schedule i.o. job_schedule

sdebruyn:
The attribute in the jobs resource seems to be called `schedule` instead of `job_schedule` like it was mentioned in the docs


stikkireddy:
LGTM thanks! 👍

Issue #21, opened 2020-04-30
[ISSUE] Azure token auth fails with service principal secret with special characters


sdebruyn:
### Terraform Version
```
➜ terraform -v
Terraform v0.12.24
```

### Affected Resource(s)
- all resources
### Terraform Configuration Files
```hcl
terraform {
  required_version = "~> 0.12"
}

provider "azurerm" {
  version = "~> 2.6.0"
  features {}
}

provider "azuread" {
  version = "~> 0.8.0"
}

data "azurerm_client_config" "current" {
}

resource "azuread_application" "aadapp" {
  name = "app"
  required_resource_access {
    resource_app_id = "e406a681-f3d4-42a8-90b6-c2b029497af1"
    resource_access {
      id   = "03e0da56-190b-40ad-a80c-ea378c433f7f"
      type = "Scope"
    }
  }
  required_resource_access {
    resource_app_id = "00000003-0000-0000-c000-000000000000"
    resource_access {
      id   = "e1fe6dd8-ba31-4d61-89e7-88639da4683d"
      type = "Scope"
    }
  }
}

resource "random_password" "aadapp_secret" {
  length = 32
  # special = false - this fixes the issue...
}

resource "azuread_service_principal" "sp" {
  application_id = azuread_application.aadapp.application_id
}

resource "azuread_service_principal_password" "sppw" {
  service_principal_id = azuread_service_principal.sp.id
  value                = random_password.aadapp_secret.result
  end_date             = "2030-01-01T00:00:00Z"
}

resource "azurerm_resource_group" "rg" {
  name     = "rg"
  location = var.region
}

resource "azurerm_role_assignment" "sprg" {
  scope                = azurerm_resource_group.rg.id
  role_definition_name = "Owner"
  principal_id         = azuread_service_principal.sp.object_id
}

resource "azurerm_databricks_workspace" "dbks" {
  name                        = "dbks"
  resource_group_name         = azurerm_resource_group.rg.name
  managed_resource_group_name = "rgdbks"
  location                    = var.region
  sku                         = "standard"
}

provider "databricks" {
  azure_auth = {
    managed_resource_group = azurerm_databricks_workspace.dbks.managed_resource_group_name
    azure_region           = azurerm_databricks_workspace.dbks.location
    workspace_name         = azurerm_databricks_workspace.dbks.name
    resource_group         = azurerm_databricks_workspace.dbks.resource_group_name
    client_id              = azuread_application.aadapp.application_id
[...]

PR #27, opened 2020-05-01
Issue #21 Fix client_secret special characters

stikkireddy:
client_secret had special characters which were not unescaped when creating the management token and platform token. This caused issues when trying to make an api call which caused the secret to be invalid. This solution swaps making the api calls manually to using the MSFT autorest adal library for authentication. This should resolve issue #21. I have ran the integration test as well as a test with a sp password of: 
`value = "!@##@%#$^$#&%^*^(^*(&()l}$GWolISz+kyXy"`


databricks/azure_auth.go
```diff
@@ -93,93 +102,88 @@ func (a *AzureAuth) getWorkspaceID(config *service.DBApiClientConfig) error {
...
+	// creation the API's will not work correctly. This may also
+	err = retry(5, 1*time.Second, func() error {
+		_, err = dbClient.Clusters().ListNodeTypes()
+		return err
+	})
+
 	return dbClient, err
 }
+
+func retry(numAttempts int, sleep time.Duration, do func() error) error {
```
stikkireddy, line 178: This retry should help fix #33 as we will attempt to retry and above we will be retrying for about 1 minute to attempt to list the node types. This should verify that the apis are ready for requests. This is a hack, but I am internally trying to identify the root cause of why the apis are not initialized properly.

sdebruyn:
After testing this a couple of times... this PR fixes #21 but I still get #33 

```
Error: status 400: err Response from server {"error_code":"INVALID_PARAMETER_VALUE","message":"Delegate unexpected exception during listing node types: com.databricks.backend.manager.util.UnknownWorkerEnvironmentException: Unknown worker environment WorkerEnvId(workerenv-2102767251748222)"}
```

lawrencegripper:
I'd expect increasing the retry number to a larger count would resolve the issue in #33. Just running a quick test now to see how long it usually takes to resolve itself in previous hacky provider. 

Hmm looks like roughly 30 seconds. Looks like current code is only going to wait about 5secs

```
Transcript started, output file is ./cluster.create.log
Starting create
> sleep 10
CLI Response: Error: b'{"error_code":"BAD_REQUEST","message":"Current organization 8479926402891530 does not have any associated worker environments"}'
PS>TerminatingError(): "Failed to execute Databricks CLI. Error response."
Create failed. Retrying after wait...
05/05/2020 15:34:26
> sleep 10
CLI Response: Error: b'{"error_code":"BAD_REQUEST","message":"Current organization 8479926402891530 does not have any associated worker environments"}'
PS>TerminatingError(): "Failed to execute Databricks CLI. Error response."
Create failed. Retrying after wait...
05/05/2020 15:34:37

> sleep 10
Found ClusterID from Terraform state: 0505-153445-lops296
cluster_id
```

stikkireddy:
@sdebruyn hey just made some changes to use a much longer time out of 30 minutes, can you please run again and verify that this works. 

Unfortunately have not gotten much better news, in regards to resolving the root cause of the apis not working correctly after workspace creation. At this point in time I dont know if the problem is caused by the databricks deployment itself or if the azurerm provider is not waiting long enough. 

And I cant seem to reproduce the issue any more with the following change made. I was able to run 5/5 times after destroying the environment without causing that issue.

sdebruyn:
It worked on the first try now :) `databricks_cluster.cluster: Creation complete after 3m35s [id=0506-080854-hums438]` Thanks! Awesome work with this provider!

sdebruyn:
I don't know if it's related, but just a heads-up, since moving to this branch, I am getting the following errors when I try to create secret scopes:

```
Error: status 503: err Response from server {"error_code": "TEMPORARILY_UNAVAILABLE", "message": "Authentication is temporarily unavailable. Please try again later."}
```

I thought it was a temporary issue, but it's been there for a few days now. This does not happen with the code that is currently on master.

stikkireddy:
@sdebruyn is this only right after you make a workspace or at all times?

sdebruyn:
@stikkireddy at all times, even hours after creating the workspace and the cluster

stikkireddy:
hmm let me look into that and see if I can recreate this scenario, @sdebruyn what region are you deploying into?

sdebruyn:
@stikkireddy eastus2

lawrencegripper:
Would it be possible to merge the auth fixes and continue to look at the root cause of #33 on a separate PR? This would be super useful to unblock us as the temporary fix, while not ideal, does work.

stikkireddy:
Hey @sdebruyn please take a look at the new changes, I have reverted the AAD token to using the PAT token. I realized there are still some issues in using AAD tokens to use the secrets, scopes, scope acls apis. So I reverted it to just use the 1 hr PAT (the tokens within Databricks) tokens. I also fixed up the acceptance tests, and split the AWS and Azure tests. The acceptance tests were failing for users & groups due to roles not existing on azure (aws iam roles). This should be all fixed now. Please give this a try, the retry functionality in the azure auth and the revert to using PAT tokens should resolve both issue #21, #33 and also avoid the following for secrets:
```Error: status 503: err Response from server {"error_code": "TEMPORARILY_UNAVAILABLE", "message": "Authentication is temporarily unavailable. Please try again later."}```.

TL;DR:
I ran the existing acceptance tests on AWS and Azure and they seem to be working. It is a WIP to integrate to travis. But for now please let me know if this branch resolves #21, #33  & TEMPORARILY_UNAVAILABLE for secrets.

@lawrencegripper this should resolve the issues. At least after the changes I cannot seem to be able to reproduce the problem.

sdebruyn:
Now all of the following succeeds:

- cluster
- token
- secret scopes
- secrets

But then it fails when creating the mounts.

Logs at https://gist.github.com/sdebruyn/e86699686b120b444b2d0be4dd10c922

sdebruyn:
I think that this might not be related to this PR, but to my service principal instead. I'll further look into it in the coming hours.

stikkireddy:
That seems to be a auth level issue you may need to make your SP a storage blob data owner. Hmm not sure what it is at the moment. But will merge this as it seems your mount is a auth issue (403) and @lawrencegripper is going to work on better logging for the mounts and help terraform propagate the error a bit better.

stikkireddy:
This will be closing #33 and #21

sdebruyn:
correct, one of the permissions was missing :)

PR #36, opened 2020-05-04
Only set new_cluster in the job when it's set in the schema

### LINKED ISSUE #25: [ISSUE] Cannot create a job while reusing an existing cluster
### Terraform Version
```
➜ terraform -v
Terraform v0.12.24
```

### Affected Resource(s)
- databricks_job

```hcl
resource "databricks_job" "transform" {
  existing_cluster_id = databricks_cluster.cluster.id
  notebook_path = databricks_notebook.transform.path
  name = "transform"

  schedule {
      quartz_cron_expression = "0 2 * * *"
      timezone_id = "UTC"
  }
}
```

### Debug Output
```
2020-04-30T18:30:30.347+0200 [DEBUG] plugin.terraform-provider-databricks_v0.1.0: 2020/04/30 18:30:30 {"Method":"POST","URI":"https://eastus2.azuredatabricks.net/api/2.0/jobs/create","Payload":{"existing_cluster_id":"0430-125417-toed833","new_cluster":{},"notebook_task":{"notebook_path":"/workspace/sample/transform.scala"},"name":"transform","schedule":{"quartz_cron_expression":"0 2 * * *","timezone_id":"UTC"}}}
2020/04/30 18:30:30 [DEBUG] databricks_job.transform: apply errored, but we're indicating that via the Error pointer rather than returning it: status 400: err Response from server {"error_code":"INVALID_PARAMETER_VALUE","message":"Missing required field: settings.cluster_spec.new_cluster.size"}
```

### Expected Behavior
Job should be created without errors.

### Actual Behavior
An error message indicating that a parameter is missing.
I think you need to leave out the `new_cluster` parameter from the HTTP request when `existing_cluster_id` is not null to avoid server-side validation of that block.

### Steps to Reproduce
1. `terraform apply`

Reuse an existing cluster when creating a job.

sdebruyn:
This fixes #25 and only sets the `new_cluster` field when it was also set in the schema

### DIFF
--- a/databricks/resource_databricks_job.go
+++ b/databricks/resource_databricks_job.go
@@ -2,13 +2,14 @@ package databricks
 
 import (
 	"fmt"
-	"github.com/databrickslabs/databricks-terraform/client/model"
-	"github.com/databrickslabs/databricks-terraform/client/service"
-	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"log"
 	"reflect"
 	"strconv"
 	"strings"
+
+	"github.com/databrickslabs/databricks-terraform/client/model"
+	"github.com/databrickslabs/databricks-terraform/client/service"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
 func resourceJob() *schema.Resource {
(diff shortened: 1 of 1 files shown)


stikkireddy:
LGTM, I merged in the linting issue, so merging this in should not really create an issue. Thanks for the PR 👍

PR #45, opened 2020-05-08
Travis ci added integration builds

stikkireddy:

Issue #38, opened 2020-05-05
[ISSUE] Azure ADAL Mount always detects diff due to extra slash


lawrencegripper:
Hi, 

A  bug we've hit that I'd like to pickup and PR a fix, looks like a super easy fix. We'd use integration tests to be added in #37 to validate it behaves correctly. 

When re-running the `adls_gen2_mount` resource it will always detect a change due to an additional slash being detected.

Think this is likely a 1 or 2 char change to fix then adding the tests to validate. 

### Terraform Version
```
# tf -v
Terraform v0.12.16
+ provider.azuread v0.8.0
+ provider.azurerm v2.8.0
+ provider.databricks v0.1.0
+ provider.random v2.2.1
```

### Affected Resource(s)

- databricks_azure_adls_gen2_mount

### Terraform Configuration Files
```hcl
resource "databricks_azure_adls_gen2_mount" "mount" {
  cluster_id           = databricks_cluster.cluster.id
  container_name       = "dev" #todo: replace with env...
  storage_account_name = azurerm_storage_account.account.name
  directory            = "/dir"
  mount_name           = "localdir"
  tenant_id            = data.azuread_client_config.current.tenant_id
  client_id            = azuread_application.datalake.application_id
  client_secret_scope  = databricks_secret_scope.terraform.name
  client_secret_key    = databricks_secret.client_secret.key
}
```

### Debug Output

```
> TF_LOG=debug tf plan -var-file vars.tfvars 2>&1 >/dev/null | grep "plugin.terraform-provider-databricks_v0.1.0"

[...]

Issue #33, opened 2020-05-04
[ISSUE] UnknownWorkerEnvironmentException when creating a cluster after creating a workspace


sdebruyn:
### Terraform Version
```
➜ terraform -v
Terraform v0.12.24
+ provider.azuread v0.8.0
+ provider.azurerm v2.6.0
+ provider.databricks v0.1.0
+ provider.http v1.2.0
+ provider.local v1.4.0
+ provider.null v2.1.2
+ provider.random v2.2.1
```

### Affected Resource(s)
- databricks_cluster

### Terraform Configuration Files
Same one as in #21 

### Debug Output
```
2020-05-04T10:45:06.969+0200 [DEBUG] plugin.terraform-provider-databricks_v0.1.0: 2020/05/04 10:45:06 {"Method":"GET","URI":"https://eastus2.azuredatabricks.net/api/2.0/clusters/list-node-types?"}
2020/05/04 10:45:07 [ERROR] eval: *terraform.EvalConfigProvider, err: status 400: err Response from server {"error_code":"INVALID_PARAMETER_VALUE","message":"Delegate unexpected exception during listing node types: com.databricks.backend.manager.util.UnknownWorkerEnvironmentException: Unknown worker environment WorkerEnvId(workerenv-3375316063940170)"}
```

```
Error: status 400: err Response from server {"error_code":"INVALID_PARAMETER_VALUE","message":"Delegate unexpected exception during listing node types: com.databricks.backend.manager.util.UnknownWorkerEnvironmentException: Unknown worker environment WorkerEnvId(workerenv-1918878560143470)"}

  on databricks.tf line 10, in provider "databricks":
  10: provider "databricks" {
```

### Expected Behavior
After creating the workspace, we should be able to create the cluster during the same apply run.

### Actual Behavior
When you create a workspace and terraform goes on to immediately create a cluster, you get the mentioned exception. It works when you apply a second time after a few seconds.

### Steps to Reproduce
1. `terraform apply`

### References
This third party databricks provider has the same issue
- https://github.com/innovationnorway/terraform-provider-databricks/issues/49

stikkireddy:
@sdebruyn this comment https://github.com/databrickslabs/databricks-terraform/pull/27/files#r419517843 should contain information about how to temporarily address this. This ties back to issue #21.

lawrencegripper:
[...]

PR #51, opened 2020-05-12
Resolve ADLS2 Mount Diff and Incorrect error messages

### LINKED ISSUE #37: [ISSUE] Azure Mount ADAL throws wrong error hiding error details
Hi, 

Great work on the provider. We've found a small bug we'd like to fix up and PR into the provider to make understanding a failure case easier. 

@stuartleeks is looking at fixing this up by changing the behavior in the `try-except` block and adding an integration test to validate.  Probably tackles #20 too as an added bonus 🎉

### Terraform Version
```
# tf -v
Terraform v0.12.16
+ provider.azuread v0.8.0
+ provider.azurerm v2.8.0
+ provider.databricks v0.1.0
+ provider.random v2.2.1
```

### Affected Resource(s)

- databricks_azure_adls_gen2_mount

```hcl
resource "databricks_azure_adls_gen2_mount" "mount" {
  cluster_id           = databricks_cluster.cluster.id
  container_name       = "dev" #todo: replace with env...
  storage_account_name = azurerm_storage_account.account.name
  directory            = "/dir"
  mount_name           = "localdir"
  tenant_id            = data.azuread_client_config.current.tenant_id
  client_id            = azuread_application.datalake.application_id
  client_secret_scope  = databricks_secret_scope.terraform.name
  client_secret_key    = databricks_secret.client_secret.key
}
```

### Expected Behavior

The errror from the `dbutils.fs.mount` should be returned by the provider. For example if a secret is misconfigured or clientID wrong the following should be returned:

> shaded.databricks.v20180920_b33d810.org.apache.hadoop.fs.azurebfs.oauth2.AzureADAuthenticator$HttpException: AADToken: HTTP connection failed for getting token from AzureAD. Http response: 401 Unauthorized

### Actual Behavior

If the mount operation fails the exception details are swallowed by the provider and another exception is instead returned as a result of `dbutils.unmount` failing. 

For example if you misconfigure the Service Principal details inputted into the resource the following output is received.

image

The actual error occurring during the mount is:
> shaded.databricks.v20180920_b33d810.org.apache.hadoop.fs.azurebfs.oauth2.AzureADAuthenticator$HttpException: AADToken: HTTP connection failed for getting token from AzureAD. Http response: 401 Unauthorized

This detail is lost as the `except` block triggers and attempts to call `dbutils.fs.unmount`. As the mount operation failed this call throws an exception. This is not caught and the `throw e` line is not reached to throw the original error. 

https://github.com/databrickslabs/databricks-terraform/blob/0a30fe8b33b9b56162c4c323cd7363f667966b84/databricks/mounts.go#L310-L323

See example repro'ing this manually in a notebook:

image

### Steps to Reproduce

1. Create an invalid SP credential and pass into the resource
1. `terraform apply`
1. Observe an exception of `Directory not mounted` instead of `Not authorized error`

lawrencegripper:
Fixes #37 #38 

Changes:
- Add additional `try-except` so `unmount` error isn't returned in place of actual error on mount failure #37 
- Remove requirement for `/` in directory name as mounting the root of the Datalake is a valid action (highlighted by the tests)
- Remove the additional `/` in the python mount command which caused a diff to be detected #38 
- Add a repeatable integration test for Azure Resources, initially implementing this resource #20 

The individual fixes for the bugs are relatively simple, majority of the work has gone into getting a repeatable end2end acceptance test setup for this Azure resource (and hopefully to use for all Azure integration tests).

Integration testing is done by running  `./integration-environment-azure/run.sh`:
1.This creates prereq from scratch (workspace, storage etc) in azure
2. These are output from the Terraform `prereq.tf` into environment variables
3. Any go tests starting with `TestAccAzure` are invoked using this envs 
4. Script cleans up the environment

> Note: We create each environment from scratch to ensure no state from previous runs breaks tests and full end2end on new workspace is successful. When testing `SKIP_CLEANUP` can be used to avoid recreating the env each time. More details in the `README.MD` file. 

Would be great to get this script run as part of travis like #45 - are you happy for us to hook it up?

This work is co-authored with @stuartleeks 

The tests seem to block on #33 occasionally when creating the cluster intermittent failures occur as seen before related to `associated worker environments`.

image


stuartleeks:
This PR ended up being a little bigger than expected as we needed to set up some infra in the code around creating the additional Azure resources needed to integration test these changes. Hopefully, this will make it easier for further tests of this nature in future :-)

stikkireddy:
hey @lawrencegripper & @stuartleeks  thanks for the PR I should be able to finish reviewing this by the afternoon today (EST)

stuartleeks:
@stikkireddy  - awesome, thanks for the update :-)

stuartleeks:
Hey @stikkireddy - any update on this? Happy to answer any questions you may have 😃

stikkireddy:
LGTM @stuartleeks this still has issues with the tests sporadically and I believe the solution for making sure that the clusters api works properly after creating a workspace is proposed to be solved via #53. I will merge this in and hopefully that #53 can help successfully run this test 10/10 times successfully.

PR #53, opened 2020-05-14
Implement retry on transient errors

stuartleeks:
As discussed in #33, there are cases where the API endpoint returns errors such as `"Succeeded{"error_code":"INVALID_PARAMETER_VALUE","message":"Unknown worker environment WorkerEnvId(workerenv-4312344789891641)"}`. 

These errors have been observed shortly after a workspace has been created. Additionally, we have seen these errors after successful API calls, so simply waiting for a successful API response isn't sufficient.

The PR changes the `service.DBApiClientConfig` to use [retryablehttp.Client](https://github.com/hashicorp/go-retryablehttp) instead of the standard `http.Client`. HTTP responses with a status code >= 400 and where the message contains any of the strings identified in #33 are treated as retryable errors.

We included logging, so if `TF_LOG=debug` is set then the log output will include `Failed request detected: Retryable type found. Attempting retry...` when a retryable error is encountered.

@lawrencegripper has run a number of test deployments where we have seen this output, demonstrating that the retry logic is kicking in.


stuartleeks:
@stikkireddy - have rebased this on master now that you have merged #51 :-)

lawrencegripper:
I've fixed up the build. I'd like to run another local test of the full integration test suite to check there isn't anything I've missed along the same lines.

lawrencegripper:
Looks like this is behaving nicely when testing with `./integration-environment-azure/run.sh` acceptance tests for mounts.

Unrelated: Should I update `make terraform-acc-azure` to run this script too?

stikkireddy:
This looks good to me @lawrencegripper I will integrate this into make terraform-acc-azure as part of the travis-ci branch to be able to run these tests along with the aws during merge to master. But if you would like to make the changes in this PR please go right ahead. Just let me know if this is ready for me to test & merge.

stikkireddy:
> **NOTE:** This PR is marked as WIP as it builds on some of the capabilities added in #55 so doesn't make sense to merge until #55 is merged.

I think thats the incorrect issue number/pr number please let me know if that is the case @stuartleeks. Issue #55 is a potential nil pointer panic due to not verifying if the pointer is not nill.

stuartleeks:
Oops! That's also not relevant now, so will just remove it!

databricks/azure_auth.go
```diff
@@ -213,53 +212,11 @@ func (a *AzureAuth) initWorkspaceAndGetClient(config *service.DBApiClientConfig)
...
 
-	// So when the workspace is initially created sometimes it fails to perform api calls so this is a simple test
-	// to verify that the workspace has been created successfully. The retry is intentional as sometimes after workspace
-	// creation the API's will not work correctly. This may also
-	err = validateWorkspaceApis(10, 30, func(attempt int) error {
-		_, err = dbClient.Clusters().ListNodeTypes()
-		return err
-	})
+	// Spin for a while while the workspace comes up and starts behaving.
+	_, err = dbClient.Clusters().ListNodeTypes()
```
stikkireddy, line 216: With the changes made it may be worth removing this arbitrary check at the provider configuration. This was pretty much a hack to verify if the workspace is setup properly. With the improvements made in this PR to use the hashicorp httpretry it may be worth removing lines 215 - 219. We may want to just let the resource object fail with an invalid token rather than the provider configure failure. 

I am not sure if was the intent of the provider configure to verify that the client object will throw a 403. Any thoughts on this @stuartleeks.

stuartleeks: We added this check back in as it got us past a few remaining errors :-)

stuartleeks:
@stikkireddy re the travis integration, it looks like you have that underway in another PR? If so, would you be happy to merge this PR and pick up the integration in the other PR?

stikkireddy:
Awesome this looks great to me! Thanks for the contribution! Merging this in.

Issue #65, opened 2020-05-26
[ISSUE] Manually deleted mount throws error on next tf plan


sebastus:
Hi folks,

If you create an azure blob mount with tf, delete it manually (via databricks notebook), then re-run terraform plan, a file not found error is thrown.

### Terraform Version
terraform -v == 0.12.19
+ azurerm == 2.11.0
+ random == 2.2.1
+ databricks == 0.1.0

### Affected Resource(s)
Please list the resources as a list, for example:
- databricks_azure_blob_mount

### Terraform Configuration Files
```hcl
variable group_name {}

variable "client_id" {
  type = string
}
variable "client_secret" {
  type = string
}
variable "tenant_id" {
  type = string
}
variable "subscription_id" {
  type = string
}
variable "dbws_name" {
  type = string
}

provider "azurerm" {
  version = "~> 2.3"
  features {}
  
  subscription_id = var.subscription_id
  client_id = var.client_id
  client_secret = var.client_secret
  tenant_id = var.tenant_id
}

provider "random" {
  version = "~> 2.2"
}

resource "random_string" "name_prefix" {
  special = false
  upper   = false
  length  = 6
}

resource "azurerm_resource_group" "example" {
  name     = var.group_name
  location = "eastus" # note must be lower without spaces not verbose style
}

resource "azurerm_databricks_workspace" "example" {
  name                = var.dbws_name
  resource_group_name = azurerm_resource_group.example.name
  location            = azurerm_resource_group.example.location
  sku                 = "standard"
}

resource "azurerm_storage_account" "account" {
  name                     = "${random_string.name_prefix.result}blob"
  resource_group_name      = azurerm_resource_group.example.name
  location                 = azurerm_resource_group.example.location
  account_tier             = "Standard"
  account_replication_type = "LRS"
  account_kind             = "StorageV2"
}

resource "azurerm_storage_container" "example" {
  name                  = "dev"
  storage_account_name  = azurerm_storage_account.account.name
  container_access_type = "private"
}

resource "databricks_secret_scope" "terraform" {
  name                     = "terraform"
  initial_manage_principal = "users"
}

resource "databricks_secret" "blob_account_key" {
  key          = "blob_account_key"
  string_value = azurerm_storage_account.account.primary_access_key
  scope        = databricks_secret_scope.terraform.name
}

provider "databricks" {
  azure_auth = {
    managed_resource_group = azurerm_databricks_workspace.example.managed_resource_group_name
[...]

PR #69, opened 2020-05-27
Manual delete of databricks_azure_blob_mount results in error on next tf plan

sebastus:
Fixes Issue #65.

Changes include:
* the bug fix
* new integration test for databricks_azure_blob_mount that replicates the scenario:
  * provision the environment (storage account, db ws, mount) with tf script
  * delete the mount (without using tf script)
  * re-run tf apply
* adds a storage account and a couple of env vars to prereqs.tf
* renames storage accounts to disambiguate


stuartleeks:
LGTM - thanks @sebastus

integration-environment-azure/prereqs.tf
```diff
@@ -62,13 +78,21 @@ output "workspace_name" {
...
+  value = azurerm_storage_account.adlsaccount.location
 }
 
 output "rg_name" {
   value = azurerm_resource_group.example.name
 }
+
+output "blob_storage_key" {
+  value = azurerm_storage_account.blobaccount.primary_access_key
+}
```
stikkireddy, line 94: This seems to be the storage account key. If we are using travis or other ci tools it will display the outputs during the apply phase. Please add the sensitive attribute so this does not get displayed in the apply phase for running the integration tests. https://www.terraform.io/docs/configuration/outputs.html#sensitive-suppressing-values-in-cli-output

sebastus: Added sensitive to the key output. re: cluster. great idea - we'll give it a think.

stikkireddy:
Thanks for the contribution @sebastus LGTM merging in.

PR #72, opened 2020-05-28
Add single_user_name to cluster resource

### LINKED ISSUE #71: [FEATURE] Add cluster.single_user_name
**Is your feature request related to a problem? Please describe.**
Currently unable to create interactive single-user clusters as the cluster resource doesn't allow seeting the `single_user_name` property

**Describe the solution you'd like**
Adding the `single_user_name` property to the cluster resource would solve this

**Describe alternatives you've considered**
None (other than falling back to CLI etc)

**Additional context**
This would support #63

stuartleeks:
Fixes #71

### DIFF
--- a/website/content/Resources/cluster.md
+++ b/website/content/Resources/cluster.md
@@ -326,7 +326,7 @@ feature requires specific AWS permissions to function correctly - refer to Autos
 cluster belongs. Refer to Instance Pools API for details.
 
 #### - `single_user_name`:
-> **(Optional)** The optional user name of the user to assign to an interactive cluster. This is required when using standard AAD Passthrough for Azure Datalake Storage (ADLS).
+> **(Optional)** The optional user name of the user to assign to an interactive cluster. This is required when using standard AAD Passthrough for Azure Datalake Storage (ADLS) with a single-user cluster (i.e. not high-concurrency clusters).
 
 #### - `idempotency_token`:
 > **(Optional)** An optional token that can be 
--- a/website/content/Resources/cluster.md
+++ b/website/content/Resources/cluster.md
@@ -325,6 +325,9 @@ feature requires specific AWS permissions to function correctly - refer to Autos
 > **(Optional)** The optional ID of the instance pool to which the 
 cluster belongs. Refer to Instance Pools API for details.
 
+#### - `single_user_name`:
+> **(Optional)** The optional user name of the user to assign to an interactive cluster.
+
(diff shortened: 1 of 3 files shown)


stuartleeks:
Oops - thanks for catching that @stikkireddy ! Updated now :-)

stuartleeks:
rebased on latest `master`

website/content/Resources/cluster.md
```diff
@@ -325,6 +325,9 @@ feature requires specific AWS permissions to function correctly - refer to Autos
 > **(Optional)** The optional ID of the instance pool to which the 
 cluster belongs. Refer to Instance Pools API for details.
 
+#### - `single_user_name`:
+> **(Optional)** The optional user name of the user to assign to an interactive cluster.
```
stikkireddy, line 329: Hey sorry to make this a bit more cumbersome but would it be possible to also mention that, this field is required when using the "standard" configuration for adls passthrough for passing the AD tokens to the cluster from the user using the cluster. I will merge this in after those changes. I will be still here for the next 20-30 minutes or so to try to get this closed for you 😄.

stuartleeks: You mean add something like "This is required when using standard AAD Passthrough for Azure Datalake Storage (ADLS)"?

stuartleeks: @stikkireddy  - just added that text :-)

stikkireddy: Yep, or document this requirement on the mount docs. Which ever is appropriate, I think it should be in at least one of those locations. I am leaning more towards the cluster docs as this field is part of the cluster resource but it only becomes relevant when trying to use the AAD passthrough feature on a "standard cluster".

On a "high concurrency cluster", this is not a required field you would just need to enable the following spark config:
```
"spark.databricks.cluster.profile": "serverless",
"spark.databricks.repl.allowedLanguages": "python,sql",
"spark.databricks.passthrough.enabled": "true",
"spark.databricks.pyspark.enableProcessIsolation": "true"
```

 (Sorry for the typo didnt mean ADLS passthrough but aad credential passthrough)

stuartleeks: Cool - I've just updated the note to call out that it is required for *single-user* clusters, not high-concurrency. Hope that is better now

stikkireddy:
Thanks for the contribution LGTM 😄Will merge right after the travis checks are done 👍

PR #76, opened 2020-05-31
Avoid accidental workspace creation

### LINKED ISSUE #54: [ISSUE] Terraform plan command actually creates the Databricks workspace when an error is encountered
Hi there,

### Terraform Version
+ terraform v0.12.24
+ provider.azurerm v2.9.0
+ provider.databricks v0.1.0

### Affected Resource(s)
- databricks_workspace

```hcl
variable "client_id" {}
variable "client_secret" {}
variable "tenant_id" {}
variable "subscription_id" {}
variable "resource_group" {}
variable "managed_resource_group_name" {}

provider "azurerm" {
  version         = ">= 2.3.0"
  client_id       = var.client_id
  client_secret   = var.client_secret
  tenant_id       = var.tenant_id
  subscription_id = var.subscription_id
  features {}
}

resource "azurerm_databricks_workspace" "demo" {
  location                    = "westeurope"
  name                        = "databricks-demo-workspace"
  resource_group_name         = var.resource_group
  managed_resource_group_name = var.managed_resource_group_name
  sku                         = "standard"
}

resource "databricks_cluster" "demo" {
  autoscale {
    min_workers = 2
    max_workers = 8
  }
  cluster_name            = "databricks-demo-cluster"
  spark_version           = "6.4.x-scala2.11"
  node_type_id            = "Standard_DS3_v2"
  autotermination_minutes = 30
}

provider "databricks" {
  version = ">= 0.1"
  azure_auth = {
    managed_resource_group = azurerm_databricks_workspace.demo.managed_resource_group_name
    azure_region           = azurerm_databricks_workspace.demo.location
    workspace_name         = azurerm_databricks_workspace.demo.name
    resource_group         = azurerm_databricks_workspace.demo.resource_group_name
    client_id              = var.client_id
    client_secret          = var.client_secret
    tenant_id              = var.tenant_id
    subscription_id        = var.subscription_id
  }
}
```

### Debug Output
[https://gist.github.com/christophecremon/f839ac7b0f342d277f0ceaba2fbab165](https://gist.github.com/christophecremon/f839ac7b0f342d277f0ceaba2fbab165)

### Expected Behavior
Terraform generates an execution plan.
Terraform will perform the following actions, upon apply command has been executed: 
- create a Databricks workspace
- create a Databricks cluster

### Actual Behavior
An error is generated: 
Error 404 The workspace with resource ID /subscriptions/<REDACTED_FOR_GITHUB>/resourceGroups/databricks-rg/providers/Microsoft.Databricks/workspaces/databricks-demo-workspace could not be found.

Terraform actually creates the Databricks workspace, with a Premium Pricing Tier, even if terraform apply command has not been executed.

### Steps to Reproduce
1. `terraform plan`

stuartleeks:
Avoid accidental workspace creation as per discussion in #54.
Update to handle `terraform plan` when the workspace doesn't yet exist by adding an `EnsureConfig` function on the client that is called before performing a query. This is set in the `azure_auth` provider to ensure that we have a token and to generate it if not.

Fixes #54

### DIFF
--- a/client/service/apis.go
+++ b/client/service/apis.go
@@ -99,9 +99,11 @@ func (c DBApiClient) Commands() CommandsAPI {
 }
 
 func (c *DBApiClient) performQuery(method, path string, apiVersion string, headers map[string]string, data interface{}, secretsMask *SecretsMask) ([]byte, error) {
-	err := c.EnsureConfig(c)
-	if err != nil {
-		return []byte{}, err
+	if c.EnsureConfig != nil {
+		err := c.EnsureConfig(c)
+		if err != nil {
+			return []byte{}, err
+		}
 	}
 	return PerformQuery(c.Config, method, path, apiVersion, headers, true, false, data, secretsMask)
 }
--- a/client/service/apis.go
+++ b/client/service/apis.go
@@ -1,12 +1,24 @@
 package service
 
+import "fmt"
+
 var scimHeaders = map[string]string{
 	"Content-Type": "application/scim+json",
 }
 
+type WorkspaceDoesNotExistError struct {
+	WorkspaceID string
+	Err         error
+}
+
+func (e *WorkspaceDoesNotExistError) Error() string {
+	return fmt.Sprintf("The workspace does not exist: %s", e.WorkspaceID)
+}
+
 // DBApiClient is the client struct that contains clients for all the services available on Databricks
 type DBApiClient struct {
-	Config *DBApiClientConfig
+	Config       *DBApiClientConfig
+	EnsureConfig func(dbClient *DBApiClient) error // used to ensure that Config has been processed, e.g. to add auth headers
 }
 
 // SetConfig initializes the client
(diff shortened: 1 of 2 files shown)


stikkireddy:
hey @stuartleeks  & @EliiseS, can we take a quick look at #78 and evaluate if we need the provider configure to pass a pointer to the client or a value, as if we pass a pointer we avoid unnecessary token creation but it increases complexity by a bit because multiple goroutines for different resources share the same client object, the only mutable component is the token creation, and it also implies that the token will only live for 1 hour so will only last an hour. I think this is a much simpler approach, only problem is for larger projects we may end up making 100+ tokens depending on how many resources there. I think it may not be so expensive as those tokens expire in an hour but I will also let you weigh in your thoughts on that.

stuartleeks:
@stikkireddy - I agree that changing it to a pointer makes sense to avoid repeated calls to retrieve tokens. I skipped making this change in this PR as I wanted to keep it as simple as possible when I was looking at this over the weekend :-)
I see that #78 includes the change to use a pointer, so it is probably the better option - it should give better performance by avoiding the repeated calls.

stuartleeks:
Just in the middle of a couple of bits but will try to get this into our CI/CD system to test later this afternoon or tomorrow

stuartleeks:
@stikkireddy - a quick test with one of our scripts ran fine and created the workspace correctly

stuartleeks:
@stikkireddy  - shall I close this in favour of #78?

stikkireddy:
@stuartleeks sorry for the delayed response but yes please. I will merge that in if it looks good to you!

stuartleeks:
Closing now - look forward to your PR being merged :-)
Have been testing more today and looking good here

PR #78, opened 2020-06-01
Fix azure workspace creation bug

stikkireddy:
Fixes accidental workspace creation in #54.

Uses the CustomAuthorizer attribute in the client config to delay the token generation till an actual api call is called via performQuery. It also changes the providerConfigure to return a pointer to the client rather than the value of the client object itself. So a mutex was added to getOrCreateToken function to ensure that multiple tokens are not generated.

### DIFF
--- a/client/service/apis.go
+++ b/client/service/apis.go
@@ -17,22 +17,22 @@ func (c *DBApiClient) SetConfig(clientConfig *DBApiClientConfig) DBApiClient {
 }
 
 // Clusters returns an instance of ClustersAPI
-func (c DBApiClient) Clusters() ClustersAPI {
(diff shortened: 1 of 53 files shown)


stikkireddy:
@stuartleeks hey sorry this took a bit of time was a lot of files changed so had to make sure everything was working. I know you made a change I was just wondering if we can take a quick look at this and determine if we need the client object by the configure function to pass by value or pass a pointer to just one client to avoid repeating token generation

stuartleeks:
@stikkireddy - a quick test with one of our scripts ran fine and created the workspace correctly

stuartleeks:
@stikkireddy  - have been using this for various deploys today and looking good here

PR #85, opened 2020-06-08
[Work In Progress] Cluster Policies Support + Databricks CLI config reads

nfx:


databricks/provider.go
```diff
@@ -197,6 +218,41 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
 	return &dbClient, nil
 }
 
+// tryDatabricksCliConfigFile sets Host and Token from ~/.databrickscfg file if it exists
+func tryDatabricksCliConfigFile(d *schema.ResourceData, config *service.DBApiClientConfig) error {
+	configFile, err := homedir.Expand(d.Get("config_file").(string))
```
dmitrij2, line 223: 1. I would let user provide the full path to the config file. In that case you do not need `homedir` package.
2. If `d.Get("config_file")` does not return string here the code is going to panic I believe.

databricks/provider.go
```diff
@@ -197,6 +218,41 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
...
+	if err != nil {
+		return err
+	}
+	cfg, err := ini.Load(configFile)
+	if err != nil {
+		return fmt.Errorf("Authentication is not configured for provider. Please configure it\n"+
+			"through one of the following options:\n"+
+			"1. DATABRICKS_HOST + DATABRICKS_TOKEN environment variables.\n"+
+			"2. host + token provider argumeents.\n"+
```
dmitrij2, line 232: `argumeents` -> `arguments`

databricks/provider.go
```diff
@@ -197,6 +218,41 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
...
+			"3. Run `databricks configure --token` that will create %s file.\n\n"+
+			"Please check https://bit.ly/2XCtuZU for details", configFile)
```
dmitrij2, line 234: I'm out of terraform context, but I would not understand that I need to create an `ini` file with host and token in it by reading this error message. Why it's not just printed to console instead of passing as an error, by the way? In other places we return internal errors, do we print those directly to the client?
Why is it an `ini` and not `json` or `yaml` by the way?

databricks/provider.go
```diff
@@ -197,6 +218,41 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
...
+	}
+	if profile, ok := d.GetOk("profile"); ok {
```
dmitrij2, line 236: Invert `ok` -> `!ok` to reduce nesting.

databricks/provider.go
```diff
@@ -197,6 +218,41 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
...
+
+		host := dbcliConfig.Key("host").String()
+		if "" == host {
+			return fmt.Errorf("Config file %s is corrupt: cannot find host in %s profile",
+				configFile, profile)
+		}
+		config.Host = host
+	}
+
+	return nil
```
dmitrij2, line 253: `schema.ResourceData` does not contain required fields to fill. Is it really a positive outcome for this function?

databricks/provider.go
```diff
@@ -209,6 +265,11 @@ func providerConfigure(d *schema.ResourceData, providerVersion string) (interfac
 		if token, ok := d.GetOk("token"); ok {
 			config.Token = token.(string)
 		}
+		if config.Host == "" || config.Token == "" {
+			if err := tryDatabricksCliConfigFile(d, &config); err != nil {
+				return nil, err
+			}
+		}
```
dmitrij2, line 272: I'd probaly go with something like this for separation of concerns and readability:
```
if config.Host == "" || config.Token == "" {
  pathRaw, ok := d.GetOk("config_file")
  if !ok { return err... }
  path, ok  := pathRaw.(string) // Not sure if ok check is needed here. I see it's never done, so maybe it doesn't panic if d.GetOk returns ok = true
  if !ok {
    return fmt.errorf(...)
  }
  profile, ok := d.Get("profile") 
... // similar to path

  host, token, err := getCredentials(path, profile) // this function would open file and try to read profile settings.
  if err {...}
  config.Host = host
  config.Token = token
}
```

databricks/provider.go
```diff
@@ -197,6 +218,41 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
...
+			return fmt.Errorf("Config file %s is corrupt: cannot find host in %s profile",
+				configFile, profile)
+		}
+		config.Host = host
+	}
+
+	return nil
+}
+
 func providerConfigure(d *schema.ResourceData, providerVersion string) (interface{}, error) {
```
dmitrij2, line 256: Name of this function is weird. It should be `buildDatabricksClient` and it should not return `interface{}`, but some proper interface or struct. Should be addressed in the separate PR though.

databricks/provider_test.go
```diff
@@ -76,3 +79,78 @@ resource "databricks_scim_group" "my-group-azure3" {
...
+	err := testAccProvider.Configure(terraform.NewResourceConfigRaw(raw))
+	assert.NotNil(t, err)
+}
+
+func TestProvider_InvalidProfileGivesError(t *testing.T) {
+	var raw = make(map[string]interface{})
+	raw["config_file"] = "testdata/.databrickscfg"
+	raw["profile"] = "invalidhost"
+	err := testAccProvider.Configure(terraform.NewResourceConfigRaw(raw))
+	assert.NotNil(t, err)
```
dmitrij2, line 141: It might also be a good idea to actually check that the error message is the one that you expect and contains all the context required.

databricks/provider.go
```diff
@@ -209,6 +265,11 @@ func providerConfigure(d *schema.ResourceData, providerVersion string) (interfac
 		if token, ok := d.GetOk("token"); ok {
 			config.Token = token.(string)
 		}
+		if config.Host == "" || config.Token == "" {
+			if err := tryDatabricksCliConfigFile(d, &config); err != nil {
+				return nil, err
```
dmitrij2, line 270: Since error does not contain stack trace as java/c# exception - it is always a good idea to wrap error into another error as you go up the stack for better debugability. You can use [errors.Wrapf](https://godoc.org/github.com/pkg/errors#Wrapf) for that.
Here it would be `return nil, errors.Wrap(err, "getting databricks credentials from config failed")`

databricks/provider.go
nfx: Terraform prints errors in it's own way, so i just rely on standard mechanism of it's framework.

Regarding INI files - 🤷, https://docs.databricks.com/dev-tools/cli/index.html#set-up-authentication, https://github.com/databricks/databricks-cli/blob/master/databricks_cli/configure/provider.py - most likely because https://docs.python.org/3/library/configparser.html is an INI parser

databricks/provider_test.go
nfx: according to certain beliefs, asserting exception message was not considered a good practice - e.g. because it may change the content of the string or be translated to other language, therefore making test "more overfitting" and fragile to insignificant changes.

databricks/provider.go
nfx: regarding 2 - terraform schemas handle the typing already, so I would assume it's safe to rely on those.

Issue #86, opened 2020-06-09
[ISSUE] ADLS Mount with cluster that has been deleted causes error in tf plan


storey247:
### Terraform Version
0.12.26

### Affected Resource(s)
Please list the resources as a list, for example:
- databricks_cluster
- databricks_azure_adls_gen2_mount

### Terraform Configuration Files
```hcl
        resource "databricks_cluster" "cluster" {
		num_workers = 1
		spark_version = "6.4.x-scala2.11"
		node_type_id = "Standard_D3_v2"
		autotermination_minutes = 15
	} 

	resource "databricks_secret_scope" "terraform" {
	  name                     = "terraform${databricks_cluster.cluster.cluster_id}"
	  initial_manage_principal = "users"
	}
	
	resource "databricks_secret" "client_secret" {
	  key          = "datalake_sp_secret"
	  string_value = "%[2]s"
	  scope        = databricks_secret_scope.terraform.name
	}
	
	resource "databricks_azure_adls_gen2_mount" "mount" {
	  cluster_id             = databricks_cluster.cluster.id
	  container_name         = "dev" # Created by prereqs.tf
	  storage_account_name   = "%[9]s"
	  directory              = ""
	  mount_name             = "localdir${databricks_cluster.cluster.cluster_id}"
	  tenant_id              = "%[3]s"
	  client_id              = "%[1]s"
	  client_secret_scope    = databricks_secret_scope.terraform.name
	  client_secret_key      = databricks_secret.client_secret.key
	  initialize_file_system = true
	}
```

### Panic Output
N/A

### Expected Behavior
When the cluster that originally created the mount has been deleted inside databricks, the `tf plan` should identify this. It should then identify that the mount most likely needs to be re-created as there is a high likelihood that the cluster is also in the same terraform configuration.

### Actual Behavior
The provider throws an error during `tf plan` and renders the state file unusable unless you manually remove the mount from state.

```
Error: status 400: err Response from server {"error_code":"INVALID_PARAMETER_VALUE","message":"Cluster <some cluster id> does not exist"}
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Log into databricks or use databricks REST api to delete the cluster associated in state with mount.
3. `terraform plan` >>> Error occurs here during refresh of state

storey247:
Have noticed that this error is also present in the blob mount resource too

stikkireddy:
closing the issue as pr #87 has been merged in

PR #87, opened 2020-06-09
mounts now handle deleted cluster correctly

storey247:
- Reworked the logic of the mounts to check for the error state that the cluster is invalid
- When the error is detected the id of the resource is cleared to indicate it must be re-created
- Tests added to check updated logic 

Ideally there is no need to re-create the cluster, but as we cannot check that the mount really does exist, this feels like the safest way to solve the issue.

This addresses issue #86


databricks/resource_databricks_azure_adls_gen2_mount_test.go
```diff
@@ -39,7 +67,7 @@ func TestAccAzureAdlsGen2Mount_capture_error(t *testing.T) {
 }
 
 func testAccAzureAdlsGen2Mount_correctly_mounts() string {
-	clientID := os.Getenv("ARM_CLIENT_ID")
+	clientID := os.Getenv("DATABRICKS_AZURE_CLIENT_ID")
```
stuartleeks, line 70: What was the reason for changing this?

stuartleeks:
@storey247 - Nice! I know that we've bumped into this issue quite a few times. Will be good to have it fixed 😄

databricks/resource_databricks_azure_adls_gen2_mount_test.go
storey247: hmmmmmm, i think this was when i was struggling to get the tests running as the .env wasn't applying and clientID was being set to null causing tests to fail.

I since figured out the issue there so can probably roll this back. The actual values of `ARM_CLIENT_ID` and `DATABRICKS_AZURE_CLIENT_ID` are the same though.

Will revert this change :)

stikkireddy:
hey quick question on the implementation, if we cant find the cluster, we are setting the id of the resource to null, but it would go through the process of mounting again if the user tries to re applies the resource correct? Then wouldnt the mount fail because it is already found?

storey247:
@stikkireddy so i tested this out and because we set the id to be `""` this causes the `tf plan` to mark this resource as needing to be destroyed. Therefore the mount will be removed and re-created.

As I said, this isn't ideal as the mount isn't exactly tightly coupled to the cluster id, but this removes the issue where you can no longer plan if the cluster id is invalid because the provider throws an error

stikkireddy:
@storey247 awesome thanks for the explanation, just a quick question, so it will get past the plan phase because during the refreshing of state the read will work as it will set it to "", but during the destroy the cluster is invalid so it will throw runtime during the apply right?

storey247:
That’s actually a good point, my acceptance tests didn’t fail on this, so not sure if the tests are missing something  🤔 

Will test the scenario manually using a local build of the provider and see what happens

storey247:
@stikkireddy ok so i double checked all this, setting the ID to be empty doesn't actually cause the old mount to be destroyed. Interestingly it just seems to make terraform forget about it during the plan.

However, there was an issue that arose because it would attempt to re-create the mount with the same configuration, and this would throw an error inside the mount create logic due to directory already being mounted. I have now added a check prior to creating the mount to check for an existing mount, and if it already exists, skip trying to re-create it

stikkireddy:
@storey247 thank you very much for contributing this fix 👍

PR #101, opened 2020-06-15
Group role member feature

stikkireddy:
Created three new resources with optional review for adding one more:

Resources: 

Pushed & Tested:
1. databricks_group (only supports display name with allow_cluster_create)
2. databricks_group_role (1 to 1 mapping between instance profile and group)
3. databricks_group_member (1 to 1 mapping between parent group [group_id] and member [group_id or user_id])

Optional & Up for Discussion:
1. databricks_group_members (1 to all mappings between one parent group and all its members which will greatly reduce stress against SCIM api for patch requests to add members)

The purpose of the above resources are to manage groups only at the name & entitlement level. 
The databricks_group only will include display name and entitlements to create or delete clusters. 

Up for discussion to also move that outside of the group object because of the permissions api. I have flagged them as deprecated. 

Example resource objects:

Creating member links:

```hcl
resource "databricks_group" "my_group" {
  display_name = "%[1]s"
}
resource "databricks_group" "my_sub_group_a" {
  display_name = "sub_a_%[1]s"
}
resource "databricks_group" "my_sub_group_b" {
  display_name = "sub_b_%[1]s"
}
resource "databricks_group_member" "my_member_a" {
  group_id = databricks_group.my_group.id
  member_id = databricks_group.my_sub_group_a.id
}
resource "databricks_group_member" "my_member_b" {
  group_id = databricks_group.my_group.id
  member_id = databricks_group.my_sub_group_b.id
}
```

Creating Roles:

```hcl
resource "databricks_instance_profile" "instance_profile" {
  instance_profile_arn = "%s"
  skip_validation = true
}
resource "databricks_group" "my_group" {
  display_name = "%s"
}
resource "databricks_group_role" "my_group_role" {
  group_id = databricks_group.my_group.id
  instance_profile_id = databricks_instance_profile.instance_profile.id
}
```


databricks/resource_databricks_group_role.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"fmt"
+	"github.com/databrickslabs/databricks-terraform/client/model"
+	"github.com/databrickslabs/databricks-terraform/client/service"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"log"
+	"strings"
+)
+
+func resourceGroupRole() *schema.Resource {
```
nfx, line 12: this resource might have an alias - `databricks_group_instance_profile` :) because here we semantically mix two things and confuse outside user

stikkireddy: Great point, I always found this quite confusing as well... I will switch it to databricks_group_instance_profile

Issue #52, opened 2020-05-13
[ISSUE] Warnings or "Error: Response from server (429)" when applying notebooks


mikemowgli:
Hi there,

### Terraform Version
```
$ terraform version
Terraform v0.12.24
+ provider.azuread v0.8.0
+ provider.azurerm v2.5.0
+ provider.databricks v0.1.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- databricks_notebook

### Terraform Configuration Files
```hcl
resource "databricks_notebook" "tamers-databricks-nb-handson-unsupervised-02" {
  content   = filebase64("notebooks/handson-unsupervised-02-end-to-end-machine-learning-project.py")
  language  = "PYTHON"
  path      = "/Shared/xxxx/handson-unsupervised-02-end-to-end-machine-learning-project.py"
  overwrite = false
  mkdirs    = true
  format    = "SOURCE"
}
```

### Debug Output
https://gist.github.com/mikemowgli/42c32bd11e21d926cadc7b05788d49d7

### Expected Behavior
No error nor warning when applying

### Actual Behavior
warning or errors

### Steps to Reproduce
`TF_LOG=DEBUG terraform apply -target databricks_notebook.tamers-databricks-nb-handson-unsupervised-02`

### Important Factoids
In an Azure Databricks workspace, a terraform apply on a specific notebook (using `-target` option) yields the log in the gist link: only a warning, and the content of the notebook not updated but left as-is.
However, when applying all my terraform plan, I get the same debug log, with only this difference at the very end:
```
databricks_notebook.tamers-databricks-nb-handson-unsupervised-02: Creation complete after 3s [id=/Shared/xxxx/handson-unsupervised-02-end-to-end-machine-learning-project.py]

Error: Response from server (429)


2020-05-13T12:58:48.240+0200 [DEBUG] plugin: plugin process exited: path=/home/mvdborne/.terraform.d/plugins/linux_amd64/terraform-provider-databricks_v0.1.0 pid=8267
2020-05-13T12:58:48.240+0200 [DEBUG] plugin: plugin exited
$ echo $?
1
```

### References
I'm impacted by [the notebook content issue](https://github.com/databrickslabs/databricks-terraform/issues/42), so this one might be a consequence of issue 42.

mikemowgli:
Please note that I can mitigate the Error 429 with `terraform apply -parallelism=1` (value of -parallelism=2` gives the same error).
Though the warnings persists of course.

[...]

Issue #42, opened 2020-05-06
[ISSUE] Notebooks content always seen as different even when no file has changed


damoodamoo:
### Terraform Version
```
Terraform v0.12.24
+ provider.azurerm v1.44.0
+ provider.databricks v0.1.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- databricks_notebook

### Terraform Configuration Files
```hcl
resource "databricks_notebook" "notebook" {
  content = filebase64("${path.module}/nb/notebook1.scala")
  path = "/Shared/Notebooks/notebook1.scala"
  overwrite = false
  mkdirs = true
  format = "SOURCE"
  language = "SCALA"
}
```

### Debug Output
On first run, the resource is seen as an add, correctly, and deploys.
On subsequent `tf apply` it sees the content has having changed:
```
  # databricks_notebook.notebook must be replaced
-/+ resource "databricks_notebook" "notebook" {
      ~ content     = "3750311991" -> "2327128740" # forces replacement
        format      = "SOURCE"
      ~ id          = "/Shared/Notebooks/notebook1.scala" -> (known after apply)
        language    = "SCALA"
        mkdirs      = true
      ~ object_id   = 4081355166030977 -> (known after apply)
      ~ object_type = "NOTEBOOK" -> (known after apply)
        overwrite   = false
        path        = "/Shared/Notebooks/notebook1.scala"
    }
```

This also exacerbates the issue #41 when deploying multiple files as it's trying to re-create all of them every time. 

### Expected Behavior
The content should be seen as the same and no-op

### Actual Behavior
The content is seen as different and a delete/create is needed

### Steps to Reproduce
Use the above hcl to deploy a notebook, then run `tf apply` again.

stikkireddy:
Hey @damoodamoo thanks for pointing this out, the issue seems to be that there is no normalization factor for the notebooks before the checksum is calculated. Unfortunately, across different clouds and potentially different Databricks deployment versions, what you import and what you export may not yield the same values.

Below is a potential solution that seems to work for me for non DBC notebooks. 

```
func normalizeB64Data(b64Content string) string {
	dataArr, _ := base64.StdEncoding.DecodeString(b64Content)
	scalaMagic := "// MAGIC "
	pythonRMagic := "# MAGIC "
	sqlMagic := "-- MAGIC "
	filteredDataArr := filter(strings.Split(string(dataArr), "\n"), func(s string) bool {
[...]

Issue #41, opened 2020-05-06
[ISSUE] 429 Conflicts when deploying multiple notebooks


damoodamoo:
### Terraform Version
```
Terraform v0.12.24
+ provider.azurerm v1.44.0
+ provider.databricks v0.1.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- databricks_notebook

### Terraform Configuration Files
```hcl
resource "databricks_notebook" "notebook" {
  for_each = fileset("${path.module}/notebooks", "*")

  content = filebase64("${path.module}/notebooks/${each.value}")
  path = "/Shared/Notebooks/${each.value}"
  overwrite = false
  mkdirs = true
  format = "SOURCE"
  language = SCALA
}
```
The above loops a number of files in a local dir and deploys them to databricks using the `databricks_notebook` provider. When more than about 3/4 files are present i'm seeing pretty regular `429` errors returned when running `tf apply`. Not sure if there's a race condition whereby the notebook is still being deleted when `tf` is trying to re-create it. 

### Debug Output
On first create, all is successful. On further runs of `tf apply` it sees each notebook as needing to be recreated (will log separate issue for this), and when trying to recreate we usually hit a `429` error.
```
databricks_notebook.notebook["notebook1.scala"]: Destroying... [id=/Shared/Notebooks/notebook1.scala]
databricks_notebook.notebook["notebook3.scala"]: Destroying... [id=/Shared/Notebooks/notebook3.scala]
databricks_notebook.notebook["notebook2.scala"]: Destroying... [id=/Shared/Notebooks/notebook2.scala]
databricks_notebook.notebook["notebook3.scala"]: Destruction complete after 0s
databricks_notebook.notebook["notebook3.scala"]: Creating...
databricks_notebook.notebook["notebook2.scala"]: Destruction complete after 0s
databricks_notebook.notebook["notebook2.scala"]: Creating...
databricks_notebook.notebook["notebook3.scala"]: Creation complete after 1s [id=/Shared/Notebooks/notebook3.scala]
databricks_notebook.notebook["notebook2.scala"]: Creation complete after 1s [id=/Shared/Notebooks/notebook2.scala]

Error: Response from server (429) 
```
### Expected Behavior
That the notebooks get re-created.

### Actual Behavior
429 errors

### Steps to Reproduce
1. Create a folder next to your tf script called "notebooks", and add a number of dummy scala files. 
[...]

Issue #19, opened 2020-04-29
[TEST] Acceptance tests for Notebooks


stikkireddy:
**Our goal is to make sure that there is acceptance tests for all resources.**
There should be an acceptance test for Azure Mounts (Blob (both SAS key and Access Key), ADLS gen 1, ADLS gen 2).

PR #111, opened 2020-06-16
Fix Notebook issues

stikkireddy:
The goal of this PR is to fix the following issues:
1. #42 
2. #41 
3. #19 
4. #52 

This should attempt to fix the following issues:
1. 429 too many requests are thrown every time multiple delete requests are called for both notebooks and folders
2. Notebooks with type source would always through a diff.

Solution:
1. Implemented a fix for the mkdirs using a mutex to lock the goroutines access to the Mkdirs method. (This is temporary till mkdirs is removed from the databricks_notebook resource)
2. Added a function to normalize the SOURCE type files and added tests
3. Added a databricks_folder resource to manage mkdir rather than a flag on the notebooks. It does not scale well when creating multiple notebooks in parallel in the same folder. Terraform attempts to make the same directory all at once.
4. Deletes are throttled so added a resource.retry for both folders and notebooks.


stikkireddy:
@damoodamoo can you please do a quick review of this?

databricks/resource_databricks_notebook_test.go
```diff
@@ -1 +1,476 @@
...
+		},
+	}, resourceNotebook, nil, func(d *schema.ResourceData, c interface{}) error {
+		d.SetId(testId)
+		return resourceNotebookDelete(d, c)
+	})
+	assert.NoError(t, err, err)
+	assert.Equal(t, testId, d.Id())
+}
+
+func TestAccAwsNotebookResource_multiple_formats(t *testing.T) {
```
nfx, line 271: this looks strange :)

docs/resources/notebook.md
```diff
@@ -0,0 +1,73 @@
...
+  language = "PYTHON"
+  format = "SOURCE"
+}
+```
+    
+## Argument Reference
+
+The following arguments are supported:
+
+#### - `content`:
```
nfx, line 30: perhaps in a separate PR, but this should be formatted according to registry docs

Issue #34, opened 2020-05-04
[FEATURE] Migrate to new unique databricks hostnames


sdebruyn:
**Is your feature request related to a problem? Please describe.**
Currently not a problem, but it's advised to start using the new unique URLs for each Databricks workspace as documented in https://docs.microsoft.com/en-us/azure/databricks/release-notes/product/2020/april#unique-urls-for-each-azure-databricks-workspace

**Describe the solution you'd like**
Replace the current code that uses https://<region>.azuredatabricks.net/<workspaceid>

**Additional context**
The current hostnames have not been deprecated (yet) so we have still time.

sdebruyn:
Tracked in the azurerm provider in https://github.com/terraform-providers/terraform-provider-azurerm/issues/6732

nfx:
@sdebruyn we'll look into this issue and see what we can do. this is not a complicated one.

lawrencegripper:
Created a PR here to add into AzureRM: https://github.com/terraform-providers/terraform-provider-azurerm/pull/6973

stikkireddy:
@lawrencegripper also noticed that in the existing code I actually can get the url with the GET request from the management API so, we may not need the workspace url fetched from the Azure RM client as the management api provides a url with the new format. 

The following get request provides it:

```
https://management.azure.com/subscriptions/" + a.TokenPayload.SubscriptionID + "/resourceGroups/" + a.TokenPayload.ResourceGroup + "/providers/Microsoft.Databricks/workspaces/" + a.TokenPayload.WorkspaceName + "" +
		"?api-version=2018-04-01
```

lawrencegripper:
Ah nice work around, in good news the change is shipping now from azurerm so can be used shortly if that saves some effort https://github.com/terraform-providers/terraform-provider-azurerm/issues/6732

stuartleeks:
Just shipped 😀

https://github.com/terraform-providers/terraform-provider-azurerm/releases/tag/v2.11.0

EliiseS:
I'll be picking up this one in an attempt that it'll fix #92

PR #114, opened 2020-06-17
Use new workspace_url format + fix azure test

EliiseS:
Implementation of the issue: #34

Co-authored-by: Lawrence.Gripper@microsoft.com
<Lawrence.Gripper@microsoft.com>


lawrencegripper:
image

Looks like we're nearly there! Just running again to check it's consistent.

stikkireddy:
again thank you @lawrencegripper @EliiseS for picking this up!

integration-environment-azure/run.sh
```diff
@@ -32,16 +32,21 @@ fi
...
+export AZURE_REGION=$(terraform output location)
+export DATABRICKS_AZURE_MANAGED_RESOURCE_GROUP=$(terraform output workspace_managed_rg_name)
+
 echo -e "----> Running Azure Acceptance Tests \n\n"
 # Output debug log to file while tests run
 export TF_LOG_PATH=$PWD/tf.log
 # Run all Azure integration tests
-TF_LOG=debug TF_ACC=1 gotestsum --format short-verbose --raw-command go test -v -json -short -coverprofile=coverage.out -test.timeout 35m -run 'TestAccAzure' ./../...
\ No newline at end of file
+TF_LOG=debug TF_ACC=1 gotestsum --format short-verbose --raw-command \
```
stikkireddy, line 50: TF_LOG=debug, can we make this optional as in remove it from the script so that way, when this does get integrated to travis to run on master branch push, we will not expose debug level information to the logs. But the end user running the tests locally can export TF_LOG=debug

databricks/provider.go
```diff
@@ -192,6 +199,12 @@ func providerConfigureAzureClient(d *schema.ResourceData, providerVersion string
 	} else if os.Getenv("DATABRICKS_AZURE_WORKSPACE_NAME") != "" {
 		tokenPayload.WorkspaceName = os.Getenv("DATABRICKS_AZURE_WORKSPACE_NAME")
 	}
+	// TODO: Can required field not be set?
+	if workspaceURL, ok := azureAuthMap["workspace_url"].(string); ok {
```
stikkireddy, line 203: terraform behaves weirdly because this azure_auth is not a set/list block but a map so env defaults variables do not propagate.

EliiseS: Is there any reason it is type map over a list? Maybe we should refactor it into a list to avoid these issues

integration-environment-azure/run.sh
EliiseS: Done :)

databricks/provider.go
stikkireddy: yep we can refactor this to a list with a min items to 1... there should be no reason to not change it, only impact is that it may break existing provider configurations.

EliiseS:
@stikkireddy  `codecov/patch — 26.31% of diff hit (target 48.22%)` is failing, but we're not sure if we need to change anything or what we need to change

lawrencegripper:
Just investigating some odd behavior with this change. Lets not merge until we've got to the bottom of it.

EliiseS:
@stikkireddy The odd behaviour we were investigating concluded in #119, which, while not great, isn't blocking this issue. Are you happy to merge this PR?

stikkireddy:
@EliiseS on the above comment regarding turning the azure_auth into a list/set rather than a map, would you like to make that another PR?

If you would like to move it outside of the scope of this PR we can merge this in please let me know.

EliiseS:
@stikkireddy I actually looked into doing that and I was unable to get the tests to work with it. Let's move that discussion to another. I'm waiting for clarification from terraform in this PR: https://github.com/hashicorp/terraform-plugin-sdk/issues/142

stikkireddy:
awesome that is fine @EliiseS , I added a comment on that issue by the way.

databricks/resource_databricks_azure_blob_mount_test.go
```diff
@@ -127,38 +127,19 @@ func testAccAzureBlobMountMountExists(n string, azureBlobMount *AzureBlobMount,
...
-		tenant_id              = "%[3]s"
-		subscription_id        = "%[4]s"
-
-		workspace_name         = "%[5]s"
-		resource_group         = "%[6]s"
-		managed_resource_group = "%[7]s"
-		azure_region           = "%[8]s"
-	  }
-	}
-
```
stikkireddy, line 155: Hey @EliiseS i am concerned that removing this block will not allow the environment variables to propagate to the databricks provider azure_auth block for SP based auth. Since this is not defined by default GetOk("azure_auth") will be skipped please let me know if I am missing something.

EliiseS: ```make`
terraform-acc-azure: lint
	@echo "==> Running Terraform Acceptance Tests for Azure..."
	@/bin/bash integration-environment-azure/run.sh
```

The `run.sh` script called in the make file sets up the environment variables

Comment (lawrencegripper) [reply databricks/resource_databricks_azure_blob_mount_test.go:155]: Yeah we unified the approach as the other Azure integration tests relied on the `init` block in `provider_test.go` to populated the provider. We update the `run.sh` for Azure integration tests to output the right Environment vars and now the mount tests can use the same path and omit the `provider` block in their tests.

image

Comment (stikkireddy) [reply databricks/resource_databricks_azure_blob_mount_test.go:155]: The weird part about this is if you are using maps the environment variables do not populate correctly unless azure_auth section exists. I am confused as to how this integration test is working. Only token and host env variables populate correctly.

Comment (lawrencegripper) [reply databricks/resource_databricks_azure_blob_mount_test.go:155]: In the [`init` section of `provider_test.go` here](https://github.com/databrickslabs/terraform-provider-databricks/blob/c343db6a0c03b98d0f038b8a6c628d3ce2962db9/databricks/provider_test.go#L31) the `providerConfigureAzureClient` is called programmatically and it [creates a new map here](https://github.com/databrickslabs/terraform-provider-databricks/blob/c343db6a0c03b98d0f038b8a6c628d3ce2962db9/databricks/provider.go#L173) then assigns to if from the environment variables. (as you say without this during a normal tf init the defaults wouldn't be applied). 

Make sense? Took me a while to work out but it's much neater than having the provider initialized in each test.

Comment (EliiseS) [reply databricks/resource_databricks_azure_blob_mount_test.go:155]: We've figured out that the providerConfigureAzureClient sets the azure_auth

**Comment by stikkireddy** [comment]
Hey @EliiseS one last comment I would like to add:

1. When we add workspace_url we break all existing provider configurations.
2. We are just mapping workspace url to host, I am wondering is it really necessary to configure workspace_url?
Two things that we can do to alleviate #2:
1. Repurpose HOST for the workspace url as introducing it as a new field just for workspace URL then only to assign it back to host.
2. The following functions api call response actually contains the workspace url and the user does not need to provide it explicitly:
```go
func (a *AzureAuth) getWorkspaceID(config *service.DBApiClientConfig) error {
	log.Println("[DEBUG] Getting Workspace ID via management token.")
	// Escape all the ids
	url := fmt.Sprintf("https://management.azure.com/subscriptions/%s/resourceGroups/%s"+
		"/providers/Microsoft.Databricks/workspaces/%s",
		urlParse.PathEscape(a.TokenPayload.SubscriptionID),
		urlParse.PathEscape(a.TokenPayload.ResourceGroup),
		urlParse.PathEscape(a.TokenPayload.WorkspaceName))
	headers := map[string]string{
		"Content-Type":  "application/json",
		"cache-control": "no-cache",
		"Authorization": "Bearer " + a.ManagementToken,
	}
	type apiVersion struct {
		APIVersion string `url:"api-version"`
	}
	uriPayload := apiVersion{
		APIVersion: "2018-04-01",
	}
	var responseMap map[string]interface{}
	resp, err := service.PerformQuery(config, http.MethodGet, url, "2.0", headers, false, true, uriPayload, nil)
	if err != nil {
		return err
	}
	err = json.Unmarshal(resp, &responseMap)
	if err != nil {
		return err
	}
	a.AdbWorkspaceResourceID = responseMap["id"].(string)
	return err
}
```

databricks/azure_auth_test.go
```diff
@@ -58,3 +54,10 @@ func TestAzureAuthCreateApiToken(t *testing.T) {
 
 	assert.NoError(t, instancePoolErr, instancePoolErr)
 }
+
+// getAndAssertEnv fetches the env for testing and also asserts that the env value is not Zero i.e ""
+func getAndAssertEnv(t *testing.T, key string) string {
+	value, present := os.LookupEnv("a")
+	assert.False(t, present, fmt.Sprintf("Env variable %s is not set", key))
+	return value
+}
```
stikkireddy, line 63: I believe this should be rather than what is currently there: 
```go
// getAndAssertEnv fetches the env for testing and also asserts that the env value is not Zero i.e ""
func getAndAssertEnv(t *testing.T, key string) string {
	value, present := os.LookupEnv(key)
	assert.True(t, present, fmt.Sprintf("Env variable %s is not set", key))
	return value
}
```

EliiseS: Done

Makefile
```diff
@@ -48,10 +48,13 @@ vendor:
 	@echo "==> Filling vendor folder with library code..."
 	@go mod vendor
 
+local-install: build
```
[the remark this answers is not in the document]
EliiseS: This pr has been merged. I can make these changes in another PR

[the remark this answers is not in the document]
stikkireddy: already fixed in another pr #129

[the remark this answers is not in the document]
EliiseS: Other providers have a script to create a binary from the local files as well. It's very useful for testing and trying out your changes. Why is the hard-coded version number a problem enough that you'd remove the addition? How would you like to set a version number instead?

Issue #117, opened 2020-06-18
[FEATURE] Automated release created in github when change on master


lawrencegripper:
**Is your feature request related to a problem? Please describe.**

Would be great to get an automated process setup which creates a release when a change is put onto `master`

**Describe the solution you'd like**

In the past I've used [goreleaser]() to build out a nice release process for [azbrowse](https://github.com/lawrencegripper/azbrowse/blob/master/.goreleaser.yml). I'd be happy to do a PR with this to integrate into the [travis build](https://goreleaser.com/ci/travis/)

**Describe alternatives you've considered**

You can build out this process yourself but it's kind of a pain to do and go releaser makes like easy.

stikkireddy:
Hey @lawrencegripper I understand the want to create a release per merge to master but since we are not using release branches at this point in time, till we get to a stable version of this provider. It will create a release & tag per pr. It may become overwhelming in terms of deciding to use and update the provider (especially since you would need to refresh every time). I was hoping we can follow tag convention that this provider uses https://github.com/terraform-providers/terraform-provider-azurerm/releases. 

I am totally for this enhancement in regards to automating the release process but I was wondering if there would be a way to do something like the azurerm and some of the other providers in terms of semver and each release being a collection of changes/fixes rather than a release tag per PR. Thats just my $0.02.

I will let other folks chime in on this as well @nfx @stuartleeks

stuartleeks:
If the small incremental releases typically just result in a change to the patch number (e.g. `0.2.1`, `0.2.2`, or `0.2.<buildnum>`) then I'm onboard with this. Users can easily pick a version (`=0.2.0`) if they want to, or just opt for latest path for a specific minor version (`~>0.2`) depending on their preferences for consuming updates. But at least by making releases available you give users a way to get fixes and new features quickly if they want them.

lawrencegripper:
As discussed in the linked PR the proposal is to move to a 7 day automated release cycle (if changes are present) potentially using GitHub Actions.

nfx:
@lawrencegripper 7-day automated release cycle using github actions is an interesting idea.

having release per PR is something we cannot yet afford maintaining.

PR #118, opened 2020-06-18
Travis publish goreleaser on master

lawrencegripper:
#117 This PR look at adding a `deploy` stage to the `travis` pipeline which will run GoReleaser on any changes to `master` and publish a new github release. 

I think this will work but it will need someone with Travis permissions to test out.

This will need the `GITHUB_TOKEN` environment set with a token that has access to the repo. This can be added here: https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings and set to only be available when on the `master` branch to stop anyone abusing this and pushing a release from a PR build. 

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,13 @@ after_success:
   - echo "travis go version='$TRAVIS_GO_VERSION'"
   - bash <(curl -s https://codecov.io/bash) -f client-coverage.txt -F client
   - bash <(curl -s https://codecov.io/bash) -f provider-coverage.txt -F provider
+  # Add a tag for this release
+  - git tag "v0.2.$TRAVIS_BUILD_NUMBER"
+
+deploy:
+- provider: script
+  skip_cleanup: true
+  script: curl -sL https://git.io/goreleaser | bash
+  on:
+    tags: true
+    branch: master


stikkireddy:
@lawrencegripper we would also need to push the tag to remote for it to work, currently the workflow has been git tag (some semver tag), git push origin <tag>. I think we should think about whether we should have a new tag per merge or a meaningful tag that has a collection of commits. There is currently a github action that will run goreleaser when a tag is pushed in. The only reason we are not using github actions for everything is, that it will not run unit tests on PRs. Please take a quick look at that.

lawrencegripper:
So currently the PR adds this call `git tag "v0.2.$TRAVIS_BUILD_NUMBER"` which will create a git tag. The act of publishing the release should push that tag up to the repo. So you'll end up with release for each PR merged into master with the version number auto-incrementing. I *think* it's been a while since I used travis. 

This is my personal preference over grouping together changes into a larger bulk release. 

If you want it to only push a release on a tag then removing `- git tag "v0.2.$TRAVIS_BUILD_NUMBER"` will mean travis will do this as the `condition: tags: true` means the deploy step only runs on tag. Would need testing but think it should work. 

Make any sense?

stikkireddy:
So the github actions does the same activity at the moment for applying this on push. I think thats where the miss communication happened. The only reason I am leaning towards github actions is the support for GPG key signing. We are trying to get this pushed up into the terraform provider beta registry. https://goreleaser.com/ci/actions/#signing This is relatively easy using github actions. I only chose to use travis due to it supporting builds for PRs. 

An additional question I had regarding this is:
1. Would it be worth managing (or generating) a high level change log between minor versions then since each PR will be creating a release tag and the change log would only contain very small set of commits and it would be hard to figure out what actually changed between minor versions?

stikkireddy:
@lawrencegripper Hmm after revisiting this, I think the end goal is to try to provide access to this repo to end users via https://registry.terraform.io/providers/databrickslabs/databricks/latest. I would like to follow some of the release patterns/cadence of some of the official providers including azurerm.

They all publish a release (major, minor or patch) once every 7 days. That seems like a reasonable release cycle. If we are okay with that we can go with a release cadence of pushing an update every 7 days. I am still trying to figure out a good way of managing a change log for the releases. Please let me know if your thoughts on this Lawrence. My major concern is the users ability versions and documentation on the docs and just using this ui will be a hassle if we publish a release every squash merge `https://registry.terraform.io/providers/databrickslabs/databricks/latest`. 

I think if we can align this project to an expected release cadence we can mitigate the fact that builds may be missing. Please let me know your thoughts on this.

lawrencegripper:
Sure that makes sense to me. With GitHub actions you can set a schedule for runs so would be easy to write one up to run every 7 days and do a release.

This is an example scheduled action we setup for another project to run a nightly build https://github.com/microsoft/k8s-cronjob-prescaler/blob/master/.github/workflows/weekly.yml

I’ll close out this PR as like the sound of this approach. Let me know if you want any help getting the GitHub action configured as I’ve played a bit with them.

PR #129, opened 2020-06-19
Add more readability for errors

nfx:
* Add resource wait retry for workspace create
* Made vscode integration testing simpler
* added read support for username/password for config files
* Made errors concise and explainable


docs/index.md
```diff
@@ -274,3 +205,35 @@ resides in. Alternatively you can provide this value as an environment variable
...
+``` hcl 
+provider "databricks" {}
+```
+
+Then run the following code and the following environment variables will be injected into the provider.    
+
+``` bash
+$ export HOST="http://databricks.domain.com"
+$ export TOKEN="dapitokenhere"
+$ terraform plan
```
stikkireddy, line 238: Changing these to DATABRICKS_HOST, DATABRICKS_TOKEN

nfx: @stikkireddy good catch!

docs/index.md
```diff
@@ -6,59 +6,73 @@ description: |-
   Terraform provider databricks.
 ---
 
-# Databricks Provider
```
stikkireddy, line 9: @nfx Headers should align to the common naming convention provided by Hashicorp which are the conventions that the other official providers follow. Guide can be found here: https://www.terraform.io/docs/registry/providers/docs.html#headers. I am reverting it to follow that guide and examples can be found here: 
Old Site:
1. https://www.terraform.io/docs/providers/aws/index.html
2. https://www.terraform.io/docs/providers/azurerm/index.html

New Registry Docs:
1. https://registry.terraform.io/providers/hashicorp/aws/latest/docs#authentication
2. https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs

This gives the user similar experience when using the docs. 

We should try to use the main headers as it is used to determine the nav on the registry on the right hand side of the page. All H2 or "##" show up on that nav bar so too many makes it confusing. I am going to follow the pattern on the AWS provider as it is very simple nav and you can generate links to specific areas if we need to send links for the H3 blocks.

PR #132, opened 2020-06-22
Add support for general permissions

nfx:


databricks/resource_databricks_permissions.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Cannot self-identify")
+	}
+	accessControl := acl.AccessControl(me.UserName)
+	err = d.Set("access_control", accessControl)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
stikkireddy, line 100: This should set the acl Id to "" if the read returns nothing, or if the parent resource is deleted i.e cluster permissions, if the cluster is deleted the permissions should be treated as deleted so state issues will not occur.

databricks/resource_databricks_permissions_test.go
```diff
@@ -0,0 +1,142 @@
...
+					func(client *service.DBApiClient, id string) error {
+						resp, err := client.Permissions().Read(id)
+						if err != nil {
+							return err
+						}
+						permissions = *resp
+						assert.Len(t, permissions.AccessControlList, 3)
+						return nil
+					}),
+			},
```
stikkireddy, line 51: We should have a test of the permissions being deleted manually, or the parent resource deleted manually and the plan should not be empty and not throw an error.

databricks/resource_databricks_permissions.go
```diff
@@ -0,0 +1,209 @@
...
+			ForceNew:      true,
+			Type:          schema.TypeString,
+			Optional:      true,
+			ConflictsWith: conflictingFields(mapping.field),
+		}
+	}
+	return &schema.Resource{
+		Create: resourcePermissionsCreate,
+		Read:   resourcePermissionsRead,
+		Delete: resourcePermissionsDelete,
```
stikkireddy, line 206: Is there a specific reason for not having an update? It seems as though the apis support patches. Is it for simplicity's sake? It seems that there is both a PUT (create/delete) and a PATCH (update) operations available.

databricks/resource_databricks_permissions.go
nfx: done

databricks/resource_databricks_permissions_test.go
nfx: you mean the isMissing in read resource?

databricks/resource_databricks_permissions.go
nfx: per se, permissions are not deletable... there'll always be inherited permission for the admin creating the resource

Makefile
```diff
@@ -44,13 +24,35 @@ fmt:
 	@gofmt -s -w main.go
 	@go fmt ./...
 
+lint: fmt
```
stikkireddy, line 27: linting should not be running with fmt, during the build it will fmt before running the lint so builds may pass but the code may not be linted

nfx: good catch

nfx: done

PR #147, opened 2020-07-02
Clusters, libraries & jobs revamp

nfx:
* started removing duplicate code
* added reflection-based exchange between resource data and struct
* added unit tests for cluster resources
* added unit tests for libraries
* started with acceptance tests for clusters


stikkireddy:
Done as part of #158

Issue #166, opened 2020-07-09
[ISSUE] Library type assertion causing panic


jjcollinge:
### Terraform Version
Terraform v0.12.26

### Affected Resource(s)
Please list the resources as a list, for example:
- databricks_job

### Panic Output
https://gist.github.com/jjcollinge/cd475da8e517446e02f702290b89c534

### Expected Behavior
The provider doesn't crash when trying to parse the library response from the Databricks API.

### Actual Behavior
The provider crashes doing a type assertion of the WHL library type.
`panic: interface conversion: interface {} is map[string]interface {}, not string`
offending line: https://github.com/databrickslabs/terraform-provider-databricks/blob/d16d18baf933336e1fea4e669267b0d97780414b/databricks/resource_databricks_job.go#L1133

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create TF config that uploads custom WHL libraries to a Databricks cluster
2. run `tf apply` - this should succeed as it is a create not update.
3. Re-run `tf apply` to cause an `update` and it will panic.

stikkireddy:
Hey @jjcollinge can you post the terraform configuration that you are using? The jobs api should not be making api calls to get cluster status as it does not need to know if the cluster is running. So I am a bit confused about this.

jjcollinge:
@stikkireddy this may be an issue on the config side then - this is the config that is the module causing the issue: https://github.com/Azure/terraform-databricks-sec-resources/blob/master/main.tf

stikkireddy:
@jjcollinge there does not seem to be any jobs in that main.tf so is it still accurate to say this:

# Affected Resource(s)
Please list the resources as a list, for example:

databricks_job

jjcollinge:
> 
> 
> @jjcollinge there does not seem to be any jobs in that main.tf so is it still accurate to say this:
> ### Affected Resource(s)
> 
> Please list the resources as a list, for example:
> 
> databricks_job

This may be a mistake on my side.  I saw the stacktrace originated from `resource_databricks_job.go` so assumed it was the job resource

stikkireddy:
Sorry this was the confusion as well it was the databricks_cluster resource that was impacted by some code used in the resource_databricks_job. The change will need to be verified as that function is shared between clusters & jobs.

PR #167, opened 2020-07-09
WIP: Fix WHL type assertion issue (#166)

jjcollinge:
This fix parses the WHL library response from the Databricks API correctly to avoid the issue #166.

This is WIP whilst I do some further testing to verify this is working as expected.


stikkireddy:
This is too old and may not be relevant anymore i believe. The logic is redone using reflection. Please raise the request again @jjcollinge if you still have this bug.

Issue #189, opened 2020-07-21
[FEATURE] Support for IP ACL Lists


jarobey:
**IP ACL Lists** is now in public preview:[IP ACL Lists API]( https://docs.databricks.com/dev-tools/api/latest/ip-access-list.html).

**The implementation might look like:**
```
resource "databricks_workspace_conf" "features" {
  enable_ip_access_lists = "true"
}

resource "databricks_ip_access_list" "keep_out" {
  label = "keep_out"
  list_type = "BLACKLIST"
  ip_addresses = [
    "10.0.10.25"
  ]
}

resource "databricks_ip_access_list" "let_in_1" {
  label = "let_in_1"
  list_type = "WHITELIST"
  ip_addresses = [
    "10.0.100.0/24"
  ]
}

resource "databricks_ip_access_list" "let_in_2" {
  label = "let_in_2"
  list_type = "WHITELIST"
  ip_addresses = [
    "10.0.200.0/24",
    "10.0.205.0/24"
  ]
}
```
**Terraform model schema:** 
```
func resourceIPAccessList() *schema.Resource {
	return &schema.Resource{
		Create: resourceIPACLCreate,
		Read:   resourceIPACLRead,
		Update: resourceIPACLUpdate,
		Delete: resourceIPACLDelete,

		Schema: map[string]*schema.Schema{
			"label": {
				Type:     schema.TypeString,
				Required: true,
			},
			"list_type": {
				Type:     schema.TypeString,
				Required: true,
				ValidateFunc: validation.StringInSlice(
					[]string{
						string(model.WhiteList),
						string(model.BlackList),
					}, false),
			},
			"ip_addresses": {
				Type:     schema.TypeList,
				Required: true,
				Elem: &schema.Schema{
					Type:         schema.TypeString,
					ValidateFunc: validation.Any(validation.IsIPv4Address, validation.IsCIDR),
				},
			},
			"enabled": {
				Type:     schema.TypeBool,
				Optional: true,
				Default:  true,
			},
		},
	}
}
```

nfx:
Code is in #257

PR #158, opened 2020-07-07
DatabricksClient hardening

nfx:
This pull request is still in progress and is meant to harden the stability of databricks api client through isolation mechanics and extended unit tests.

PR #196, opened 2020-07-23
Feature add ipacl

jarobey:
Implementation of Issue #189 for adding IP ACL List support.


databricks/resource_databricks_workspace_conf.go
```diff
@@ -0,0 +1,198 @@
...
+
+func resourceWorkspaceConf() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceWorkspaceConfCreateOrUpdate,
+		Read:   resourceWorkspaceConfRead,
+		Update: resourceWorkspaceConfCreateOrUpdate,
+		Delete: resourceWorkspaceConfDelete,
+
+		Schema: map[string]*schema.Schema{
+			"enable_ip_access_lists": {
```
nfx, line 22: please rewrite to `schema.TypeMap` unless you have explicit OK from all PMs to make these properties public :) with this approach we can properly combine public & private configurations (which may change it's format)

databricks/resource_databricks_workspace_conf_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Preview feature: https://docs.databricks.com/security/network/ip-access-list.html
+// REST API: https://docs.databricks.com/dev-tools/api/latest/ip-access-list.html#operation/create-list
+
+import (
+	"testing"
+)
+
+func TestWorkspaceConfCreateOrUpdate(t *testing.T) {
+	// d, err := ResourceTester(t, []HTTPFixture{
```
nfx, line 11: no large commented chunks of code in this repository, please :)

examples/enterprise-azure-deployment/enterprise-workspace-deploy/apply.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+terraform apply -var-file='secret.tfvars' -var-file='dbws.tfvars'
```
nfx, line 2: most likely this file will not be needed after #158 is merged. for this repository, we aim to use environment variables, so that integration tests can run. usage of `*.tfvars` in production workloads might be less preferable than remote terraform state, though vars file is completely okay for local PoC's.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/main.tf
```diff
@@ -0,0 +1,91 @@
+provider "azurerm" {
```
nfx, line 1: please let's discuss internally how we do manage examples :)

my_lame_ent_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
nfx, line 1: please remove this file

client/model/workspace_configration.go
```diff
@@ -0,0 +1,145 @@
...
+type sidebarLogoInactive string
+type homePageLogo string
+type homePageLogoWidth string
+type productName string
+type loginLogo string
+type loginLogoWidth string
+type customReferences string
+type enableIpAccessLists string
+
+func (c sidebarLogoActive) apply(conf *WorkspaceConfRequestMapBuilder) {
```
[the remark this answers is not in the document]
stikkireddy: Just be aware that this api requests will not work well with structs that use json tags & omit empty. It requires you to use maps. Removing the builder pattern is fine but, but the request should be able to send any combination of "keys" with option to send a value of "" which go treats as empty. So structs & struct tags will not play nicely here.

databricks/resource_databricks_workspace_conf.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return resourceWorkspaceConfRead(d, m)
+}
+
+func resourceWorkspaceConfRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*service.DBApiClient)
+	resp, err := client.WorkspaceConfigurations().Read(model.AllWorkspaceConfKeys)
+	if err != nil {
+		return err
+	}
```
stikkireddy, line 115: Can we verify that this api does not have 404s and add that as a comment. This can identify that there is no configuration drift.

databricks/resource_databricks_ipaccesslist.go
```diff
@@ -0,0 +1,135 @@
...
+		return err
+	}
+
+	log.Println("IPACLLists:  Updating interface")
+	err = updateFromStatus(d, status)
+	log.Printf("IPACLLists:  Updated interface %v\n", d)
+	if err != nil {
+		log.Printf("IPACLLists:  Error updating interface %v\n", err)
+		return err
+	}
```
stikkireddy, line 76: Unless the fields are static and not generated by the "read" api , we should not set anything but the id in the terraform create: example here- https://github.com/terraform-providers/terraform-provider-aws/blob/34482925e20774621ab80fa0b03054514795460e/aws/resource_aws_appmesh_route.go#L293-L313.
Terraform actually uses the set to identify differences. So it should be done after a read action and not a create action. This can identify inconsistencies in what was created and what the hcl schema expects.

client/model/workspace_configration_test.go
```diff
@@ -0,0 +1,32 @@
...
+	opts := []WorkspaceConfRequestOption{
+		WithCustomReferences(""),
+	}
+	wc := WorkspaceConfRequestMapBuilder{}
+
+	wc.Build(opts...)
+
+	t.Log(wc.configurationReqMap)
+	json, _ := json.Marshal(wc.configurationReqMap)
+	t.Log(string(json))
```
stikkireddy, line 31: maybe an assertion here but this test is not doing anything.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/apply.sh
stikkireddy: I believe this is for non databricks provider variables. This also seems to be an example. But again terraform supports arbitrary variables to be declared by environment variables using TF_VAR_* prefix. Depending on your ci/cd tool, some many not support encrypted secrets files, you can use environment variables. Find more information here: https://www.terraform.io/docs/configuration/variables.html#assigning-values-to-root-module-variables

databricks/resource_databricks_ipaccesslist.go
```diff
@@ -0,0 +1,135 @@
+package databricks
+
+// Preview feature: https://docs.databricks.com/security/network/ip-access-list.html
+// REST API: https://docs.databricks.com/dev-tools/api/latest/ip-access-list.html#operation/create-list
```
stikkireddy, line 4: This needs to be added to the doc somewhere: 
```For all allow lists and block lists combined, the workspace supports a maximum of 1000 IP/CIDR values, where one CIDR counts as a single value.```

We do not capture this as a validation step so should be documented somewhere or link to the original documentation pointing towards this direction.

File: examples/enterprise-azure-deployment/enterprise-workspace-deploy/main.tf [hunk 11]
```diff
@@ -0,0 +1,91 @@
...
+  sku                           = "premium"
+  custom_parameters {
+    no_public_ip = "true"
+    public_subnet_name = var.subnet_public
+    private_subnet_name = var.subnet_private
+    virtual_network_id = var.vnet_id
+  }
+  tags = {
+    Owner = "jason.robey@databricks.com"
+  }
```
stikkireddy, line 23: you may want to move this out of the example

databricks/resource_databricks_ipaccesslist.go
nfx: @jarobey @stikkireddy i think we can validate this by listing all ip access lists? :)

client/model/workspace_configration.go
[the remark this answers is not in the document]
nfx: yeah... we need to make two requests before sending the update

databricks/resource_databricks_ipaccesslist.go
stikkireddy: Yep ideally we can do this as a validation step, during terraform validate but there is no way to capture that as the whitelists are defined separately. There is the option of doing it at runtime, but I am not sure I see the value in doing that before every create/update call as that puts stress on the api and would be no different user experience than the UI yelling back at that given instance and every subsequent instance.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/main.tf
jarobey: doh!

client/model/workspace_configration.go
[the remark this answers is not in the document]
jarobey: I updated to remove the builder.  Interestingly enough a "" is bad for enableIpAccessLists.  To clear or delete, you have to set false.  I think we'll have to have some logic per property.  I have currently stripped this down to just enableIpAccessLists.  Happy to add logic for white-label in the future.

client/service/ip_access_list.go
```diff
@@ -0,0 +1,92 @@
...
+	wrapper := model.IPAccessListStatusWrapper{}
+	if err != nil {
+		return wrapper.IPAccessList, err
+	}
+
+	err = json.Unmarshal(resp, &wrapper)
+	return wrapper.IPAccessList, err
+}
+
+func (a IPAccessListsAPI) Update(objectID string, label string, listType model.IPAccessListType, ipAddresses []string, enabled bool) (model.IPAccessListStatus, error) {
```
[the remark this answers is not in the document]
jarobey: got it, will update

databricks/resource_databricks_workspace_conf.go
jarobey: Based on the API, this will always be a map.  I might need a little hint, here.  Delete logic for each key is going to be dependent on the key.  enableIpAccessLists is in the public docs, so I was thinking this was ok.

databricks/resource_databricks_workspace_conf_test.go
jarobey: heh, np, I was explicitly asking for help, there.  It was difficult to mix the ResourceTester with the Option/Builder pattern.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/apply.sh
jarobey: I also have a number of enterprises that are not comfortable setting things as environment variables.  This is a convenience wrapper for me.  I think it makes sense to have the example in there, but happy to remove the example or just the wrapper shells.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/main.tf
jarobey: Sounds good.  Lol, yeah, this example is specific to setting up on azure.  Let me drop that and think about a better enterprise example.  This is more of an end-to-end anyway which would have to call the specific cloud provider to setup the workspace in the first place.

my_lame_ent_test.sh
jarobey: done

databricks/resource_databricks_workspace_conf.go
jarobey: Adding comment, only 200 or 400 for Get and 200, 400, 401, and 500 on patch

databricks/resource_databricks_ipaccesslist.go
jarobey: Interresting, ok.  Removed that section.

databricks/resource_databricks_ipaccesslist.go
jarobey: I wasn't sure how to make this constraint across lists in terraform.  After a quick consult, I left it for documentation (which I'll be adding).  I think we'll get an error on the api so it will fail to apply, but might be good to check.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/apply.sh
stikkireddy: Ya i think its worth having an example with a vars file as it is quite common to use encrypted secret files rather than OS environment variables. But a readme for the example should probably be there to explain the commands at a high level. Does not too detailed.

client/service/workspace_configrations.go
```diff
@@ -0,0 +1,34 @@
...
+// Update will handle creation of new values as well as deletes. Deleting just implies that a value of "" is sent with
+// the appropriate key
+func (a WorkspaceConfAPI) Update(workspaceConfMap map[string]string) error {
+	_, err := a.Client.performQuery(http.MethodPatch, "/preview/workspace-conf", "2.0", nil, workspaceConfMap, nil)
+	return err
+}
+
+// Read just returns back a map of keys and values which keys are the configuration items and values are the settings
+func (a WorkspaceConfAPI) Read(keys string) (map[string]string, error) {
+	var wsConfResp map[string]string
```
stikkireddy, line 22: This may need a todo just to verify that the maps are map[string]string and not map[string]interface{} as json supports string -> string, string -> float64, string -> bool and string -> struct... I think it is string -> string but worth verifying and documenting that.

examples/enterprise-azure-deployment/enterprise-workspace-deploy/apply.sh
jarobey: Cool, I can add a readme for sure.

client/service/workspace_configrations.go
jarobey: I tested that.  If you pass anything but a string in for any of the keys you get: 
{
    "error_code": "BAD_REQUEST",
    "message": "Values must be strings"
}

jarobey: Adding some comments with a TODO marked.  The API returns an error for any value passed in that isn't of type string in json.

databricks/resource_databricks_workspace_conf.go
stikkireddy: I think what he might be saying is that the field is something like "configuration" where the type is TypeMap of all string values. The user can the pick and choose what workspace conf items get added. This obscures all the "keys" available but I am not sure if I totally agree with that approach as we cant take advantage of schema constructs such as validation, etc.

client/model/workspace_configration_test.go
jarobey: Updated to move some troubleshooting code out.

databricks/resource_databricks_workspace_conf_test.go
jarobey: Rewrote

databricks/resource_databricks_workspace_conf.go
jarobey: Right, what do you think @nfx ?  We only add features here that show up in the public docs.  Each option has some special handling around how you delete and such.  I think we'll need to keep updating this as the platform keeps using.  i put special code in this resource based on the specific option of "enable_ip_access_lists" in this case.

client/service/ip_access_list.go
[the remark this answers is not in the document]
jarobey: This might take a bit of work.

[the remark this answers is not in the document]
jarobey: done

nfx:
@jarobey can you merge in latest master onto this branch?

client/model/ip_access_list.go
```diff
@@ -0,0 +1,46 @@
+package model
```
nfx, line 1: in preparation of refactoring in #224, please move this to `workspace` package. `access` package might be good as well. (because it's `access` for `workspace`...)

client/model/ip_access_list.go
```diff
@@ -0,0 +1,46 @@
...
+type CreateIPAccessListRequest struct {
+	Label       string           `json:"label,omitempty"`
+	ListType    IPAccessListType `json:"list_type,omitempty"`
+	IPAddresses []string         `json:"ip_addresses,omitempty"`
+}
+
+// List: Get all IP access lists
+// Norequest needed--just a get
+type ListIPAccessListsResponse struct {
+	ListIPAccessListsResponse []IPAccessListStatus `json:"ip_access_lists,omitempty"`
```
nfx, line 37: please rename field to `IPAccessLists` here, because it's reflecting API response more. Besides, `HTTPFixture` framework auto-suggests stubs with some naming heuristics

client/model/ip_access_list.go
```diff
@@ -0,0 +1,46 @@
...
+	AddressCount  int              `json:"address_count,omitempty"`
+	CreatedAt     int64            `json:"created_at,omitempty"`
+	CreatorUserID int64            `json:"creator_user_id,omitempty"`
+	UpdatedAt     int64            `json:"updated_at,omitempty"`
+	UpdatorUserID int64            `json:"updator_user_id,omitempty"`
+	Enabled       bool             `json:"enabled,omitempty"`
+}
+
+type IPAccessListStatusWrapper struct {
+	IPAccessList IPAccessListStatus `json:"ip_access_list,omitempty"`
```
nfx, line 24: rename to `IPAccessList`

client/service/ip_access_list.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/databrickslabs/databricks-terraform/client/model"
+)
+
+// IPAccessListsAPI exposes the instance pools api
+type IPAccessListsAPI struct {
+	Client *DBApiClient
+}
+
+// Create creates the IP Access List to given the instance pool configuration
+func (a IPAccessListsAPI) Create(ipAddresses []string, label string, listType model.IPAccessListType) (status model.IPAccessListStatus, err error) {
```
nfx, line 16: move this file to `access` (or `workspace`) package

client/service/workspace_configrations.go
```diff
@@ -0,0 +1,39 @@
+package service
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// WorkspaceConfAPI exposes the workspace configurations API
+type WorkspaceConfAPI struct {
```
nfx, line 9: Move this file to `workspace` package, because it's making configuration of workspace

databricks/resource_databricks_workspace_conf.go
nfx: well... in order not to maintain "public" & "private" forks, we need to make these flexible. so, instead of `enable_ip_access_list` we should have `features = ["ip_access_lists"]` for "workspace feature flags"

docs/resources/ip_acl_lists.md
```diff
@@ -0,0 +1,40 @@
...
+
+```hcl
+resource "databricks_ip_access_list" "allowed-list" {
+  label = "allow_in"
+  list_type = "WHITELIST"
+  ip_addresses = [
+    "1.2.3.0/24",
+    "1.2.5.0/24"
+  ]
+  depends_on = [<Replace with terraform workspace id syntax>]
```
nfx, line 17: why do we need explicit `depends_on`? resource internally must retry until necessary state is reached (e.g. workspace being in a runnable state). this way we're having more smooth user journey, compared to creating azure databricks workspace with vnet injection :)

docs/resources/ip_acl_lists.md
```diff
@@ -0,0 +1,40 @@
...
+
+* `ip_addresses` -  **(Required)** This is a field to allow the group to have instance pool create priviliges.
+
+* `enabled` - **(Optional)** Boolean `true` or `false` indicating whether this list should be active.  Defaults to `true`
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `list_id` - Canonical unique identifier for the IP Access List.
```
nfx, line 36: why do we need to expose `list_id` if `id` is exposed by default?

examples/enterprise-azure-deployment/enterprise-workspace-deploy/main.tf
```diff
@@ -0,0 +1,90 @@
...
+  enable_ip_access_lists = "true"
+}
+
+resource "databricks_ip_access_list" "naughty_list" {
+  label = "lumps_of_coal"
+  list_type = "BLACKLIST"
+  ip_addresses = [
+    "10.0.10.25","10.0.10.0/24"
+  ]
+  depends_on = [azurerm_databricks_workspace.enterprise-demo-1]
```
nfx, line 80: this should not be needed. Implement retries with `resource.Retry`. also in this case you forgot to mention `databricks_workspace_conf.features` dependency as well. which is confusing to end-user.

jarobey:
Refactored following the refactors.  Opened a new PR.

PR #199, opened 2020-07-27
Increase API connection timeouts and support instance profile import

johnkeates:
Does what is says on the tin


johnkeates:
Tested by: me

PR #222, opened 2020-08-06
FIX node type of auto generated cluster for mounts

sdebruyn:
* Use one of the available node types from the API
* Rename method so that it is clear that Azure auth does not automatically mean the environment is located on Azure

### DIFF
--- a/client/service/apis.go
+++ b/client/service/apis.go
@@ -200,7 +200,7 @@ func (c *DatabricksClient) configureHTTPCLient() {
 
 // IsAzure returns true if Azure is configured
 func (c *DatabricksClient) IsAzure() bool {
-	return c.AzureAuth.resourceID() != ""
+	return strings.Contains(strings.ToLower(c.Host), "azure")
 }
 
 // Clusters returns an instance of ClustersAPI
--- a/client/service/apis.go
+++ b/client/service/apis.go
@@ -198,9 +198,9 @@ func (c *DatabricksClient) configureHTTPCLient() {
 	}
 }
 
-// IsAzure returns true if Azure is configured
-func (c *DatabricksClient) IsAzure() bool {
-	return strings.Contains(strings.ToLower(c.Host), "azure")
+// UsingAzureAuth returns true if Azure authentication is configured, this is not a valid check to determine if we are running on Azure
+func (c *DatabricksClient) UsingAzureAuth() bool {
+	return c.AzureAuth.resourceID() != ""
 }
 
 // Clusters returns an instance of ClustersAPI
--- a/client/service/clusters.go
+++ b/client/service/clusters.go
@@ -236,10 +236,15 @@ func (a ClustersAPI) GetOrCreateRunningCluster(name string, custom ...model.Clus
 			return cl, nil
 		}
 	}
-	instanceType := "m4.large"
-	if a.client.IsAzure() {
-		instanceType = "Standard_DS3_v2"
+
+	nodeTypes, err := a.ListNodeTypes()
+	if err != nil {
+		return
 	}
+
+	instanceType := nodeTypes[0].NodeTypeID
+	log.Printf("[INFO] Creating an autoterminating cluster with node type %s", instanceType)
+
 	r := model.Cluster{
 		NumWorkers:             1,
 		ClusterName:            name,
--- a/client/service/instance_pools_test.go
+++ b/client/service/instance_pools_test.go
@@ -249,7 +249,7 @@ func TestAccInstancePools(t *testing.T) {
 			"6.3.x-scala2.11",
 		},
 	}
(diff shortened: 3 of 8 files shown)


client/service/clusters.go
```diff
@@ -236,10 +236,15 @@ func (a ClustersAPI) GetOrCreateRunningCluster(name string, custom ...model.Clus
...
-	instanceType := "m4.large"
-	if a.client.IsAzure() {
-		instanceType = "Standard_DS3_v2"
+
+	nodeTypes, err := a.ListNodeTypes()
+	if err != nil {
+		return
 	}
+
+	instanceType := nodeTypes[0].NodeTypeID
```
nfx, line 245: one cannot pick just any node type - it must be the one without the need to attach block storage & cheap :)

client/service/apis.go
```diff
@@ -198,8 +198,8 @@ func (c *DatabricksClient) configureHTTPCLient() {
 	}
 }
 
-// IsAzure returns true if Azure is configured
-func (c *DatabricksClient) IsAzure() bool {
+// UsingAzureAuth returns true if Azure authentication is configured, this is not a valid check to determine if we are running on Azure
+func (c *DatabricksClient) UsingAzureAuth() bool {
 	return c.AzureAuth.resourceID() != ""
```
nfx, line 203: i really wonder how this changes things?..

client/service/clusters.go
nfx: my advice will be: query node types, sort all nodes, pick
```
{
      "memory_mb": <= lowest memory
      "num_cores": <= lowest cores
      "description": "m5a.4xlarge (beta)",
      "instance_type_id": "m5a.4xlarge",
      "is_deprecated": <= non-deprecated
      "category": "General Purpose",
...
      "num_gpus": 0,
      "node_instance_type": {
        "local_disks":                  <= with a local disk
        "local_disk_size_gb": 0
      },
    },
```

client/service/apis.go
sdebruyn: For what this method is currently used for (the tests), this check would work fine. Renaming it would avoid that anyone uses it for anything else. What would you propose?

client/service/clusters.go
sdebruyn: Alright, sorted by deprecation, memory and CPU. But I don't see how to sort by the other parameters you mentioned.

sdebruyn: Found it, they were just not mapped yet, on it

nfx:
@sdebruyn have you run `make test-azcli` or `make test-azsp` (~30 minutes to run)? Interested in passing tests related to cluster creation.

sdebruyn:
> @sdebruyn have you run `make test-azcli` or `make test-azsp` (~30 minutes to run)? Interested in passing tests related to cluster creation.

Not yet, I'll get right to it. However I've used this for my own project a couple of times without any issues.

sdebruyn:
How can I use those tests? Looks like I need to set some env vars first, where can I find the ones that I need to set?

sdebruyn:
Ok so I figured it out and ran the tests. A lot of them are failing, but those can't all be failing because of my changes? The unit tests succeed and all that is changed is the logic for the node type, which couldn't even work before.

nfx:
@sdebruyn would you be able to fix the tests today?

sdebruyn:
> @sdebruyn would you be able to fix the tests today?

Tomorrow probably, I have my Databricks Spark 3 cert exam later today 😁

sdebruyn:
@nfx unit test fixed, ready for re-review

common/client.go
```diff
@@ -202,7 +202,7 @@ func (c *DatabricksClient) configureHTTPCLient() {
 	}
 }
 
-// IsAzure returns true if Azure is configured
-func (c *DatabricksClient) IsAzure() bool {
+// UsingAzureAuth returns true if Azure authentication is configured, this is not a valid check to determine if we are running on Azure
```
stikkireddy, line 205: Only comment I have is the rename of this makes this not so readable. I think this should be prefixed with an Is or Has. IMO Using implies that you will be returning the pointer of the parent object in method chaining. In this case it makes it seem UsingAzureAuth returns an instance of the client.

```
client.UsingAzureAuth().do(....)
```

where as being prefixed with Is or Has (preferably Is) implies that it will return a boolean value.

```
client.IsUsingAzureAuth() or client.HasUsingAzureAuth()
```

nfx (approved):
`GetSmallestNodeType` should be package-private. it has no use outside of it. Will change that after merging.

sdebruyn:
> `GetSmallestNodeType` should be package-private. it has no use outside of it. Will change that after merging.

Can you still unit test it if it's private? I think that was the reason I made it public.

PR #233, opened 2020-08-11
Make test generator working with new template

nfx:
Have test generator working with new template. until the following resources have compliant tests, we cannot turn this on by default:

* resource_scim_user
* resource_scim_group
* resource_group_member


nfx:
+ @jarobey

PR #250, opened 2020-08-17
Migrate docs to registry-compatible structure

nfx:
* Consolidate all documentation under `docs/` folder.
* Add more extensive examples for all of documentation items.
* Remove legacy website, further removing third-party code from this repository.


nfx:
add deprecation message to each of the `gh-pages` with re-linking to tf registry & revert github workflow.

nfx:
image
added migration notice

docs/resources/job.md
```diff
@@ -0,0 +1,91 @@
...
+        num_workers   = 300
+        spark_version = "6.6.x-scala2.11
+        node_type_id  = "i3.xlarge"
+    }
+    
+    notebook_task {
+        notebook_path = "/Production/MakeFeatures"
+    }
+    
+    libraries {
```
nfx, line 24: @stikkireddy i think we'd need one more PR to change `libraries` to `library` in jobs & clusters. There's also a bit of resource inconsistency - `library` blocks logically should be within `new_cluster` block. though, JAR jobs on existing clusters may think otherwise.

stikkireddy: @nfx first statement makes sense its awkward to say libraries but only define one. We should make the change before pushing in this release then. The second statement is inaccurate as you can define jobs onto an existing cluster but the jobs can still install libraries. So it does not make sense for library construct to be built into the new_cluster.

PR #253, opened 2020-08-18
mounts patch

stikkireddy:
changes made
1. fixed contexts/status to get request
2. refactored EnvironmentTemplate and added unit tests for it
3. added acc test for IamMount
4. awsiammount struct config method returns empty map
5. NewMount creates an instance of the CommandsAPI if executor in the client is nil (both tests and prod path works) IMO this suffices for now. Not sure if we would ever mock some calls and let it actually "production" apis in the other half but dont really need that at the moment.

### DIFF
--- a/compute/commands.go
+++ b/compute/commands.go
@@ -144,7 +144,7 @@ func (a CommandsAPI) deleteContext(contextID, clusterID string) error {
 
 func (a CommandsAPI) getContext(contextID, clusterID string) (string, error) {
 	var contextStatus Command // internal hack, yes
-	err := a.client.OldAPI("POST", "/contexts/status", genericCommandRequest{
+	err := a.client.OldAPI("GET", "/contexts/status", genericCommandRequest{
 		ContextID: contextID,
 		ClusterID: clusterID,
 	}, &contextStatus)
--- a/internal/qa/testing.go
+++ b/internal/qa/testing.go
@@ -281,10 +281,19 @@ func fixHCL(v interface{}) interface{} {
 }
 
 // For writing a unit test to intercept the errors (t.Fatalf literally ends the test in failure)
-func environmentTemplate(t *testing.T, template string) (string, error) {
+func environmentTemplate(t *testing.T, template string, otherVars ...map[string]string) (string, error) {
 	vars := map[string]string{
 		"RANDOM": acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum),
 	}
+	if len(otherVars) > 1 {
+		return "", errors.New("Cannot have more than one customer variable map!")
+	}
+	if len(otherVars) == 1 {
+		for k, v := range otherVars[0] {
+			vars[k] = v
+		}
+	}
+	// pullAll otherVars
 	missing := 0
 	var varType, varName, value string
 	r := regexp.MustCompile(`{(env|var).([^{}]*)}`)
@@ -312,8 +321,8 @@ func environmentTemplate(t *testing.T, template string) (string, error) {
 }
 
 // EnvironmentTemplate asserts existence and fills in {env.VAR} & {var.RANDOM} placeholders in template
-func EnvironmentTemplate(t *testing.T, template string) string {
-	resp, err := environmentTemplate(t, template)
+func EnvironmentTemplate(t *testing.T, template string, otherVars ...map[string]string) string {
+	resp, err := environmentTemplate(t, template, otherVars...)
 	if err != nil {
 		t.Fatal(err.Error())
 	}
--- a/internal/qa/testing.go
+++ b/internal/qa/testing.go
@@ -280,8 +280,8 @@ func fixHCL(v interface{}) interface{} {
 	}
 }
 
-// EnvironmentTemplate asserts existence and fills in {env.VAR} & {var.RANDOM} placeholders in template
-func EnvironmentTemplate(t *testing.T, template string) string {
+// For writing a unit test to intercept the errors (t.Fatalf literally ends the test in failure)
+func environmentTemplate(t *testing.T, template string) (string, error) {
 	vars := map[string]string{
 		"RANDOM": acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum),
 	}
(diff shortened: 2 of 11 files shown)


compute/commands.go
```diff
@@ -144,7 +144,7 @@ func (a CommandsAPI) deleteContext(contextID, clusterID string) error {
 
 func (a CommandsAPI) getContext(contextID, clusterID string) (string, error) {
 	var contextStatus Command // internal hack, yes
-	err := a.client.OldAPI("POST", "/contexts/status", genericCommandRequest{
+	err := a.client.OldAPI("GET", "/contexts/status", genericCommandRequest{
```
nfx, line 147: whoops😊

internal/qa/testing.go
```diff
@@ -280,8 +280,8 @@ func fixHCL(v interface{}) interface{} {
 	}
 }
 
-// EnvironmentTemplate asserts existence and fills in {env.VAR} & {var.RANDOM} placeholders in template
-func EnvironmentTemplate(t *testing.T, template string) string {
+// For writing a unit test to intercept the errors (t.Fatalf literally ends the test in failure)
+func environmentTemplate(t *testing.T, template string) (string, error) {
```
nfx, line 284: where this private method is going to be used?

storage/acceptance/aws_s3_mount_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/databrickslabs/databricks-terraform/internal/acceptance"
+	. "github.com/databrickslabs/databricks-terraform/storage"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/databrickslabs/databricks-terraform/internal/qa"
+)
+
+func getAwsS3MountConfig(t *testing.T, mountName string) string {
+	if os.Getenv("TEST_AWS_MOUNT_CLUSTER_ID") != "" {
```
nfx, line 19: 1) add in `otherVars` change from https://github.com/databrickslabs/terraform-provider-databricks/blob/6290ae42e6bd1c6e0152228208e8c9f58e55c5da/internal/qa/testing.go#L286 and remove `TEST_AWS_MOUNT_CLUSTER_ID`
2) `EnvironmentTemplate` is intended to be used unconditionally for test step configs - so that it's readable in single place what this step config does. if there is conditional - it should be two tests. so in this case it's two different tests.

storage/mounts.go
```diff
@@ -92,9 +93,14 @@ func commonMountResource(tpl Mount, s map[string]*schema.Schema) *schema.Resourc
 
 // NewMountPoint returns new mount point config
 func NewMountPoint(client *common.DatabricksClient, name, clusterID string) MountPoint {
+	// TODO: the executor is mutable probably needs to be refactored
+	executor := client.CommandExecutor()
```
nfx, line 97: why can't this change be done within DatabricksClient#CommandExecutor method? it's more logical to have it there, mounts should not know anything about existence of CommandsAPI, it should only know about CommandExecutor interface.

storage/mounts_test.go
```diff
@@ -21,3 +25,92 @@ func TestValidateMountDirectory(t *testing.T) {
 		assert.Lenf(t, errs, tc.errorCount, "directory '%s' does not generate the expected error count", tc.directory)
 	}
 }
+
+var expectedCommandResp = "done"
+
+func testMountFuncHelper(t *testing.T, mountName, expectedCommand string, mount Mount,
```
nfx, line 31: please use `ResourceFixture#CommandMock`.

storage/mounts_test.go
```diff
@@ -21,3 +25,92 @@ func TestValidateMountDirectory(t *testing.T) {
...
+		Host:  ".",
+		Token: ".",
+	}
+	err := c.Configure()
+	assert.NoError(t, err)
+
+	var called bool
+
+	c.WithCommandMock(func(commandStr string) (s string, e error) {
+		called = true
```
nfx, line 43: it may be called few different times with different commands, so keep this consideration in tests

storage/mounts_test.go
```diff
@@ -21,3 +25,92 @@ func TestValidateMountDirectory(t *testing.T) {
 		assert.Lenf(t, errs, tc.errorCount, "directory '%s' does not generate the expected error count", tc.directory)
 	}
 }
+
+var expectedCommandResp = "done"
+
+func testMountFuncHelper(t *testing.T, mountName, expectedCommand string, mount Mount,
+	mountFunc func(mp MountPoint, mount Mount) (string, error)) {
```
nfx, line 32: please move mountFunc to beginning of the arg list, after `t`, because it's the most important thing that this function does. During reviewing tests everyone should read it first, not the assertion arguments.

storage/mounts_test.go
```diff
@@ -21,3 +25,92 @@ func TestValidateMountDirectory(t *testing.T) {
 		assert.Lenf(t, errs, tc.errorCount, "directory '%s' does not generate the expected error count", tc.directory)
 	}
 }
+
+var expectedCommandResp = "done"
```
nfx, line 29: this is rather `const`, than var. please change

storage/mounts_test.go
```diff
@@ -21,3 +25,92 @@ func TestValidateMountDirectory(t *testing.T) {
...
+		raise Exception("Mount not found")
+	`, randomMountName)
+	testMountFuncHelper(t, randomMountName, expectedCommand, nil,
+		func(mp MountPoint, mount Mount) (s string, e error) {
+			return mp.Source()
+		})
+}
+
+func TestMountPoint_Delete(t *testing.T) {
+	randomMountName := qa.RandomName()
```
nfx, line 102: why do unit tests have to be having random names? static value for them will do.

internal/qa/testing.go
stikkireddy: it was refactored to write the unit test otherwise t.Fatal makes it untestable to test the unpopulated logic. It is only used in the EnvironmentTemplate so it is kept as a private method to the package. No one else should be calling it. It is only used to unit test the error logic.

storage/acceptance/aws_s3_mount_test.go
stikkireddy: sure I will make a test that requires cluster id and another one that does not.

storage/mounts_test.go
```diff
@@ -21,3 +25,92 @@ func TestValidateMountDirectory(t *testing.T) {
...
+		Token: ".",
+	}
+	err := c.Configure()
+	assert.NoError(t, err)
+
+	var called bool
+
+	c.WithCommandMock(func(commandStr string) (s string, e error) {
+		called = true
+		assert.Equal(t, expectedCommand, commandStr)
```
nfx, line 44: `assert.Equal(t, internal.TrimLeadingWhitespace(expectedCommand), commandStr)`

storage/mounts.go
stikkireddy: it creates a cyclical import:

import cycle not allowed
package github.com/databrickslabs/databricks-terraform/compute/acceptance (test)
	imports github.com/databrickslabs/databricks-terraform/common
	imports github.com/databrickslabs/databricks-terraform/compute
	imports github.com/databrickslabs/databricks-terraform/common

storage/mounts_test.go
stikkireddy: refactored tests to only test mp.Mount, Source and Delete which only invoke this once.

nfx (approved):

PR #254, opened 2020-08-18
Fix tests

nfx:


access/acceptance/permissions_test.go
```diff
@@ -16,103 +16,41 @@ import (
 func TestAccDatabricksPermissionsResourceFullLifecycle(t *testing.T) {
```
stikkireddy, line 16: what is the reason for removing these tests?

provider/generate_test.go
```diff
@@ -192,8 +192,10 @@ func (stub *resourceTestStub) Deletes(t *testing.T) {
 }
 
 func TestGenerateTestCodeStubs(t *testing.T) {
-	t.Skip()
 	funcs := getExistingUnitTests()
+	t.Logf("Got %d unit tests in total. %v",
```
nfx, line 196: // nolint

PR #256, opened 2020-08-18
FIX instance_pool max_capacity should be optional

### LINKED ISSUE #195: [ISSUE] Instance_pool resource max_capacity is optional
Max_capacity is currently required.
This should be optional as the user has the option to use an unlimited number of servers in the pool

sdebruyn:
Max capacity is now optional as in the API. Added this case to the acceptance test and validated that the test succeeds.
Fixes #195 

Sidenote:
It looks like this is not using the new reflection based approach. It might be useful to have some roadmap somewhere (on CONTRIBUTING.md or in the issues) to mark the resources that should be converted.


nfx:
Closing in favour of #260

PR #259, opened 2020-08-19
Libraries Rename

stikkireddy:
Renamed libraries to library as it is awkward and it was introduced in this release.


nfx (changes requested):
also change docs :)

nfx (approved):

PR #268, opened 2020-08-25
Provider Lazy Authentication

### LINKED ISSUE #267: [ISSUE] Provider Lazy Authentication
### Terraform Version
Terraform 0.12.29
latest master version

### Affected Resource(s)
all

The following environment variables are set (copy from my azure-pipelines.yml):
```yml
echo "##vso[task.setvariable variable=ARM_SUBSCRIPTION_ID]$(subscription-id)"
echo "##vso[task.setvariable variable=ARM_CLIENT_ID]$(client-id)"
echo "##vso[task.setvariable variable=ARM_CLIENT_SECRET]$(client-secret)"
echo "##vso[task.setvariable variable=ARM_TENANT_ID]$(tenant-id)"
echo "##vso[task.setvariable variable=AD_CLIENT_ID]$(azuread-client)"
echo "##vso[task.setvariable variable=AD_CLIENT_SECRET]$(azuread-secret)"
```

```hcl
provider "azurerm" {
  features {}
}

resource "azurerm_databricks_workspace" "dbks" {
  name                        = "dbks${local.name}"
  resource_group_name         = azurerm_resource_group.rg.name
  managed_resource_group_name = "rgdbks${local.name}"
  location                    = local.region
  sku                         = "standard"
  tags                        = local.tags
}

provider "databricks" {
  azure_workspace_resource_id = azurerm_databricks_workspace.dbks.id
}

resource "databricks_instance_pool" "pool" {
  instance_pool_name                    = "dl-pool"
  min_idle_instances                    = 0
  node_type_id                          = "Standard_DS3_v2"
  idle_instance_autotermination_minutes = 10
  enable_elastic_disk                   = true
  preloaded_spark_versions              = [var.databricks_version]
}
```

### Debug Output
```
Error: Authentication is not configured for provider. Please configure it
through one of the following options:
1. DATABRICKS_HOST + DATABRICKS_TOKEN environment variables.
2. host + token provider arguments.
3. azure_databricks_workspace_id + AZ CLI authentication.
4. azure_databricks_workspace_id + azure_client_id + azure_client_secret + azure_tenant_id for Azure Service Principal authentication.
5. Run `databricks configure --token` that will create ~/.databrickscfg file.

Please check https://github.com/databrickslabs/terraform-provider-databricks/blob/master/docs/index.md#authentication for details

  in databricks.tf, in provider "databricks":
  provider "databricks" {

```

### Expected Behavior
Succesful authentication

### Actual Behavior
Error above

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
I am using a child module that has a different provider configuration set, maybe they are conflicting?

sdebruyn:
Fixes #267 

This moves the validation of the authentication completely to the CRUD operations. This way a plan can succeed where Terraform does not know the values of some of the provider configuration arguments yet.

The changes themselves are not complex. Fixing the tests was another story... Feel free to just take the PR and change it if that can help to get 0.2.4 out of the door faster. I can usually make some time in the evening (UTC+2).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -26,14 +26,6 @@
                "TF_ACC": "1",
            },
            "showLog": true
-       },
-       {
-
-            "name": "Attach to Process",
-            "type": "go",
-            "request": "attach",
-            "mode": "local",
-            "processId": 89935
        }
    ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -26,6 +26,14 @@
                "TF_ACC": "1",
            },
            "showLog": true
+       },
+       {
+
+            "name": "Attach to Process",
+            "type": "go",
+            "request": "attach",
+            "mode": "local",
+            "processId": 89935
        }
    ]
 }
--- a/common/client.go
+++ b/common/client.go
@@ -47,6 +47,7 @@ func (c *DatabricksClient) Configure() error {
 
 // ConfigureWithAuthentication configures the client to work
 // and validates if an authentication mechanism is present
+// TODO: remove me
 func (c *DatabricksClient) ConfigureWithAuthentication() error {
 	err := c.Configure()
 	if err != nil {
--- a/common/client.go
+++ b/common/client.go
@@ -45,17 +45,6 @@ func (c *DatabricksClient) Configure() error {
 	return nil
 }
 
-// ConfigureWithAuthentication configures the client to work
-// and validates if an authentication mechanism is present
-// TODO: remove me
-func (c *DatabricksClient) ConfigureWithAuthentication() error {
-	err := c.Configure()
-	if err != nil {
-		return err
-	}
-	return c.ensureAuthenticationIsSetup()
-}
-
 func (c *DatabricksClient) findAndApplyAuthorizer() error {
 	authorizers := []func() (func(r *http.Request) error, error){
 		c.configureAuthWithDirectParams,
--- a/common/client.go
+++ b/common/client.go
@@ -38,24 +38,12 @@ func (c *DatabricksClient) Configure() error {
 	c.configureHTTPCLient()
 	c.AzureAuth.databricksClient = c
 	c.userAgent = UserAgent()
-	err := c.findAndApplyAuthorizer()
-	if err != nil {
-		return err
-	}
-	if c.Host != "" && !(strings.HasPrefix(c.Host, "https://") || strings.HasPrefix(c.Host, "http://")) {
-		// azurerm_databricks_workspace.*.workspace_url is giving URL without scheme
-		// so that is why this line is here
-		c.Host = "https://" + c.Host
-	}
 	// TODO: fix it once bigger boom is done
 	//c.commandExecutor = c.Commands()
 	return nil
 }
 
 func (c *DatabricksClient) findAndApplyAuthorizer() error {
-	if c.authVisitor != nil {
-		return nil
-	}
 	authorizers := []func() (func(r *http.Request) error, error){
 		c.configureAuthWithDirectParams,
 		c.AzureAuth.configureWithClientSecret,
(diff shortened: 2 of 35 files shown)


.vscode/launch.json
```diff
@@ -26,6 +26,14 @@
...
            },
            "showLog": true
+       },
+       {
+
+            "name": "Attach to Process",
+            "type": "go",
+            "request": "attach",
+            "mode": "local",
+            "processId": 89935
```
nfx, line 36: why do we need this change?

common/client.go
```diff
@@ -38,24 +38,22 @@ func (c *DatabricksClient) Configure() error {
...
 	//c.commandExecutor = c.Commands()
 	return nil
 }
 
-func (c *DatabricksClient) findAndApplyAuthorizer() error {
-	if c.authVisitor != nil {
-		return nil
+// ConfigureWithAuthentication configures the client to work
+// and validates if an authentication mechanism is present
+func (c *DatabricksClient) ConfigureWithAuthentication() error {
```
nfx, line 48: Do not introduce a public method just to be used in singe unit test file. 

Please revert this `Configure` method rename, because it also deviates behaviour from the one specified in `Provider#Configure` - e.g. never called in production flow.

common/http.go
```diff
@@ -210,8 +210,9 @@ func (c *DatabricksClient) checkHTTPRetry(ctx context.Context, resp *http.Respon
 
 // Get on path
 func (c *DatabricksClient) Get(path string, request interface{}, response interface{}) error {
-	if c.authVisitor == nil {
-		return fmt.Errorf("Authentication not initialized")
+	err := c.ensureAuthenticationIsSetup()
```
nfx, line 213: 1) create method `authenticatedQuery` with same signature as `genericQuery` 
2) do error check there
3) replace `genericQuery` with `authenticatedQuery` in `Get/Post/Put/Patch/Delete/Scim/OldAPI`, but don't pass in authVisitor, because it should be handled in `authenticatedQuery`.
4) this is happening from multiple threads, so add a mutex as field of `DatabricksClient` struct, that would be managed within either `authenticatedQuery` or `findAndApplyAuthorizer`. otherwise even more unexplainable bugs will happen

this way we'll have less copy-pasted lines and single place of control.

common/client_test.go
```diff
@@ -16,7 +16,7 @@ func TestDatabricksClientConfigure_BasicAuth_NoHost(t *testing.T) {
 		Username: "foo",
 		Password: "bar",
 	}
-	err := dc.Configure()
+	err := dc.ConfigureWithAuthentication()
```
nfx, line 19: assert for error effect by calling `.Get("/dummy", nil, nil)`, because this is where change actually is in production flow, not in `ConfigureWithAuthentication`, which is not called by provider itself and should be reverted.

provider/provider_test.go
```diff
@@ -266,7 +266,7 @@ func TestProviderConfigurationOptions(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(fmt.Sprintf("config:%v env:%v", tt.rawConfig(), tt.env), func(t *testing.T) {
-			c, err := configureProviderAndReturnClient(t, tt)
+			c, err := configureProviderWithAuthenticationAndReturnClient(t, tt)
```
nfx, line 269: assert for error effect by calling `.Get("/dummy", nil, nil)`, because this is where change actually is in production flow, not in `ConfigureWithAuthentication`, which is not called by provider itself and should be reverted.

scripts/azcli-integration/main.tf
```diff
@@ -48,7 +48,7 @@ output "test_gen1_name" {
 }
 
 output "test_storage_account_key" {
-  value = module.this.test_storage_account_key
+  value     = module.this.test_storage_account_key
```
nfx, line 51: formatting changes are irrelevant for contents of this PR. Please revert and make a separate PR for formatting changes if you think they are needed.

nfx (changes requested):
1. revert irrelevant changes to this PR
2. `authenticatedQuery` method
3. revert `ConfigureWithAuthentication` renames
4. assert for error effect by calling `.Get("/dummy", nil, nil)`

nfx:
Also mention in documentation on auth differences with planning & actual reading phases https://github.com/databrickslabs/terraform-provider-databricks/blob/master/docs/index.md#empty-provider-block

.vscode/launch.json
sdebruyn: We didn't, this was unintentional.

sdebruyn:
@nfx In some places I also replaced assert.Nil with assert.NoError and assert.NotNil with assert.Error, would you like me to put that back as well?

common/client_test.go
sdebruyn: Alright, got it. This will work for this test, but for `TestDatabricksClientConfigure_BasicAuth` and `TestDatabricksClientConfigure_HostWithoutScheme` (basically all tests where we expect a success state) we'll need the mock server because the get request will fail otherwise. Is that okay?

provider/provider_test.go
sdebruyn: @nfx Same as above, we'd probably need a mock HTTP server here so that cases with successful authentication can pass through without errors. Can you confirm this is the way to go?

nfx:
merged from a bit earlier fork of it. closing this one. thanks, @sdebruyn !

PR #274, opened 2020-08-27
Add retry on HTTP 429

### LINKED ISSUE #192: [ISSUE] Retry on HTTP 429
Hi,

Notebooks are a resource type on databricks that are large in volume in terms of content being created, modified and deleted. These types are resources are an easy way to trigger 429s. This should be a retryable error rather than a failure point. Similar to how delete is managed.

```
error:   13: resource "databricks_notebook" 
error: Error: Response from server (429) : unexpected end of JSON input
error: Please consult API docs at https://docs.databricks.com/dev-tools/api/latest/workspace.html#import for details.
```

nfx:
Fixes #192


common/http.go
```diff
@@ -207,6 +207,12 @@ func (c *DatabricksClient) checkHTTPRetry(ctx context.Context, resp *http.Respon
 		// In this case don't retry and return the original error from httpclient
 		return false, err
 	}
+	if resp.StatusCode == 429 {
+		return true, APIError{
+			ErrorCode: "TOO_MANY_REQUESTS",
+			Message:   "Current request has to be retried",
+		}
+	}
```
stikkireddy, line 215: Awesome! This makes it so much easier if its handled at the client level rather than retrying at the resource levels. I think in most places having a linear retry with jitter should be enough. Do not think any of them would require exponential backoff.

nfx: whatever is standard is the best :)

PR #313, opened 2020-09-10
Fix #296 - Added support for local disk encryption

### LINKED ISSUE #296: [FEATURE] Support for cluster local disk encryption
Request the support to enable local disk encryption on clusters, which is in public preview and available in the API.  [Enable local disk encryption](https://docs.databricks.com/clusters/configure.html#enable-local-disk-encryption)
[Azure Enable Disk Encryption](https://docs.microsoft.com/en-us/azure/databricks/clusters/configure#enable-local-disk-encryption)

##### This implementation may look like:
```hcl
  spark_version           = "6.6.x-scala2.11"
  node_type_id            = "i3.xlarge"
  autotermination_minutes = 20
  enable_local_disk_encryption = true
  autoscale {
    min_workers = 1
    max_workers = 50
  }
}

nfx:


nfx:
Manually verified it on Azure

Issue #235, opened 2020-08-11
[ISSUE] Remove deprecated testing framework functions with 10 arguments and replace them with `ResourceFixture` or `HttpFixtureClient`


nfx:
There's `ResourceFixture` (or `HttpFixtureClient`) as convenience method for all tests.

* AssertRequestWithMockServer has 10 arguments, which makes it extremely difficult to work with those tests
* AssertMultipleRequestsWithMockServer has 10 arguments, which makes it extremely difficult to work with those tests. issue is aggravated by "columns" of values passed in, which makes adding/removing a request 8 times more difficult than it should :)

nfx:
@alexott, can you look at this one as well?

PR #326, opened 2020-09-17
[bugfix] "Secret scope ACL is MANAGE for all users by default"

### LINKED ISSUE #322: [ISSUE] Secret scope ACL is set to MANAGE for all users by default and no way to override
### Terraform Version
```
Terraform v0.13.2
+ provider registry.terraform.io/databrickslabs/databricks v0.2.4
```

### Affected Resource(s)
- `databricks_secret_scope`

```hcl
resource "databricks_secret_scope" "test" {
  name = "test"
}
```

### Expected Behavior
A secret scope is created with only the user whose databricks authentication is being used having MANAGE access

### Actual Behavior
All users are given MANAGE access to the secret scope

### Steps to Reproduce
1. `terraform apply`

The underlying issue appears to be that the Databricks API is somewhat limited and requires that `initial_manage_principal` be omitted OR be set to `users`. Surprisingly it does not allow a caller to specify a specific desired manage principal, even if you set the principal to match the calling user.

In the provider code I see that there's a default value of `Users` set for this field: https://github.com/databrickslabs/terraform-provider-databricks/blob/master/access/resource_secret_scope.go#L83

There should be no default for this field if it's not present, because the API sets the value to the current user only if the parameter is not present in the API call. Any other value is rejected by the API with a 400 error code.

alexott:
If the scope is defined with `initial_manage_principal = ""`, then the corresponding field is omitted from the request body, and as result, `users` group won't get the `MANAGE` permission for created scope as [described in Secrets API docs](https://docs.databricks.com/dev-tools/api/latest/secrets.html#create-secret-scope).

This fixes #322

### DIFF
--- a/access/acceptance/secret_acl_test.go
+++ b/access/acceptance/secret_acl_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	. "github.com/databrickslabs/databricks-terraform/access"
+	"github.com/databrickslabs/databricks-terraform/identity"
 
 	"github.com/databrickslabs/databricks-terraform/common"
 	"github.com/databrickslabs/databricks-terraform/internal/acceptance"
@@ -26,6 +27,9 @@ func TestAccSecretAclResource(t *testing.T) {
 	scope := fmt.Sprintf("tf-scope-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
 	principal := "users"
 	permission := "READ"
+	client := common.CommonEnvironmentClient()
+	me, _ := identity.NewUsersAPI(client).Me()
+	userName := me.UserName
 
 	acceptance.AccTest(t, resource.TestCase{
 		CheckDestroy: testSecretACLResourceDestroy,
@@ -35,6 +39,8 @@ func TestAccSecretAclResource(t *testing.T) {
 				Config: testSecretACLResource(scope, principal, permission),
 				// compose a basic test, checking both remote and local values
 				Check: resource.ComposeTestCheckFunc(
+					// test scope permissions - it should be current user
+					testSecretScopeHasPrincipal(t, scope, userName, "MANAGE"),
 					// query the API to retrieve the tokenInfo object
 					testSecretACLResourceExists("databricks_secret_acl.my_secret_acl", &secretACL, t),
 					// verify remote values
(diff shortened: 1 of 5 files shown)


nfx:
not to forget adding this comment later:

1. entry within `changelog.md`
2. updated documentation for markdown file
3. integration and/or acceptance test

alexott:
@nfx updated with a new version

access/resource_secret_scope.go
```diff
@@ -80,7 +84,6 @@ func ResourceSecretScope() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 				// or creator...
```
nfx, line 86: please also remove this comment, that's not pointing to any line :)

access/acceptance/secret_acl_test.go
```diff
@@ -132,3 +147,18 @@ func testSecretACLResource(scopeName, principal, permission string) string {
...
 }
+
+// testAccTokenResource returns an configuration for an Example Widget with the provided name
+func testSecretACLResourceWithDefaultPrincipal(scopeName, default_principal, principal, permission string) string {
+	return fmt.Sprintf(`
+		resource "databricks_secret_scope" "my_scope" {
+			name = "%s"
+            initial_manage_principal = "%s"
+		}
+		resource "databricks_secret_acl" "my_secret_acl" {
```
nfx, line 158: so to test this, you should only create `databricks_secret_scope` and check what ACLs are returned through `NewSecretAclsAPI` in a custom resource check. In this integration test you're changing existing resource, which is not really testing the case of the bug, so i'd ask to create new `TestAcc...` for it.

access/resource_secret_scope.go
```diff
@@ -21,10 +21,14 @@ type SecretScopesAPI struct {
 
 // Create creates a new secret scope
 func (a SecretScopesAPI) Create(scope string, initialManagePrincipal string) error {
-	return a.C.Post("/secrets/scopes/create", map[string]string{
-		"scope":                    scope,
-		"initial_manage_principal": initialManagePrincipal,
-	}, nil)
+	paramsMap := map[string]string{
+		"scope": scope,
```
nfx, line 25: if `initialManagePrincipal`, isn't it omitting it on marshaller side?..

docs/resources/secret_scope.md
```diff
@@ -23,7 +23,7 @@ resource "databricks_secret_scope" "my-scope" {
 The following arguments are supported:
 
 * `name` - (Required) Scope name requested by the user. Scope names are unique. This field is required.
-* `initial_manage_principal` - (Optional) The principal that is initially granted MANAGE permission to the created scope.
+* `initial_manage_principal` - (Optional) The principal that is initially granted `MANAGE` permission to the created scope.  If it's omitted, then the initial ACL with `MANAGE` permission applied to the scope is assigned to the API request issuer's user identity (see [documentation](https://docs.databricks.com/dev-tools/api/latest/secrets.html#create-secret-scope)).
```
nfx, line 26: you can get `API request issuer's user identity` within the test by calling `Me()` function from UsersAPI

nfx (changes requested):
Refactor test step into it's own integration test, so that we can test the creation of a secret scope.

access/resource_secret_scope.go
alexott: it's omitted only when used in struct fields, annotated with `omitempty`

nfx: Thanks for confirming. For GET requests, where map turns into a query string - it’s omitted, just in case.

alexott:
@nfx all comments were addressed... The code coverage change - maybe because of new functions where we don't have test cases fir `if err != nil` cases.

access/acceptance/secret_acl_test.go
nfx: to rephrase previous comment - can you remove `databricks_secret_acl.my_secret_acl`, because this test should verify if there's no ACL for `users` when `initial_manage_principal` is not set.

access/acceptance/secret_acl_test.go
```diff
@@ -69,6 +70,44 @@ func TestAccSecretAclResource(t *testing.T) {
...
 
+func TestAccSecretAclResourceDefaultPrincipal(t *testing.T) {
+	// TODO: refactor for common instance pool & AZ CLI
+	if _, ok := os.LookupEnv("CLOUD_ENV"); !ok {
+		t.Skip("Acceptance tests skipped unless env 'CLOUD_ENV' is set")
+	}
+	//var secretScope Secre
+	var secretACL ACLItem
+	scope := fmt.Sprintf("tf-scope-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+	initialPrincipal := "users"
```
nfx, line 81: we need two different automated tests for this PR:

1) `initial_manage_principal` is not set and verify, that only admins can manage this scope.
2) `initial_manage_principal` set to group created specifically for this test and verify that only that group and admins can manage this scope.

nfx (changes requested):
we need two different automated tests for this PR:

* initial_manage_principal is not set and verify, that only admins can manage this scope.
* initial_manage_principal set to group created specifically for this test and verify that only that group and admins can manage this scope.

access/acceptance/secret_acl_test.go
alexott: we can't use arbitrary group for `initial_principal` as per [documentation](https://docs.databricks.com/dev-tools/api/latest/secrets.html#create-secret-scope): 

> The only supported principal for this option is the group users, which contains all users in the workspace.

and if I create any other group, and use it as initial principal, then Secrets API returns error saying that I can't use that group as initial principal

nfx: > and if I create any other group, and use it as initial principal, then Secrets API returns error saying that I can't use that group as initial principal

that one is strange. what about user?

alexott: same - you can't explicitly set user as `initial_principal` as well, but this happens implicitly when no `initial_principal` is provided - I've added corresponding test that checks that user has MANAGE permission for scope if `initial_principal` is not provided.

But I think that we need to talk with developers to understand that design decision, I also wondering why this happens...

access/acceptance/secret_acl_test.go
```diff
@@ -26,6 +27,9 @@ func TestAccSecretAclResource(t *testing.T) {
 	scope := fmt.Sprintf("tf-scope-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
 	principal := "users"
 	permission := "READ"
+	client := common.CommonEnvironmentClient()
+	me, _ := identity.NewUsersAPI(client).Me()
```
nfx, line 31: this is called "ignoring the error". we must handle errors everywhere, assert.NoError(t,err) in tests as well.

nfx (approved):

access/acceptance/secret_acl_test.go
alexott: just pushed the modified version, but it was already merged - will fix in the next iteration, when will do #235

PR #330, opened 2020-09-18
Identity refactor

nfx:


nfx:
@stikkireddy consider this as almost finished one. Still need to add few tests and perhaps `databricks_user_instance_profile` resource

identity/resource_user.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		err = internal.StructToData(entity, s, d)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		return nil
+	}
+}
```
stikkireddy, line 30: Hmm seems a bit obscure to make this a public method and only use it in the readContext. The method definition is also quite confusing to read in the readContext.

maybe something like:
```go
entity = NewUsersAPI(m).ReadR(d.Id())
return Bind(entity)(d, userSchema)
```

stikkireddy (changes requested):

identity/resource_user.go
nfx: removed this method

nfx (commented):
applied some review comments

stikkireddy (approved):
LGTM, need to run the integration tests for the release.

PR #334, opened 2020-09-22
Fix #323 - add DiffSuppressFunc for legacy conf

nfx:

PR #335, opened 2020-09-22
Improve library await

### LINKED ISSUE #315: [FEATURE] Add option to skip waiting of library installation
Library resource always waits for library installation. I would like an option to skip this because it is not important to have a library installed for some use cases. Otherwise, it would be great if there is a way to set the timeouts instead of using default timeouts.

The issue/feature is related to this function https://github.com/databrickslabs/terraform-provider-databricks/blob/732b3f01224658d9ee01a594f3be1e5d27a613f8/compute/resource_cluster.go#L183

nfx:
Fix #315


compute/model.go
```diff
@@ -443,7 +443,7 @@ type JobEmailNotifications struct {
 type CronSchedule struct {
 	QuartzCronExpression string `json:"quartz_cron_expression"`
 	TimezoneID           string `json:"timezone_id"`
-	PauseStatus          string `json:"pause_status,omitempty"`
+	PauseStatus          string `json:"pause_status,omitempty" tf:"computed"`
```
alexott, line 446: This will conflict with #334

compute/resource_cluster.go
```diff
@@ -183,7 +183,7 @@ func resourceClusterRead(d *schema.ResourceData, m interface{}) error {
 func waitForLibrariesInstalled(
 	libraries LibrariesAPI, clusterInfo ClusterInfo) (result *ClusterLibraryStatuses, err error) {
 	// nolint should be a bigger refactor
-	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
+	err = resource.Retry(30*time.Minute, func() *resource.RetryError {
```
alexott, line 186: Will it make sense to move this to constant, defined at the beginning?

alexott (changes requested):
there will be conflict in the CronSchedule struct - this change is not required for this PR

compute/model.go
nfx: why? it's the same change

PR #342, opened 2020-09-28
Initial commit for Table ACL support

nfx:
Refactoring Executor interface to support table results

### DIFF
--- a/access/resource_table_acl.go
+++ b/access/resource_table_acl.go
@@ -1,0 +1,303 @@
+package access
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/databrickslabs/databricks-terraform/common"
+	"github.com/databrickslabs/databricks-terraform/compute"
+	"github.com/databrickslabs/databricks-terraform/internal"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// https://docs.databricks.com/security/access-control/table-acls/object-privileges.html#operations-and-privileges
+
+// TableACL defines table access control
+type TableACL struct {
+	Table     string   `json:"table,omitempty"`
+	View      string   `json:"view,omitempty"`
+	Database  string   `json:"database,omitempty"`
+	Principal string   `json:"principal"`
+	Grants    []string `json:"grants,omitempty" tf:"slice_set"`
+	Denies    []string `json:"denies,omitempty" tf:"slice_set"`
+	Owner     string   `json:"owner,omitempty"`
+	ClusterID string   `json:"cluster_id,omitempty"`
+	// SanityCheck bool     `json:"sanity_check,omitempty"`
+}
+
+func (ta *TableACL) actualDatabase() string {
+	if ta.Database == "" {
+		return "default"
+	}
+	return ta.Database
+}
+
+// TypeAndKey returns ACL object type and key
+func (ta *TableACL) TypeAndKey() (string, string) {
+	if ta.Table != "" {
+		return "TABLE", fmt.Sprintf("`%s`.`%s`", ta.actualDatabase(), ta.Table)
+	}
+	if ta.View != "" {
+		return "VIEW", fmt.Sprintf("`%s`.`%s`", ta.actualDatabase(), ta.View)
+	}
+	if ta.Database != "" {
+		return "DATABASE", ta.Database
+	}
+	return "", ""
+}
+
+// ID returns Terraform resource ID
+func (ta *TableACL) ID() string {
+	objectType, key := ta.TypeAndKey()
+	noBackticks := strings.ReplaceAll(key, "`", "")
+	return fmt.Sprintf("%s/%s/%s", objectType, noBackticks, ta.Principal)
+}
+
+func loadTableACL(id string) (TableACL, error) {
+	ta := TableACL{}
+	split := strings.SplitN(id, "/", 3)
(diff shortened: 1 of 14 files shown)


access/resource_table_acl.go
```diff
@@ -0,0 +1,303 @@
...
+)
+
+// https://docs.databricks.com/security/access-control/table-acls/object-privileges.html#operations-and-privileges
+
+// TableACL defines table access control
+type TableACL struct {
+	Table     string   `json:"table,omitempty"`
+	View      string   `json:"view,omitempty"`
+	Database  string   `json:"database,omitempty"`
+	Principal string   `json:"principal"`
```
databricks-david-lewis, line 23: Is this the principal who has the granted and denied permissions?
If I want to give multiple people permission on the same object and I specify a different owner in different blocks, which owner is actually applied?

It seems like a SQL object should have a list of (principal, granted permissions, denied permissions) tuples.

This way all of the permissions for a single object are defined together.

access/resource_table_acl.go
```diff
@@ -0,0 +1,303 @@
...
+	Grants    []string `json:"grants,omitempty" tf:"slice_set"`
+	Denies    []string `json:"denies,omitempty" tf:"slice_set"`
+	Owner     string   `json:"owner,omitempty"`
+	ClusterID string   `json:"cluster_id,omitempty"`
+	// SanityCheck bool     `json:"sanity_check,omitempty"`
+}
+
+func (ta *TableACL) actualDatabase() string {
+	if ta.Database == "" {
```
databricks-david-lewis, line 32: Allowing database to not be set and defaulting to `default` seems like handing users a foot-gun. I think forcing them to explicitly specify the Database is a good idea.

access/resource_table_acl.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/databrickslabs/databricks-terraform/internal"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// https://docs.databricks.com/security/access-control/table-acls/object-privileges.html#operations-and-privileges
+
+// TableACL defines table access control
+type TableACL struct {
```
databricks-david-lewis, line 19: There are also three other SQL objects that can have acls.
`CATALOG`
`ANY_FILE`
`ANONYMOUS_FUNCTION`
https://docs.databricks.com/security/access-control/table-acls/object-privileges.html

access/resource_table_acl.go
nfx: @databricks-david-lewis, right now it is managing grants/denies on intersection between the object & principal. e.g.

```hcl
resource "datbaricks_table_acl" "data_scientists_can_read_production_database" {
	database = "production"
	principal = "Data Scientists"
	grants = ["SELECT", "READ_METADATA"]
}

resource "datbaricks_table_acl" "data_scientists_cannot_select_very_sensitive_table" {
	database = "production"
	table = "very_sensitive_table"
	principal = "Sub Group of Data Scientists"
	denies = ["SELECT"]
}

resource "datbaricks_table_acl" "circle_of_trust_can_select_very_sensitive_table" {
	database = "production"
	table = "very_sensitive_table"
	principal = "Circle Of Trust"
	grants = ["SELECT"]
}
```

Other options i was considering were:
```
// option #1 - easiest to code, but then users will have same difficulties as with SHOW GRANT statement
resource "datbaricks_table_acl" "data_scientists_cannot_read_things_table" {
	// but can modify it :D to deny modify we need a separate thing
        database = "production"
	table = "things"
	principal = "Data Scientists"
	allow = false
	action = "SELECT"
}
// option #2 - direct equivalent of SHOW GRANT with one-hot encoding for priviliges
resource "datbaricks_table_acl" "permissions_of_data_scientist_on_things_table" {
	// and destroying resource will REVOKE ALL PRIVILEGES ON TABLE things FROM foo
       database = "production"
	table = "things"
	principal = "Data Scientists"
	select = true
	modify = true
	read_metadata = true
}
```

Design constraints here are related to time spent on the client. Each resource in the best case scenario is at least one context creation and query execution. The more resource definition people would have to make, the more time it'll spend planning the change. 

Alternatively, we can also make resource definition the same way it's done for [databricks_permissions](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/docs/resources/permissions.md) (see docs - that's why i was checking with you if it's going to make sense to expose things like Table ACLs to General Permissions API), which may simplify one things for users and complicate others:

```
resource "datbaricks_table_acl" "permissions_of_data_scientist_on_things_table" {
	// and destroying resource will REVOKE ALL PRIVILEGES ON TABLE things FROM foo
        database = "production"
	table = "things"

        grant {
             principal = "Data Scientists"
             permission_level = "SELECT"
        }

        grant {
             principal = "Automation"
             // or this...
             permissions = ["SELECT", "READ_METADATA", "MODIFY"]
        }

        deny {
             principal = "Data Science Interns (subgroup of data scientists)"
             permissions = ["SELECT"]
        }
}
```

Owner support is not yet implemented.

access/resource_table_acl.go
nfx: take a look at `TypeAndKey` & `loadTableACL ` methods for clarification

access/resource_table_acl.go
nfx: are those often used?.. i wonder how to create a resource id for those.

access/resource_table_acl.go
databricks-david-lewis: I see, that makes sense about minimizing the number of contexts.

I like that last option. To me it makes sense to group all of the permissions for one object together in the same place.

PR #348, opened 2020-09-30
Implement Pinned Cluster Support (#113)

### LINKED ISSUE #113: [FEATURE] Pinned Cluster Support
SDK has support for pinning/unpinning the cluster, though it's not available through the resource. 

We should add `is_pinned = true/false` to resource and make sure cluster deletion would unpin the cluster before doing so, as `Cannot delete pinned cluster 0000-111222-abcd0. Please ask an admin to unpin the cluster first.` errors are appearing in current version.

alexott:
This should fix #113


nfx (changes requested):
Missing api call in read method to get if status is actually pinned. This is why you’re always unpinning cluster in unit tests (e.g. libs create), even though state changes were not part of those tests. State change must not be applied if it didn’t happen.

alexott:
@nfx I don't see API to check if the cluster is pinned or not - it's not returned in the call to `Get`.  The only place we get information about pin/unpin is the events API - we need to get all events, and filter out if cluster was pinned or unpinned.
Maybe we could go the way of the Diff function, or something like.

docs/resources/cluster.md
```diff
@@ -33,6 +33,7 @@ disable automatic termination. _It is highly recommended to have this setting pr
...
 * `custom_tags` - (Optional) Additional tags for cluster resources. Databricks will tag all cluster resources (e.g., AWS instances and EBS volumes) with these tags in addition to `default_tags`.
 * `spark_conf` - (Optional) Map with key-value pairs to fine tune Spark clusters, where you can provide custom pSpark configuration properties](https://spark.apache.org/docs/latest/configuration.html) in a cluster configuration. You can also pass in a string of extra JVM options to the driver and the executors via `spark.driver.extraJavaOptions` and `spark.executor.extraJavaOptions` respectively. It is advised to keep all common configurations in [Cluster Policies](cluster_policy.md) to maintain control of the environments launched.
+* `is_pinned` - (Optional) boolean value specifying if cluster is pinned (not pinned by default)
```
nfx, line 36: please also add a link to pinning documentation and it's limitations.

nfx:
> The only place we get information about pin/unpin is the events API - we need to get all events, and filter out if cluster was pinned or unpinned.

@alexott that's the only place so far 🤷‍♂️

docs/resources/cluster.md
alexott: done

alexott:
For clusters that exist for a long time, the events API could produce a lot of data, and that could be quite slow to go through.  I need to read about Diff function

stikkireddy:
just a quick comment there is a limit to how many clusters you can have pinned. It may fail to pin, may want to look into how to handle that error state and what that would do during the update process/create process. May want a unit test for failed pin to make sure the right path is taken when it fails. 

Right now in the create phase when it fails it will taint the resource. In update it will just fail the resource. I think this is the expected behavior. The api docs do not state what error will be thrown. I am assuming it throws a 400+. https://docs.databricks.com/clusters/clusters-manage.html#pin-a-cluster has the pinned cluster limitations but no explanation on how the api will behave when we try to pin the 21st cluster.

alexott:
@stikkireddy @nfx what do you think will be the best strategy for handling this - log the error, and continue, or fail the task?

nfx:
@alexott @stikkireddy unit test on failed pin is definitely needed. please create 22 clusters and try to pin them all and see how failure mode looks like :)

expected behaviour should be setting pin back to false (default option). 

> For clusters that exist for a long time, the events API could produce a lot of data, and that could be quite slow to go through. I need to read about Diff function

https://docs.databricks.com/dev-tools/api/latest/clusters.html#events has `event_types` filter, which you can set to `PINNED` and `UNPINNED` with DESC sorting. this way iteration is going to be quick - by simply checking first element status to `PINNED` equality or presence.

alexott:
Yep, I've started to work on supporting events API as well...

compute/resource_cluster.go
```diff
@@ -107,6 +107,18 @@ func resourceClusterSchema() map[string]*schema.Schema {
 			Optional: true,
 			Computed: true,
 		}
+		s["is_pinned"] = &schema.Schema{
+			Type:     schema.TypeBool,
+			Optional: true,
+			Default:  false,
+			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
stikkireddy, line 114: can we add a comment on why this diffsuppressfunc is needed @alexott afaik this should not be required right?

stikkireddy: ahh i see the problem now there is no get pinned or get cluster does not yield pinned status.

alexott:
@nfx @stikkireddy I've pushed the implementation that uses Events API to fetch pinned/unpinned status

compute/clusters.go
```diff
@@ -226,6 +226,39 @@ func (a ClustersAPI) Unpin(clusterID string) error {
...
+		curPos += len(eventsResponse.Events)
+		copy(events[startPos:curPos], eventsResponse.Events)
+	}
+
+	return events, err
+}
+
+// Cluster Events API - only using Cluster ID string to get all events
+// https://docs.databricks.com/dev-tools/api/latest/clusters.html#events
+func (a ClustersAPI) EventsForCluster(clusterID string) ([]ClusterEvent, error) {
```
nfx, line 258: is it used anywhere else than in tests?

compute/resource_cluster.go
```diff
@@ -250,6 +287,19 @@ func resourceClusterUpdate(d *schema.ResourceData, m interface{}) error {
 	if err != nil {
 		return err
 	}
+	oldPinned, newPinned := d.GetChange("is_pinned")
+	// log.Printf("[DEBUG] Update: is_pinned. Old: %#v, New: %#v", oldPinned, newPinned)
```
nfx, line 291: log statement can be moved into condition and uncommented

compute/resource_cluster.go
```diff
@@ -181,6 +201,23 @@ func resourceClusterRead(d *schema.ResourceData, m interface{}) error {
 	if err != nil {
 		return err
 	}
+	events, err := clusterAPI.Events(EventsRequest{
```
nfx, line 204: could you factor this out to separate function?

nfx (changes requested):
small changes requested

alexott:
> small changes requested

addressed...

nfx (approved):
LGTM

PR #375, opened 2020-10-21
Support "single-node" clusters

sabino:
I was doing some tests using the community version of Databricks and the provider is not able to create the cluster:
image

This is also true in case you want to create a [Single Node cluster](https://docs.databricks.com/clusters/single-node.html)

~Not sure if this will actually fix it or even if this is the best approach, but I'm proposing the change based on this post
https://www.sohamkamani.com/golang/2018-07-19-golang-omitempty/#the-difference-between-0--and-nil~

I've set a default value in case it doesn't exist, that way it is possible to remove the `omitempty`.
And since all the tests passed I'm assuming it shouldn't break existing resources definitions.


sabino:
I could use some help.
Is this actually something sane to do?

nfx:
@sabino `omitempty` is the option you're looking for. Current PR change causes unit tests to fail.

sabino:
@nfx I saw that option but changing it will make the field required, which is not ideal since it is only required when not using autoscale (not supported on Community edition).

Having `omitempty` actually ignores the value if it is 0.
Removing it will make it required always, but the value won't be returned from the GET cluster endpoint, thus triggering an update always.

That's why I was trying to allow 0 by using a pointer.

https://stackoverflow.com/questions/38486564/unmarshal-marshal-json-with-int-set-to-0-does-not-seem-to-work

nfx:
i'm not sure if we want to support provider to work for all community edition corner cases :)

sabino:
@nfx this should fix it

compute/resource_cluster.go
```diff
@@ -101,6 +101,11 @@ func resourceClusterSchema() map[string]*schema.Schema {
 			p.Sensitive = true
 		}
 		s["autotermination_minutes"].Default = 60
+		s["num_workers"] = &schema.Schema{
+			Type:     schema.TypeInt,
+			Optional: true,
```
nfx, line 106: what do acceptance tests say about this change?

sabino: I've added a test, is it enough?

nfx: @sabino no, it's not enough :)

1. please run `make test-azcli` and attach the output of `scripts/azcli-integration/test-report.log`. You're on Azure, right?
2. please modify documentation for cluster resource

I a have strange hunch that this might break some functionality and want to have it properly tested before merging.

nfx (changes requested):

compute/resource_cluster.go
sabino: I'm actually running on AWS - Single Tenant.

Anyway, I'll update the documentation.
Btw, thanks for the follow up!

sabino: I'll create a trial account and run the tests.

sabino: @nfx FYI looks like it is not possible to use a trial account, which is a bummer. I've tried Pay-As-You-Go as well and they still have limits for those. I've requested a quota increase, but I'm not sure how long it will take.

image

nfx: @sabino, just in case, i'm interested in output of only `TestAcc*Cluster*` tests. They are independent of a cloud. See contributing docs for guidelines how to run tests

sabino: @nfx is this it?

I've used this command:
`DATABRICKS_CONFIG_PROFILE="staging" ./run.sh awsst "TestAcc*Cluster*" --debug --tee`

- `test-report.log`

```
awsst
---

 * [x] compute/acceptance.TestAccClusterPolicyResourceFullLifecycle (8.14s)
 * [x] compute/acceptance.TestAccClusterResource_CreateClusterWithLibraries (158.32s)
```

- `out.log`
```
EMPTY .
testing: warning: no tests to run
coverage: 0.0% of statements
PASS common
testing: warning: no tests to run
coverage: 16.3% of statements
PASS internal
testing: warning: no tests to run
coverage: 0.0% of statements
PASS identity
testing: warning: no tests to run
coverage: 0.0% of statements
PASS internal/qa
testing: warning: no tests to run
coverage: 0.0% of statements
PASS internal/util
testing: warning: no tests to run
coverage: [no statements]
PASS access/acceptance
testing: warning: no tests to run
coverage: 5.8% of statements
PASS compute
testing: warning: no tests to run
coverage: 0.0% of statements
PASS access
testing: warning: no tests to run
coverage: [no statements]
PASS identity/acceptance
testing: warning: no tests to run
coverage: 0.0% of statements
PASS internal/sanity
testing: warning: no tests to run
coverage: 0.8% of statements
PASS mws
testing: warning: no tests to run
coverage: [no statements]
PASS mws/acceptance
testing: warning: no tests to run
coverage: 0.0% of statements
PASS provider
testing: warning: no tests to run
coverage: 0.0% of statements
PASS storage
testing: warning: no tests to run
coverage: [no statements]
PASS storage/acceptance
testing: warning: no tests to run
coverage: 0.0% of statements
PASS workspace
testing: warning: no tests to run
coverage: [no statements]
PASS workspace/acceptance
PASS compute/acceptance.TestAccClusterPolicyResourceFullLifecycle (8.14s)
PASS compute/acceptance.TestAccClusterResource_CreateClusterWithLibraries (158.32s)
coverage: [no statements]
PASS compute/acceptance
EMPTY internal/acceptance

DONE 2 tests in 170.845s
```

nfx (approved):

nfx:
@sabino i'll revert the changes from this PR, because job-related tests do fail. don't have time fixing this. please provide full test output of `make test-awsmt` or `make test-awsst`. `TestAccJobResource` fails. others are failing too.

Issue #346, opened 2020-09-29
Data source for cloud-dependent node types


nfx:
Make terraform configuration less cloud-dependent by allowing to specify node characteristics instead of node type names

```hcl
data "databricks_node_type" "ds" {
    min_memory = 32
    min_memory_per_core = 2
    min_cores = 8
    local_disk = true
    min_gpus = 1
}
```

* by default should return smallest possible instance with local disk
* min_memory_per_core conflicts with min_memory
* id is set to the first matching node type
* error is returned if no match

nfx:
already implemented it, PR pending

gilShin:
I'm working on Azure and for some reason I sometimes get i3.xlarge
Looks like a bug to me

PR #376, opened 2020-10-21
Add databricks_node_type data source

nfx:
Implements issue #346 to help customers pick node types easier:

```hcl
data "databricks_node_type" "with_gpu" {
    local_disk  = true
    min_cores   = 16
    gb_per_core = 1
    min_gpus    = 1
}
resource "databricks_cluster" "research" {
    cluster_name            = "Research Cluster"
    spark_version           = "6.6.x-scala2.11"
    node_type_id            = databricks_node_type.with_gpu.id
    autotermination_minutes = 20
    autoscale {
        min_workers = 1
        max_workers = 50
    }
}
```

### DIFF
--- a/compute/clusters.go
+++ b/compute/clusters.go
@@ -3,7 +3,6 @@ package compute
 import (
 	"fmt"
 	"log"
-	"sort"
 	"strings"
 	"time"
 
@@ -235,11 +234,10 @@ func (a ClustersAPI) List() ([]ClusterInfo, error) {
 	return clusterList.Clusters, err
 }
 
-// ListNodeTypes returns a list of supported Spark node types
-func (a ClustersAPI) ListNodeTypes() ([]NodeType, error) {
-	var nodeTypeList = &NodeTypeList{}
-	err := a.client.Get("/clusters/list-node-types", nil, nodeTypeList)
-	return nodeTypeList.NodeTypes, err
+// ListNodeTypes returns a sorted list of supported Spark node types
+func (a ClustersAPI) ListNodeTypes() (l NodeTypeList, err error) {
+	err = a.client.Get("/clusters/list-node-types", nil, &l)
+	return
 }
 
 // GetOrCreateRunningCluster creates an autoterminating cluster if it doesn't exist
@@ -270,7 +268,9 @@ func (a ClustersAPI) GetOrCreateRunningCluster(name string, custom ...Cluster) (
 			}
 		}
 	}
-	smallestNodeType := a.GetSmallestNodeTypeWithStorage()
+	smallestNodeType := a.GetSmallestNodeType(NodeTypeRequest{
+		LocalDisk: true,
+	})
 	log.Printf("[INFO] Creating an autoterminating cluster with node type %s", smallestNodeType)
 	r := Cluster{
 		NumWorkers:             1,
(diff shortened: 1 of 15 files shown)


compute/clusters.go
```diff
@@ -290,59 +290,54 @@ func (a ClustersAPI) GetOrCreateRunningCluster(name string, custom ...Cluster) (
...
+func (a ClustersAPI) GetSmallestNodeType(r NodeTypeRequest) string {
+	list, _ := a.ListNodeTypes()
+	// error is explicitly ingored here, because Azure returns
+	// apparently too big of a JSON for Go to parse
+	if len(list.NodeTypes) == 0 {
 		if a.client.IsAzure() {
 			return "Standard_D3_v2"
 		}
 		return "i3.xlarge"
 	}
```
alexott, line 313: I'm not sure that it's a correct behavior to return default value in case if people specified some requirements that need to be fulfilled...
maybe we need to look to the other libraries that can parse large JSONs? I've seen references to https://github.com/tamerh/jsparser, for example

compute/clusters.go
```diff
@@ -290,59 +290,54 @@ func (a ClustersAPI) GetOrCreateRunningCluster(name string, custom ...Cluster) (
...
-		return sortedNodeTypes[i].InstanceTypeID < sortedNodeTypes[j].InstanceTypeID
-	})
-	nodeType := sortedNodeTypes[0]
-	return nodeType
+		return nt.NodeTypeID
+	}
+	if a.client.IsAzure() {
+		return "Standard_D3_v2"
+	}
+	return "i3.xlarge"
```
alexott, line 342: similar to above - not sure that we need to return the default instance

alexott (commented):
code wise it looks good, but I have some doubts about fallback to default instance types if we can't parse JSON or can't find instances with matching parameters

compute/clusters.go
```diff
@@ -290,59 +290,54 @@ func (a ClustersAPI) GetOrCreateRunningCluster(name string, custom ...Cluster) (
...
-func (a ClustersAPI) GetSmallestNodeTypeWithStorage() string {
-	nodeTypes, err := a.ListNodeTypes()
-	if err != nil || len(nodeTypes) == 0 {
+// NodeTypeRequest is a wrapper for local filtering of node types
+type NodeTypeRequest struct {
+	MinMemoryGB int32  `json:"min_memory_gb,omitempty"`
+	GBPerCore   int32  `json:"gb_per_core,omitempty"`
+	MinCores    int32  `json:"min_cores,omitempty"`
+	MinGPUs     int32  `json:"min_gpus,omitempty"`
+	LocalDisk   bool   `json:"local_disk,omitempty"`
```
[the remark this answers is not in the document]
nfx: i'd call it "advanced usage" and probably if administrators are willing to get nodes with SSD's, they'd know about them already (or open a bug)

compute/clusters.go
nfx: i've documented it as a known issue -
https://github.com/databrickslabs/terraform-provider-databricks/blob/fabfcdda485050e444e97a74809a1a669a7000bf/docs/data-sources/node_type.md - does it seem reasonable? :) 

https://github.com/tamerh/jsparser is immature.

alexott: that's ok - I went top-down, leaving comments, so the documentation was aftewards...

alexott (approved):
with clarification, looks good

PR #384, opened 2020-10-23
Allow to specify max items to return in the Cluster Events API call (issue #380)

### LINKED ISSUE #380: [ISSUE] Allow to consume only N items from the Events API
Right now, `cluster.Events` call specifies the limit, but still collects all events from the REST API, as the `limit` parameter specifies the number of the items per page, not the total limit.  It would be useful to add a parameter to the call that will specify the maximal number of events to retrieve...

alexott:
Right now, Cluster Events API doesn't allow to specify maximal number of items to fetch -
the existing `limit` parameter specifies the number of the items per page, not max count,
so this limit should enforced on the caller's site.

this fixes #380


nfx (approved):
LGTM

PR #385, opened 2020-10-23
Don't restart cluster on changes that doesn't affect it (issue #379)

### LINKED ISSUE #379: [ISSUE] Changing Pinned status causes cluster restart
The change in the pinned status for existing cluster causes the cluster restart because `/clusters/edit` REST API is not idempotent. We need to implement more sophisticated state comparison, and call `/clusters/edit` only if actual cluster definition has changed.  The cluster restart may also happen if we just add library, etc.

### Terraform Version

Terraform v0.13.4, Databricks provider is built from Git today

### Affected Resource(s)
- databricks_cluster

Initial:

```hcl
resource "databricks_cluster" "test_libs" {
  cluster_name            = "Test Maven"
  spark_version           = "7.3.x-scala2.12"
  node_type_id            = "Standard_D3_v2"
  autotermination_minutes = 20

  autoscale {
    min_workers = 1
    max_workers = 2
  }
}
```
modified:

```hcl
resource "databricks_cluster" "test_libs" {
  cluster_name            = "Test Maven"
  spark_version           = "7.3.x-scala2.12"
  node_type_id            = "Standard_D3_v2"
  autotermination_minutes = 20
  is_pinned               = true

  autoscale {
    min_workers = 1
    max_workers = 2
  }
}
```

### Debug Output

Debug output: https://gist.github.com/alexott/77d1e43786d6b9981299b6682d0dcfd3

### Expected Behavior

Cluster shouldn't be restarted, only pinned status change happens

### Actual Behavior

Cluster is restarted, and pinned status changes

### Steps to Reproduce

1. change the pinned status of the existing cluster
1. execute `terraform apply`

alexott:
This PR adds detection of what cluster parameters were changed, and call `/clusters/edit`
only we changed parameters that are related to cluster directly, not things like
`is_pinned` or `library`...

This fixes #379

### DIFF
--- a/compute/resource_cluster.go
+++ b/compute/resource_cluster.go
@@ -13,6 +13,7 @@ import (
 
 var clusterSchema = resourceClusterSchema()
 
+// ResourceCluster - returns Cluster resource description
 func ResourceCluster() *schema.Resource {
 	return &schema.Resource{
 		SchemaVersion: 2,
@@ -282,6 +283,19 @@ func legacyReadLibraryListFromData(d *schema.ResourceData) (cll ClusterLibraryLi
 	return
 }
 
+func hasClusterConfigChanged(d *schema.ResourceData) bool {
+	for k := range clusterSchema {
+		// TODO: create a map if we'll add more non-cluster config parameters in the future
+		if k == "library" || k == "is_pinned" {
+			continue
+		}
+		if d.HasChange(k) {
+			return true
+		}
+	}
+	return false
+}
+
 func resourceClusterUpdate(d *schema.ResourceData, m interface{}) error {
 	client := m.(*common.DatabricksClient)
 	clusters := NewClustersAPI(client)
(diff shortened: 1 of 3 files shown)


compute/resource_cluster.go
```diff
@@ -277,6 +279,18 @@ func legacyReadLibraryListFromData(d *schema.ResourceData) (cll ClusterLibraryLi
 	return
 }
 
+func hasClusterConfigChanged(d *schema.ResourceData) bool {
+	for k := range clusterSchema {
+		if strings.HasPrefix(k, "library") || k == "is_pinned" {
+			continue
+		}
+		if d.HasChange(k) {
```
nfx, line 287: can you add a unit test to it? :)

nfx (changes requested):
can you add a unit-test?

compute/resource_cluster.go
```diff
@@ -277,6 +279,18 @@ func legacyReadLibraryListFromData(d *schema.ResourceData) (cll ClusterLibraryLi
 	return
 }
 
+func hasClusterConfigChanged(d *schema.ResourceData) bool {
+	for k := range clusterSchema {
+		if strings.HasPrefix(k, "library") || k == "is_pinned" {
```
nfx, line 284: + `libraries` key itself.

nfx: because `library_` is deprecated and would be removed in few weeks

alexott:
it's interesting question - about unit testing of "incremental" change.  I've looked through existing testing infrastructure, and couldn't figure out how to do a "single" change to the existing state - right now you need to provide complete state (with Spark version, etc.), but it will trigger the `/clusters/edit` as before...

nfx (approved):
I'll change it to 0.2.8 :)

PR #398, opened 2020-11-03
Added databricks_workspace_conf resource

nfx:
# databricks_workspace_conf Resource

Manages workspace configuration for expert usage. Currently, more than one instance of resource can exist in Terraform state, though there's no deterministic behavior, when they manage same property. We strongly recommend to use single `databricks_workspace_conf` per workspace.

## Example Usage

Allows specification of custom configuration properties for expert usage:

 * `enableIpAccessLists` - enables the use of `databricks_ip_access_list` resources

```hcl
resource "databricks_workspace_conf" "this" {
    custom_config = {
        "enableIpAccessLists": true
    }
}
```

## Argument Reference

The following arguments are available:

* `custom_config` - (Required) Key-value map of strings, that represent workspace configuration. Upon resource deletion, properties that start with `enable` or `enforce` will be reset to `false` value, regardless of initial default one.

## Import

This resource cannot support import.


jarobey (approved):
I like the updates.  Much more flexible map with keyjoin for values and proper for loops for setting :).  Dubious commented out acceptance test.  Not sure why the delete in the test framework tries to use the IP Access List API to delete a Workspace Conf object.

Issue #403, opened 2020-11-06
[ISSUE] databricks_cluster use of policy_id does not set defaults


kydos93:
When setting the policy_id attribute with use of the databricks_cluster resource, terraform will raise validation errors but is not setting any defaults as specified in the underlying cluster policy

### Terraform Version
0.12.29

### Affected Resource(s)
- databricks_cluster
- databricks_cluster_policy

### Expected Behavior
databricks_cluster resource should use the defaults set in the databricks_cluster_policy when policy_id is supplied with the databricks_cluster resource and report validation errors where overrides that do not comply are set

### Actual Behavior
No defaults were set by the cluster policy and every field raised a validation error

```
Validation failed for spark_conf, spark.databricks.hive.metastore.glueCatalog.enabled must be present; 
Validation failed for spark_conf, spark.hadoop.fs.s3a.acl.default must be present; 
Validation failed for spark_conf, spark.hadoop.fs.s3a.canned.acl must be present; 
Validation failed for driver_node_type_id, the value must be i3.2xlarge (is "i3.xlarge"); 
Validation failed for enable_elastic_disk, the value must be true (is "false"); 
Validation failed for custom_tags, environment must be present; 
Validation failed for custom_tags, request_id must be present; 
Validation failed for custom_tags, business_unit must be present; 
Validation failed for custom_tags, resource_group must be present; 
Validation failed for custom_tags, function must be present; 

Validation failed for custom_tags, project_name must be present; 
Validation failed for custom_tags, requestor must be present; 
Validation failed for custom_tags, project_code must be present; 
Validation failed for custom_tags, service_owner_support must be present; 
Validation failed for custom_tags, terraform must be present; 
Validation failed for custom_tags, application_id must be present; 
Validation failed for aws_attributes.first_on_demand, the value must be 1 (is "0"); 
Validation failed for aws_attributes.spot_bid_price_percent, the value needs to be in range [30, 90] (is "100"); 
Validation failed for cluster_log_conf.path, the value must be present; 
Validation failed for cluster_log_conf.region, the value must be present; 
Validation failed for cluster_log_conf.type, the value must be S3 (is "NONE")
```

nfx:
@kydos93 please share your full relevant configuration & debug log.

nfx:
Will update documentation, but TLDR  - don't use `policy_id` from terraform.

PR #405, opened 2020-11-09
Improve small issues with documentation

nfx:
Also removes confusion of #403


docs/resources/cluster.md
```diff
@@ -148,20 +147,19 @@ cluster_log_conf {
...
+* `endpoint` - (Optional) S3 endpoint, e.g. https://s3-us-west-2.amazonaws.com. Either `region` or `endpoint` needs to be set. If both are set, the endpoint is used.
+* `enable_encryption` - (Optional) Enable server-side encryption, false by default.
+* `encryption_type` - (Optional) The encryption type, it could be `sse-s3` or `sse-kms`. It is used only when encryption is enabled, and the default type is `sse-s3`.
+* `kms_key` - (Optional) KMS key used if encryption is enabled and encryption type is set to `sse-kms`.
+* `canned_acl` - (Optional) Set canned access control list, e.g. bucket-owner-full-control. If `canned_cal` is set, the cluster instance profile must have `s3:PutObjectAcl` permission on the destination bucket and prefix. The full list of possible canned ACL can be found [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl). By default, only the object owner gets full control. If you are using cross-account role for writing data, you may want to set bucket-owner-full-control to make bucket owners able to read the logs.
```
alexott, line 158: "The full list of possible canned ACL" -> add s to ACL.

also wrap bucket-owner-full-control into backquotes: `bucket-owner-full-control`

docs/resources/cluster.md
```diff
@@ -182,11 +180,11 @@ init_scripts {
 }
 ```
 
-Attributes are the same as for `cluster_log_conf` configuration block.
+Attributes are the same as for the `cluster_log_conf` configuration block.
 
 ## aws_attributes
 
-`aws_attributes` optional configuration block contains attributes related to [clusters running on Amazon Web Services](https://docs.databricks.com/clusters/configure.html#aws-configurations). If not specified at cluster creation, a set of default values will be used. It is advised to keep all common configurations in [Cluster Policies](cluster_policy.md) to maintain control of the environments launched.
+`aws_attributes` optional configuration block contains attributes related to [clusters running on Amazon Web Services](https://docs.databricks.com/clusters/configure.html#aws-configurations). If not specified at cluster creation, a set of default values will be used.
```
alexott, line 187: Can we specify the partial list of the attributes? Is it all or nothing?  Also, documentation doesn't talk about specific configuration parameters, so it may not be clear what could be specified here.

docs/resources/cluster.md
```diff
@@ -213,10 +211,10 @@ The following options are available:
...
+* `first_on_demand` - (Optional) The first first_on_demand nodes of the cluster will be placed on on-demand instances. If this value is greater than 0, the cluster driver node will be placed on an on-demand instance. If this value is greater than or equal to the current cluster size, all nodes will be placed on on-demand instances. If this value is less than the current cluster size, first_on_demand nodes will be placed on on-demand instances, and the remainder will be placed on availability instances. This value does not affect cluster size and cannot be mutated over the lifetime of a cluster.
```
alexott, line 214: first_on_demand -> into backquotes

alexott (commented):
looks good, there are few small comments that would be nice to address

docs/resources/cluster.md
nfx: it does cover specific params below

PR #415, opened 2020-11-18
Experimental Resource Importer tool

nfx:
Generates HCL code according to schemas defined in resources and generates a file with shell commands to import the state with each individual resource

```                                                     
                                    +----------+      +--------------------+  
                                    | resource +------+ stateApproximation |  
                                    +--|-------+      +----|-------------|-+  
                                       |                   |             |    
  +------------------------+           |                   |             |    
  | "normal provider flow" |    +------|-----+       +-----|-----+   +---|---+
  +------------|-----------+    | importable --------- reference |   | scope |
               |                +------|-----+       +--------|--+   +---|---+
+--------------+--------------+        |                      |          |    
|terraform-provider-databricks|        |                      |          |    
+--------------+--------------+        |                      |          |    
               |                       |                   +--|----------|---+
    +----------|---------+        +----|------------+      |                 |
    |                    |        |                 |      |    Generated    |
    |  importer command  ----------  importContext  |      |      Files      |
    |                    |        |                 |      |                 |
    +--------------------+        +-----------------+      +-----------------+
```            

Invocation would be something like

```
./terraform-provider-databricks importer -directory /tmp/data-group -services identity
```


nfx:
@johnkeates @favoretti you may find this one useful

importer/importables.go
```diff
@@ -0,0 +1,483 @@
...
+			}
+			lastActiveMs := ic.lastActiveDays * 24 * 60 * 60 * 1000
+			for offset, c := range clusters {
+				if c.ClusterSource == "JOB" {
+					log.Printf("[INFO] Skipping job cluster %s", c.ClusterID)
+					continue
+				}
+				if c.ClusterID != "0328-184258-tempi2" {
+					continue
+				}
```
alexott, line 148: what is the magic ID here?

alexott (changes requested):
That's really huge help for customers who already have existing environments!

The only problem that I've found so far is the handling of the list of services to import...

importer/command.go
```diff
@@ -0,0 +1,68 @@
...
+		"Directory to generate sources in. Defaults to current directory.")
+	flags.Int64Var(&ic.lastActiveDays, "last-active-days", 3650,
+		"Items with older than activity specified won't be imported.")
+	flags.BoolVar(&ic.debug, "debug", false, "Print extra debug information")
+
+	services := ""
+	for _, ir := range ic.Importables {
+		if strings.Contains(services, ir.Service) {
+			continue
+		}
```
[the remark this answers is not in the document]
nfx: cluster and cluster policy are parts of `compute` service. e.g. package in provider. things will be documented.

importer/context.go
```diff
@@ -0,0 +1,499 @@
...
+	p := provider.DatabricksProvider()
+	return &importContext{
+		Module:      "",
+		Client:      c,
+		State:       stateApproximation{},
+		Importables: resourcesMap,
+		Resources:   p.ResourcesMap,
+		Files:       map[string]*hclwrite.File{},
+		Scope:       []*resource{},
+		Directory:   "/tmp/importer",
```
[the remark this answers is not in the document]
nfx: the idea here is that it's $CWD and folks generate code in that directory. I'll remove defaults from code in a bit.

importer/importables.go
nfx: will be gone

johnkeates:
@nfx Neat! Would make terraforming a lot easier considering the huge amount of existing resources.

nfx:
@johnkeates yes, you do have huge number of resources indeed. Do you want to give it test run or two next weeks? I'm testing it on internal demo workspace almost as big as yours, but willing to see what it will miss on other ones.

favoretti:
That is neat indeed. One remark though, by design TF providers are plugins, so shouldn't be called directly. Would a separate binary make sense here?

nfx:
@favoretti this is why i call this one "**experimental**". By design, with having single binary, i don't need to care about additional installation/downloading/repositories/approvals/processes/qa/cross-version compatibility issues that come with multiple binaries. it's isolated, anyway. Like microservices in monorepos :) 

Once Terraform would support this mechanism natively, i'll migrate to one.

importer/command.go
```diff
@@ -0,0 +1,82 @@
...
+	flags.StringVar(&ic.services, "services", services,
+		"Coma-separated list of services to import. By default all services are imported.")
+	flags.StringVar(&ic.listing, "listing", listing,
+		"Coma-separated list of services to be listed and further passed on for importing. "+
+			"`-services` parameter controls which transitive dependencies will be processed. "+
+			"We recommend limiting services with `-listing` more often, than `-services`.")
+	flags.StringVar(&ic.match, "match", "", "Match resource names during listing operation. "+
+		"This filter applies to all resources that are getting listed, so if you want to import "+
+		"all dependencies of just one cluster, specify -listing=compute")
+	err = flags.Parse(args)
```
alexott, line 74: Should we also allow to specify the profile name in the `~/.databrickscfg` ?

nfx: flags here have nothing to do with authentication. authentication is done purely from environment variables:

- https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs#environment-variables
- https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs#empty-provider-block

alexott: yep, my point was that sometimes it's easier to use command-line flags vs environment variables

nfx: @alexott it might be, but ~/.databrickscfg profile is one of 6 authentication ways we have here... if we support one through CLI flag, people would start opening bugs why other 5 don't work. Azure users are not very fond of `~/.databrickscfg` file, because they only use azure cli... besides, then we need to support another authentication conf verification layer, duplicate the docs, make sure that all of combinations work, fix bugs in those. 

to be honest, this is not something i'd like our resources to be invested in, especially for experimental feature 😄

alexott:
Some comments, just not to forget:

* I see multiple calls to the same object, like, pulling the `admins` group multiple times, secrets, etc. - maybe we can cache that
* `spark_options` could be formatted slightly better - right now they are shown as:

```
  spark_conf = {
  "spark.databricks.delta.preview.enabled" = "true" }
```

nfx:
@alexott pulling as in Emit or in Import? Emit has a call to Has, which prevents duplicate generation. At least it’s supposed to.

alexott:
More comments when using:
* `Error: resource databricks_dbfs_file doesn't support import`
* `Error: resource databricks_secret_acl doesn't support import`

nfx:
> More comments when using:
> 
> * `Error: resource databricks_dbfs_file doesn't support import`
> * `Error: resource databricks_secret_acl doesn't support import`

looks like you've got an old build - https://github.com/databrickslabs/terraform-provider-databricks/pull/405

importer/context.go
```diff
@@ -139,6 +140,26 @@ func (ic *importContext) Run() error {
 	// nolint
 	sh.WriteString("#!/bin/sh\n\n")
 
+	dcfile, err := os.Create(fmt.Sprintf("%s/databricks.tf", ic.Directory))
```
nfx, line 143: i think you've forgot to surround it with flag :)

alexott: everybody needs it! :-)

nfx: nope. that would break tf0.12

alexott: I think that we can generate it based on the Terraform version - I need to check if it's possible to get it from the inside of the provider

PR #443, opened 2020-12-15
Minor fixes to importer

pietern:
I tried using the importer against a set of mounts and needed this fix for it to work.


pietern:
cc @nfx

nfx (approved):

PR #465, opened 2021-01-24
Add global rate limit to HTTP calls

### LINKED ISSUE #156: [ISSUE] Add global rate limit to API calls
There needs to be a rate limit configuration present to limit number of requests to reasonable number per minute to ensure smooth operations. The most complicated part is testing.

nfx:
Fixes #156

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/time/rate"
+
 	"github.com/hashicorp/go-retryablehttp"
 	"github.com/mitchellh/go-homedir"
 	"gopkg.in/ini.v1"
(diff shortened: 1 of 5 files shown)


common/client.go
```diff
@@ -177,6 +181,10 @@ func (c *DatabricksClient) configureHTTPCLient() {
 	if c.TimeoutSeconds == 0 {
 		c.TimeoutSeconds = 60
 	}
+	if c.RateLimit == 0 {
+		c.RateLimit = 1200
```
alexott, line 185: maybe it makes sense to declare a constant somewhere, so it will be used in the provider conf as well?

alexott (approved):
lgtm

common/client.go
nfx: good idea

PR #470, opened 2021-01-26
Azure environment patch 1

stikkireddy:
The major change in this PR is only azure_auth, it makes changes to ensureWorkspaceUrl by using the azure.EnvironmentFromName generated Environment. It would be great if we can have this patch for 0.3.0.

Introduced an interface for getting the environment object to make use of the interface for mocking the environment for testing purposes.


common/azure_auth.go
```diff
@@ -276,14 +285,20 @@ func (aa *AzureAuth) ensureWorkspaceURL(ctx context.Context,
...
+		return err
 	}
+	managementUrl, err := url.Parse(env.ResourceManagerEndpoint)
+	if err != nil {
+		return fmt.Errorf("Somehow unable to parse correct azure environment")
+	}
+	managementUrl.Path = path.Join(managementUrl.Path, resourceID)
+	managementResourceUrl := managementUrl.String()
+
 	var workspace azureDatabricksWorkspace
```
nfx, line 299: Add it here:
```
azureEnv, err := aa.getAzureEnvironment()
...
endpoint  := azureEnv.ResourceManagerEndpoint()
```

common/azure_auth.go
```diff
@@ -68,19 +69,27 @@ type TokenInfo struct {
...
+type environmentFetcherInterface interface {
+	getAzureEnvironment(environment string) (azure.Environment, error)
+}
+
+type environmentFetcher struct{}
+
+func (e environmentFetcher) getAzureEnvironment(environment string) (azure.Environment, error) {
+	if environment == "" {
 		return azure.PublicCloud, nil
 	}
```
nfx, line 82: add
```
if aa.azureManagementEndpoint != "" {
   return azure.Environment{
       ResourceManagerEndpoint: aa.azureManagementEndpoint,
   }
}
```

nfx (changes requested):
Too many changes very late in the release :) and introducing an interface is rather unnecessary. Gave a hint how to make changes with smallest amount of the least risky code.

common/azure_auth_test.go
```diff
@@ -272,6 +272,23 @@ func TestAzureAuth_configureWithClientSecret(t *testing.T) {
 	assert.Len(t, zi.Zones, 3)
 }
 
+func TestAzureEnvironment_WithAzureManagementEndpoint(t *testing.T) {
```
nfx, line 275: This test is not testing anything new :) please remove

common/azure_auth.go
```diff
@@ -276,14 +279,19 @@ func (aa *AzureAuth) ensureWorkspaceURL(ctx context.Context,
 		return fmt.Errorf("Somehow resource id is not set")
 	}
 	log.Println("[DEBUG] Getting Workspace ID via management token.")
-	endpoint := "https://management.azure.com"
-	if aa.azureManagementEndpoint != "" {
-		// sets endpoint specified in unit test
-		endpoint = aa.azureManagementEndpoint
+	env, err := aa.getAzureEnvironment()
+	if err != nil {
+		return err
```
nfx, line 284: Please cover this line with a test

nfx (changes requested):
Remove unnecessary url parsing and add unit test for missed error case

common/azure_auth.go
```diff
@@ -276,14 +279,19 @@ func (aa *AzureAuth) ensureWorkspaceURL(ctx context.Context,
...
+	}
+	managementUrl, err := url.Parse(env.ResourceManagerEndpoint)
```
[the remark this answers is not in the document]
stikkireddy: so the issue is all urls end with a trailing slash, and in the future that may or may not change, and all resource ids begin with a slash so it breaks the unit tests. so i can remove this and use substrings if that is preferred. But either way there needs to be some logic here that fixes that.

common/azure_auth_test.go
stikkireddy: done

common/azure_auth.go
stikkireddy: added

common/azure_auth.go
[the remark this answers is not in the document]
stikkireddy: i opted to go with the native url parsing library that can resolve trailing slashes or no trailing slashes for urls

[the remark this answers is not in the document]
nfx: management point urls are hardcoded into Azure SDK. they will always end with a slash. let's not add complexity :)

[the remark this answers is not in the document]
stikkireddy: replaced the block with a strings.TrimSuffix for trimming trailing slashes on the url since it is hardcoded. Unit tests work.

common/azure_auth_test.go
```diff
@@ -167,7 +165,8 @@ func TestAzureAuth_ensureWorkspaceURL(t *testing.T) {
 	defer server.Close()
 
 	aa.ResourceID = "/subscriptions/a/resourceGroups/b/providers/Microsoft.Databricks/workspaces/c"
-	aa.azureManagementEndpoint = server.URL
+	// resource management endpoints end with a trailing slash in url
+	aa.azureManagementEndpoint = fmt.Sprintf("%s/", server.URL)
```
alexott, line 169: why not simple `server.URL + "/"` ?

common/azure_auth_test.go
```diff
@@ -249,7 +248,8 @@ func TestAzureAuth_configureWithClientSecret(t *testing.T) {
 	aa.ClientID = "a"
 	aa.ClientSecret = "b"
 	aa.TenantID = "c"
-	aa.azureManagementEndpoint = server.URL
+	// resource management endpoints end with a trailing slash in url
+	aa.azureManagementEndpoint = fmt.Sprintf("%s/", server.URL)
```
alexott, line 252: ditto

alexott (approved):
lgtm

nfx (approved):

common/azure_auth_test.go
stikkireddy: noticed it after it was merged. I use had more stuff there before forgot to change it to a +.

common/azure_auth_test.go
stikkireddy: noticed it after it was merged. I use had more stuff there before forgot to change it to a +.

PR #495, opened 2021-02-10
Fix token creation without comment or expiration

### LINKED ISSUE #494: [ISSUE] Error: requirement failed: Invalid timeout: 0 on token creation
### Terraform Version
Terraform 0.14.6
Databricks provider 0.3.0

### Affected Resource(s)
- databricks_token

```hcl
resource "databricks_token" "token" {}
```

### Debug Output
Cannot be disclosed publicly, mail me at _sam.debruyn at lynxcare.eu_

### Expected Behavior
Token created like it did in previous versions

### Actual Behavior
```
databricks_token.token: Creating...

Error: requirement failed: Invalid timeout: 0
Please consult API docs at https://docs.databricks.com/dev-tools/api/latest/token.html#create for details.
```

sdebruyn:
fixes #494 

- allow tokens without comment or expiration
- add acceptance test for tokens without comment & expiration

### DIFF
--- a/identity/resource_token.go
+++ b/identity/resource_token.go
@@ -14,8 +14,8 @@ import (
 
 // TokenRequest asks for a token
 type TokenRequest struct {
-	LifetimeSeconds int32  `json:"lifetime_seconds"`
-	Comment         string `json:"comment"`
+	LifetimeSeconds int32  `json:"lifetime_seconds,omitempty"`
+	Comment         string `json:"comment,omitempty"`
 }
 
 // TokenResponse is a struct that contains information about token that is created from the create tokens api
@@ -50,10 +50,16 @@ type TokensAPI struct {
 
 // Create creates a api token given a expiration duration and a comment
 func (a TokensAPI) Create(tokenLifetime time.Duration, comment string) (r TokenResponse, err error) {
-	err = a.client.Post(a.context, "/token/create", TokenRequest{
-		LifetimeSeconds: int32(tokenLifetime.Seconds()),
-		Comment:         comment,
-	}, &r)
+	request := TokenRequest{}
+	seconds := int32(tokenLifetime.Seconds())
+	if seconds > 0 {
+		request.LifetimeSeconds = seconds
+	}
+	if comment != "" {
+		request.Comment = comment
+	}
+
+	err = a.client.Post(a.context, "/token/create", request, &r)
 	return
 }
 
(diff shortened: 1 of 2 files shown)


identity/resource_token.go
```diff
@@ -50,10 +50,16 @@ type TokensAPI struct {
...
 func (a TokensAPI) Create(tokenLifetime time.Duration, comment string) (r TokenResponse, err error) {
-	err = a.client.Post(a.context, "/token/create", TokenRequest{
-		LifetimeSeconds: int32(tokenLifetime.Seconds()),
-		Comment:         comment,
-	}, &r)
+	request := TokenRequest{}
+	seconds := int32(tokenLifetime.Seconds())
+	if seconds > 0 {
+		request.LifetimeSeconds = seconds
+	}
```
nfx, line 57: can you introduce a constant in the top of the file, `DefaultTokenLifetimeSeconds`, and make it to 3600 and reference it here?...

identity/resource_token.go
```diff
@@ -98,13 +104,11 @@ func ResourceToken() *schema.Resource {
 			Type:     schema.TypeInt,
 			Optional: true,
 			ForceNew: true,
-			Default:  0,
 		},
 		"comment": {
 			Type:     schema.TypeString,
 			Optional: true,
 			ForceNew: true,
-			Default:  "",
```
nfx, line 107: Maybe just make it "Terraform" by default and have no if/else in the code?...

nfx (changes requested):
small comments regarding default values.

identity/resource_token.go
sdebruyn: I prefer not to.  How would you be able to create tokens without an expiration then? Seconds == 0 should mean that the token doesn't expire?

nfx: I think it will fail, no?

sdebruyn: No, I posted the API docs in #494, you should be able to create tokens without comments and without expiration

nfx: @sdebruyn Oh, okay. then i'll ask for one last thing - can you add `TestResourceTokenCreate_NoExpiration`, that would verify it on the resource level. I just suspect we'll have the same issue as with single node clusters. When i'll see that one new unit test (you've added integration test) with creating token with no arguments - i'll merge :)

identity/resource_token.go
sdebruyn: As discussed above, empty comments should be allowed

nfx (approved):

nfx:
@sdebruyn can you rebase it on master? or merge in the master? then good to go :)

sdebruyn:
@nfx done :)

PR #512, opened 2021-02-15
More coverage for exporter

nfx:
make tests run faster and add more coverage


exporter/util.go
```diff
@@ -142,18 +143,102 @@ func (ic *importContext) refreshMounts() error {
...
+import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+
+val readableMounts = dbutils.fs.mounts.par.map { mount =>
+try {
+	Await.result(Future {
+		dbutils.fs.ls(mount.mountPoint)
+		(mount.mountPoint
+			.replace("/mnt/", "")
```
alexott, line 217: it won't handle correctly the mount points that doesn't start with `/mnt`, especially built-in like, `/databricks-datasets`, etc.

```
[MountInfo(mountPoint='/databricks-datasets', source='databricks-datasets', encryptionType='sse-s3'),
 MountInfo(mountPoint='/databricks/mlflow-tracking', source='databricks/mlflow-tracking', encryptionType='sse-s3'),
 MountInfo(mountPoint='/databricks-results', source='databricks-results', encryptionType='sse-s3'),
 MountInfo(mountPoint='/databricks/mlflow-registry', source='databricks/mlflow-registry', encryptionType='sse-s3'),
 MountInfo(mountPoint='/', source='DatabricksRoot', encryptionType='sse-s3')]
```

alexott (changes requested):

exporter/util.go
nfx: but those are system-mounts, not customer-mounts

alexott: yep, but they are in the output of the `dbutils.fs.mounts`, so they will go into the result

nfx: `.filter(_.mountPoint.startsWith("/mnt"))` solves it :)

alexott (approved):
lgtm

PR #537, opened 2021-02-24
fix applying changes in absence of the email_notifications block

### LINKED ISSUE #530: [ISSUE] Plan shows change for email_notifications block in databricks_job when not set
### Terraform Version
Terraform v0.14.6
+ provider registry.terraform.io/databrickslabs/databricks v0.3.0

### Affected Resource(s)
- databricks_job

### Environment variable names
None

### Expected Behavior
Plan shows no change between applies when email_notification block is not provided

### Actual Behavior
Plan shows a change for email_notifications between applies when email_notifications block is not provided
```
  # module.databricks_jobs["data_export"].databricks_job.job will be updated in-place
  ~ resource "databricks_job" "job" {
        id                        = "XXX"
        name                      = "XXX"
        # (6 unchanged attributes hidden)

      - email_notifications {
          - no_alert_for_skipped_runs = false -> null
          - on_failure                = [] -> null
          - on_start                  = [] -> null
          - on_success                = [] -> null
        }



        # (3 unchanged blocks hidden)
    }
```

### Steps to Reproduce
1. Create a databricks_job resource with no email_notifications block
2. run `terraform apply`
3. run `terraform plan` or `terraform apply` to see the plan with erroneous change

alexott:
it also happened when `max_concurrent_runs` wasn't provided

this fixes #530

### DIFF
--- a/compute/resource_job.go
+++ b/compute/resource_job.go
@@ -126,6 +126,14 @@ var jobSchema = common.StructToSchema(JobSettings{},
 				return false
 			}
 		}
+		s["email_notifications"].DiffSuppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+			log.Printf("[INFO] k='%v', old='%v', new='%v'", k, old, new)
+			if old == "1" && new == "0" {
+				return true
+			}
+			return false
+		}
+
 		s["name"].Description = "An optional name for the job. The default value is Untitled."
 		s["library"].Description = "An optional list of libraries to be installed on " +
 			"the cluster that will execute the job. The default value is an empty list."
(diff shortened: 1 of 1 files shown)


compute/resource_job.go
```diff
@@ -126,6 +126,14 @@ var jobSchema = common.StructToSchema(JobSettings{},
 				return false
 			}
 		}
+		s["email_notifications"].DiffSuppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+			log.Printf("[INFO] k='%v', old='%v', new='%v'", k, old, new)
+			if old == "1" && new == "0" {
```
nfx, line 131: `no_alert_for_skipped_runs` is an integer?

alexott: No. This field is a list, so it receives length of that list, and diff on individual fields isn’t called (maybe it’s called if the predefined value changes, but it should be handled by default parameters).

nfx (approved):

PR #538, opened 2021-02-25
add name when importing secret scope

brandonmbanks:
Ran into an issue when importing the secret scope. The import succeeds but then the next plan wants to replace the secret scope.

Steps to recreate:
1. Create secret scope in Databricks
2. Import secret scope
3. Run plan

```
module.databricks_secret_scope.databricks_secret_scope.scope: Import prepared!
  Prepared databricks_secret_scope for import
module.databricks_secret_scope.databricks_secret_scope.scope: Refreshing state... [id=observability-keys]

Import successful!
```

But then the next plan failed because the name attribute was not set.
```
Error: Required attribute is not set

  on modules/secret_scope/main.tf line 8, in resource "databricks_secret" "secret":
  18:    scope        = databricks_secret_scope.scope.name
```
I updated the secret to reference `databricks_secret_scope.scope.id` and the plan succeeded but wanted to replace the secret scope.

```
  # module.databricks_secret_scope.databricks_secret_scope.scope must be replaced
-/+ resource "databricks_secret_scope" "scope" {
        ~ backend_type = "DATABRICKS" -> (known after apply)
        ~ id           = "observability-keys" -> (known after apply)
        + name         = "observability-keys" -> # forces replacement
    }
```

This fix just adds the `name` attribute to state so the next plan will not replace the secret scope.


nfx:
@brandonmbanks this PR won't fix this issue, as you've touched code generator, not resource importing part.

brandonmbanks:
@nfx Gotcha. Could you point me in the right direction?
Also is there a test for the import?

brandonmbanks:
I'm wondering why the refresh isn't completing the resource?
```
// The returned InstanceState only requires ID be set. Importing
// will always call Refresh after the state to complete it.
```

nfx:
@brandonmbanks importing happens [generically in common.Resourse](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/common/resource.go#L91), as making importers by hand for all 40 resources is a support nightmare :)  

so this generic importer calls the Read method of the resource: https://github.com/databrickslabs/terraform-provider-databricks/blob/master/access/resource_secret_scope.go#L145

though it is strange that `name` value was not imported. Please take a closer look with TF_LOG=DEBUG when you plan/apply.

technically, [read resource fixture](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/qa/testing.go#L93) is enough for testing the imports, but you may want to experiment with that.

[the change you've committed](https://github.com/databrickslabs/terraform-provider-databricks/pull/538/commits/353cbc22178afd13d99907ae2460e4a94f12da02) relates to [resource name generation](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/exporter/context.go#L354) in [very experimental resource exporter](https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs/guides/experimental-exporter)

brandonmbanks:
Ok thanks for the pointers. Back to the drawing board.

nfx (changes requested):
this change has no effect, please see pointers in the previous comment.

PR #545, opened 2021-03-02
[RESOURCE] databricks_sql_permissions to manage data object access control lists

nfx:
---
subcategory: "Security"
---
# databricks_sql_permissions Resource

-> **Note** This resource has an evolving API, which may change in the upcoming versions.

This resource manages data object access control lists in Databricks workspaces for things like tables, views, databases, and [more](https://docs.databricks.com/security/access-control/table-acls/object-privileges.html). In order to enable Table Access control, you have to login to the workspace as administrator, go to `Admin Console`, pick `Access Control` tab, click on `Enable` button in `Table Access Control` section, and click `Confirm`. The security guarantees of table access control **will only be effective if cluster access control is also turned on**. Please make sure that no users can create clusters in your workspace and all [databricks_cluster](cluster.md) have approximately the following configuration:

```hcl
resource "databricks_cluster" "cluster_with_table_access_control" {
  // ...

  spark_conf = {
    "spark.databricks.acl.dfAclsEnabled": "true",
    "spark.databricks.repl.allowedLanguages": "sql,python,r",
    "spark.databricks.cluster.profile": "serverless"
  }

  custom_tags = {
    "ResourceClass" = "Serverless"
  }
}  
```

## Example Usage

The following resource definition will enforce access control on a table by executing the following SQL queries on a special auto-terminating cluster it would create for this operation:

* ```SHOW GRANT ON TABLE `default`.`foo` ```
* ```REVOKE ALL PRIVILEGES ON TABLE `default`.`foo` FROM ... every group and user that has access to it ...```
* ```GRANT MODIFY, SELECT ON TABLE `default`.`foo` TO `serge@example.com` ```
* ```GRANT SELECT ON TABLE `default`.`foo` TO `special group` ```

```hcl
resource "databricks_sql_permissions" "foo_table" {
    table = "foo"

    privilege_assignments {
        principal = "serge@example.com"
        privileges = ["SELECT", "MODIFY"]
    }

    privilege_assignments {
        principal = "special group"
        privileges = ["SELECT"]
    }
}
```

## Argument Reference

The following arguments are available to specify the data object you need to enforce access controls on. You must specify only one of those arguments (except for `table` and `view`), otherwise resource creation will fail.

* `database` - Name of the database. Has default value of `default`.
* `table` - Name of the table. Can be combined with `database`. 
* `view` - Name of the view. Can be combined with `database`. 
* `catalog` - (Boolean) If this access control for the entire catalog. Defaults to `false`.
* `any_file` - (Boolean) If this access control for reading any file. Defaults to `false`.
* `anonymous_function` - (Boolean) If this access control for using anonymous function. Defaults to `false`.

### `privilege_assignments` blocks

You must specify one or many `privilege_assignments` configuration blocks to declare `privileges` to a `principal`, which corresponds to `display_name` of [databricks_group](group.md#display_name) or [databricks_user](user.md#display_name). Terraform would ensure that only those principals and privileges defined in the resource are applied for the data object and would remove anything else. It would not remove any transitive privileges. `DENY` statements are intentionally not supported. Every `privilege_assignments` has the following required arguments:

* `principal` - `display_name` of [databricks_group](group.md#display_name) or [databricks_user](user.md#display_name).
* `privileges` - set of available privilege names in upper case.

[Available](https://docs.databricks.com/security/access-control/table-acls/object-privileges.html) privilege names are:

* `SELECT` - gives read access to an object.
* `CREATE` - gives the ability to create an object (for example, a table in a database).
* `MODIFY` - gives the ability to add, delete, and modify data to or from an object.
* `USAGE` - do not give any abilities, but is an additional requirement to perform any action on a database object.
* `READ_METADATA` - gives the ability to view an object and its metadata.
* `CREATE_NAMED_FUNCTION` - gives the ability to create a named UDF in an existing catalog or database.
* `MODIFY_CLASSPATH` - gives the ability to add files to the Spark class path.
* `ALL PRIVILEGES` - gives all privileges (is translated into all the above privileges).

## Import

The resource can be imported using a synthetic identifier. Examples of valid synthetic identifiers are:

* `table/default.foo` - table `foo` in a `default` database. Database is always mandatory.
* `view/bar.foo` - view `foo` in `bar` database.
* `database/bar` - `bar` database.
* `catalog/` - entire catalog. `/` suffix is mandatory.
* `any file/` - direct access to any file. `/` suffix is mandatory.
* `anonymous function/` - anonymous function. `/` suffix is mandatory.

```bash
$ terraform import databricks_sql_permissions.foo /<object-type>/<object-name>
```


alexott (commented):
overall looks good, but test is still failing

nfx:
@alexott it's that SP test - started failing after ADAL updated to a more recent version. The problem is that it tries now to go and see if there's MSI endpoint available and fail otherwise. Does not really affect the normal case scenarios with Azure SP, though. Will test with `make test-azsp`.

access/resource_table_acl.go
```diff
@@ -0,0 +1,339 @@
...
+		if !strings.EqualFold(objectKey, key) {
+			continue
+		}
+		target := ta.Grants
+		var isDeny = strings.HasPrefix(actionType, "DENIED_")
+		if isDeny {
+			actionType = strings.Replace(actionType, "DENIED_", "", 1)
+			target = ta.Denies
+		}
+		var privileges *[]string
```
nfx, line 147: @alexott can you tell me if i'm not doing anything stupid here? you've worked with pointers more than me, so you know for sure :P

alexott: from looking to the code, it looks ok...

stikkireddy:
hey @nfx does this use a shared cluster and what is the throughput of the changes? Quick question also around does parallelism cause any funky behavior when you are applying grants/denys? Typically in notebook everything is run sequentially.

nfx:
@stikkireddy it would use the shared cluster (or we can specify one as parameter). resource will work on table level, so there should be no conflicts

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+		for i, tp := range target {
+			if tp.Principal == principal {
+				if isDeny {
+					privileges = &ta.Denies[i].Privileges
+				} else {
+					privileges = &ta.Grants[i].Privileges
+				}
+			}
+		}
+		if privileges == nil {
```
pietern, line 158: And an accompanying comment:
```
// Create new permission object if privileges for current principal were not found.
```

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+			return
+		}
+	}
+	clusterInfo, err := clustersAPI.StartAndGetInfo(ta.ClusterID)
+	if err != nil {
+		return
+	}
+	if v, ok := clusterInfo.SparkConf["spark.databricks.acl.dfAclsEnabled"]; !ok || v != "true" {
+		err = fmt.Errorf("cluster_id: not a High-Concurrency cluster: %s (%s)",
+			clusterInfo.ClusterName, clusterInfo.ClusterID)
```
pietern, line 240: What does high concurrency have to do with table ACLs?

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	ta.exec = c.CommandExecutor(ctx)
+	return nil
+}
+
+func (ta *TableACL) getOrCreateCluster(clustersAPI compute.ClustersAPI) (string, error) {
+	sparkVersion := clustersAPI.LatestSparkVersionOrDefault(compute.SparkVersionRequest{
+		Latest: true,
+	})
+	nodeType := clustersAPI.GetSmallestNodeType(compute.NodeTypeRequest{LocalDisk: true})
```
pietern, line 251: Local disk needed?

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+
+		// validateGrants := validation.StringInSlice([]string{
+		// 	"SELECT",
+		// 	"CREATE",
+		// 	"MODIFY",
+		// 	"READ_METADATA",
+		// 	"CREATE_NAMED_FUNCTION",
+		// }, false)
+		// s["grants"].ValidateFunc = validateGrants
+		// s["denies"].ValidateFunc = validateGrants
```
pietern, line 310: Fix?

pietern (commented):
Did a quick pass. Looks good!

access/resource_table_acl.go
nfx: You won't believe it - "when a high number of users use the same cluster concurrently, we can provide an enhanced security for that cluster using table acls". E.g. it is synonymous.

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+	})
+	nodeType := clustersAPI.GetSmallestNodeType(compute.NodeTypeRequest{LocalDisk: true})
+	aclCluster, err := clustersAPI.GetOrCreateRunningCluster("terrraform-table-acl", compute.Cluster{
+		ClusterName:            "terrraform-table-acl",
+		SparkVersion:           sparkVersion,
+		NodeTypeID:             nodeType,
+		AutoterminationMinutes: 10,
+		SparkConf: map[string]string{
+			"spark.databricks.acl.dfAclsEnabled": "true",
+			"spark.master":                       "local[*]",
```
alexott, line 259: I think that you also need: `"spark.databricks.cluster.profile" : "singleNode"` in the `SparkConf`: https://github.com/databrickslabs/terraform-provider-databricks/blob/master/storage/mounts.go#L126

docs/resources/table_acl.md
```diff
@@ -0,0 +1,83 @@
...
+    deny {
+        principal = "users"
+        privileges = ["SELECT", "READ"]
+    }
+}
+```
+
+## Argument Reference
+
+The following arguments are available to specify the data object you need to enfore access controls on. You must specify only one of those arguments (except for `table` and `view`), otherwise resource creation will fail.
```
alexott, line 43: `enforce` instead of `enfore`

docs/resources/table_acl.md
```diff
@@ -0,0 +1,83 @@
...
+* `database` - Name of the database. Has default value of `default`.
+* `table` - Name of the table. Can be combined with `database`. 
+* `view` - Name of the view. Can be combined with `database`. 
+* `catalog` - (Boolean) If this access control for the entire catalog. Defaults to `false`.
+* `any_file` - (Boolean) If this access control for reading any file. Defaults to `false`.
+* `anonymous_function` - (Boolean) If this access control for using anonymous function. Defaults to `false`.
+
+### `grant` and `deny` blocks
+
+You must specify one or many `grant` and/or `deny` configuration blocks to declare `privileges` to a `principal`, which corresponds to `display_name` of [databricks_group](group.md#display_name) or [databricks_user](user.md#display_name). Terraform would ensure that only those principals and priviliges defined in the resource are applied for the data object and would remove anything else. It would not remove any transitive privileges. Every `grant` or `deny` has the following required arguments:
```
alexott, line 54: `privileges` instead of `priviliges`

docs/resources/table_acl.md
```diff
@@ -0,0 +1,83 @@
...
+
+* `principal` - `display_name` of [databricks_group](group.md#display_name) or [databricks_user](user.md#display_name).
+* `privileges` - set of available privilege names in upper case.
+
+[Available](https://docs.databricks.com/security/access-control/table-acls/object-privileges.html) privilege names are:
+
+* `SELECT` - gives read access to an object.
+* `CREATE` - gives ability to create an object (for example, a table in a database).
+* `MODIFY` - gives ability to add, delete, and modify data to or from an object.
```
alexott, line 63: `give ability` -> `gives the ability`

docs/resources/table_acl.md
```diff
@@ -0,0 +1,83 @@
...
+* `USAGE` - does not give any abilities, but is an additional requirement to perform any action on a database object.
+* `READ_METADATA` - gives ability to view an object and its metadata.
+* `CREATE_NAMED_FUNCTION` - gives ability to create a named UDF in an existing catalog or database.
+* `MODIFY_CLASSPATH` - gives ability to add files to the Spark class path.
+* `ALL PRIVILEGES` - gives all privileges (is translated into all the above privileges).
```
alexott, line 68: same here

alexott (changes requested):

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+	if ta.Catalog {
+		return "CATALOG", ""
+	}
+	if ta.AnyFile {
+		return "ANY FILE", ""
+	}
+	if ta.AnonymousFunction {
+		return "ANONYMOUS FUNCTION", ""
+	}
+	return "", ""
```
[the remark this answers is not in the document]
nfx: it won't look well on the apply side.

access/resource_table_acl.go
```diff
@@ -0,0 +1,351 @@
...
+	ta.Grants = []TablePermissions{}
+	ta.Denies = []TablePermissions{}
+	var principal, actionType, objType, objectKey string
+	for result.Scan(&principal, &actionType, &objType, &objectKey) {
+		if !strings.EqualFold(objectType, objType) {
+			continue
+		}
+		if !strings.EqualFold(objectKey, key) {
+			continue
+		}
```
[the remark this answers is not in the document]
nfx: agreed, this method got too long for short variable names.

nfx:
```
resource "databricks_sql_permissions" "x" {
    privilege_assignments {
        principal = "serge@example.com"
        privileges = ["SELECT", "READ", "MODIFY"]
    }
}
```

adamcain-db:
I looked through the changes and no obvious issues jump out.  Thanks for this work, @nfx!

alexott (approved):
lgtm ;-)

PR #553, opened 2021-03-09
Resources for SQL Analytics queries and dashboards

pietern:
New resources are:
* `databricks_sql_query`
* `databricks_sql_visualization`
* `databricks_sql_dashboard`
* `databricks_sql_widget`


sqlanalytics/resource_visualization.go
```diff
@@ -0,0 +1,195 @@
...
+type Visualization struct {
+	ID          string `json:"id,omitempty" tf:"computed"`
+	QueryID     string `json:"query_id"`
+	Type        string `json:"type"`
+	Name        string `json:"name"`
+	Description string `json:"description,omitempty"`
+	Options     string `json:"options"`
+}
+
+type visualizationResource struct {
```
nfx, line 25: to be consistent with the rest of codebase, remove this struct and inline methods

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func (r *queryResource) delete(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+	aq, err := r.toAPIObject(d)
+	if err != nil {
+		return err
+	}
+
+	var w = api.NewWrapper(ctx, c)
+	err = w.DeleteQuery(aq)
```
nfx, line 321: refactor it to just `return api.NewWrapper(ctx, c).DeleteQuery(d.Id())`, don't make this method longer than it has to be :-)

sqlanalytics/api/query.go
```diff
@@ -0,0 +1,158 @@
...
+type QueryParameterEnum struct {
+	QueryParameter
+	Options string                               `json:"enumOptions"`
+	Multi   *QueryParameterMultipleValuesOptions `json:"multiValuesOptions,omitempty"`
+}
+
+// MarshalJSON sets the type before marshaling.
+func (p QueryParameterEnum) MarshalJSON() ([]byte, error) {
+	p.QueryParameter.Type = "enum"
+	type localQueryParameter QueryParameterEnum
```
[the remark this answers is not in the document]
pietern: Because we need the default `MarshalJSON` implementation to do the actual marshaling. We use it here to intercept marshaling, set the type field, and let the default marshaler do its job.

sqlanalytics/resource_visualization.go
```diff
@@ -0,0 +1,195 @@
...
+func ResourceVisualization() *schema.Resource {
+	r := visualizationResource{
+		common.StructToSchema(
+			Visualization{},
+			func(m map[string]*schema.Schema) map[string]*schema.Schema {
+				// We care only about logical changes to the JSON payload in `options`.
+				m["options"].DiffSuppressFunc = func(_, old, new string, d *schema.ResourceData) bool {
+					oldp, err := jsonRemarshal([]byte(old))
+					if err != nil {
+						panic(err)
```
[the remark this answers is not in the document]
pietern: I ran into issues here when creating new resources. The `old` string would be empty and the `new` string would contain a UUID (not valid JSON) and make the process crash. I wasn't able to figure out where this UUID came from. I suspect it has something to do with `computed` fields.

sqlanalytics/resource_visualization.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+
+	var w = api.NewWrapper(ctx, c)
+	avNew, err := w.CreateVisualization(av)
+	if err != nil {
+		return err
+	}
+
+	err = r.fromAPIObject(avNew, d)
```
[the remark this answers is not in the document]
pietern: Done.

sqlanalytics/resource_visualization.go
```diff
@@ -0,0 +1,195 @@
...
+func (r *visualizationResource) toAPIObject(d *schema.ResourceData) (*api.Visualization, error) {
+	var v Visualization
+
+	// Transform from ResourceData.
+	if err := common.DataToStructPointer(d, r.schema, &v); err != nil {
+		return nil, err
+	}
+
+	// Transform to API object.
+	var av api.Visualization
```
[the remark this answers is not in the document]
pietern: Done. This pattern is now applied across all new resources.

sqlanalytics/resource_visualization.go
```diff
@@ -0,0 +1,195 @@
...
+	"strings"
+
+	"github.com/databrickslabs/terraform-provider-databricks/common"
+	"github.com/databrickslabs/terraform-provider-databricks/sqlanalytics/api"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Visualization ...
+type Visualization struct {
+	ID          string `json:"id,omitempty" tf:"computed"`
```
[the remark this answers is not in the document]
pietern: Done.

sqlanalytics/resource_visualization.go
```diff
@@ -0,0 +1,195 @@
...
+					if err != nil {
+						panic(err)
+					}
+					return bytes.Compare(oldp, newp) == 0
+				}
+				return m
+			}),
+	}
+	return common.Resource{
+		Schema: r.schema,
```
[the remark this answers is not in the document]
pietern: Done.

sqlanalytics/resource_query_test.go
```diff
@@ -0,0 +1,310 @@
+package sqlanalytics
+
+import (
+	"testing"
+
+	"github.com/databrickslabs/terraform-provider-databricks/qa"
+	"github.com/databrickslabs/terraform-provider-databricks/sqlanalytics/api"
```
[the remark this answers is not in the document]
pietern: Discussed this over chat. There is duplication between the TF resource structs (suffix `Entity`) under the `sqlanalytics` package and the API structs, with subtle differences. These differences stem from the fact that the resources aren't an exact 1:1 copy of the API objects. Stuffing both in the same namespace will lead to confusion between TF resource structs and API structs.

sqlanalytics/resource_query.go
pietern: Done.

sqlanalytics/api/wrapper.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+// Wrapper is an API wrapper for the SQL Analytics queries, visualizations, and dashboards API.
+type Wrapper struct {
+	client  *common.DatabricksClient
+	context context.Context
+}
+
+// NewWrapper creates and returns an API wrapper.
+func NewWrapper(ctx context.Context, m interface{}) Wrapper {
```
nfx, line 19: Please split `Wrapper` into `VisualizationAPI`, `QueryAPI`, `WidgetAPI` structs and move entities and API wrappers to respective resource files, e.g. resource_widget.go will have `NewWidgetsAPI` factory, `WidgetsAPI` struct, `WidgetEntity` terraform entity for reflection, `Widget` api entity, and `ResourceWidget`. The main reason for this is that we have 82%+ code coverage requirement for all code that is written and `gotestsum` calculates coverage on the package level. I've figured it, when i was confident that [reflect_resource](https://app.codecov.io/gh/databrickslabs/terraform-provider-databricks/compare/553/tree/common/reflect_resource.go) is pretty well tested by other packages, but the coverage report didn't show any good results for it. [api package coverage](https://app.codecov.io/gh/databrickslabs/terraform-provider-databricks/compare/553/tree/sqlanalytics) is missing from the report, meaning that there are long-term risks of changes going untested to production. This may not be ideal, but it is consistent with the rest of the codebase. Even the special package, dedicated to [Accounts API only](https://github.com/databrickslabs/terraform-provider-databricks/tree/master/mws), follows the same conventions as packages that create resources within the workspace. _We may reconsider package layout conventions if we want to have databricks-go-sdk and refactor all 7 packages in a **separate** pull request, adjusting `resource.RetryContext` along the way._

nfx (changes requested):
This PR drops coverage by 19% - https://app.codecov.io/gh/databrickslabs/terraform-provider-databricks/compare/553/tree/sqlanalytics, please bring code coverage to the level of `resource_sql_endpoint.go`.

The other 3 things:
* comment in api package
* please add an acceptance test for each resource
* please add documentation for each resource

nfx:
acceptance tests should have `TestPreviewAcc` prefix, so that we can test those separately.

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+		return nil, err
+	}
+
+	return &qp, nil
+}
+
+// Update ...
+func (a QueryAPI) Update(q *api.Query) (*api.Query, error) {
+	var qp api.Query
+	err := a.client.Post(a.context, a.buildPath(q.ID), q, &qp)
```
nfx, line 415: i just wonder if we could have `q, &q`? e.g. if we cannot - then what is the reason? struct is the same, entity is the same, parameter accepts the pointer, so there's no immutability, that asks for a new instance in return 🤷

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+			// }
+		}
+		qp.Visualizations = []json.RawMessage{}
+	}
+
+	return &qp, err
+}
+
+// Read ...
+func (a QueryAPI) Read(q *api.Query) (*api.Query, error) {
```
nfx, line 402: named return arguments would remove the need to cover error return statements here with tests ;)

nfx: also.. it's very strange to Read with entire struct type to be passed. string ID must be enough to read a resource. (in terraform id is always a string)

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+func (a QueryAPI) buildPath(path ...string) string {
+	out := "/preview/sql/queries"
+	if len(path) == 1 {
+		out = out + "/" + strings.Join(path, "/")
+	}
+	return out
+}
+
+// Create ...
+func (a QueryAPI) Create(q *api.Query) (*api.Query, error) {
```
nfx, line 373: same question as about the Update

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+			default:
+				log.Fatalf("Don't know what to do for type: %#v", reflect.TypeOf(apv).String())
+			}
+
+			q.Parameter = append(q.Parameter, p)
+		}
+	}
+
+	// Transform to ResourceData.
+	if err := common.StructToData(*q, schema, data); err != nil {
```
nfx, line 346: why not immediately returning? :P

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+		aq.Options = &api.QueryOptions{}
+		for _, p := range q.Parameter {
+			ap := api.QueryParameter{
+				Name:  p.Name,
+				Title: p.Title,
+			}
+
+			var iface interface{}
+
+			switch {
```
nfx, line 169: and it's all covered with tests? O_O

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+	Value string `json:"value,omitempty"`
+	// Values iff `multiple != nil`
+	Values []string `json:"values,omitempty"`
+
+	QueryID  string                       `json:"query_id"`
+	Multiple *QueryParameterAllowMultiple `json:"multiple,omitempty"`
+}
+
+// QueryParameterDate ...
+type QueryParameterDate struct {
```
nfx, line 83: why do we need so many types with literally the same structure?...

nfx (changes requested):

sqlanalytics/resource_query.go
pietern: I used a new instance here to ensure you don't pass along input parameters that got removed as output.

sqlanalytics/resource_query.go
pietern: The sequence here was:
1. Need multiple IDs to read a visualization/widget.
2. Pass API object since it already captures both.
3. Use the same style everywhere and stick to it.

With the use of composite IDs in the latest commit, this is no longer needed, and read/update/delete take string parameters.

sqlanalytics/resource_query.go
pietern: Yes :)

sqlanalytics/resource_query.go
pietern: I started with a single struct per query parameter type.

Checked it out and was able to collapse the non-range and range date-like types into a single struct.

sqlanalytics/resource_query.go
```diff
@@ -0,0 +1,495 @@
...
+		Create: func(ctx context.Context, data *schema.ResourceData, c *common.DatabricksClient) error {
+			var q QueryEntity
+			aq, err := q.toAPIObject(s, data)
+			if err != nil {
+				return err
+			}
+
+			aqNew, err := NewQueryAPI(ctx, c).Create(aq)
+			if err != nil {
+				return err
```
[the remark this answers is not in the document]
pietern: Added for queries/dashboards. Doesn't work with composite IDs yet.

Submitting a PR allowing override of the ID so those can be tested for bad HTTP responses as well.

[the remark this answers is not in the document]
pietern: Turns out no PR needed as this function is only used in `sqlanalytics/`.

sqlanalytics/resource_widget.go
```diff
@@ -0,0 +1,312 @@
...
+	return a.client.Delete(a.context, fmt.Sprintf("/preview/sql/widgets/%s", widgetID), nil)
+}
+
+// ResourceWidget ...
+func ResourceWidget() *schema.Resource {
+	p := common.NewPairSeparatedID("dashboard_id", "widget_id", "/")
+	s := common.StructToSchema(
+		WidgetEntity{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["text"].ConflictsWith = []string{"visualization_id"}
```
nfx, line 246: don't forget to add ForceNew to dashboard_id & widget_id

nfx (approved):

sqlanalytics/resource_widget.go
pietern: Done for both widget and visualization (both use composite ID).

pietern:
Output of `make test-preview`:

```
preview
---
 * [ ] TestPreviewAccDashboard (593.86s)
 * [ ] TestPreviewAccPipelineResource_CreatePipeline (2.15s)
 * [ ] TestPreviewAccQuery (865.81s)
 * [ ] TestPreviewAccSQLEndpoint (519.09s)
 * [x] sqlanalytics.TestPreviewAccSQLEndpoints (443.40s)
```

The pipeline failure is a known problem (erroneous template).

The SQLA failures are all caused by a non-empty plan for the SQL endpoint:

```
    acceptance.go:160: Step 1/1 error: After applying this test step, the plan was not empty.
        stdout:


        An execution plan has been generated and is shown below.
        Resource actions are indicated with the following symbols:
          ~ update in-place

          # databricks_sql_endpoint.this will be updated in-place
          ~ resource "databricks_sql_endpoint" "this" {
              - auto_stop_mins   = 120 -> null
                id               = "024de5b61abeb9c3"
              - min_num_clusters = 1 -> null
                name             = "tf-thcbecidgbdii"
              - num_clusters     = 1 -> null
                # (6 unchanged attributes hidden)


              - tags {
                }
                # (1 unchanged block hidden)
            }

```

PR #554, opened 2021-03-09
Ensure required fields in nested objects are set

pietern:
We have polymorphic parameters in SQLA queries. For `text` and `number` parameters, there isn't a whole lot to configure, except for the default value (which is differently typed, depending on the parameter type). If the default value isn't set (or is set to be empty), the whole type wrapper object isn't set, and no longer shows up in `*schema.ResourceData`. I'm fixing this by making the default value a required field.

However, if the value was empty, it still didn't show up in `*schema.ResourceData`. This was caused by the reflection code ignoring all empty strings in nested types, instead of checking the schema first, as is done for top level fields.

Intuitively, if a field is required, it should always show up in `*schema.ResourceData`.

### DIFF
--- a/common/reflect_resource.go
+++ b/common/reflect_resource.go
@@ -263,6 +263,7 @@ func collectionToMaps(v interface{}, s *schema.Schema) ([]interface{}, error) {
 			path []string, valueField *reflect.Value) error {
 			fieldName := path[len(path)-1]
 			fieldValue := valueField.Interface()
+			fieldPath := strings.Join(path, ".")
 			switch fieldSchema.Type {
 			case schema.TypeList, schema.TypeSet:
 				nv, err := collectionToMaps(fieldValue, fieldSchema)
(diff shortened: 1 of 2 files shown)


nfx (approved):

pietern:
Well...

So `GetOkExists` is deprecated and `GetOk` checks if the value was ever non-zero...

pietern:
Discussion at https://github.com/hashicorp/terraform-plugin-sdk/pull/350

nfx (changes requested):
Build fails with

> Error: common/reflect_resource_test.go:361:12: SA1019: d.GetOkExists is deprecated: usage is discouraged due to undefined behaviors and may be removed in a future version of the SDK  (staticcheck)

pietern:
Added `nolint`. Let's see what happens on the tests...

nfx (approved):

PR #559, opened 2021-03-10
Fix databricks_permissions resource for SQLA entities

pietern:


pietern:
@nfx Looking at the test failures I think the Azure token has expired.

nfx:
@pietern not quite. Issue is in overriding HOME and PATH environment variables, I think. It cannot call az cli dummy.

All of the vendor dependencies are pinned, though some of the upstream ones may have done something to break the overriding part. Will have investigative cycles on Friday. By then I’ll just give you pointers to the right places.

nfx (changes requested):
On the second thought - could you add a unit test for one of SQLA permissions? we may forget about this subtle detail the next time we decide to refactor this complexity of a resource.

Code is not bad if it has a unit test (c)

pietern:
+1, will do.

nfx (approved):

PR #561, opened 2021-03-11
Troubleshoot failing az login on GitHub Actions

nfx:
TLDR: @alexott and @pietern , you didn't break the build, GitHub did.

* Github actions switched to Ubuntu 20.04, which they warned about few months ago.
* Github actions didn't use to have Az CLI installed
* Now AZ cli is installed by default in `/bin`
* `/bin` was removed from PATH mockery. 
* `az` dummy had to be rewritten to use full paths.
* fixed, tested and verified in Docker dev container. @alexandruanghel now you can finally contribute :P


alexott (approved):
lgtm

PR #570, opened 2021-03-22
Add Delta Pipelines Support

mgyucht:
This PR adds a `databricks_pipeline` resource to the Databricks Terraform provider.

Smaller changes:
* Updates to CONTRIBUTING.md with a walkthrough of how to add a new resource to the provider, complete with example code.
* Adding support for local files in init scripts for regular clusters.


compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+
+type PipelinesInitScriptInfo struct {
+	Dbfs *PipelinesDbfsStorageInfo `json:"dbfs,omitempty" tf:"optional"`
+	S3   *PipelinesS3StorageInfo   `json:"s3,omitempty" tf:"optional"`
+	File *PipelinesLocalFileInfo   `json:"file,omitempty" tf:"optional"`
+}
+
+type PipelinesClusterAttributes struct {
+	SparkConf        map[string]string        `json:"spark_conf,omitempty"`
+	AwsAttributes    *PipelinesAwsAttributes  `json:"aws_attributes,omitempty"`
```
alexott, line 49: Will we have attributes for other clouds as well?

alexott (changes requested):
Please add tests for new resource - both unit & acceptance.
Also, `make test` is failing.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	return LiveTablesAPI{m.(*common.DatabricksClient), ctx}
+}
+
+func (a LiveTablesAPI) Create(s PipelineSpec, allowDuplicateNames bool) (LiveTableId, error) {
+	var id LiveTableId
+	err := a.client.Post(a.ctx, "/pipelines", map[string]interface{}{
+		"spec":                  s,
+		"allow_duplicate_names": allowDuplicateNames,
+	}, &id)
+	return id, err
```
nfx, line 162: i'm wondering if we should add `a.waitForStart` method here, as general expectation for people, who use terraform is to have infrastructure, that works after apply. This is the only way to validate if code we've deployed for the pipeline is actually working or cluster config attributes are not broken.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// Live Tables API Client
+
+type LiveTablesAPI struct {
+	client *common.DatabricksClient
+	ctx    context.Context
+}
+
+type LiveTableId string
```
nfx, line 150: why do we need a type, if we don't overload it with additional methods? it results in overhead of casting it to string everywhere, without any added benefit.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	var id LiveTableId
+	err := a.client.Post(a.ctx, "/pipelines", map[string]interface{}{
+		"spec":                  s,
+		"allow_duplicate_names": allowDuplicateNames,
+	}, &id)
+	return id, err
+}
+
+func (a LiveTablesAPI) Read(id LiveTableId) (p PipelineInfo, err error) {
+	err = a.client.Get(a.ctx, "/pipelines/"+string(id), nil, &p)
```
nfx, line 166: will `/pipelines` eventually change to `/live-tables`?

go.mod
```diff
@@ -15,11 +15,18 @@ require (
 	github.com/hashicorp/hcl/v2 v2.9.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.4.0
 	github.com/mitchellh/go-homedir v1.1.0
+	github.com/onsi/ginkgo v1.8.0 // indirect
```
nfx, line 18: please remove any changes to go.mod/go.sum files, as they are not needed.

provider/provider.go
```diff
@@ -77,6 +77,8 @@ func DatabricksProvider() *schema.Provider {
 			"databricks_global_init_script": workspace.ResourceGlobalInitScript(),
 			"databricks_notebook":           workspace.ResourceNotebook(),
 			"databricks_workspace_conf":     workspace.ResourceWorkspaceConf(),
+
+			"databricks_live_table": compute.ResourceLiveTable(),
```
nfx, line 81: what do we have in databricks cli? `databricks pipeline deploy` or `databricks live-tables deploy`? preferably we should have consistent naming with databricks cli, when it's possible. technically, `databricks_live_table` would define 1+ table in the artifact, so i wonder about the naming.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	})
+}
+
+func (a LiveTablesAPI) Delete(id LiveTableId) error {
+	return a.client.Delete(a.ctx, "/pipelines/"+string(id), nil)
+}
+
+// Live Tables Resource Definition
+
+var pipelineSchema = common.StructToSchema(PipelineSpec{}, func(m map[string]*schema.Schema) map[string]*schema.Schema { return m })
```
nfx, line 183: we probably should define it inside of `ResourceLiveTable` function scope, as schema is not used elsewhere in the package.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+
+func ResourceLiveTable() *schema.Resource {
+	return common.Resource{
+		Schema:        pipelineSchema,
+		SchemaVersion: 2,
```
nfx, line 188: `SchemaVersion` is used when resource schema introduces incompatible changes from previous releases. this is a new resource, so it doesn't need `SchemaVersion` field :)

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+func NewLiveTablesAPI(ctx context.Context, m interface{}) LiveTablesAPI {
+	return LiveTablesAPI{m.(*common.DatabricksClient), ctx}
+}
+
+func (a LiveTablesAPI) Create(s PipelineSpec, allowDuplicateNames bool) (LiveTableId, error) {
+	var id LiveTableId
+	err := a.client.Post(a.ctx, "/pipelines", map[string]interface{}{
+		"spec":                  s,
+		"allow_duplicate_names": allowDuplicateNames,
+	}, &id)
```
nfx, line 161: it really does return just a string? :) wow.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	Exclusions  []string `json:"exclusions,omitempty"`
+}
+
+type NotebookLibrary struct {
+	Path string `json:"path"`
+}
+
+type PipelineLibrary struct {
+	Jar      string                 `json:"jar,omitempty"`
+	Maven    *PipelinesMavenLibrary `json:"maven,omitempty"`
```
nfx, line 91: definitely reuse `Maven` type

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+type PipelinesClusterAttributes struct {
+	SparkConf        map[string]string        `json:"spark_conf,omitempty"`
+	AwsAttributes    *PipelinesAwsAttributes  `json:"aws_attributes,omitempty"`
+	NodeTypeId       string                   `json:"node_type_id,omitempty"`
+	DriverNodeTypeId string                   `json:"driver_node_type_id,omitempty"`
+	SshPublicKeys    string                   `json:"ssh_public_keys,omitempty"`
+	CustomTags       map[string]string        `json:"custom_tags,omitempty"`
+	ClusterLogConf   *PipelinesClusterLogConf `json:"cluster_log_conf,omitempty"`
```
nfx, line 54: reuse `StorageInfo`

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	S3   *PipelinesS3StorageInfo   `json:"s3,omitempty" tf:"optional"`
+	File *PipelinesLocalFileInfo   `json:"file,omitempty" tf:"optional"`
+}
+
+type PipelinesClusterAttributes struct {
+	SparkConf        map[string]string        `json:"spark_conf,omitempty"`
+	AwsAttributes    *PipelinesAwsAttributes  `json:"aws_attributes,omitempty"`
+	NodeTypeId       string                   `json:"node_type_id,omitempty"`
+	DriverNodeTypeId string                   `json:"driver_node_type_id,omitempty"`
+	SshPublicKeys    string                   `json:"ssh_public_keys,omitempty"`
```
nfx, line 52: shouldn't it be `[]string`?..

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	KmsKey           string `json:"kms_key,omitempty" tf:"optional"`
+	CannedAcl        string `jsono:"canned_acl,omitempty" tf:"optional"`
+}
+
+type PipelinesClusterLogConf struct {
+	Dbfs *PipelinesDbfsStorageInfo `json:"dbfs,omitempty" tf:"optional"`
+	S3   *PipelinesS3StorageInfo   `json:"s3,omitempty" tf:"optional"`
+}
+
+type PipelinesInitScriptInfo struct {
```
nfx, line 41: type is unused. nevertheless, we should use `StorageInfo` instead.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	NodeTypeId       string                   `json:"node_type_id,omitempty"`
+	DriverNodeTypeId string                   `json:"driver_node_type_id,omitempty"`
+	SshPublicKeys    string                   `json:"ssh_public_keys,omitempty"`
+	CustomTags       map[string]string        `json:"custom_tags,omitempty"`
+	ClusterLogConf   *PipelinesClusterLogConf `json:"cluster_log_conf,omitempty"`
+	SparkEnvVars     map[string]string        `json:"spark_env_vars,omitempty"`
+	InstancePoolId   string                   `json:"instance_pool_id,omitempty"`
+}
+
+type PipelinesAutoScale struct {
```
nfx, line 59: reuse AutoScale

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	MaxWorkers int `json:"max_workers"`
+}
+
+type PipelinesClusterSize struct {
+	NumWorkers int                 `json:"num_workers,omitempty"`
+	Autoscale  *PipelinesAutoScale `json:"autoscale,omitempty"`
+}
+
+type PipelinesNewCluster struct {
+	PipelinesClusterAttributes
```
nfx, line 70: interesting idea about merging types. i wonder how would it affect http fixtures when testing it all.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	PipelinesClusterSize
+}
+
+type PipelineCluster struct {
+	Label string `json:"label,omitempty"`
+	PipelinesNewCluster
```
nfx, line 76: why do we need to create many types without extra methods, if they are used only in one place?...

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	Include []string `json:"include,omitempty"`
+	Exclude []string `json:"exclude,omitempty"`
+}
+
+type PipelineSpec struct {
+	Id            string            `json:"id,omitempty"`
+	Name          string            `json:"name,omitempty"`
+	Storage       string            `json:"storage,omitempty"`
+	Configuration map[string]string `json:"configuration,omitempty"`
+	Clusters      []PipelineCluster `json:"clusters,omitempty"`
```
nfx, line 106: i'd advise to use `Cluster` type here and delete all fields that are not supported.

compute/resource_live_table.go
```diff
@@ -0,0 +1,221 @@
...
+	Coordinates string   `json:"coordinates"`
+	Repo        string   `json:"repo,omitempty"`
+	Exclusions  []string `json:"exclusions,omitempty"`
+}
+
+type NotebookLibrary struct {
+	Path string `json:"path"`
+}
+
+type PipelineLibrary struct {
```
nfx, line 89: if not for the `NotebookLibrary`, i'd ask to reuse `Library` type.

nfx (changes requested):

compute/resource_live_table.go
mgyucht: True. Wasn't sure how much we would benefit from type safety when I started, to prevent this API being called with another kind of ID. but now I see it is mainly self-contained, so can go back to plain string.

compute/resource_live_table.go
mgyucht: Interesting idea. We may need to see about paused/idle pipelines, but I could get behind this.

compute/resource_live_table.go
mgyucht: Not sure, asked: https://databricks.slack.com/archives/CE2JKAKQC/p1616667231083300

provider/provider.go
mgyucht: Currently, it is `pipelines`, but the plan is to rename it. I suppose we can start with this. We can always register another provider name with the same resource definition.

compute/resource_live_table.go
mgyucht: Good suggestion :)

compute/resource_live_table.go
mgyucht: Thanks, wasn't sure if it had to do with the api/2.0

compute/resource_live_table.go
mgyucht: nope, I'm just dumb.

compute/resource_live_table.go
mgyucht: It will not.

go.mod
mgyucht: removed.

provider/provider.go
mgyucht: Final answer: the resource will be called `pipeline`, and it will manage one or many live tables.

compute/resource_pipeline.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	log.Printf("[DEBUG] Pipeline %s is in state %s", id, *lastState)
+	if *lastState == desiredState {
+		return nil
+	}
+	return resource.RetryableError(fmt.Errorf("Pipeline %s is in state %s, not yet in state %s", id, *lastState, desiredState))
+}
+
+func (a pipelinesAPI) waitForState(id string, desiredState PipelineState) (PipelineState, error) {
+	var lastState PipelineState
```
nfx, line 157: please remove this lastState, as you're ignoring the result anyway :) and please merge retryFunc and waitForState - it would be more readable that way. 

please also add `timeout time.Duration` argument, as you'll have to propagate down a user-specified timeout as `d.Timeout(schema.TimeoutCreate)`. System default is 20 minutes. You can look at `ResourceWorkspace` for an example. That one is probably the most readable, along with sql endpoint.

compute/resource_pipeline.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (a pipelinesAPI) defaultTimeout() time.Duration {
+	return 30 * time.Minute
+}
+
+func (a pipelinesAPI) retryFunc(id string, desiredState PipelineState, lastState *PipelineState) *resource.RetryError {
+	s, err := a.read(id)
+	if err != nil {
+		if e, ok := err.(common.APIError); ok && e.ErrorCode == "RESOURCE_DOES_NOT_EXIST" {
+			*lastState = StateDeleted
```
nfx, line 142: once you remove `lastState`, you should simplify these if conditions and don't handle the 404 here (it's properly handled in CommonResource anyway)

provider/provider.go
```diff
@@ -77,6 +77,8 @@ func DatabricksProvider() *schema.Provider {
 			"databricks_global_init_script": workspace.ResourceGlobalInitScript(),
 			"databricks_notebook":           workspace.ResourceNotebook(),
 			"databricks_workspace_conf":     workspace.ResourceWorkspaceConf(),
+
+			"databricks_pipeline": compute.ResourcePipeline(),
```
nfx, line 81: can you move it near all `compute` resources and place it alphabetically?

nfx (changes requested):

compute/resource_pipeline.go
```diff
@@ -0,0 +1,228 @@
...
+			err := common.DataToStructPointer(d, pipelineSchema, &s)
+			if err != nil {
+				return err
+			}
+			api := newPipelinesAPI(ctx, c)
+			id, err := api.create(s, false)
+			if err != nil {
+				return err
+			}
+			_, err = api.waitForState(id, StateRunning)
```
nfx, line 196: if rewritten this way, it's a bit more readable:

```go
if err = api.waitForState(id, StateRunning); err != nil {
  return err
}
```

nfx: i usually prefer those for statements, that have only an error in return.

compute/resource_pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+				return err
+			}
+			return newPipelinesAPI(ctx, c).update(d.Id(), s, false)
+		},
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			api := newPipelinesAPI(ctx, c)
+			if err := api.delete(d.Id()); err != nil {
+				return err
+			}
+			return api.waitForDeleted(d.Id(), d.Timeout(schema.TimeoutDelete))
```
nfx, line 225: on the second thought - can you put `waitForDeleted` code into `delete` method? this way we always have api-wrapper-level consistency for deleting the entity. which will make a good GoSDK in the future if we decide to

compute/resource_pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+			err := common.DataToStructPointer(d, pipelineSchema, &s)
+			if err != nil {
+				return err
+			}
+			api := newPipelinesAPI(ctx, c)
+			id, err := api.create(s, false)
+			if err != nil {
+				return err
+			}
+			if err = api.waitForState(id, d.Timeout(schema.TimeoutCreate), StateRunning); err != nil {
```
nfx, line 200: and can you put `waitForState` into `create` method itself. TF resource would be thinner this way and we can do more integration tests in the future.

compute/resource_pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+			return resource.RetryableError(fmt.Errorf("Pipeline %s is in state %s, not yet in state %s", id, state, desiredState))
+		})
+}
+
+func (a pipelinesAPI) waitForDeleted(id string, timeout time.Duration) error {
+	return resource.RetryContext(a.ctx, timeout,
+		func() *resource.RetryError {
+			i, err := a.read(id)
+			if err != nil {
+				if e, ok := err.(common.APIError); ok && e.IsMissing() {
```
nfx, line 156: technically, we can simplify this the following way:

* move `IsMissing` check to `delete` method
* just return error as non-retryable.

when you return a retryable error, please also log something, as this will tell the one debugging it with TF_LOG=DEBUG that something is still working.

compute/resource_pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+			state := *i.State
+			if state == desiredState {
+				return nil
+			}
+			return resource.RetryableError(fmt.Errorf("Pipeline %s is in state %s, not yet in state %s", id, state, desiredState))
+		})
+}
+
+func (a pipelinesAPI) waitForDeleted(id string, timeout time.Duration) error {
+	return resource.RetryContext(a.ctx, timeout,
```
nfx, line 152: i'd advice to move this logic to `delete` method

compute/resource_pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (a pipelinesAPI) update(id string, s pipelineSpec, allowDuplicateNames bool) error {
+	return a.client.Put(a.ctx, "/pipelines/"+id, s)
+}
+
+func (a pipelinesAPI) delete(id string) error {
+	return a.client.Delete(a.ctx, "/pipelines/"+id, map[string]string{})
+}
+
+func (a pipelinesAPI) waitForState(id string, timeout time.Duration, desiredState PipelineState) error {
```
nfx, line 136: i'd advise to move this logic to `create` method.

nfx (changes requested):

compute/resource_pipeline.go
mgyucht: Oh interesting, that message is not logged?

nfx: so that message is logged as http request on DEBUG level, but i think every state await should be logged as INFO as well.

nfx (approved):

PR #571, opened 2021-03-23
Support for spot instances on Azure

alexott:
This PR adds support for Azure spot instances for clusters & instance pools. It also adds an additional support for preemptible executors on GCP (only for clsuters)

### DIFF
--- a/compute/resource_cluster.go
+++ b/compute/resource_cluster.go
@@ -68,8 +68,9 @@ func resourceClusterSchema() map[string]*schema.Schema {
 			Optional: true,
 			Computed: true,
 		}
-		s["aws_attributes"].ConflictsWith = []string{"azure_attributes"}
-		s["azure_attributes"].ConflictsWith = []string{"aws_attributes"}
+		s["aws_attributes"].ConflictsWith = []string{"azure_attributes", "gcp_attributes"}
+		s["azure_attributes"].ConflictsWith = []string{"aws_attributes", "gcp_attributes"}
+		s["gcp_attributes"].ConflictsWith = []string{"aws_attributes", "azure_attributes"}
 		s["is_pinned"] = &schema.Schema{
 			Type:     schema.TypeBool,
 			Optional: true,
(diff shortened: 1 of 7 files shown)


compute/resource_cluster.go
```diff
@@ -321,6 +324,12 @@ func modifyClusterRequest(clusterModel *Cluster) {
 		}
 		clusterModel.AwsAttributes = &awsAttributes
 	}
+	if clusterModel.AzureAttributes != nil {
+		clusterModel.AzureAttributes = nil
```
nfx, line 328: nit: can we add a unit test here

compute/resource_cluster.go
```diff
@@ -321,6 +324,12 @@ func modifyClusterRequest(clusterModel *Cluster) {
...
+	}
+	if clusterModel.GcpAttributes != nil {
+		clusterModel.GcpAttributes = nil
```
nfx, line 331: nit: can we add a unit test here

compute/resource_instance_pool.go
```diff
@@ -61,6 +61,8 @@ func ResourceInstancePool() *schema.Resource {
 		s["custom_tags"].ForceNew = true
 		s["enable_elastic_disk"].ForceNew = true
 		s["enable_elastic_disk"].Default = true
+		s["aws_attributes"].ConflictsWith = []string{"azure_attributes"}
+		s["azure_attributes"].ConflictsWith = []string{"aws_attributes"}
```
nfx, line 65: no gcp attributes for instance pools?

nfx (approved):

compute/resource_instance_pool.go
alexott: there is no instance pools support yet, so I can't test it - will add later

compute/resource_cluster.go
alexott: let me add it...

alexott: added

compute/resource_cluster.go
nfx: But does it make sense to nil the field if it's not nil?..

alexott: all Azure attributes now are related to the Spot instances only, and if we're using instance pools, then we shouldn't override instance types?

nfx (approved):

PR #575, opened 2021-03-24
added support for job to have pause_status

stikkireddy:
Changed pause_status from the schedule config block in jobs to a optional+computed field rather than a computed field (least amount of disruption to existing state). 
(Users can choose to omit it if the do not want terraform to nitpick that it has changed. Changing it to just optional will give the behavior of always tracking state and always must be managed by terraform.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 * Fix incorrect escaping of notebook names ([#566](https://github.com/databrickslabs/terraform-provider-databricks/pull/566))
 * Added support for spot instances on Azure ([#571](https://github.com/databrickslabs/terraform-provider-databricks/pull/571))
 
+**Behavior changes**
+
+* Job schedules now have their schedule configuration block to support pause_status as a optional field. ([#575](https://github.com/databrickslabs/terraform-provider-databricks/pull/575))
+
 ## 0.3.1
 
 * Added `databricks_global_init_script` resource to configure global init scripts ([#487](https://github.com/databrickslabs/terraform-provider-databricks/issues/487)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,7 @@
 
 * Fix incorrect escaping of notebook names ([#566](https://github.com/databrickslabs/terraform-provider-databricks/pull/566))
 * Added support for spot instances on Azure ([#571](https://github.com/databrickslabs/terraform-provider-databricks/pull/571))
-
-**Behavior changes**
-
-* Job schedules now have their schedule configuration block to support pause_status as a optional field. ([#575](https://github.com/databrickslabs/terraform-provider-databricks/pull/575))
+* Job schedules support pause_status as a optional field. ([#575](https://github.com/databrickslabs/terraform-provider-databricks/pull/575))
 
 ## 0.3.1
 
--- a/compute/model.go
+++ b/compute/model.go
@@ -520,7 +520,7 @@ type JobEmailNotifications struct {
 type CronSchedule struct {
 	QuartzCronExpression string `json:"quartz_cron_expression"`
 	TimezoneID           string `json:"timezone_id"`
-	PauseStatus          string `json:"pause_status,omitempty" tf:"computed"`
+	PauseStatus          string `json:"pause_status,omitempty"`
 }
 
 // JobSettings contains the information for configuring a job on databricks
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 * Fix incorrect escaping of notebook names ([#566](https://github.com/databrickslabs/terraform-provider-databricks/pull/566))
 * Added support for spot instances on Azure ([#571](https://github.com/databrickslabs/terraform-provider-databricks/pull/571))
 
+**Behavior changes**
```
nfx, line 8: This is not a behavior change, just field that is exposed now

compute/model.go
```diff
@@ -520,7 +520,7 @@ type JobEmailNotifications struct {
 type CronSchedule struct {
 	QuartzCronExpression string `json:"quartz_cron_expression"`
 	TimezoneID           string `json:"timezone_id"`
-	PauseStatus          string `json:"pause_status,omitempty" tf:"computed"`
```
nfx, line 523: What happens if you keep tf computed flag? I'd rather keep the schema compatible

compute/resource_job.go
```diff
@@ -115,6 +115,14 @@ var jobSchema = common.StructToSchema(JobSettings{},
 			p.Required = false
 		}
 
+		if p, err := common.SchemaPath(s, "schedule", "pause_status"); err == nil {
+			p.Optional = true
```
nfx, line 119: Omitempty is setting optional, so this line is not needed

compute/resource_job.go
```diff
@@ -115,6 +115,14 @@ var jobSchema = common.StructToSchema(JobSettings{},
...
+			p.Computed = true
+			p.Type = schema.TypeString
+			p.Required = false
```
nfx, line 122: Boolean is false by default so this line is redundant

compute/resource_job.go
```diff
@@ -115,6 +115,14 @@ var jobSchema = common.StructToSchema(JobSettings{},
 			p.Required = false
 		}
 
+		if p, err := common.SchemaPath(s, "schedule", "pause_status"); err == nil {
+			p.Optional = true
+			p.Computed = true
```
nfx, line 120: Tf:computed sets it

compute/resource_job.go
```diff
@@ -115,6 +115,14 @@ var jobSchema = common.StructToSchema(JobSettings{},
...
+			p.Type = schema.TypeString
```
nfx, line 121: Type is discovered by reflect_resource already

nfx (changes requested):
Minor edits required

CHANGELOG.md
stikkireddy: changed it to just a line entry

docs/resources/job.md
```diff
@@ -69,6 +69,7 @@ The following arguments are required:
 
 * `quartz_cron_expression` - (Required) A [Cron expression using Quartz syntax](http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html) that describes the schedule for a job. This field is required.
 * `timezone_id` - (Required) A Java timezone ID. The schedule for a job will be resolved with respect to this timezone. See Java TimeZone for details. This field is required.
+* `pause_status` - (Optional) Indicate whether this schedule is paused or not. Either “PAUSED” or “UNPAUSED”.
```
alexott, line 72: maybe write `UNPAUSED (default)`?

alexott (approved):
just minor comment

docs/resources/job.md
nfx: is it default, though?

nfx (approved):

docs/resources/job.md
stikkireddy: hmm... @alexott i dont think we can provide a default if the field itself is computed. So in theory we dont provide a value when we make the rest call. So it is what ever our rest api chooses to do when that field is not omitted. And our rest api docs dont seem to indicate that there is a default. Should I check with them to verify if there is a built in default?

alexott: maybe I incorrectly formulated - we need to say that this specific will be used if it's omitted. Of course it's enforced on API level... but that was just a comment

stikkireddy: ahh gotcha i can add the create behavior if it is omitted.

compute/model.go
stikkireddy: removed

docs/resources/job.md
stikkireddy: added this: `When the pause_status field is omitted and a schedule is provided, the server will default to using "UNPAUSED" as a value for pause_status.`

alexott (approved):

PR #587, opened 2021-04-08
[DOCS] Move non-auth provider configs into separate section

alexott:
also documented missing parameters


docs/index.md
```diff
@@ -239,10 +239,18 @@ resource "databricks_user" "my-user" {
...
 * `pat_token_duration_seconds` - The current implementation of the azure auth via sp requires the provider to create a temporary personal access token within Databricks. The current AAD implementation does not cover all the APIs for Authentication. This field determines the duration in which that temporary PAT token is alive. It is measured in seconds and will default to `3600` seconds. 
+
+There are multiple environment variable options, the `DATABRICKS_AZURE_*` environment variables take precedence, and the `ARM_*` environment variables provide a way to share authentication configuration using the `databricks` provider alongside the `azurerm` provider.
+
+## Miscellaneous configuration parameters
+
+This section covers configuration parameters not related to authentication.  They could be used when debugging problems, or do an additional tuning of provider's behaviour:
+
+* `rate_limit` - defines maximum number of requests per second made to Databricks REST API by Terraform. Default is *15*.
```
nfx, line 249: This is the case where I'm not sure we should surface `rate_limit` and `skip_verify` in public documentation. @mgyucht @pietern for second opinion on this :)

nfx (approved):

docs/index.md
pietern: It's an advanced knob though, so it should be marked as such, like the others in this section.

PR #611, opened 2021-04-26
Fix handling of empty blocks for clusters/jobs/instance pools

alexott:
We don't need to specify empty blocks for aws_attributes/azure_attributes/gcp_attributes to prevent edit/re-creation of the clusters/jobs/instance pools.

For instance pools even specification of empty blocks didn't work because we were need to specify availability & other attributes as well to prevent recreation of objects

### DIFF
--- a/common/version.go
+++ b/common/version.go
@@ -3,7 +3,7 @@ package common
 import "context"
 
 var (
-	version = "0.3.3"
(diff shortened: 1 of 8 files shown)


nfx (approved):
This is great

PR #613, opened 2021-04-27
chore(docs): fix link to point to correct path

steve148:
Minor change, fixes a link in the docs to go to the correct link. Looking through git history, doesn't look like `data-source` ever existed as a directory under `docs`, so probably just a spelling mistake.


nfx (approved):
Thanks for these even minor changes to docs. What might have happened is that Google docs or Grammarly spellcheckers did wrongfully autocorrect those in the first doc proofreading batch.

PR #615, opened 2021-04-28
Bucket region cannot be set by specifying region

tomarv2:
Tried this, we cannot set bucket `region`, it throws an error, also changed the bucket name for end users understanding:

```
38:   region = "eu-west-1"

Computed attributes cannot be set, but a value was set for "region".
```


nfx:
@tomarv2 can you look at couple more places where `aws_s3_bucket` is defined and also change those there?

tomarv2:
> @tomarv2 can you look at couple more places where `aws_s3_bucket` is defined and also change those there?

verified there are 10 references to s3 resource and all look ok except this. Will keep a check if I notice something.

nfx (approved):

PR #621, opened 2021-04-29
Check for mount readability on Refresh

nfx:


nfx:
this may create more problems that benefits, removing it.

PR #631, opened 2021-05-03
chore(docs): only one runtime per pool

steve148:
Another day, another PR.

Didn't trigger the issue, but I was looking at the [instance pool docs](https://docs.databricks.com/clusters/instance-pools/configure.html#preload-databricks-runtime-version) and the [instance pool create  endpoint](https://docs.databricks.com/dev-tools/api/latest/instance-pools.html#create) and noticed the documentation there specified that only 1 spark version was allowed to be preloaded onto a machine. This PR updates the line related to the preloaded_spark_versions to match the databricks docs.

I wonder if there's a way to make sure the databricks docs and these docs stay aligned. It's not a big deal, but it can lead to confusion when there's differences between the two.


docs/resources/instance_pool.md
```diff
@@ -45,8 +45,7 @@ The following arguments are required:
...
 * `enable_elastic_disk` - (Optional) (Bool) Autoscaling Local Storage: when enabled, the instances in the pool dynamically acquire additional disk space when they are running low on disk space.
-
-* `preloaded_spark_versions` - (Optional) (List) A list with the runtime version the pool installs on each instance. Pool clusters that use a preloaded runtime version start faster as they do have to wait for the image to download.  You can retrieve them via [databricks_spark_version](../data-sources/spark-version.md) data source or via  [Runtime Versions API](https://docs.databricks.com/dev-tools/api/latest/clusters.html#clusterclusterservicelistsparkversions) call.
+* `preloaded_spark_versions` - (Optional) (List) A list with at most one runtime version the pool installs on each instance. Pool clusters that use a preloaded runtime version start faster as they do have to wait for the image to download. You can retrieve them via [databricks_spark_version](../data-sources/spark-version.md) data source or via  [Runtime Versions API](https://docs.databricks.com/dev-tools/api/latest/clusters.html#clusterclusterservicelistsparkversions) call.
```
steve148, line 48: I removed the newline above so the changes isn't as clear, but in essence I changed the following.

```diff
- A list with the runtime version the pool installs on each instance.
+ A list with at most one runtime version the pool installs on each instance.
```

ileshko: `as they do have` - > `as they do not have`

steve148: Change made. Note that the databricks docs also have this issue, but I don't think there's a place I can contribute a fix to.

Screen Shot 2021-05-03 at 1 49 48 PM

steve148:
Looks like the checks are stuck for LGTM. Otherwise I think the change is good to go.

PR #633, opened 2021-05-04
patch the entitlements to support allow_workspace_access

stikkireddy:
this pr allows users to use the `allow_workspace_access` entitlement as a field in the `databricks_user` and `databricks_group` resources.

### DIFF
--- a/identity/resource_group.go
+++ b/identity/resource_group.go
@@ -33,6 +33,9 @@ func ResourceGroup() *schema.Resource {
 		if err = d.Set("allow_instance_pool_create", isGroupInstancePoolCreateEntitled(&group)); err != nil {
 			return diag.FromErr(err)
 		}
+		if err = d.Set("allow_workspace_access", isGroupWorkspaceAccessEntitled(&group)); err != nil {
+			return diag.FromErr(err)
+		}
 		d.Set("url", m.(*common.DatabricksClient).FormatURL("#setting/accounts/groups/", d.Id()))
 		return nil
 	}
@@ -42,6 +45,7 @@ func ResourceGroup() *schema.Resource {
 			allowClusterCreate := d.Get("allow_cluster_create").(bool)
 			allowInstancePoolCreate := d.Get("allow_instance_pool_create").(bool)
 			allowSQLAnalyticsAccess := d.Get("allow_sql_analytics_access").(bool)
+			allowWorkspaceAccess := d.Get("allow_workspace_access").(bool)
 
 			// If entitlement flags are set to be true
 			var entitlementsList []string
@@ -53,6 +57,9 @@ func ResourceGroup() *schema.Resource {
 			}
 			if allowInstancePoolCreate {
 				entitlementsList = append(entitlementsList, string(AllowInstancePoolCreateEntitlement))
+			}
+			if allowWorkspaceAccess {
+				entitlementsList = append(entitlementsList, string(AllowWorkspaceAccessEntitlement))
 			}
 			group, err := NewGroupsAPI(ctx, m).Create(groupName, nil, nil, entitlementsList)
 			if err != nil {
@@ -98,6 +105,17 @@ func ResourceGroup() *schema.Resource {
 					entitlementsRemoveList = append(entitlementsRemoveList, string(AllowInstancePoolCreateEntitlement))
 				}
 			}
+			// If allow_instance_pool_create has changed
+			if d.HasChange("allow_workspace_access") {
+				allowClusterCreate := d.Get("allow_workspace_access").(bool)
+				// Changed to true
+				if allowClusterCreate {
+					entitlementsAddList = append(entitlementsAddList, string(AllowWorkspaceAccessEntitlement))
+				} else {
+					// Changed to false
+					entitlementsRemoveList = append(entitlementsRemoveList, string(AllowWorkspaceAccessEntitlement))
+				}
+			}
 			// TODO: not currently possible to update group display name
 			if entitlementsAddList != nil || entitlementsRemoveList != nil {
 				if err := NewGroupsAPI(ctx, m).Patch(d.Id(),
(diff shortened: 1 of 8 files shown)


identity/resource_group.go
```diff
@@ -98,6 +105,17 @@ func ResourceGroup() *schema.Resource {
 					entitlementsRemoveList = append(entitlementsRemoveList, string(AllowInstancePoolCreateEntitlement))
 				}
 			}
+			// If allow_workspace_access has changed
+			if d.HasChange("allow_workspace_access") {
+				allowClusterCreate := d.Get("allow_workspace_access").(bool)
```
alexott, line 110: I would rename `allowClusterCreate` to `allowWorkspaceAccess`

identity/resource_group.go
```diff
@@ -136,6 +154,10 @@ func ResourceGroup() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"allow_workspace_access": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
alexott, line 160: Should we support existing behaviour when users get workspace access by default? Otherwise they will lose it as we don't specify default for new field.

nfx (changes requested):
Please rewrite with generic implementation instead of copying if d.HasChanges blocks, which are harder to review - logic is copied several times already.

identity/resource_group.go
stikkireddy: hmm its an optional field we can also make it computed and let the server take care of it? but ya these fields dont exist on workspaces without sqlA enabled.

identity/resource_group.go
```diff
@@ -124,49 +101,38 @@ func ResourceGroup() *schema.Resource {
...
 	}
 	return false
 }
 
-func isGroupInstancePoolCreateEntitled(group *ScimGroup) bool {
-	for _, entitlement := range group.Entitlements {
-		if entitlement.Value == AllowInstancePoolCreateEntitlement {
-			return true
-		}
+func getSortedKeys(m map[string]Entitlement) []string {
```
nfx, line 129: remove this method. just have a constant for entitlement field order.

identity/resource_group.go
```diff
@@ -124,49 +101,38 @@ func ResourceGroup() *schema.Resource {
...
 		},
 	}
-}
-
-func isGroupClusterCreateEntitled(group *ScimGroup) bool {
-	for _, entitlement := range group.Entitlements {
-		if entitlement.Value == AllowClusterCreateEntitlement {
-			return true
+	//Add entitlements to the schema
+	for _, entitlementField := range entitlementFields {
```
nfx, line 111: 1) do not introduce a variable for resource, where you can introduce a variable just for schema :)
2) make sure that workspace access has true by default. otherwise thousands of users won't have access to workspaces anymore.

identity/resource_group.go
```diff
@@ -11,6 +12,14 @@ import (
 
 // ResourceGroup manages user groups
 func ResourceGroup() *schema.Resource {
+	entitlementsMap := map[string]Entitlement{
```
nfx, line 15: remove this map and encapsulate this logic in a type.

identity/resource_group.go
```diff
@@ -124,49 +101,38 @@ func ResourceGroup() *schema.Resource {
...
 		}
 	}
-	return false
+	return groupResource
 }
 
-func isGroupSQLAnalyticsAccessEntitled(group *ScimGroup) bool {
-	for _, entitlement := range group.Entitlements {
-		if entitlement.Value == AllowSQLAnalyticsAccessEntitlement {
+func groupEntitlementExists(group *ScimGroup, entitlement Entitlement) bool {
```
nfx, line 120: make `type Entitlements []entitlementsListItem` and add make this as a method of that type, probably call it as `Has(fieldName string) bool`. you can replace the underscore in the field name to match it with an entitlement.

identity/resource_group.go
```diff
@@ -65,37 +62,17 @@ func ResourceGroup() *schema.Resource {
...
-			// If allow_instance_pool_create has changed
-			if d.HasChange("allow_instance_pool_create") {
-				allowClusterCreate := d.Get("allow_instance_pool_create").(bool)
-				// Changed to true
-				if allowClusterCreate {
-					entitlementsAddList = append(entitlementsAddList, string(AllowInstancePoolCreateEntitlement))
-				} else {
-					// Changed to false
-					entitlementsRemoveList = append(entitlementsRemoveList, string(AllowInstancePoolCreateEntitlement))
+			for _, entitlementField := range entitlementFields {
```
nfx, line 65: move this logic to `Entitlements` type.

identity/resource_user_test.go
```diff
@@ -20,6 +20,20 @@ func TestResourceUserRead(t *testing.T) {
...
+						{
+							Value: AllowSQLAnalyticsAccessEntitlement,
+						},
+						{
+							Value: AllowInstancePoolCreateEntitlement,
+						},
+						{
+							Value: AllowClusterCreateEntitlement,
+						},
+					},
```
nfx, line 36: why this much was added?.. refactoring should have added only an item with `AllowWorkspaceAccessEntitlement`.

nfx (changes requested):
code is not abstract enough yet

identity/resource_user_test.go
stikkireddy: was missing coverage in lines so i added it i can remove it if needed.

nfx: ah, okay. it's more that refactoring the code should never change the tests.

nfx:
closing this PR for now, as this will require a code change for all customers using user/group resources in the current situation of things. we can revisit this, once SQLA is available on all environments.

alexott:
@nfx  @stikkireddy I think that we need to revisit it - SQLA should go in the public ungated preview very soon

Issue #635, opened 2021-05-05
[ISSUE] Create/Import databricks_mws_log_delivery JSON error


andywinslow:
### Terraform Version
v0.13.5 

### Affected Resource(s)
databricks_mws_log_delivery

### Expected Behavior
databricks_mws_log_delivery resource should have been created or imported

### Actual Behavior
terraform apply or import fail with something like the following
`Error: Invalid JSON received (518 bytes): {"log_delivery_configuration":{"config_id":"11aa11a1-aaaa-11aa-1a1a-111a111aa111","config_name":"Audit Logs","log_type":"AUDIT_LOGS","output_format":"JSON","account_id":"a111aa1a-a11a-1111-aaaa-111a111a1111","credentials_id":"111a1111-1a11-1a11-11a1-a1aaa111a111","storage_configuration_id":"1a11a111-a1a1-111a-111a-1a111a11aa11","workspace_ids_filter":[1111111111111111],"delivery_path_prefix":"audit-logs","delivery_start_time":"2021-05","status":"ENABLED","creation_time":1620237433000,"update_time":1620237433000}}`

### Steps to Reproduce
`terraform apply` or `terraform import` a databricks_mws_log_delivery resource

### Important Factoids
I believe this will happen when any databricks_mws_log_delivery resource is created or imported as the provider code seems to be expecting workspace_ids_filter in the JSON response to be an array of strings rather than the array of numbers that the Databricks API actually returns.

PR #636, opened 2021-05-05
Fix for issue #635 

andywinslow:
Fix for issue #635... switching WorkspaceIdsFilter from []string to []int (the Databricks API returns an array of integers rather than strings)

### DIFF
--- a/mws/resource_log_delivery.go
+++ b/mws/resource_log_delivery.go
@@ -18,17 +18,17 @@ type LogDelivery struct {
 
 // LogDeliveryConfiguration describes log delivery
 type LogDeliveryConfiguration struct {
-	AccountID              string   `json:"account_id"`
-	ConfigID               string   `json:"config_id,omitempty" tf:"computed"`
-	CredentialsID          string   `json:"credentials_id"`
-	StorageConfigurationID string   `json:"storage_configuration_id"`
-	WorkspaceIdsFilter     []string `json:"workspace_ids_filter,omitempty"`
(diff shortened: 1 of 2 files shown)


mws/resource_log_delivery.go
```diff
@@ -18,17 +18,17 @@ type LogDelivery struct {
...
-	Status                 string   `json:"status,omitempty" tf:"computed"`
-	LogType                string   `json:"log_type"`
-	OutputFormat           string   `json:"output_format"`
-	DeliveryPathPrefix     string   `json:"delivery_path_prefix,omitempty"`
-	DeliveryStartTime      string   `json:"delivery_start_time,omitempty" tf:"computed"`
+	AccountID              string `json:"account_id"`
+	ConfigID               string `json:"config_id,omitempty" tf:"computed"`
+	CredentialsID          string `json:"credentials_id"`
+	StorageConfigurationID string `json:"storage_configuration_id"`
+	WorkspaceIdsFilter     []int  `json:"workspace_ids_filter,omitempty"`
```
alexott, line 25: please change it to `[]int64` to match API documentation: 

> You can optionally set this field to an array of workspace IDs (each one is an **int64**) to which log delivery should apply to

alexott (changes requested):

mws/resource_log_delivery.go
nfx: yep, very long integer over there

nfx:
will change int64 in a separate PR

PR #639, opened 2021-05-10
Recreate mounting clusters when they are deleted

### LINKED ISSUE #637: [ISSUE] Refreshing mount for cluster, that does not exist anymore fails the plan
Reproduce:

1. Create mount without any cluster specified
2. Manually remove `terraform-mount` cluster
3. terraform apply

Can be worked around by removing the cluster_id field from mount resource state, but this is rather a dark magic and horrible workaround.

nfx:
Fixed #637


alexott (approved):
looks good except the formatting warning

PR #648, opened 2021-05-13
add data source user

psg2:
Issue: https://github.com/databrickslabs/terraform-provider-databricks/issues/647

### DIFF
--- a/identity/data_user.go
+++ b/identity/data_user.go
@@ -1,0 +1,56 @@
+package identity
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/databrickslabs/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+
+// DataSourceUser returns information about user specified by user name
+func DataSourceUser() *schema.Resource {
+	type entity struct {
+		UserName             	string   `json:"user_name"`
+		Home             	string   `json:"home" tf:"computed"`
+		Alphanumeric             	string   `json:"alphanumeric" tf:"computed"`
+	}
+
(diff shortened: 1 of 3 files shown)


identity/data_user.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+
+// DataSourceUser returns information about user specified by user name
+func DataSourceUser() *schema.Resource {
+	type entity struct {
+		UserName             	string   `json:"user_name"`
+		Home             	string   `json:"home" tf:"computed"`
```
nfx, line 19: Please run make fmt

identity/data_user.go
```diff
@@ -0,0 +1,56 @@
...
+		Alphanumeric             	string   `json:"alphanumeric" tf:"computed"`
+	}
+
+	s := common.StructToSchema(entity{}, func(
```
nfx, line 23: By the way, if you feel that schema map is shorter - you can do it as well :) just a side thought

identity/data_user.go
```diff
@@ -0,0 +1,56 @@
...
+		s map[string]*schema.Schema) map[string]*schema.Schema {
+		// nolint once SDKv2 has Diagnostics-returning validators, change
+		s["user_name"].ValidateFunc = validation.StringIsNotEmpty
```
nfx, line 26: We've migrated to other linter, so comment above might not work. You need ValidateDiagFunc and wrap this sdkv1 one

nfx (changes requested):
Small fixes and we're ready

identity/data_user.go
psg2: Done.

identity/data_user.go
psg2: Done.

identity/data_user.go
psg2: I guess it's done now that I've changed to schema map.

nfx (approved):
Thank you

PR #652, opened 2021-05-14
Allow instance profiles with paths

alangmdb:
Change that fixes: https://github.com/databrickslabs/terraform-provider-databricks/issues/651

I tried to see if I can write a test, but my knowledge of go projects is too limited. The existing tests obviously pass.

### DIFF
--- a/storage/aws_s3_mount.go
+++ b/storage/aws_s3_mount.go
@@ -121,10 +121,7 @@ func GetOrCreateMountingClusterWithInstanceProfile(
 		return i, err
 	}
 	instanceProfileParts := strings.Split(ia.Resource, "/")
-	if len(instanceProfileParts) != 2 {
-		return i, fmt.Errorf("should have gotten two parts: %v", instanceProfileParts)
-	}
-	clusterName := fmt.Sprintf("terraform-mount-%s", instanceProfileParts[1])
+	clusterName := fmt.Sprintf("terraform-mount-%s", strings.Join(instanceProfileParts[1:], "-"))
 	return clustersAPI.GetOrCreateRunningCluster(clusterName, compute.Cluster{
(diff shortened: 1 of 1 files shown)


storage/aws_s3_mount.go
```diff
@@ -121,10 +121,7 @@ func GetOrCreateMountingClusterWithInstanceProfile(
 		return i, err
 	}
 	instanceProfileParts := strings.Split(ia.Resource, "/")
-	if len(instanceProfileParts) != 2 {
-		return i, fmt.Errorf("should have gotten two parts: %v", instanceProfileParts)
-	}
-	clusterName := fmt.Sprintf("terraform-mount-%s", instanceProfileParts[1])
+	clusterName := fmt.Sprintf("terraform-mount-%s", strings.Join(instanceProfileParts[1:], "-"))
```
nfx, line 124: This change is going to break things when multiple profiles share the same root path. Please change

nfx (changes requested):

storage/aws_s3_mount.go
alangmdb: Could you explain that a bit more? What is the behaviour you are expecting?

Before the change:
"arn:aws:iam::1234566789:instance-profile/myprofile" -> works, cluster name: terraform-mount-myprofile
"arn:aws:iam::1234566789:instance-profile/path/to/myprofile" -> fails due to two parts check (parts: instance-profile, path, to, myprofile)

With the change:
"arn:aws:iam::1234566789:instance-profile/myprofile" -> works, cluster name: `terraform-mount-myprofile`
"arn:aws:iam::1234566789:instance-profile/path/to/myprofile" -> works, cluster name: `terraform-mount-path-to-myprofile`

PR #658, opened 2021-05-21
Fix handling of is_no_public_ip_enabled

pietern:
I tried to create a workspace with public IPs for all workers and found it didn't work.

Turns out the `omitempty` on the `is_no_public_ip_enabled` field means a value of `false` doesn't make it to the API, where it then defaults to `true`. Therefore, it can't be `omitempty`, but it's not a required field either. To make matters more complicated, it's not returned as part of the GET response. This PR sets a `DiffSuppressFunc` ignores value changes and only sets the field on creation.

Removed the field from all responses in tests because the API doesn't return it.

### DIFF
--- a/mws/mws.go
+++ b/mws/mws.go
@@ -87,7 +87,7 @@ type Workspace struct {
 	PricingTier                         string `json:"pricing_tier,omitempty" tf:"computed"`
 	PrivateAccessSettingsID             string `json:"private_access_settings_id,omitempty"`
 	NetworkID                           string `json:"network_id,omitempty"`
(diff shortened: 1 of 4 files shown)


alexott (approved):
lgtm

nfx:
@abhinavg6 ^ can you also look at the logic?

nfx:
running `make test-mws`...

nfx:
`TestMwsAccWorkspaces` failed with `Error: MALFORMED_REQUEST: Failed storage configuration validation checks: PUT, LIST, DELETE`... looking at the policies...

nfx (approved):

PR #681, opened 2021-06-06
Fix ActiveDirectoryEndpoint in adls gen1-2 mounts

burak-db:
For all azure environments we have been using public cloud endpoint
which fails to deploy in other environments like govcloud and china. Use
environment variables to fetch the correct ActiveDirectoryEndpoint per
Azure Environment.

### DIFF
--- a/common/azure_auth_test.go
+++ b/common/azure_auth_test.go
@@ -336,7 +336,7 @@ func TestAzureAuth_configureWithClientSecretAAD(t *testing.T) {
 func TestAzureEnvironment_WithAzureManagementEndpoint(t *testing.T) {
 	fakeEndpoint := "http://google.com"
 	aa := AzureAuth{azureManagementEndpoint: fakeEndpoint}
-	env, err := aa.getAzureEnvironment()
+	env, err := aa.GetAzureEnvironment()
 	assert.Nil(t, err)
 	// This value should be populated with azureManagementEndpoint for testing
 	assert.Equal(t, env.ResourceManagerEndpoint, fakeEndpoint)
@@ -345,40 +345,40 @@ func TestAzureEnvironment_WithAzureManagementEndpoint(t *testing.T) {
 
 	// Making the azureManagementEndpoint empty should yield PublicCloud
 	aa.azureManagementEndpoint = ""
-	env, err = aa.getAzureEnvironment()
+	env, err = aa.GetAzureEnvironment()
 	assert.Nil(t, err)
 	assert.Equal(t, azure.PublicCloud, env)
 }
 
 func TestAzureEnvironment(t *testing.T) {
 	aa := AzureAuth{}
-	env, err := aa.getAzureEnvironment()
+	env, err := aa.GetAzureEnvironment()
 
 	assert.Nil(t, err)
 	assert.Equal(t, azure.PublicCloud, env)
 
 	aa.Environment = "public"
-	env, err = aa.getAzureEnvironment()
+	env, err = aa.GetAzureEnvironment()
 	assert.Nil(t, err)
 	assert.Equal(t, azure.PublicCloud, env)
 
 	aa.Environment = "china"
-	env, err = aa.getAzureEnvironment()
+	env, err = aa.GetAzureEnvironment()
 	assert.Nil(t, err)
 	assert.Equal(t, azure.ChinaCloud, env)
 
 	aa.Environment = "german"
-	env, err = aa.getAzureEnvironment()
+	env, err = aa.GetAzureEnvironment()
 	assert.Nil(t, err)
 	assert.Equal(t, azure.GermanCloud, env)
 
 	aa.Environment = "usgovernment"
-	env, err = aa.getAzureEnvironment()
+	env, err = aa.GetAzureEnvironment()
 	assert.Nil(t, err)
 	assert.Equal(t, azure.USGovernmentCloud, env)
 
 	aa.Environment = "xyzdummy"
-	_, err = aa.getAzureEnvironment()
+	_, err = aa.GetAzureEnvironment()
 	assert.NotNil(t, err)
 }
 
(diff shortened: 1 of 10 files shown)


alexott (approved):
lgtm

common/env.go
```diff
@@ -16,6 +18,26 @@ var (
 	commonClient *DatabricksClient
 )
 
+// GetAzureEnvironment returns azure environment from giving environment string
+func GetAzureEnvironment(environment string) (azure.Environment, error) {
```
nfx, line 22: 1) I'd advise to move this function back to azure_auth.go, as this file has broader scope. E.g. provider.go works a lot with env, but it doesn't mean it should be defined here :)

2) merge these two functions. GetAzureEnvironment is supposed to return error, but it does never return it. And the other function logs the error instead of returning it. Priority for error is to propagate it up (as Go doesn't have exceptions...), not just logging.

storage/adls_gen1_mount.go
```diff
@@ -30,7 +31,7 @@ func (m AzureADLSGen1Mount) Config() map[string]string {
 
 		m.PrefixType + ".oauth2.client.id":   m.ClientID,
 		m.PrefixType + ".oauth2.credential":  fmt.Sprintf("{secrets/%s/%s}", m.SecretScope, m.SecretKey),
-		m.PrefixType + ".oauth2.refresh.url": fmt.Sprintf("https://login.microsoftonline.com/%s/oauth2/token", m.TenantID),
+		m.PrefixType + ".oauth2.refresh.url": fmt.Sprintf("%s/%s/oauth2/token", common.AzureEnvironmentFromARMEnvironment().ActiveDirectoryEndpoint, m.TenantID),
```
nfx, line 34: Don't just rely on environment, because we can specify azure environment in provider block through "azure_env" property (or something like that one) and this code will respect only the ARM_ENVIRONMENT. 

This will require adding DatabricksClient argument to Config method, so you can retrieve correctly configured c.AzureAuth. getAzureEnvironment. And modifying the mount interface.

nfx (changes requested):
Hi @burak-db , the change should be a bit bigger :) I can explain more detail tomorrow

storage/adls_gen1_mount.go
burak-db: updated, unit tests are missing I will look into that as well.

storage/mounts_test.go
```diff
@@ -180,7 +184,8 @@ func TestMountPoint_Mount(t *testing.T) {
 		dbutils.notebook.exit(mount_source)
 	`, mountName, expectedMountSource, expectedMountConfig)
 	testMountFuncHelper(t, func(mp MountPoint, mount Mount) (s string, e error) {
-		return mp.Mount(mount)
+		client := common.CommonEnvironmentClient()
```
nfx, line 187: this test is a unit test - e.g. not talking to the real APIs. `common.CommonEnvironmentClient()` is making a real HTTP client, which may be calling azure cli / ~/.databrickscfg configured workspace. not desired, please change.

storage/mounts_test.go
```diff
@@ -86,6 +87,7 @@ func TestAccSourceOnInvalidMountFails(t *testing.T) {
 
 func TestAccInvalidSecretScopeFails(t *testing.T) {
```
nfx, line 88: and `TestAcc` prefix is for acceptance tests, so this means that `common.CommonEnvironmentClient()` could be used :)

storage/adls_gen2_mount.go
```diff
@@ -25,13 +27,20 @@ func (m AzureADLSGen2Mount) Source() string {
...
+	} else {
+		clientEndpoint = azureEnvironment.ActiveDirectoryEndpoint
+	}
 	return map[string]string{
 		"fs.azure.account.auth.type":                          "OAuth",
 		"fs.azure.account.oauth.provider.type":                "org.apache.hadoop.fs.azurebfs.oauth2.ClientCredsTokenProvider",
 		"fs.azure.account.oauth2.client.id":                   m.ClientID,
 		"fs.azure.account.oauth2.client.secret":               fmt.Sprintf("{secrets/%s/%s}", m.SecretScope, m.SecretKey),
-		"fs.azure.account.oauth2.client.endpoint":             fmt.Sprintf("https://login.microsoftonline.com/%s/oauth2/token", m.TenantID),
+		"fs.azure.account.oauth2.client.endpoint":             fmt.Sprintf("%s/%s/oauth2/token", clientEndpoint, m.TenantID),
```
nfx, line 43: this is great :) eventually we can make `tenant_id` optional, if its present in AzureAuth! :)

storage/adls_gen1_mount.go
```diff
@@ -24,13 +26,20 @@ func (m AzureADLSGen1Mount) Source() string {
 }
 
 // Config ...
-func (m AzureADLSGen1Mount) Config() map[string]string {
+func (m AzureADLSGen1Mount) Config(client *common.DatabricksClient) map[string]string {
+	azureEnvironment, err := client.AzureAuth.GetAzureEnvironment()
+	var clientEndpoint string
+	if err != nil {
+		clientEndpoint = azure.PublicCloud.ActiveDirectoryEndpoint
```
nfx, line 33: hm... `Config` function has no error returned. and you're not logging this error anywhere.

Ideally, you can add `AzureEnvironment` field to `AzureAuth`, which you'll set from `configureWithClientSecret` and `configureWithAzureCLI` calling the `getAzureEnvironment`. this way you'll have guarantees to propagate errors early on and have no side-effects for this function.

nfx (changes requested):

storage/mounts_test.go
burak-db: will change this to use a mock client

common/azure_auth_test.go
```diff
@@ -382,29 +392,6 @@ func TestAzureEnvironment(t *testing.T) {
...
-
-	assert.Equal(t, envErr, err)
-
-	_, err = aa.acquirePAT(ctx, mockFunc)
-	assert.Equal(t, envErr, err)
-
-	_, err = aa.getClientSecretAuthorizer("")
-	assert.Equal(t, envErr, err)
-}
-
```
burak-db, line 407: these functions are not calling getAzureEnvironment anymore making this test kind of pointless to have.

PR #686, opened 2021-06-11
Generate workspace hostname based on account console hostname

pietern:
It not always ends in `.cloud.databricks.com`. We can reasonably assume the workspace hostname to equal the account console hostname with the `accounts` prefix replaced by the deployment name of the workspace.


pietern:
I also needed to comment out https://github.com/databrickslabs/terraform-provider-databricks/blob/14f74eb006a40b38e57eaf50014ee2e53ae8c26b/common/http.go#L131-L161 to make this work.

@nfx Thoughts on removing the hardcoded accounts console URL there?

nfx:
@pietern https://github.com/databrickslabs/terraform-provider-databricks/pull/686#issuecomment-859364265 - eh... technically we can, but i don't know if we should. We have to **explicitly** tell users that they must use provider aliasing. or when they are not using the correct hostname. logically it should be backend's responsibility, but our rest api doesn't give such a detail :( it's very annoying to have those issues.

do you think that making an env variable to specify our internal dev/staging accounts hosts for `strings.Contains(c.Host, accountsHost)` will work?...

nfx:
in order to merge this, we have to get passing tests

alexott (approved):
lgtm

nfx (approved):

pietern:
@nfx Can you merge? I'm not authorized to.

PR #690, opened 2021-06-18
add resource_directory

psg2:
Issue: https://github.com/databrickslabs/terraform-provider-databricks/issues/672

### DIFF
--- a/docs/resources/permissions.md
+++ b/docs/resources/permissions.md
@@ -1,9 +1,10 @@
 ---
 subcategory: "Security"
 ---
+
 # databricks_permissions Resource
 
-This resource allows you to generically manage permissions for other resources in Databricks workspace. It would guarantee, that only *admins*, *authenticated principal* and those declared within `access_control` blocks would have specified access. It is not possible to remove management rights from *admins* group.
+This resource allows you to generically manage permissions for other resources in Databricks workspace. It would guarantee, that only _admins_, _authenticated principal_ and those declared within `access_control` blocks would have specified access. It is not possible to remove management rights from _admins_ group.
 
 ## Cluster usage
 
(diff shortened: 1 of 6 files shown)


workspace/resource_directory.go
```diff
@@ -0,0 +1,56 @@
...
+		"delete_recursive": {
+			Type:     schema.TypeBool,
+			Default:  false,
+			Optional: true,
+		},
+	}
+
+	return common.Resource{
+		Schema:        s,
+		SchemaVersion: 1,
```
nfx, line 29: Schema version should be zero or omitted

workspace/resource_directory.go
```diff
@@ -0,0 +1,56 @@
...
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			notebooksAPI := NewNotebooksAPI(ctx, c)
+			path := d.Get("path").(string)
+			if err := notebooksAPI.Mkdirs(path); err != nil {
+				// TODO: handle RESOURCE_ALREADY_EXISTS
```
nfx, line 34: For directory you probably have to remove this todo

workspace/resource_directory.go
```diff
@@ -0,0 +1,56 @@
...
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			notebooksAPI := NewNotebooksAPI(ctx, c)
+			objectStatus, err := notebooksAPI.Read(d.Id())
+			if err != nil {
+				return err
+			}
+			if objectStatus.ObjectType != Directory {
+				// TODO: better error message
+				return errors.New("different object type on this path other than a directory")
```
nfx, line 48: And error message should tell which is that different object type

workspace/resource_directory.go
```diff
@@ -0,0 +1,56 @@
...
+			}
+			return common.StructToData(objectStatus, s, d)
```
nfx, line 50: You just need to d.Set("path", objectStatus.Path), right?.. otherwise expose object_id as computed

nfx (changes requested):
Small adjustments required

workspace/resource_directory.go
psg2: I think it still needs to be handled. If we try to create the path `/a/b/c` but `a/b` is already a notebook or other type of object it will fail.

image

workspace/resource_directory.go
psg2: Great, will do that later today.

workspace/resource_directory.go
psg2: Great, will do that later today.

workspace/resource_directory.go
psg2: Right, will do this later today as well.

psg2:
Hi, @nfx .

Just to be sure, are you ok with the `delete_recursive` flag? I thought about setting this up using the AWS provider as a reference, they have a similar flag that affects the deletion behavior for databases.

image

psg2:
Will add documentation later as well.

docs/resources/permissions.md
```diff
@@ -241,15 +242,13 @@ resource "databricks_group" "eng" {
...
-    language = "PYTHON"
+resource "databricks_directory" "this" {
+    path = "/Production/ETL"
 }
 
 resource "databricks_permissions" "folder_usage" {
-    directory_path = "/Production/ETL"
-    depends_on = [databricks_notebook.this]
+    directory_path = databricks_directory.this.path
+    depends_on = [databricks_directory.this]
```
nfx, line 251: Depends_on is not needed as long as there's an explicit dependency.

nfx (approved):

PR #706, opened 2021-07-01
validate permission levels relative to the object

psg2:
Hi, folks.

This change is so we can validate during the plan relation between different resource attributes which is yet not possible on Terraform other than ConflictsWith.

The idea here is to block resources like:

```hcl
resource "databricks_permissions" "cluster_permissions" {
  cluster_id = "0000"

  access_control {
    user_name = "me"

    permission_level = "CAN_USE"
  }
}
```

With an error like this one:

```
Error: permission_level CAN_USE is not supported with cluster_id objects

  on main.tf line 10, in resource "databricks_permissions" "cluster_permissions":
  10: resource "databricks_permissions" "cluster_permissions" {
```

References:
- https://github.com/hashicorp/terraform-plugin-sdk/issues/354
- https://github.com/hashicorp/terraform-plugin-sdk/issues/233


access/resource_permissions.go
```diff
@@ -353,7 +363,87 @@ func ResourcePermissions() *schema.Resource {
...
-		Schema:      s,
+		Schema: s,
+		CustomizeDiff: customdiff.Sequence(
+			func(_ context.Context, diff *schema.ResourceDiff, v interface{}) error {
+				// Plan time validation for object permission levels
+				objectPermissionLevelMappingList := []struct {
+					objectKey               string
+					allowedPermissionLevels []string
+				}{
+					{
```
nfx, line 374: Please incorporate with generic permissions mapping:

https://github.com/databrickslabs/terraform-provider-databricks/blob/993bc17c22cd32569fb3f65d071956c9722386a7/access/resource_permissions.go#L230

And perhaps fetch permission levels from API, so that provider won't have to be changed once new permissions are added on the platform

nfx (changes requested):
@psg2 , great addition. Please see the comment

nfx:
And we need to incorporate custom diff to couple of other resources, thanks for taking the first step

access/resource_permissions.go
psg2: Good idea, forgot that we had those on the API. Will do the improvement.

psg2:
> And we need to incorporate custom diff to couple of other resources, thanks for taking the first step

Do we have issues with those? I can take some others as well.

access/resource_permissions.go
psg2: However, one question just saw that the [API to get cluster permissions](https://docs.databricks.com/dev-tools/api/latest/permissions.html#operation/get-clusters-permission-levels) for instance requires a `cluster_id`. I'm thinking if this won't be a problem, since that the CustomizeDiff runs during the plan if I understood correctly there might be the case that the cluster_id is still unresolved and we won't be able to call the API. What are your thoughts here?

nfx: @psg2  you're right. let's then keep them hardcoded. but in `permissionsResourceIDFields`

psg2: Done.

nfx (approved):
Code looks good. Have to double check permissions mapping next week

nfx (approved):

PR #721, opened 2021-07-12
Azure auth with SPN now uses AAD token by default instead of PAT

alexott:
* PAT token still could be used by setting `azure_use_pat_for_spn` to `true`
* Check was added to make sure that Azure KeyVault + SPN is detected during plan, not during execution

### DIFF
--- a/access/resource_permissions.go
+++ b/access/resource_permissions.go
@@ -14,7 +14,6 @@ import (
 	"github.com/databrickslabs/terraform-provider-databricks/workspace"
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/pkg/errors"
 )
@@ -366,32 +365,30 @@ func ResourcePermissions() *schema.Resource {
 	}
 	return &schema.Resource{
 		Schema: s,
-		CustomizeDiff: customdiff.Sequence(
-			func(ctx context.Context, diff *schema.ResourceDiff, m interface{}) error {
-				me, err := identity.NewUsersAPI(ctx, m).Me()
-				if err != nil {
-					return err
+		CustomizeDiff: func(ctx context.Context, diff *schema.ResourceDiff, m interface{}) error {
+			me, err := identity.NewUsersAPI(ctx, m).Me()
+			if err != nil {
+				return err
+			}
+			// Plan time validation for object permission levels
+			for _, mapping := range permissionsResourceIDFields(ctx) {
+				if _, ok := diff.GetOk(mapping.field); !ok {
+					continue
 				}
-				// Plan time validation for object permission levels
-				for _, mapping := range permissionsResourceIDFields(ctx) {
-					if _, ok := diff.GetOk(mapping.field); !ok {
-						continue
+				access_control_list := diff.Get("access_control").(*schema.Set).List()
+				for _, access_control := range access_control_list {
+					m := access_control.(map[string]interface{})
+					permission_level := m["permission_level"].(string)
+					if !stringInSlice(permission_level, mapping.allowedPermissionLevels) {
+						return fmt.Errorf(`permission_level %s is not supported with %s objects`, permission_level, mapping.field)
 					}
-					access_control_list := diff.Get("access_control").(*schema.Set).List()
-					for _, access_control := range access_control_list {
-						m := access_control.(map[string]interface{})
-						permission_level := m["permission_level"].(string)
-						if !stringInSlice(permission_level, mapping.allowedPermissionLevels) {
-							return fmt.Errorf(`permission_level %s is not supported with %s objects`, permission_level, mapping.field)
-						}
-						if m["user_name"].(string) == me.UserName {
-							return fmt.Errorf("it is not possible to decrease administrative permissions for the current user: %s", me.UserName)
-						}
+					if m["user_name"].(string) == me.UserName {
+						return fmt.Errorf("it is not possible to decrease administrative permissions for the current user: %s", me.UserName)
 					}
 				}
-				return nil
-			},
-		),
+			}
+			return nil
+		},
 		ReadContext: readContext,
 		CreateContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 			var entity PermissionsEntity
--- a/access/resource_secret_scope.go
+++ b/access/resource_secret_scope.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/databrickslabs/terraform-provider-databricks/common"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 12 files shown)


access/resource_secret_scope.go
```diff
@@ -150,4 +153,20 @@ func ResourceSecretScope() *schema.Resource {
 			return NewSecretScopesAPI(ctx, c).Delete(d.Id())
 		},
 	}.ToResource()
+
+	resource.CustomizeDiff = customdiff.Sequence(
```
nfx, line 157: why do we need a sequence for one function?

access/resource_secret_scope.go
```diff
@@ -150,4 +153,20 @@ func ResourceSecretScope() *schema.Resource {
...
+		func(ctx context.Context, diff *schema.ResourceDiff, v interface{}) error {
+			kvLst := diff.Get("keyvault_metadata").([]interface{})
+			if len(kvLst) == 0 {
+				return nil
```
nfx, line 161: is it still possible to create databricks scopes?

common/azure_auth.go
```diff
@@ -270,6 +277,19 @@ func (aa *AzureAuth) patRequest() tokenRequest {
 	}
 }
 
+func maybeExtendAuthzError(err error) error {
+	var ae APIError
+	fmtString := "Azure authorization error.  Does your SPN  have Contributor access to Databricks workspace. %v"
```
nfx, line 282: nit: double-spaces 

missing question mark

common/azure_auth.go
```diff
@@ -270,6 +277,19 @@ func (aa *AzureAuth) patRequest() tokenRequest {
...
+	if errors.As(err, &ae) {
+		if ae.ErrorCode == "403" {
```
nfx, line 284: isn't it `ae.StatusCode == 403` instead?

access/resource_secret_scope.go
```diff
@@ -123,11 +124,13 @@ func ResourceSecretScope() *schema.Resource {
 		s["name"].ValidateFunc = validScope
 		s["initial_manage_principal"].ForceNew = true
 		s["keyvault_metadata"].ForceNew = true
+
 		return s
 	})
-	return common.Resource{
+	resource := common.Resource{
```
nfx, line 130: probably make CustomizeDiff package-local function and add unit tests for corner cases

provider/provider_test.go
```diff
@@ -350,6 +356,7 @@ func configureProviderAndReturnClient(t *testing.T, tt providerConfigTest) (*com
 	if err != nil {
 		return nil, err
 	}
+	client.AzureAuth.UsePATForSPN = tt.usePATForSPN
```
nfx, line 359: probably this line should go before `client.Authenticate()`

common/azure_auth.go
```diff
@@ -270,6 +277,19 @@ func (aa *AzureAuth) patRequest() tokenRequest {
...
+			return fmt.Errorf(fmtString, err)
+		}
+	} else if strings.Contains(err.Error(), "does not have authorization to perform action") {
+		return fmt.Errorf(fmtString, err)
```
nfx, line 288: maybe add unit tests for these two corner cases? :)

common/azure_auth.go
```diff
@@ -270,6 +277,19 @@ func (aa *AzureAuth) patRequest() tokenRequest {
 	}
 }
 
+func maybeExtendAuthzError(err error) error {
+	var ae APIError
+	fmtString := "Azure authorization error.  Does your SPN  have Contributor access to Databricks workspace. %v"
+	if errors.As(err, &ae) {
```
nfx, line 283: just in case, there might be shorter version:

```go
if e, ok := err.(APIError); ok && e.StatusCode == 403 {
  // ..
}
```

nfx (changes requested):

access/resource_secret_scope.go
alexott: yes

common/azure_auth.go
alexott: yeah....

access/resource_permissions.go
```diff
@@ -366,32 +365,30 @@ func ResourcePermissions() *schema.Resource {
 	}
 	return &schema.Resource{
 		Schema: s,
-		CustomizeDiff: customdiff.Sequence(
-			func(ctx context.Context, diff *schema.ResourceDiff, m interface{}) error {
-				me, err := identity.NewUsersAPI(ctx, m).Me()
-				if err != nil {
-					return err
+		CustomizeDiff: func(ctx context.Context, diff *schema.ResourceDiff, m interface{}) error {
```
nfx, line 368: can you revert the formatting for this file not to interfere with diff?

access/resource_secret_scope.go
```diff
@@ -114,6 +114,20 @@ var validScope = validation.StringMatch(regexp.MustCompile(`^[\w\.@_-]{1,128}$`)
 	"Must consist of alphanumeric characters, dashes, underscores, and periods, "+
 		"and may not exceed 128 characters.")
 
+func kvDiffFunc(ctx context.Context, diff *schema.ResourceDiff, v interface{}) error {
+	if diff != nil {
```
nfx, line 118: `if diff == nil { return nil }` would be better

access/resource_secret_scope_test.go
```diff
@@ -301,3 +302,39 @@ func TestResourceSecretScopeDelete_Error(t *testing.T) {
...
+		HCL: `
+			name = "Boom"
+			keyvault_metadata {
+				resource_id = "bcd"
+				dns_name = "def"
+			}`,
+		AzureAuth: &common.AzureAuth{ClientID: "123", ClientSecret: "123", TenantID: "123",
+			ResourceID: "/subscriptions/a/resourceGroups/b/providers/Microsoft.Databricks/workspaces/c"},
+		Create: true,
+	}.Apply(t)
```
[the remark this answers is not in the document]
nfx: you can also use `qa.HTTPFixturesApply` if you want to test just the function and not do the whole fixture

provider/provider.go
```diff
@@ -211,6 +211,12 @@ func DatabricksProvider() *schema.Provider {
 				Default:     false,
 				Description: "Create ephemeral PAT tokens also for AZ CLI authenticated requests",
 			},
+			"azure_use_pat_for_spn": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
nfx, line 217: could you, just in case, make it as EnvDefaultFunc?

nfx (changes requested):
last round of small comments

access/resource_permissions.go
alexott: it's not a pure formatting, it's removal of `customdiff.Sequence`, just applying the earlier comment to multiple places.  But I can remove it

qa/testing.go
```diff
@@ -102,6 +103,9 @@ func (f ResourceFixture) Apply(t *testing.T) (*schema.ResourceData, error) {
 	if f.Azure {
 		client.AzureAuth.ResourceID = "/subscriptions/a/resourceGroups/b/providers/Microsoft.Databricks/workspaces/c"
 	}
+	if f.AzureAuth != nil {
+		client.AzureAuth = *f.AzureAuth
```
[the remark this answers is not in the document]
alexott: I didn't want to make it more complex than necessary - with `f.Azure` it's easy, just set it to `true`, and it's all. with `AzureAuth` it's slightly more complex

nfx (approved):

PR #725, opened 2021-07-14
Small fixes for DLT pipelines

alexott:
* updated documentation
* added support for missing attribute `target` that was documented, but not implemented


nfx (approved):
Lgtm

PR #728, opened 2021-07-15
bugfix on documentation: networks must be private.

dfanesiDB:


docs/resources/mws_workspaces.md
```diff
@@ -50,7 +50,7 @@ resource "databricks_mws_networks" "this" {
   account_id   = var.databricks_account_id
   network_name = "${var.prefix}-network"
   vpc_id       = var.vpc_id
-  subnet_ids = [var.subnet_public, var.subnet_private]
+  subnet_ids = [var.subnet_private]
```
nfx, line 53: workspace requires at least two subnets, did you verify this change to work?

dfanesiDB: actually, in my tests I was using multiple private subnets since the best practices are to always use multiple AZs. 

do you want me to change the whole VPC definition?

dfanesiDB: changing expected variable type to list (variable definition not present, so just changing the name and the syntax).

docs/resources/mws_workspaces.md
```diff
@@ -50,7 +50,7 @@ resource "databricks_mws_networks" "this" {
   account_id   = var.databricks_account_id
   network_name = "${var.prefix}-network"
   vpc_id       = var.vpc_id
-  subnet_ids = [var.subnet_public, var.subnet_private]
+  subnet_ids = var.subnets_private
   security_group_ids = [var.security_group]
```
nkvuong, line 54: should this be `var.security_groups` as well and expects a list variable, similar to above, as Databricks can have access to more than a single security group?

nfx:
@dfanesiDB please change every markdown file, where subnet_ids are referred, not just one - https://github.com/databrickslabs/terraform-provider-databricks/search?l=Markdown&q=subnet_ids

nfx (approved):

PR #735, opened 2021-07-21
Skip common error clarity for alternative accounts console deployments

nfx:
This PR makes it possible to use different URLs for accounts console when `development_mode = true` is added to provider. Reserved for internal use.


alexott (approved):
lgtm

pietern (approved):
Nice!

PR #742, opened 2021-07-26
Unify HTTP 404 handling

### LINKED ISSUE #740: [ISSUE] databricks_permission can't be destroyed when job doesn't exist anymore
### Terraform Version
v0.15.4
Databricks provider version:; 0.3.6

### Affected Resource(s)
- databricks_permission

### Expected Behavior
We are trying to remove a databricks_permission resources for a job that has been manually deleted via the console. We would expect the provider to simply skip this step and delete the resource cleanly.

### Actual Behavior
We are getting the following error instead:

```
databricks_permissions.job_permission: Destroying... [id=/jobs/<id>]
╷
│ Error: Job <id> does not exist.
```

Our terraform project is completely unusable now because, as far as we can tell, we can't restore/create a job with a specific id.

### Steps to Reproduce
1. Create a databricks_permissions for a job via Terraform
2. Delete the job via the Databricks console
3. Try to alter or remove the databricks_permissions in Terraform

### Important Factoids
This Terraform project is deployed by gitlab-ci/cd, the statefile isn't readily accessible to be altered (it's not good practice anyway).

nfx:
This commit also fixes #740

### DIFF
--- a/access/resource_permissions.go
+++ b/access/resource_permissions.go
@@ -339,6 +339,7 @@ func ResourcePermissions() *schema.Resource {
 		id := d.Id()
 		objectACL, err := NewPermissionsAPI(ctx, m).Read(id)
 		if common.IsMissing(err) {
+			log.Printf("[INFO] %s is removed on backend", d.Id())
 			d.SetId("")
 			return nil
 		}
@@ -433,7 +434,7 @@ func ResourcePermissions() *schema.Resource {
 		DeleteContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 			err := NewPermissionsAPI(ctx, m).Delete(d.Id())
 			if common.IsMissing(err) {
-				log.Printf("[INFO] databricks_permissions[id=%s] dependency is arelady removed on backend", d.Id())
+				log.Printf("[INFO] %s is already removed on backend", d.Id())
 				return nil
 			}
 			if err != nil {
--- a/access/resource_permissions.go
+++ b/access/resource_permissions.go
@@ -3,6 +3,7 @@ package access
 import (
 	"context"
 	"fmt"
+	"log"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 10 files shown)


access/resource_permissions.go
```diff
@@ -431,6 +432,10 @@ func ResourcePermissions() *schema.Resource {
 		},
 		DeleteContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 			err := NewPermissionsAPI(ctx, m).Delete(d.Id())
+			if common.IsMissing(err) {
+				log.Printf("[INFO] databricks_permissions[id=%s] dependency is arelady removed on backend", d.Id())
```
alexott, line 436: arelady -> already

access/resource_permissions.go
```diff
@@ -337,7 +338,7 @@ func ResourcePermissions() *schema.Resource {
 	readContext := func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 		id := d.Id()
 		objectACL, err := NewPermissionsAPI(ctx, m).Read(id)
-		if aerr, ok := err.(common.APIError); ok && aerr.IsMissing() {
+		if common.IsMissing(err) {
```
alexott, line 341: Would it make sense to add log output here as well?

common/http.go
```diff
@@ -60,6 +60,17 @@ func (apiError APIError) Error() string {
 	return apiError.Message
 }
 
+// IsMissing tells if error is about missing resource
+func IsMissing(err error) bool {
+	if err == nil {
+		return false
+	}
+	if e, ok := err.(APIError); ok && e.IsMissing() {
```
alexott, line 68: Simplify it to the 

e, ok := ….
return ok && e.IsMissing()

?

alexott (approved):
Lgtm overall, small changes would be useful

common/http.go
nfx: Good one

PR #746, opened 2021-07-27
Don't error if widget or visualization doesn't exist

pietern:
It's fine if the remote resource was deleted. Terraform will attempt to recreate.

### DIFF
--- a/sqlanalytics/resource_visualization.go
+++ b/sqlanalytics/resource_visualization.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


sqlanalytics/resource_visualization.go
```diff
@@ -187,6 +190,12 @@ func ResourceVisualization() *schema.Resource {
 				return err
 			}
 
+			// Handle if visualization was not found.
+			if av == nil {
```
nfx, line 194: this is handled generically already - every APIError with status code 404 is removing the resource... you probably need to return error the right way from Read method.

https://github.com/databrickslabs/terraform-provider-databricks/blob/master/common/resource.go#L48-L54

nfx (changes requested):

sqlanalytics/resource_visualization.go
pietern: Fixed!

sqlanalytics/resource_widget_test.go
```diff
@@ -500,6 +500,40 @@ func TestWidgetCreateWithPositionAndAutoheight(t *testing.T) {
...
+		Resource:    ResourceWidget(),
+		Read:        true,
+		Removed:     true,
+		RequiresNew: true,
+		ID:          "some-uuid/12344",
+		InstanceState: map[string]string{
+			"dashboard_id":     "some-uuid",
+			"visualization_id": "678",
+		},
+	}.Apply(t)
```
nfx, line 531: and for these simple tests that don't have to check for `ResourceData`, please just use the `.ApplyNoError(t)` instead of `.Apply(t)`

nfx (approved):

PR #753, opened 2021-07-28
Google Cloud Platform Service Account Auth Support

nfx:
Added Google Cloud Platform Service Account authentication for Accounts API


common/gcp.go
```diff
@@ -0,0 +1,86 @@
...
+		TargetPrincipal: c.GoogleServiceAccount,
+		IncludeEmail:    true,
+	}, c.googleAuthOptions...)
+	if err != nil {
+		err = fmt.Errorf("could not obtain OIDC token. %w Running 'gcloud auth application-default login' may help", err)
+		return
+	}
+	// TODO: verify that refreshers work...
+	ts = oauth2.ReuseTokenSource(nil, ts)
+	return
```
iamsgarg-db, line 24: return ts, nil ?

common/gcp.go
```diff
@@ -0,0 +1,86 @@
...
+func (c *DatabricksClient) getGoogleOIDCSource() (ts oauth2.TokenSource, err error) {
+	// source for generateIdToken
+	ts, err = impersonate.IDTokenSource(c.InitContext, impersonate.IDTokenConfig{
+		Audience:        c.Host,
+		TargetPrincipal: c.GoogleServiceAccount,
+		IncludeEmail:    true,
+	}, c.googleAuthOptions...)
+	if err != nil {
+		err = fmt.Errorf("could not obtain OIDC token. %w Running 'gcloud auth application-default login' may help", err)
+		return
```
iamsgarg-db, line 20: return nil, err?

common/gcp.go
iamsgarg-db: I understand you are using named return arguments, but personally I think saying `return ts,nil` makes the code nicer to read

nfx: On shorter methods we use named returns 🤷🏻‍♂️

mws/mws.go
```diff
@@ -73,6 +75,30 @@ const (
...
+}
+
+type GCPCommonNetworkConfig struct {
+	GKEConnectivityType     string `json:"gke_connectivity_type"`
+	GKEClusterMasterIPRange string `json:"gke_cluster_master_ip_range"`
+}
+
+type GCPNetwork struct {
+	GCPManagedNetworkConfig *GCPManagedNetworkConfig `json:"gcp_managed_network_config"`
+	GCPCommonNetworkConfig  *GCPManagedNetworkConfig `json:"gcp_common_network_config"`
```
grayluck, line 99: Why is GCPCommonNetworkConfig with the type of *GCPManagedNetworkConfig?

mws/mws.go
```diff
@@ -95,6 +121,34 @@ type Workspace struct {
 	CreationTime                        int64  `json:"creation_time,omitempty" tf:"computed"`
 
 	ExternalCustomerInfo *externalCustomerInfo `json:"external_customer_info,omitempty" tf:"computed"`
+
+	CloudResourceBucket *CloudResourceBucket `json:"cloud_resource_bucket,omitempty"`
+	Network             *GCPNetwork          `json:"network,omitempty"`
```
grayluck, line 126: Don't see tests for the network field?

mws/mws.go
nfx: This is a good one. Will be fixed in prepare/0.3.7 branch

PR #771, opened 2021-08-11
Add support for Databricks Repos

alexott:
This PR adds `databricks_repo` resource to manage [Databricks Repos](https://docs.databricks.com/repos.html)


exporter/importables.go
```diff
@@ -856,4 +856,57 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			if ic.meAdmin {
+				ic.Emit(&resource{
+					Resource: "databricks_permissions",
+					ID:       fmt.Sprintf("/repos/%s", r.ID),
+					Name:     "repo_" + ic.Importables["databricks_repo"].Name(r.Data),
+				})
+			}
+			return nil
+		},
+		Body: func(ic *importContext, body *hclwrite.Body, r *resource) error {
```
nfx, line 897: Why do we need custom body exporter?

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// NewReposAPI creates ReposAPI instance from provider meta
+func NewReposAPI(ctx context.Context, m interface{}) ReposAPI {
+	return ReposAPI{m.(*common.DatabricksClient), ctx}
+}
+
+// ReposResponse provides information about given repository
+type ReposResponse struct {
+	Id           int64  `json:"id"`
```
nfx, line 26: nit: ID

(Somewhere in the old checkstyle it was enforced)

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+	Url          string `json:"url"`
+	Provider     string `json:"provider"`
+	Path         string `json:"path"`
+	Branch       string `json:"branch"`
+	HeadCommitId string `json:"head_commit_id"`
```
nfx, line 31: And ID here

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+	provider := ""
+	u, err := url.Parse(uri)
+	if err == nil {
+		provider = gitProvidersMap[strings.ToLower(u.Host)]
+	}
+	return provider
+}
+
+func ResourceRepo() *schema.Resource {
+	s := map[string]*schema.Schema{
```
nfx, line 107: Please use reflect resource

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+		},
+	}
+	return common.Resource{
+		Schema:        s,
+		SchemaVersion: 1,
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			reposAPI := NewReposAPI(ctx, c)
+			url := d.Get("url").(string)
+			provider := d.Get("git_provider").(string)
+			if provider == "" { // trying to infer Git Provider from the URL
```
nfx, line 151: Could you hide it in Create method of repos api?

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+			req := createRequest{Path: d.Get("path").(string), Provider: provider, Url: url}
+			resp, err := reposAPI.Create(req)
+			if err != nil {
+				return err
+			}
+			d.SetId(fmt.Sprintf("%d", resp.Id))
+			branch := d.Get("branch").(string)
+			tag := d.Get("tag").(string)
+			if tag != "" {
```
nfx, line 165: Can you hide it in Update method of repos api? It's reused :)

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+	context context.Context
+}
+
+// NewReposAPI creates ReposAPI instance from provider meta
+func NewReposAPI(ctx context.Context, m interface{}) ReposAPI {
+	return ReposAPI{m.(*common.DatabricksClient), ctx}
+}
+
+// ReposResponse provides information about given repository
+type ReposResponse struct {
```
nfx, line 25: Probably it's better named as Repo or RepoSnapshot, then it fits onto reflect resource

workspace/resource_repo.go
```diff
@@ -0,0 +1,216 @@
...
+			// for URL/provider force re-create as there are limits on what could be done for changing URL/provider
+			reposAPI := NewReposAPI(ctx, c)
+			// Not working yet, wait until API is ready
+			// if d.HasChange("path") {
+			// 	err := reposAPI.Update(d.Id(), map[string]string{"path": d.Get("path").(string)})
+			// 	if err != nil {
+			// 		return err
+			// 	}
+			// }
+			if d.HasChange("branch") || d.HasChange("tag") {
```
nfx, line 201: If all non-branch/tag fields are forcenew, then this method is way shorter

workspace/resource_repo_test.go
```diff
@@ -0,0 +1,405 @@
...
+		Resource: ResourceRepo(),
+		Read:     true,
+		Removed:  true,
+		ID:       fmt.Sprintf("%d", repoID),
+	}.Apply(t)
+}
+
+func TestResourceRepoDelete(t *testing.T) {
+	repoID := 48155820875912
+	repoIdStr := fmt.Sprintf("%d", repoID)
```
nfx, line 85: Nit: repoID

workspace/resource_repo_test.go
```diff
@@ -0,0 +1,405 @@
...
+		},
+		Resource: ResourceRepo(),
+		Delete:   true,
+		ID:       repoIdStr,
+	}.Apply(t)
+	assert.NoError(t, err, err)
+	assert.Equal(t, repoIdStr, d.Id())
+}
+
+func TestResourceRepoCreateNoBranch(t *testing.T) {
```
nfx, line 102: Can you run "make fmt" or how did you manage to get that two-space identation? :)

nfx (changes requested):
Please adjust repos resource to reflect_resource

workspace/resource_repo.go
alexott: it's temporary. there are plans to extend API to allow to change other parameters without recreation

exporter/importables.go
alexott: Because by default it omits fields that are marked as optional, i.e. exporting only URL

workspace/resource_repo_test.go
alexott: I did run it, interesting how it got formatted that way

nfx: ah, it was github mobile app

workspace/resource_repo.go
nfx: then ForceNews will be removed?..

alexott: yes, it will be - I put the comment for that field to remove when it's available.

exporter/importables.go
nfx: @alexott probably a separate exporter bug.

alexott: I think it’s a combination of optional and computed properties. From exporter side we don’t know everything, we need to think on how to handle that correctly

workspace/resource_repo.go
```diff
@@ -0,0 +1,203 @@
...
+	u, err := url.Parse(uri)
+	if err == nil {
+		provider = gitProvidersMap[strings.ToLower(u.Host)]
+	}
+	return provider
+}
+
+func ResourceRepo() *schema.Resource {
+	s := common.StructToSchema(ReposInformation{}, func(s map[string]*schema.Schema) map[string]*schema.Schema {
+		s["url"].ForceNew = true
```
nfx, line 133: i wonder if we should add `force_new` to reflect_resource

workspace/resource_repo.go
```diff
@@ -0,0 +1,203 @@
...
+	return a.client.Delete(a.context, fmt.Sprintf("/repos/%s", id), nil)
+}
+
+func (a ReposAPI) Update(id string, r map[string]string) error {
+	if len(r) == 0 {
+		return nil
+	}
+	// TODO: update may change ONE OF (url AND provider (optional)), (path), or (branch OR tag).
+	// for URL/provider force re-create as there are limits on what could be done for changing URL/provider
+	if path, ok := r["path"]; ok {
```
nfx, line 69: i think the code would be cleaner if this method accepts this struct

```
type updateRequest struct {
  Branch       string `json:"branch,omitempty"`
  Tag.            string `json:"tag,omitempty"`
}
```

because it's going to be easier to use `common.DataToStructPointer` & `common.StructToData` and there'll be less errors to handle. 

and then the schema might be composed of calling StructToSchema twice?.. usage of reflection simplifies resource maintenance, because we have to deal just with structs...

workspace/resource_repo.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+			return reposAPI.Update(d.Id(), updateReq)
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			reposAPI := NewReposAPI(ctx, c)
+			resp, err := reposAPI.Read(d.Id())
+			if err != nil {
+				return err
+			}
+			d.SetId(resp.RepoID())
```
nfx, line 181: not needed in Read

workspace/resource_repo.go
```diff
@@ -0,0 +1,203 @@
...
+		SchemaVersion: 1,
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			reposAPI := NewReposAPI(ctx, c)
+			req := createRequest{Path: d.Get("path").(string), Provider: d.Get("git_provider").(string), Url: d.Get("url").(string)}
+			resp, err := reposAPI.Create(req)
+			if err != nil {
+				return err
+			}
+			d.SetId(resp.RepoID())
+			branch := d.Get("branch").(string)
```
nfx, line 165: technically, you can re-use the update method here, right?..

nfx (changes requested):

workspace/resource_repo.go
alexott: Addition of path, and necessity of handling it differently makes things harder - setting branch/tag here is different from Update method. Although maybe we’ll need th handle it later, when it’s implemented

workspace/resource_repo.go
alexott: I thought about that, just didn't want to pollute the PR for specific functionality

workspace/resource_repo.go
alexott: update logic is different here compared to Update method - I don't want to re-checkout branch if it's already checked out. Correct way would be if the Create API had the `branch` and `tag` parameters, but it's not implemented there - that's why two calls.  On other side, I can specify `branch` parameter set to the default branch of repository, that's why we have `branch != resp.Branch` condition

workspace/resource_repo.go
alexott: Update of path-only will come in the near future, so it will be easier to change code with current approach

nfx (approved):

Issue #781, opened 2021-08-18
[ISSUE] Azure authentication stopped working with azurerm v2.71.0+


mark-greene:
## Terraform Version
terraform v1.04
databricks v0.3.7
azurerm v2.71.0+ (works on ≤2.70.0) cc @favoretti

### Terraform Configuration Files
```hcl
provider "databricks" {
  azure_workspace_resource_id = azurerm_databricks_workspace.this.id
  azure_client_id             = var.client_id
  azure_client_secret         = var.client_secret
  azure_tenant_id             = var.tenant_id
}
```

or another example

```hcl
provider "databricks" {
  azure_workspace_resource_id = azurerm_databricks_workspace.databricks_ws.id
  azure_client_id             = azuread_application.databricks.application_id
  azure_client_secret         = "${random_string.password.result}"
  azure_tenant_id             = "${data.azurerm_client_config.current.tenant_id}"
}
```

### Actual Behavior
```
Error: authentication is not configured for provider. Please configure it
│ through one of the following options:
│ 1. DATABRICKS_HOST + DATABRICKS_TOKEN environment variables.
│ 2. host + token provider arguments.
│ 3. azure_databricks_workspace_id + AZ CLI authentication.
│ 4. azure_databricks_workspace_id + azure_client_id + azure_client_secret + azure_tenant_id for Azure Service Principal authentication.
│ 5. Run `databricks configure --token` that will create ~/.databrickscfg file.
│ 
│ Please check https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs#authentication for details
```

and logs show `Explicit and implicit attributes: azure_client_id, azure_client_secret, azure_tenant_id`, though `azure_databricks_workspace_id` is empty, but it should not be empty.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`

### Important Factoids
Azure methods 3 and 4 were working fine until recently.  Now only 1 and 2 are working.  I have gone back to v0.3.5 with the same results so I'm guessing this is a change on the Azure side but I'm not positive.

The text `azure_databricks_workspace_id` only appears in the string for this error message, so it would seem that:
1. the error message is wrong
2. there is some other issue that is the cause of the error message

I hit this issue while trying to upgrade our databricks provider from 0.3.2 to 0.3.7. However, that caused TF to also pull in this upgrade:
```
 provider "registry.terraform.io/hashicorp/azurerm" {
-  version     = "2.67.0"
[...]

PR #791, opened 2021-08-20
Improve error reporting

nfx:
This pull request aims at presenting clearer authentication error messages - specifically returning the names of all non-empty provider attributes and environment variables, so that cases like #781 are triaged without looking at the logs

**Deprecations**
* `azure_workspace_name`, `azure_resource_group`, `azure_subscription_id`, and `azure_workspace_resource_id` are deprecated and would be removed in v0.4.0. Please rewrite provider configuration with `host = data.azurerm_databricks_workspace.example.workspace_url` to achieve the same effect. Please check [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/databricks_workspace#workspace_url) resource documentation for details.
* `azure_use_pat_for_spn`, `azure_use_pat_for_cli`, and `azure_pat_token_duration_seconds` are deprecated to fully switch to AAD token authentication in the near future.
* `DATABRICKS_AZURE_CLIENT_SECRET` environment variable is deprecated in favor of just using `ARM_CLIENT_SECRET`.
* `DATABRICKS_AZURE_CLIENT_ID` environment variable is deprecated in favor of just using `ARM_CLIENT_ID`.
* `DATABRICKS_AZURE_TENANT_ID` environment variable is deprecated in favor of just using `ARM_TENANT_ID`.


common/client.go
```diff
@@ -97,15 +167,13 @@ func (c *DatabricksClient) Authenticate() error {
...
-		"2. host + token provider arguments.\n" +
-		"3. azure_databricks_workspace_id + AZ CLI authentication.\n" +
-		"4. azure_databricks_workspace_id + azure_client_id + azure_client_secret + azure_tenant_id " +
-		"for Azure Service Principal authentication.\n" +
-		"5. Run `databricks configure --token` that will create ~/.databrickscfg file.\n\n" +
-		"Please check https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs#authentication for details")
+	info := ""
+	if len(c.configAttributesUsed) > 0 {
+		// TODO: add env vars
+		info = fmt.Sprintf("Attributes used: %s - %d", strings.Join(c.configAttributesUsed, ", "), len(configOptions))
```
alexott, line 173: I would maybe changed the message to something like `%d attributes used: %s` - easier to read & understand what the number means

common/client.go
```diff
@@ -97,15 +167,13 @@ func (c *DatabricksClient) Authenticate() error {
...
+	}
+	docUrl := "https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs#authentication"
```
alexott, line 175: Can we put provider version instead of latest?

common/azure_auth.go
```diff
@@ -69,87 +39,81 @@ type tokenInfo struct {
...
 	return azure.EnvironmentFromName(envName)
 }
 
-func (aa *AzureAuth) resourceID() string {
-	if aa.ResourceID != "" {
-		if aa.SubscriptionID == "" || aa.ResourceGroup == "" {
-			res, err := azure.ParseResourceID(aa.ResourceID)
+func (aa *DatabricksClient) resourceID() string {
+	if aa.AzureDatabricksResourceID != "" {
+		if aa.AzureSubscriptionID == "" || aa.AzureResourceGroup == "" {
```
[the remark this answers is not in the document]
nfx: It does

common/client.go
nfx: I'd remove the count, it's a dummy for static check to see variable usage

common/client.go
nfx: I'll see if it's available at this moment - most likely it should.

alexott: yep, just version number instead of `latest`: https://registry.terraform.io/providers/databrickslabs/databricks/0.3.6/docs

common/azure_auth.go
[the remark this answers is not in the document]
nfx: Actually, i'm killing this field, because it makes more maintenance

common/client.go
nfx: latest version links will be broken...

nfx:
@favoretti, what do you think about `azure_workspace_resource_id` deprecation?

favoretti:
@nfx Dunno really how you were using it. If URL is enough for provider to authenticate I don't see a point supplying Azure resource ID if it's not used.

nfx:
@favoretti url became enough couple of months ago :) there are still some workspaces with regional URLs, but those are no longer created

favoretti:
Could do a 2-step approach here then? Deprecate it with a deprecation warning first, remove it completely in version N+1?

nfx:
> Could do a 2-step approach here then? Deprecate it with a deprecation warning first, remove it completely in version N+1?

@favoretti it's a 2-step approach :) v0.4.0 will remove those fields, for now they'll keep working as expected, but give warnings.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 * Added `databricks_repo` resource to manage [Databricks Repos](https://docs.databricks.com/repos.html) ([#771](https://github.com/databrickslabs/terraform-provider-databricks/pull/771))
 
+**Deprecations**
+* `azure_workspace_name`, `azure_resource_group`, `azure_subscription_id`, and `azure_workspace_resource_id` are deprecated and would be removed in v0.4.0. Please rewrite provider configuration with `host = data.azurerm_databricks_workspace.example.workspace_url` to achieve the same effect. Please check [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/databricks_workspace#workspace_url) resource documentation for details.
```
alexott, line 8: should we change it to `host = azurerm_databricks_workspace.example.workspace_url (or host = data.azurerm_databricks_workspace.example.workspace_url)` ?

the first variant is primarily used when chaining azurerm with our provider

provider/provider.go
```diff
@@ -264,111 +100,73 @@ func DatabricksProvider() *schema.Provider {
...
+		if len(attr.EnvVars) > 0 {
+			fieldSchema.DefaultFunc = schema.MultiEnvDefaultFunc(attr.EnvVars, nil)
+		}
+	}
+
+	ps["token"].Sensitive = true
+	ps["azure_client_secret"].Sensitive = true
+
+	azCoordinatesDeprecation := "`%s` is deprecated and would be removed in v0.4.0. Please rewrite provider configuration " +
+		"with `host = data.azurerm_databricks_workspace.example.workspace_url` to achieve the same effect. Please check " +
```
alexott, line 127: remove `data.` ?

alexott (approved):
lgtm

provider/provider.go
nfx: Data is data source. Probably I'll add another couple of examples into main doc later

PR #819, opened 2021-09-03
Add service principal in permissions doc

NastasiaSaby:
# What

Add a service principal in the doc example for Databricks permissions for jobs creation.

# Why

To get somewhere an example.

### DIFF
--- a/docs/resources/permissions.md
+++ b/docs/resources/permissions.md
@@ -151,6 +151,10 @@ resource "databricks_group" "eng" {
     display_name = "Engineering"
 }
 
+resource "databricks_service_principal" "main" {
+  display_name = "main"
+}
+
 resource "databricks_job" "this" {
     name = "Featurization"
     max_concurrent_runs = 1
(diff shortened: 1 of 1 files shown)


docs/resources/permissions.md
```diff
@@ -151,6 +151,10 @@ resource "databricks_group" "eng" {
     display_name = "Engineering"
 }
 
+resource "databricks_service_principal" "main" {
+  display_name = "main"
```
nfx, line 155: Please also explicitly mention for which cloud is it. The API is rather confusing, unfortunately.

nfx (changes requested):
Thank you for polishing our documentation.

Small comment to be added, as SPs have different definitions on AWS and Azure.

docs/resources/permissions.md
NastasiaSaby: I did it by changing the name of the "databricks service principal".

nfx: Great!

Issue #823, opened 2021-09-07
[ISSUE] Destroying databricks_workspace_conf fails when using storeInteractiveNotebookResultsInCustomerAccount


laurensknoll:
Hi,

Databricks currently allows us to store all interactive notebook results in the Azure/AWS account. Also see https://docs.databricks.com/administration-guide/workspace/storage.html#configure-the-storage-location-for-all-interactive-notebook-results

When this feature is configured using the `databricks_workspace_conf`-resource. The feature is enabled accordingly after deployment. On destroy, however, the provider fails with the following error: `Some values are not allowed: {"storeInteractiveNotebookResultsInCustomerAccount":""}`

Apparently, the provider sends out a empty string value instead of false value.

### Configuration
```hcl
resource "databricks_workspace_conf" "this" {
  custom_config = {
    "storeInteractiveNotebookResultsInCustomerAccount" : true
  }
}
```

### Expected Behavior
On destroy, the databricks workspace configuration should be destroyed without any error. 

### Actual Behavior
The destroy fails, due to a bad request:  `Some values are not allowed {"storeInteractiveNotebookResultsInCustomerAccount":""}`

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `terraform destroy`

### Terraform and provider versions
Terraform v1.0.2
AzureRM 2.61
Databricks v0.3.6

alexott:
Fixed by #826

SudoSpartanDan:
@alexott Seems this issue exists still with some attributes:

```
Error: Some values are not allowed: {"maxTokenLifetimeDays":""}
```

This gets passed in as a string and the API for the configurations doesn't seems to like that it's just a blank string.

https://github.com/databrickslabs/terraform-provider-databricks/blob/57d334d0173ba7b70ae95de068df0a6958fc2f49/workspace/resource_workspace_conf.go#L104-L118

alexott:
why are you passing empty value?

SudoSpartanDan:
I’m not, this is on destruction.

SudoSpartanDan:
I think the issue is the API for this particular value expects ```"0"``` rather than just a blank string, so on destruction it fails.


https://docs.databricks.com/dev-tools/api/latest/token-management.html#operation/set-configuration

alexott:
yep, I'll look later today or tmorrow morning

abij:
[...]

PR #826, opened 2021-09-15
Reset custom workspace configuration using value

laurensknoll:
Created this PR to address issue #823.

The issue shows that a key is added which doesn't comply to the expected naming conventions. As a consequence, the user is unable to remove the key, because it defaults to an invalid type according to the Databricks API.

This change replaces the key-based default value lookup with a value-based lookup. The lookup first checks the value type, and second tries to parse boolean strings.

Like the old implementation, this defaults to the `"false"`-string for boolean values, and default to the empty `""` string for all other values.

### DIFF
--- a/workspace/resource_workspace_conf.go
+++ b/workspace/resource_workspace_conf.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strconv"
 	"strings"
 
 	"github.com/databrickslabs/terraform-provider-databricks/common"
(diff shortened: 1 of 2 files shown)


nfx (commented):
@laurensknoll the issue with workspace conf is that there are currently no conventions about workspace conf keys. how many different valid keys did you test this with? I assume you've found a way to get their names ;)

laurensknoll:
There indeed is no key naming convention @nfx .. And while an exhaustive list of keys would be great, it doesn't help us to support new keys. Therefore, this change pushes the problem of new/inconveniently named keys to the end user.

If the end user specifies a boolean value, we handle it as such. This means that when resetting the value, we send out `"false"`. For any other value, we reset it to `""`.

As a result, we don't need to know which key is meant to be boolean/ other. The end user just tells us.. and when they fail, they can change the code and continue development.

nfx (approved):

PR #828, opened 2021-09-16
Change scheduler.mode to FIFO for terraform-mounts

epa095:
We are trying out this change, and so far it seems to work nicely. If we have multiple PRs doing validate at the same time they would often all timeout on mounts-validation. With this change the first ones will at least be allowed to finnish withouth beeing influenced by new ones. 

If you want we can try it out a bit longer and report back if we are still pleased with the change in a week or two ;-)


nfx:
@epa095 Let's see what our storage team thinks about whether or not this will help

nfx:
@epa095 I think it might be worth also trying to add standard terraform's timeout propagation. Could you try it out locally as well? You need to propagate timeout handling from ResourceData to CommandAPI. Theoretically timeout could be propagated down with context.WithTimeout with less change surface, but still has to be researched.

https://www.sohamkamani.com/golang/context-cancellation-and-values/#context-timeouts

epa095:
Changing the timeout handling seems like a good idea as well! But since I don't know Go I don't think I am the guy to do it :-p Also, after changing to FIFO we no longer get timeouts where we did before (34 mounts, 2-3 concurrent calls to terraform-validate).  

If you have access to databricks developers it would be really interesting to know if it is possible to somehow make the cluster run more than nr-of-cores tasks at the same time. In a less managed setting I think we could make that happen by tweaking `spark.executor.cores` or `spark.executor.instances`, but it seems like databricks ignores/overwrites those settings when I try to make the 4-core machine run e.g. 16 concurrent tasks.

epa095:
I updated the tests to account for the new config, but PR pipeline needs approval before running :-S

Issue #888, opened 2021-11-02
[ISSUE] `databricks_secret_acl` scope property validation does not allow `/` characters


tylangesmith:
Hey there, using the `databricks_secret_acl` resource I'm coming across a few scope validation issues.

The Databricks CLI / API allows for scopes to be created with a `/` character, this behaviour however is not consistent with the validation observed against this terraform `databricks_secret_acl` resource.

### Configuration
```hcl
resource "databricks_secret_acl" "scope_acl1" {
  principal  = databricks_group.developers.display_name
  permission = "READ"
  scope      = "myapplication/master"
}
```

### Expected Behavior

I expect the validation of the scope `"myapplication/master"` to be valid which is inline with the API.

### Actual Behavior

An error is thrown by the resource indicating the scopes `"myapplication/master"` is invalid.

> Error: invalid value for scope (Must consist of alphanumeric characters, dashes, underscores, and periods, and may not exceed 128 characters.)

### Potential Fix

The behaviour looks to be coming from this function https://github.com/databrickslabs/terraform-provider-databricks/blob/master/access/resource_secret_scope.go#L113 - which validates against the regex `^[\w\.@_-]{1,128}$`.

This can be updated to `^[\w\.@_/-]{1,128}$` to allow the `/` character.

nfx:
@tylangesmith just keep in mind we're separating scope and key with `|||` within resource ID, which is highly unlikely to occur in the name of real-world secret scope

nfx:
@tylangesmith where did PR go?

PR #892, opened 2021-11-02
Update secret scope validation to include `/` characters

tylangesmith:
This PR fixes issue #888

### DIFF
--- a/access/resource_secret_acl_test.go
+++ b/access/resource_secret_acl_test.go
@@ -79,74 +79,6 @@ func TestSecretsScopesAclsIntegration(t *testing.T) {
 	assert.NoError(t, err, err)
 }
 
-func TestSecretsScopesAclsIntegration_ScopeWithSlash(t *testing.T) {
-	cloud := os.Getenv("CLOUD_ENV")
-	if cloud == "" {
-		t.Skip("Acceptance tests skipped unless env 'CLOUD_ENV' is set")
-	}
-	client := common.NewClientFromEnvironment()
-	ctx := context.Background()
-	scopesAPI := NewSecretScopesAPI(ctx, client)
-	secretsAPI := NewSecretsAPI(ctx, client)
-	secretAclsAPI := NewSecretAclsAPI(ctx, client)
-
-	testScope := "myapplication/branch"
-	testKey := "my-test-key"
-	testSecret := "my-test-secret"
-	initialManagePrincipal := ""
-	// TODO: on random group
-	testPrincipal := "users"
-
-	err := scopesAPI.Create(SecretScope{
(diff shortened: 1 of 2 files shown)


access/resource_secret_acl_test.go
```diff
@@ -79,6 +79,74 @@ func TestSecretsScopesAclsIntegration(t *testing.T) {
 	assert.NoError(t, err, err)
 }
 
+func TestSecretsScopesAclsIntegration_ScopeWithSlash(t *testing.T) {
```
nfx, line 82: Test prefix must be "TestAcc"

nfx: This test doesn't add value to provider - SDK layer doesn't perform validation, Terraform resource does.

Create a test in acceptance package instead and add a scope with a slash in name.

access/resource_secret_acl_test.go
```diff
@@ -121,6 +211,26 @@ func TestResourceSecretACLRead_NotFound(t *testing.T) {
 	}.ApplyNoError(t)
 }
 
+func TestResourceSecretACLRead_ScopeWithSlash_NotFound(t *testing.T) {
```
nfx, line 214: This test is not necessary

access/resource_secret_acl_test.go
```diff
@@ -142,6 +252,27 @@ func TestResourceSecretACLRead_Error(t *testing.T) {
 	assert.Equal(t, "global|||something", d.Id(), "Id should not be empty for error reads")
 }
 
+func TestResourceSecretACLRead_ScopeWithSlash_Error(t *testing.T) {
```
nfx, line 255: This test is not necessary as well - url encoding of a slash was tested by the first test added

access/resource_secret_acl_test.go
```diff
@@ -174,6 +305,38 @@ func TestResourceSecretACLCreate(t *testing.T) {
 	assert.Equal(t, "global|||something", d.Id())
 }
 
+func TestResourceSecretACLCreate_ScopeWithSlash(t *testing.T) {
```
nfx, line 308: Probably just keep this test, as it covers both read and create

nfx (changes requested):
Please remove redundant unit tests and keep only one for creating. Url encoding and validation is the test focus, not already tested functionality.

nfx (approved):
Lgtm

nfx:
@tylangesmith please rebase onto latest master

tylangesmith:
Forgot to tag you this morning @nfx but I believe that should be rebased now

nfx (approved):

PR #899, opened 2021-11-04
Resource for global configuration of SQL Endpoints

### LINKED ISSUE #855: Support Data Access configuration for SQL endpoint (Azure)
I cannot find the option to terraform the Service Principal configuration in Databricks SQL on Azure.  This is in the SQL Endpoint Settings of the admin console:

image

Can support for this be added to the provider please?

alexott:
This fixes #855


common/client.go
```diff
@@ -411,19 +411,39 @@ func (c *DatabricksClient) configureHTTPCLient() {
 	}
 }
 
+func (c *DatabricksClient) GetHost() string {
+	if c.Host == "" {
+		err := c.Authenticate(context.TODO())
+		if err != nil {
+			panic(err)
```
nfx, line 418: No panics in production flow. It'll crash provider and customers would loose faith again

sqlanalytics/resource_sql_global_config.go
```diff
@@ -0,0 +1,117 @@
...
+	context context.Context
+}
+
+// Set ...
+func (a SQLEndpointsGlobalConfigAPI) Set(gc SQLEndpointGlobalConfig) error {
+	data := map[string]interface{}{
+		"security_policy":           gc.SecurityPolicy,
+		"enable_serverless_compute": gc.EnableServerlessCompute,
+	}
+	if a.client.IsAws() {
```
nfx, line 50: Why not relying on backend to throw error if arn is specified for azure?

sqlanalytics/resource_sql_global_config.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/databrickslabs/terraform-provider-databricks/common"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type SQLEndpointConfPair struct {
```
nfx, line 12: We already have a similar entity for tags in sql endpoint. Though dunno if it's good idea to reuse it

nfx (changes requested):
No panic in production flow

common/client.go
alexott: so just return empty string?

sqlanalytics/resource_sql_global_config.go
alexott: I prefer to do detect this earlier

common/client.go
alexott: fixed

sqlanalytics/resource_sql_global_config.go
nfx: can we just eagerly call `err = a.client.Authenticate(a.context)` in this method, so that we propagate errors correctly and don't need to change `IsAzure` / `IsAws` / `IsGcp` methods? this will reduce the risks of uncaught error in lazy authenticated client and assert hostname presence. starting from v0.4.x there'll always be a host, anyway.

sqlanalytics/resource_sql_global_config.go
```diff
@@ -0,0 +1,117 @@
...
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// SQLEndpointGlobalConfig ...
+type SQLEndpointGlobalConfig struct {
```
nfx, line 18: probably remove `SQLEndpoint` prefix from all types in this file - names make lines too long for something this small :) it's `sql` package anyway, no name clashes.

nfx (changes requested):
please replace risky changes of `IsAzure`/`IsGcp` with less risky `err = a.client.Authenticate(a.context)` in `SQLEndpointsGlobalConfigAPI.Set`

sqlanalytics/resource_sql_global_config.go
alexott: I'll look tomorrow it's not the first time where we hit lazy auth issue - previous was in the generic mounts

pietern (commented):
LGTM!

sqlanalytics/resource_sql_global_config.go
```diff
@@ -0,0 +1,123 @@
...
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// GlobalConfig ...
+type GlobalConfig struct {
+	SecurityPolicy          string            `json:"security_policy,omitempty" tf:"default:DATA_ACCESS_CONTROL"`
+	DataAccessConfig        map[string]string `json:"data_access_config,omitempty"`
+	InstanceProfileARN      string            `json:"instance_profile_arn,omitempty"`
+	EnableServerlessCompute bool              `json:"enable_serverless_compute,omitempty" tf:"default:false"`
```
nfx, line 22: aren't all bools false by default? why do we need `tf:"default:false"`

sqlanalytics/resource_sql_global_config.go
```diff
@@ -0,0 +1,123 @@
...
+	EnableServerlessCompute bool       `json:"enable_serverless_compute,omitempty"`
+}
+
+// NewSqlGlobalConfigAPI ...
+func NewSqlGlobalConfigAPI(ctx context.Context, m interface{}) sGlobalConfigAPI {
+	return sGlobalConfigAPI{m.(*common.DatabricksClient), ctx}
+}
+
+// sAPI ...
+type sGlobalConfigAPI struct {
```
nfx, line 39: what is `s`? :) maybe `globalConfigAPI`?

nfx (changes requested):
Please also add `TestPreviewAcc` test for this one in `sqlanalytics/acceptance` package.

It would also be nice, if you add either a SQL endpoint data resource or something like `newServerlessReusedEnpointForTesting`, so that `make test-preview` doesn't run for 30+ minutes

https://github.com/databrickslabs/terraform-provider-databricks/blob/master/compute/common_instances.go#L98-L116

sqlanalytics/resource_sql_global_config.go
alexott: artifact of find/replace :-)

sqlanalytics/resource_sql_global_config.go
alexott: Without it, there is a warning when executing:

```
      - .enable_serverless_compute: was null, but now cty.False
```

so I decided to remove it

alexott:
Ok, I've added the Acc test. Let's do the performance optimization as a separate PR

nfx (approved):
lgtm

PR #900, opened 2021-11-04
Allow to disable auto-stop for SQL Endpoints

alexott:


common/reflect_resource.go
```diff
@@ -85,7 +85,7 @@ func StructToSchema(v interface{}, customize func(map[string]*schema.Schema) map
 }
 
 func handleOptional(typeField reflect.StructField, schema *schema.Schema) {
-	if strings.Contains(typeField.Tag.Get("json"), "omitempty") {
+	if strings.Contains(typeField.Tag.Get("json"), "omitempty") || strings.Contains(typeField.Tag.Get("tf"), "optional") {
```
nfx, line 88: This may cause schema consistency validation errors, need to run also integration tests just to be sure

alexott: If you afraid that something will break, then I can get back to original fix:

```
m["auto_stop_mins"].Required = false
m["auto_stop_mins"].Optional = true
```

nfx: Required is false by default anyway. 

This is literally the same annoying case as with `num_workers`, where we have a lot of hacks around.

alexott: No. If I remove omitempty then it’s implicitly set to Required = true during analysis of the struct… I already hit this issue :-)

alexott: Yes. I need to check, maybe we can simplify other places with that

nfx: there should be no struct tags for single place use

sqlanalytics/resource_sql_endpoint.go
```diff
@@ -186,6 +186,8 @@ func ResourceSQLEndpoint() *schema.Resource {
 	s := common.StructToSchema(SQLEndpoint{}, func(
 		m map[string]*schema.Schema) map[string]*schema.Schema {
 		m["auto_stop_mins"].Default = 120
+		m["auto_stop_mins"].Required = false
+		m["auto_stop_mins"].Optional = true
```
nfx, line 190: `.Optional = true` and no `omitempty` creates a schema consistency conflict.

nfx (changes requested):

nfx:
this PR is of low priority, given the percentage of resource usage across all resources.

sqlanalytics/resource_sql_endpoint.go
alexott: that's why it has `.Required = false`

PR #902, opened 2021-11-09
convert job ID to int64 instead of int32

### LINKED ISSUE #901: [ISSUE] Job deployment failing with strconv.ParseInt: parsing [...] value out of range
Starting suddenly today, deployment of new jobs using terraform is failing with the exception
```
strconv.ParseInt: parsing "1087122280397219": value out of range
```
When I log in to the databricks webpage I see that the job has actually been made (but terraform crashed), and it has been made with the ID "1087122280397219 ". All old jobs have ID in the thousands-range. Validation of old jobs are working. If I create a new job manually in the GUI it also gets a huge ID (233342562969030). 
It really seems like the issue is that databricks suddenly have started to provide job-IDs which no longer fit in an int, making this provider crash when it tries to parse the ID. 

Unfortunately I am on version 0.3.5 and at the moment unable to upgrade. If this issue is known and fixed in a later version I appologize for wasting your time.

### Expected Behavior
terraform databricks provider should be able to handle the job-ID returned by databricks.

### Actual Behavior
Terraform crashes when it tries to parse an int which is too large. 

### Steps to Reproduce
Create a new job

provider "registry.terraform.io/databrickslabs/databricks" {
  version     = "0.3.5"

alexott:
fixes #901

### DIFF
--- a/compute/resource_job.go
+++ b/compute/resource_job.go
@@ -101,7 +101,7 @@ func (a JobsAPI) Start(jobID int64, timeout time.Duration) error {
 }
 
 func (a JobsAPI) Restart(id string, timeout time.Duration) error {
-	jobID, err := strconv.ParseInt(id, 10, 32)
+	jobID, err := strconv.ParseInt(id, 10, 64)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


nfx (approved):
Lgtm:)

nfx:
Thank you 🙏🏻

PR #927, opened 2021-11-18
Add optional external_id field to SCIM users and groups

alexott:
External ID could be propagated from linked identity provider, and is used for things like Unity Catalog


nfx (approved):
please also add an entry to changelog

PR #937, opened 2021-11-30
Ignore autoaz changes

kuritsu:
This change will prevent Terraform to restart a cluster which AwsAttributes.zone_id = "auto", as this is an unneeded/unwanted behavior.

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -3,6 +3,7 @@ package clusters
 import (
 	"context"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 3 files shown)


clusters/resource_cluster.go
```diff
@@ -47,9 +48,19 @@ func SparkConfDiffSuppressFunc(k, old, new string, d *schema.ResourceData) bool
 	return false
 }
 
+func AwsAttribsDiffSuppressFunc(k, old, new string, d *schema.ResourceData) bool {
+	if strings.HasPrefix(k, "aws_attributes.") &&
```
nfx, line 52: Just check for key as a string - I've made GoLang custom struct fields to translate as always-single-element-lists, so you can be sure of first element being zero

clusters/resource_cluster.go
```diff
@@ -47,9 +48,19 @@ func SparkConfDiffSuppressFunc(k, old, new string, d *schema.ResourceData) bool
...
+		strings.HasSuffix(k, ".zone_id") && new == "auto" {
+		log.Printf("[DEBUG] Suppressing diff for k=%#v old=%#v new=%#v", k, old, new)
```
nfx, line 54: Old value must be checked, otherwise this will ignore "auto" for new clusters. Very subtle, but important detail.

clusters/resource_cluster_test.go
```diff
@@ -858,6 +858,100 @@ func TestResourceClusterUpdate_Error(t *testing.T) {
...
+				ExpectedRequest: Cluster{
+					AutoterminationMinutes: 15,
+					ClusterID:              "abc",
+					NumWorkers:             100,
+					ClusterName:            "Shared Autoscaling",
+					SparkVersion:           "7.1-scala12",
+					NodeTypeID:             "i3.xlarge",
+					AwsAttributes: &AwsAttributes{
+						Availability:  "SPOT",
+						FirstOnDemand: 1,
```
nfx, line 923: Shouldn't zone id auto be sent here?

nfx (changes requested):
Please see comments

clusters/resource_cluster.go
kuritsu: done

clusters/resource_cluster_test.go
kuritsu: done

clusters/resource_cluster.go
kuritsu: done

PR #944, opened 2021-12-07
Adding experimental resources

nfx:


catalog/resource_metastore.go
```diff
@@ -0,0 +1,125 @@
...
+	client  *common.DatabricksClient
+	context context.Context
+}
+
+func NewMetastoresAPI(ctx context.Context, m interface{}) MetastoresAPI {
+	return MetastoresAPI{m.(*common.DatabricksClient), ctx}
+}
+
+type MetastoreInfo struct {
+	Name         string  `json:"name"`
```
alexott, line 22: Should it be force new? Together with storage root?

nfx: metastore is identified by UUID, so changing name is just fine. storage root, on the other hand, cannot be updated.

alexott: yeah, I saw ForceNew later in the code, after I left the comment. Why not annotation? ;-)

nfx: Oh, forgot about it! Will change for annotation.

PR #946, opened 2021-12-08
Rename `allow_sql_analytics_access` to `databricks_sql_access`

### LINKED ISSUE #844: [ISSUE] allow_sql_analytics_access should be renamed
It's confusing that the property name for enabling Databricks SQL is `allow_sql_analytics_access`, whereas the product name is Databricks SQL. The property name should be `allow_db_sql_access`.

nfx:
Now `databricks_user`, `databricks_group`, and `databricks_service_principal` resources are more in line with Databricks SQL product naming.

Fixes #844


identity/scim.go
```diff
@@ -41,7 +41,7 @@ func (cv complexValues) HasValue(value string) bool {
 var entitlementMapping = map[string]string{
 	"allow-cluster-create":       "allow_cluster_create",
 	"allow-instance-pool-create": "allow_instance_pool_create",
-	"databricks-sql-access":      "allow_sql_analytics_access", // TODO: state change
+	"databricks-sql-access":      "databricks_sql_access",
```
alexott, line 44: State migration?

docs/resources/group.md
```diff
@@ -50,7 +50,7 @@ The following arguments are supported:
...
 * `allow_instance_pool_create` -  (Optional) This is a field to allow the group to have [instance pool](instance_pool.md) create privileges. More fine grained permissions could be assigned with [databricks_permissions](permissions.md#Instance-Pool-usage) and [instance_pool_id](permissions.md#instance_pool_id) argument.
-* `allow_sql_analytics_access` - (Optional) This is a field to allow the group to have access to [Databricks SQL](https://databricks.com/product/databricks-sql) feature through [databricks_sql_endpoint](sql_endpoint.md).
+* `databricks_sql_access` - (Optional) This is a field to allow the group to have access to [Databricks SQL](https://databricks.com/product/databricks-sql) feature in User Interface and through [databricks_sql_endpoint](sql_endpoint.md).
```
alexott, line 53: does it make sense to specify default values? same for SP & user resources

identity/scim.go
nfx: Not required, tested locally. Just field rename is enough.

PR #955, opened 2021-12-11
Added `databricks_clusters` data source

nfx:
# databricks_clusters Data Source

-> **Note** If you have a fully automated setup with workspaces created by [databricks_mws_workspaces](../resources/mws_workspaces.md) or [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace), please make sure to add [depends_on attribute](../index.md#data-resources-and-authentication-is-not-configured-errors) in order to prevent _authentication is not configured for provider_ errors.

Retrieves a list of [databricks_cluster](../resources/cluster.md#cluster_id) ids, that were created by Terraform or manually, with or without [databricks_cluster_policy](../resources/cluster_policy.md).

## Example Usage

Retrieve all clusters on this workspace on AWS or GCP:

```hcl
data "databricks_clusters" "all" {
    depends_on = [databricks_mws_workspaces.this]
}
```

Retrieve all clusters with "Shared" in their cluster name on this Azure Databricks workspace:

```hcl
data "databricks_clusters" "all_shared" {
    depends_on = [azurerm_databricks_workspace.this]
    cluster_name_contains = "shared"
}
```

## Argument Reference

* `cluster_name_contains` - (Optional) Only return [databricks_cluster](../resources/cluster.md#cluster_id) ids that match the given name string.

## Attribute Reference

This data source exports the following attributes:

* `ids` - list of [databricks_cluster](../resources/cluster.md#cluster_id) ids

### DIFF
--- a/clusters/data_clusters.go
+++ b/clusters/data_clusters.go
@@ -1,0 +1,47 @@
+package clusters
+
+import (
+	"context"
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceClusters() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: func(ctx context.Context, d *schema.ResourceData, i interface{}) diag.Diagnostics {
+			clusters, err := NewClustersAPI(ctx, i).List()
+			if err != nil {
+				return diag.FromErr(err)
+			}
+			ids := []string{}
+			name_contains := strings.ToLower(d.Get("cluster_name_contains").(string))
+			for _, v := range clusters {
+				match_name := strings.Contains(strings.ToLower(v.ClusterName), name_contains)
+				if name_contains != "" && !match_name {
+					continue
+				}
+				ids = append(ids, v.ClusterID)
+			}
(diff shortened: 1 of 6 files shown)


clusters/data_clusters.go
```diff
@@ -0,0 +1,47 @@
...
+		},
+		Schema: map[string]*schema.Schema{
+			"ids": {
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"cluster_name_contains": {
```
alexott, line 41: Does it make sense to add more selectors, like, creator name, tags, etc.?

alexott (approved):
Code looks good, just question on should we have more selectors? Also, should we expose more information, like name, etc.

clusters/data_clusters.go
nfx: when people would need them, they'll send a PR :) until then it makes no sense to make it more complex any further.

nfx:
@alexott adding names of the clusters would make this resource more complicated - for now it returns a list of ids. if we start exposing names - we should rather add `databricks_cluster` data source (and only when we'd need it for some useful scenario).

PR #956, opened 2021-12-11
Remove failed libraries on running clusters

### LINKED ISSUE #599: Remove managed library that failed to install on refresh
Remove managed library that failed to install on refresh

nfx:
Whenever a library fails to get installed on a running cluster, we automatically remove it, so that the clean state of managed libraries is properly maintained. Without this fix users had to manually go to Clusters UI and remove library from a cluster, where it failed to install. Libraries add up to CREATE and UPDATE timeouts of `databricks_cluster` resource.

Fixes #599

### DIFF
--- a/libraries/libraries_api.go
+++ b/libraries/libraries_api.go
@@ -45,7 +45,14 @@ func (a LibrariesAPI) ClusterStatus(clusterID string) (cls ClusterLibraryStatuse
 	return
 }
 
-func (a LibrariesAPI) UpdateLibraries(clusterID string, add, remove ClusterLibraryList, isActive bool) error {
+type Wait struct {
+	ClusterID string
+	Timeout   time.Duration
+	IsRunning bool
+	IsRefresh bool
+}
+
+func (a LibrariesAPI) UpdateLibraries(clusterID string, add, remove ClusterLibraryList, timeout time.Duration) error {
 	if len(remove.Libraries) > 0 {
 		err := a.Uninstall(remove)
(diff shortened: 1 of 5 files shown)


libraries/libraries_api.go
```diff
@@ -243,10 +281,10 @@ func (cll *ClusterLibraryList) String() string {
 
 // LibraryStatus is the status on a given cluster when using the libraries status api
 type LibraryStatus struct {
-	Library                         *Library `json:"library,omitempty"`
-	Status                          string   `json:"status,omitempty"`
-	IsLibraryInstalledOnAllClusters bool     `json:"is_library_for_all_clusters,omitempty"`
-	Messages                        []string `json:"messages,omitempty"`
+	Library  *Library `json:"library,omitempty"`
+	Status   string   `json:"status,omitempty"`
+	IsGlobal bool     `json:"is_library_for_all_clusters,omitempty"`
```
alexott, line 286: should we start deprecation of this flag? Because per [documentation](https://docs.databricks.com/libraries/cluster-libraries.html#library):
>  This option does not install the library on clusters running Databricks Runtime 7.0 and above.

alexott (approved):
lgtm

libraries/libraries_api.go
nfx: It's not changeable by users anymore 🤷🏻‍♂️ but there still might be some workspaces with dbr7.0, that have this

PR #965, opened 2021-12-14
Improve Databricks CLI auth

### LINKED ISSUE #915: [ISSUE] Error creating SP on Azure when using profile instead of explicit host
When trying to create SP in Azure workspace and using profile instead of explicit host, the creation fails with following error:

```
╷
│ Error: display_name is required for service principals in Databricks on AWS
│
│   with databricks_service_principal.sp,
│   on vm.tf line 44, in resource "databricks_service_principal" "sp":
│   44: resource "databricks_service_principal" "sp" {
│
```

It looks like another place affected by lazy initialization of authentication part.

nfx:
* eagerly resolve `host`, `username`, `password`, and `token` from the specified `profile`
* add explicit logging of auth parameters in debug logs
* make sure that `ClientForHost()` eagerly authenticates before passing parameters to a child client
* possibly fix #915

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -250,7 +250,7 @@ func (c *DatabricksClient) Authenticate(ctx context.Context) error {
 		if authorizer == nil {
 			continue
 		}
-		log.Printf("[INFO] Configured %s auth: %s", auth.name, c.configDebugString())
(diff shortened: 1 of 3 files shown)


alexott (approved):
lgtm

PR #969, opened 2021-12-15
Added `token` block to `databricks_mws_workspaces` 

### LINKED ISSUE #957: [FEATURE] Create refreshable workspace PAT on from workspace resource
When creating E2 workspaces, provider is authorized against accounts.cloud.databricks.com with username and password. When creating entities within that workspace, same username and password would work, as it's from account admin.

It may not be the best option to use same credentials with account level privileges for the purposes of environment segregation: production workspace and development workspace are recommended to use different pat tokens. This could be done currently with provider aliasing, though configuration experience is not ideal.

The proposal is to add optional "token block, that would allow specifying PAT lifetime seconds to be created within that workspace and manage token refresh for it.

nfx:
Fixes #957

### DIFF
--- a/.vscode/testing.code-snippets
+++ b/.vscode/testing.code-snippets
@@ -24,5 +24,14 @@
 		  "}"
 		],
 		"description": "Method with single error result"
-	  }
+	  },
+	"if err != nil return err": {
+		"prefix": "ife",
+		"body": [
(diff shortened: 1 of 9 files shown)


alexott (approved):
lgtm

Issue #954, opened 2021-12-11
Ability to install library on all clusters within workspace


nfx:
There are cases, when a certain set of libraries has to be installed on all of the clusters. One of the ways of doing it is through `databricks_global_init_scirpt`, though this is more of a workaround. Terraform-friendly solution is to create a data source to return a list of clusters and `databricks_cluster_library` resource (#904).

```hcl
data "databricks_clusters" "all" {
}

resource "databricks_library" "cli" {
  for_each = data.databricks_clusters.all.ids
  cluster_id = each.key
  pypi {
    package = "databricks-cli"
  }
}
```

PR #971, opened 2021-12-15
Added `databricks_library` resource

### LINKED ISSUE #904: [FEATURE] `databricks_cluster_library` resource
Adding libraries to clusters may impose certain challenges mentioned in #820, #468, and #599. These issues become harder, when artifacts are installed outside of terraform workflow.

Scope of the issue is to modify `databricks_cluster` to ignore any changes to libraries, if they are either managed by new `databricks_cluster_library` resource or outside of TF workflow.

nfx:
```hcl
data "databricks_clusters" "all" {
}

resource "databricks_library" "cli" {
  for_each = data.databricks_clusters.all.ids
  cluster_id = each.key
  pypi {
    package = "databricks-cli"
  }
}
```

Fix #904 #954

### DIFF
--- a/clusters/databricks_library.go
+++ b/clusters/databricks_library.go
@@ -18,6 +18,13 @@ func ResourceLibrary() *schema.Resource {
 		}
 		return m
 	})
+	parseId := func(id string) (string, string) {
+		split := strings.SplitN(id, "/", 2)
+		if len(split) != 2 {
+			return "unknown", "unknown"
+		}
+		return split[0], split[1]
+	}
 	return common.Resource{
 		Schema: s,
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
@@ -51,11 +58,7 @@ func ResourceLibrary() *schema.Resource {
 			return nil
 		},
 		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			split := strings.SplitN(d.Id(), "/", 2)
-			if len(split) != 2 {
-				split = []string{"unknown", "unknown"}
-			}
-			clusterID, libraryRep := split[0], split[1]
+			clusterID, libraryRep := parseId(d.Id())
 			cll, err := libraries.NewLibrariesAPI(ctx, c).WaitForLibrariesInstalled(libraries.Wait{
 				ClusterID: clusterID,
 				Timeout:   d.Timeout(schema.TimeoutRead),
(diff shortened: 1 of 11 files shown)


clusters/databricks_library.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			d.SetId(fmt.Sprintf("%s/%s", clusterID, lib.String()))
+			return nil
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			split := strings.SplitN(d.Id(), "/", 2)
+			if len(split) != 2 {
+				split = []string{"unknown", "unknown"}
+			}
+			clusterID, libraryRep := split[0], split[1]
```
alexott, line 58: move this into a separate function to avoid repetition? I think that was the same code in the cluster resource?

docs/resources/library.md
```diff
@@ -0,0 +1,115 @@
+---
+subcategory: "Compute"
+---
+# databricks_library resource
+
+Installs a library on [databricks_cluster](cluster.md). Each different type of library has a slightly different syntax. It's possible to set only one type of library within one resource. Otherwise, the plan will fail with an error. 
+
+-> **Note** `databricks_library` resource would always start the associated cluster if it's not running, so make sure to have auto-termination configured. It's not possible to atomically change the version of the same library without cluster restart.
```
alexott, line 8: not only change the version, but also uninstall requires restart

docs/resources/library.md
```diff
@@ -0,0 +1,115 @@
...
+  cluster_id = each.key
+  pypi {
+    package = "databricks-cli"
+  }
+}
+```
+
+## Java/Scala JAR
+
+Installing JAR artifacts on a cluster. Location can be anything, that is DBFS or mounted object store (s3, adls, ...)
```
alexott, line 29: not only mounted - newer runtimes can load directly from S3/ADLS, ...

docs/resources/library.md
```diff
@@ -0,0 +1,115 @@
...
+
+resource "databricks_library" "app" {
+  cluster_id = databricks_cluster.this.id
+  jar = databricks_dbfs_file.app.dbfs_path
+}
+```
+
+## Java/Scala Maven
+
+Installing artifacts from Maven repository. You can also optionally specify a `repo` parameter for custom Maven-style repository, that should be accessible without any authentication for the network that cluster runs in. It can even be properly configured [maven s3 wagon](https://github.com/seahen/maven-s3-wagon), [AWS CodeArtifact](https://aws.amazon.com/codeartifact/) or [Azure Artifacts](https://azure.microsoft.com/en-us/services/devops/artifacts/).
```
alexott, line 45: Not sure about "without any authentication for the network that cluster runs in" -> Maven libraries are resolved in the control plane, so repo should be accessible from it

docs/resources/library.md
```diff
@@ -0,0 +1,115 @@
...
+
+```hcl
+resource "databricks_library" "deequ" {
+  cluster_id = databricks_cluster.this.id
+  maven {
+    coordinates = "com.amazon.deequ:deequ:1.0.4"
+    // exlusions block is optional
```
alexott, line 52: `exlusions` -> `exclusions`

alexott (approved):
mostly good, but would be useful to fix  docs comments

PR #987, opened 2021-12-16
Update examples for Azure configuration

a2m1:
I've faced with an issue that azurerm provider was passing an empty value to databricks provider. I was able to fix it only by using datasources. Noticed that similar information is provided in the guide after searching the source code [azure-workspace.md](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/docs/guides/azure-workspace.md#data-resources-and-authentication-is-not-configured-errors) but it's not available while browsing the registry.

Related issue: https://github.com/hashicorp/terraform/issues/2430

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -210,8 +210,17 @@ resource "azurerm_databricks_workspace" "this" {
   sku                           = "premium"
 }
 
+data "azurerm_databricks_workspace" "this" {
+  name                = azurerm_databricks_workspace.this.name
+  resource_group_name = var.resource_group
+
+  depends_on = [
+    azurerm_databricks_workspace.this
+  ]
+}
+
 provider "databricks" {
-  host = azurerm_databricks_workspace.this.workspace_url
+  host = data.azurerm_databricks_workspace.this.workspace_url
 }
 
 resource "databricks_user" "my-user" {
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -210,8 +210,17 @@ resource "azurerm_databricks_workspace" "this" {
   sku                           = "premium"
 }
 
+data "azurerm_databricks_workspace" "this" {
```
nfx, line 213: the examples do work most of the times like they were before. This PR seems to suggest a workaround with data resource, but does not explain why.

Please do not modify these examples and instead add a new paragraph with detailed explanation, issue reference (https://github.com/hashicorp/terraform/issues/2430), and code example. Make sure to modify azure workspace guide as well.

nfx (changes requested):

nfx:
hi @a2m1 , any update on review comments?

a2m1:
Hey @nfx,
I will update it as a separate paragraph. The current example worked for me as well until one day when I didn't change any configuration and it just stopped working. To be honest I didn't find the root cause of my issue. But the issue was that `azurerm` wasn't providing information to the `databricks` provider and I was able to fix it by hardcoding static values (as workspace was already deployed) and later via the data source.

alexott:
@a2m1 you just need to put `depends_on = [azurerm_databricks_workspace.this]` into databricks resources - this will force them wait until workspace is created

a2m1:
@alexott I wish it was that simple. Unfortunately, it didn't help

PR #992, opened 2021-12-21
Increase coverage for `storage` package

nfx:


nfx:
@alexott theoretically, we can make role mounts very generic:

```go
type roleMeta struct {
	path []string
	clusterNamer func(roleName string) string
}

var mountWithRole = map[string]roleMeta{
	"s3.0.instance_profile": roleMeta{
		path: []string{"aws_attributes", "instance_profile_arn"},
		clusterNamer: func(roleName string) string {
			return "."
		},
	},
	"gs.0.service_account": roleMeta{
		path: []string{"gcp_attributes", "google_service_account"},
		clusterNamer: func(roleName string) string {
			return fmt.Sprintf("terraform-mount-gcs-%x", md5.Sum([]byte(roleName)))
		},
	},
}

func preprocessSpecial(gm GenericMount, ctx context.Context, d *schema.ResourceData, 
	m interface{}) error {
	r2 := clusters.ResourceCluster()
	dummyClusterData := r2.TestResourceData()
	dummyClusterData.MarkNewResource()
	for k, v := range mountWithRole {
		block, attr := v.path[0], v.path[1]
		split := strings.SplitN(k, ".", 2)
		specialStorageType := split[0]
		_, ok := d.GetOk(specialStorageType)
		// if there's s3 or gs block or mount source url starts with it
		if !ok || !strings.HasPrefix(gm.URI, specialStorageType) {
			continue
		}
		clustersAPI := clusters.NewClustersAPI(ctx, m)
		if gm.ClusterID != "" {
			clusterInfo, err := clustersAPI.Get(gm.ClusterID)
			if err != nil {
				return err
			}
			common.StructToData(clusterInfo, r2.Schema, dummyClusterData)
			roleName := dummyClusterData.Get(fmt.Sprintf("%s.0.%s", block, attr)).(string)
			if roleName == "" {
				return fmt.Errorf("cluster %s must have %s attached", gm.ClusterID, attr)
			}
		}
		roleName, ok := d.GetOk(k)
		if !ok {
			return fmt.Errorf("must have %s attached", attr)
		}
		dummyClusterData.Set(block, []interface{}{
			map[string]interface{}{
				attr: roleName,
			},
		})
		clusterName := v.clusterNamer(roleName.(string))
		cluster := getCommonClusterObject(clustersAPI, clusterName)
		common.DataToStructPointer(dummyClusterData, r2.Schema, &cluster)
		clusterInfo, err := clustersAPI.GetOrCreateRunningCluster(clusterName, cluster)
		if err != nil {
			return err
		}
		return d.Set("cluster_id", clusterInfo.ClusterID)
	}
	return nil
}
```

alexott:
I really wanted to wait until we remove deprecated resources, and do the refactoring

nfx:
@alexott ok, not touching that part yet then

PR #1000, opened 2021-12-22
Added `auth_type` for multi-cloud states

nfx:
Added optional `auth_type` provider conf to enforce specific auth type to be used in very rare cases, where a single Terraform state manages Databricks workspaces on more than one cloud and `More than one authorization method configured` error is a false positive. Valid values are `pat`, `basic`, `azure-client-secret`, `azure-msi`, `azure-cli`, and `databricks-cli`.

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -61,6 +61,10 @@ type DatabricksClient struct {
 	AzureTenantID      string `name:"azure_tenant_id" env:"ARM_TENANT_ID" auth:"azure"`
 	AzurermEnvironment string `name:"azure_environment" env:"ARM_ENVIRONMENT"`
 
+	// When multiple auth attributes are available in the environment, use the auth type
+	// specified by this argument. This argument also holds currently selected auth.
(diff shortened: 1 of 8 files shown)


common/client.go
```diff
@@ -233,28 +237,39 @@ func (c *DatabricksClient) Authenticate(ctx context.Context) error {
...
+			continue
+		}
 		authorizer, err := auth.configure(ctx)
 		if err != nil {
 			return c.niceAuthError(fmt.Sprintf("cannot configure %s auth: %s", auth.name, err))
 		}
 		if authorizer == nil {
 			continue
 		}
+		// even though this may complain about clear text loggin, passwords are replaced with `***`
```
alexott, line 263: `loggin` -> `logging`

alexott (approved):
lgtm

common/client.go
nfx: thanks, in the next pr

PR #1010, opened 2021-12-29
[exporter] interactive cli and notebooks

nfx:
Exporter has low adoption so far and it's related to the difficulty working with it. it's target use is interactive one-off export of PVC and ST deployments, so admin persona most likely will launch it in interactive mode without any env vars configured upfront. let's cater for this persona to roll out export with minimal SA assistance.

[asciicast](https://asciinema.org/a/Rv8ZFJQpfrfp6ggWddjtyXaOy)


exporter/importables.go
```diff
@@ -817,4 +817,65 @@ var resourcesMap map[string]importable = map[string]importable{
...
+		},
+		List: func(ic *importContext) error {
+			notebooksAPI := workspace.NewNotebooksAPI(ic.Context, ic.Client)
+			notebookList, err := notebooksAPI.List("/", true)
+			if err != nil {
+				return err
+			}
+			// TODO: emit permissions for notebook folders if non-default,
+			// as per-notebook permission entry would be a noise in the state
+			for offset, notebook := range notebookList {
```
alexott, line 840: We need to exclude `/Repos` from that list.

exporter/importables.go
```diff
@@ -817,4 +817,65 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			}
+			return nil
+		},
+		Body: func(ic *importContext, body *hclwrite.Body, r *resource) error {
+			notebooksAPI := workspace.NewNotebooksAPI(ic.Context, ic.Client)
+			status, err := notebooksAPI.Read(r.ID)
+			if err != nil {
+				return err
+			}
+			contentB64, err := notebooksAPI.Export(r.ID, "SOURCE")
```
alexott, line 858: I would prefer to try to export as DBC first as it will keep results & things like dashboards/graphs that are lost when you use SOURCE mode.

alexott:
I'm really not sure about interactivity by default...

exporter/importables.go
nfx: but how much DBC exports is helping with migration to TF?.. the other concern is granularity of DBC - per user? what about the non-/Users folders, that are sometimes used?

exporter/importables.go
nfx: is it listed in workspace api?... mkay...

exporter/importables.go
alexott: You can get individual notebooks as DBC

alexott: Maybe export format should be configurable?

nfx:
@alexott why not interactivity by default? it's too difficult as it is now. breaking change - of course it is, but exporter is experimental, so we can do that.

exporter/importables.go
alexott: Yep, just re-checked with databricks cli

exporter/importables.go
nfx: if we're solving the problem of "re-architect this workspace for disaster recovery" or "migrate ST or PVC into MT" problems via TF, then we should provide good use-case by use-case migration tactics. If notebooks are considered application code, then it should be exported as source. repos are for everything else.

DBC is only meant for exporting bulks of notebooks, like Databricks Academy or Spark Summit cases.

exporter/importables.go
nfx: thanks!

alexott:
I just thought about this PR again. I see two additional issues:
* Permissions aren't emitted for notebooks
* We need to export workspace directories as well, and permissions for them. Otherwise, when we import everything back (for example, into new workspace), then permissions for directories will be set to admin

exporter/exporter_test.go
```diff
@@ -223,6 +224,11 @@ func TestImportingUsersGroupsSecretScopes(t *testing.T) {
 			meAdminFixture,
 			repoListFixture,
 			emptyGitCredentials,
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/workspace/list?path=%2F",
+				Response: workspace.ObjectList{},
+			},
```
alexott, line 231: move into a separate fixture, like it's done for `emptyGitCredentials`, etc.?

exporter/importables.go
```diff
@@ -148,7 +148,7 @@ var resourcesMap map[string]importable = map[string]importable{
 			}
 			// libraries installed with init scripts won't be exported.
 			b := body.AppendNewBlock("resource", []string{r.Resource, r.Name}).Body()
-			relativeFile := fmt.Sprintf("${path.module}/files/%s", fileName)
+			relativeFile := fmt.Sprintf("${path.module}/%s", fileName)
```
alexott, line 151: won't it create a mix of too many files in the output location?

exporter/importables.go
```diff
@@ -896,4 +903,68 @@ var resourcesMap map[string]importable = map[string]importable{
...
+		Service: "notebooks",
+		Name: func(d *schema.ResourceData) string {
+			name := d.Get("path").(string)
+			if name == "" {
+				return d.Id()
+			} else {
+				name = strings.TrimPrefix(name, "/")
+			}
+			re := regexp.MustCompile(`[^0-9A-Za-z_]`)
+			return strings.ToLower(re.ReplaceAllString(name, "_"))
```
alexott, line 916: isn't it done already by normalizer in the context? it was your comment yesterday ;-)

exporter/importables.go
```diff
@@ -896,4 +903,68 @@ var resourcesMap map[string]importable = map[string]importable{
...
+				return err
+			}
+			// libraries installed with init scripts won't be exported.
+			b := body.AppendNewBlock("resource", []string{r.Resource, r.Name}).Body()
+			relativeFile := fmt.Sprintf("${path.module}/%s", fileName)
+			b.SetAttributeValue("path", cty.StringVal(r.ID))
+			b.SetAttributeRaw("source", hclwrite.Tokens{
+				&hclwrite.Token{Type: hclsyntax.TokenOQuote, Bytes: []byte{'"'}},
+				&hclwrite.Token{Type: hclsyntax.TokenQuotedLit, Bytes: []byte(relativeFile)},
+				&hclwrite.Token{Type: hclsyntax.TokenCQuote, Bytes: []byte{'"'}},
```
alexott, line 965: Create a function for tokens generation, as we have the same code in the DBFS file?

exporter/importables.go
nfx: There'll be two folders - files and notebooks.

exporter/importables.go
nfx: ToLower definitely not done, good idea to move it there tomorrow 😜

alexott: We need to remove that explicit normalization from more places, I see it in `databricks_global_init_script` & `databricks_repo`

exporter/exporter_test.go
nfx: a bit of overkill for 2 repetitions, but fine

alexott (approved):
looks good to me

PR #1011, opened 2021-12-30
Added explanation to HTTP 403 errors

### LINKED ISSUE #821: Handle HTTP 403 responses nicer
Debugging authentication with 20 different parameters is frequently frustrating, especially for TF first-time users. 

Enrich every HTTP 403 error with [niceError](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/common/client.go#L243-L254) wrapper. As well as adding what authorizer was used for specific request, as people frequently forget about `~/.databrickscfg` having configured profiles.

cc @pohlposition

nfx:
Added `Using XXX auth` explanation to HTTP 403 errors, which should help troubleshooting misconfigured authentication or provider aliasing. Example error message now looks like: *cannot create group: /2.0/preview/scim/v2/Groups is only accessible by admins. Using databricks-cli auth: host=https://XXX.cloud.databricks.com/, token=`***REDACTED***`, profile=demo.* All sensitive configuration parameters (`token`, `password`, and `azure_client_secret`) are redacted and replaced with `***REDACTED***`

Fixes #821

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.4.3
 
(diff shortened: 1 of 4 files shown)


alexott (approved):
Lgtm

PR #1019, opened 2021-12-31
Add debugger mode for the binary

nfx:
You can [run provider in a debug mode](https://www.terraform.io/plugin/sdkv2/debugging#running-terraform-with-a-provider-in-debug-mode) from VScode IDE by launching `Debug Provider` run configuration and invoking `terraform apply` with `TF_REATTACH_PROVIDERS` environment variable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,9 @@ After this, you should be able to run `make coverage` to run the tests and see t
 
 ## Debugging
 
-**TF_LOG=DEBUG terraform apply** allows you to see the internal logs from `terraform apply`.
+**TF_LOG_PROVIDER=DEBUG terraform apply** allows you to see the internal logs from `terraform apply`.
+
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -69,7 +69,9 @@ After this, you should be able to run `make coverage` to run the tests and see t
 
 ## Debugging
 
-**TF_LOG=DEBUG terraform apply** allows you to see the internal logs from `terraform apply`.
+**TF_LOG_PROVIDER=DEBUG terraform apply** allows you to see the internal logs from `terraform apply`.
```
alexott, line 72: maybe add a note that it's TF 0.15.0+ variable?

alexott (approved):
lgtm

CONTRIBUTING.md
nfx: more than 50% of customers use 1.0+, so it's safe enough. Also i don't suspect any contributors to run something older than 1.0 or 1.1

PR #1021, opened 2022-01-04
Adding new experimental resources

nfx:


catalog/resource_external_location.go
```diff
@@ -0,0 +1,84 @@
...
+	return a.client.Post(a.context, "/unity-catalog/external-locations", el, &el)
+}
+
+func (a ExternalLocationsAPI) get(id string) (el ExternalLocationInfo, err error) {
+	err = a.client.Get(a.context, "/unity-catalog/external-locations/"+id, nil, &el)
+	return
+}
+
+func (a ExternalLocationsAPI) update(name string, el ExternalLocationInfo) error {
+	return a.client.Patch(a.context, "/unity-catalog/external-locations/"+name, el)
```
alexott, line 38: URLencode name? and is it really name in the API? :confused:

catalog/resource_external_location.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (a ExternalLocationsAPI) delete(name string) error {
+	return a.client.Delete(a.context, "/unity-catalog/external-locations/"+name, nil)
```
alexott, line 42: same here

catalog/resource_external_location.go
```diff
@@ -0,0 +1,84 @@
...
+	Owner          string `json:"owner,omitempty" tf:"computed"`
+	MetastoreID    string `json:"metastore_id,omitempty" tf:"computed"`
+}
+
+func (a ExternalLocationsAPI) create(el *ExternalLocationInfo) error {
+	return a.client.Post(a.context, "/unity-catalog/external-locations", el, &el)
+}
+
+func (a ExternalLocationsAPI) get(id string) (el ExternalLocationInfo, err error) {
+	err = a.client.Get(a.context, "/unity-catalog/external-locations/"+id, nil, &el)
```
alexott, line 33: here as well, if id == name

catalog/resource_external_location.go
nfx: yes, it's name. alphanumeric validation may be a better option than urlencoding. i don't think we allow emojis in names here...

alexott: External locations support any SQL identifier and any character is allowed if you put them into back quotes ([doc](https://docs.databricks.com/sql/language-manual/sql-ref-identifiers.html#parameters))

alexott (commented):
Lgtm, can’t check privileges mapping right now

alexott (approved):
Lgtm

PR #1024, opened 2022-01-05
Skip lib refresh for clusters with no libs

nfx:
Allowed managing of libraries on `databricks_cluster` outside of Terraform state for resources without any `library` configuration blocks.

**Behavior changes**

* Whenever library is installed on `databricks_cluster` without any [`library` configuration blocks](https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs/resources/cluster#library-configuration-block), it won't be removed anymore.


alexott (approved):
lgtm

PR #1043, opened 2022-01-18
Explicitly set charset for SCIM requests

### LINKED ISSUE #1034: Creating databricks groups with æøå does not work
**databricks provider:** 0.4.4
**Issue:** creating groups containing special characters (for example æøå ) gets wrong/different encoding in Databricks Workspace.

Example:
resource "databricks_group" "æøå-test" {
  display_name          = "æøå"
}

causes issue:
```
{
    "@level": "warn",
    "@message": "Provider \"provider[\\\"registry.terraform.io/databrickslabs/databricks\\\"]\" produced an unexpected new value for databricks_group.æøå-test, but we are tolerating it because it is using the legacy plugin SDK.
      - .display_name: was cty.StringVal(\"æøå\"), but now cty.StringVal(\"Ã¦Ã¸Ã¥\")",
    "@timestamp": "2022-01-13T12:30:04.589586Z"
}
```

Making the name not match between configuration and Databricks Workspace deployed groups and configuration providing changes on every plan:
```
  # databricks_group.æøå-test must be replaced
-/+ resource "databricks_group" "æøå-test" {
      ~ display_name               = "Ã¦Ã¸Ã¥" -> "æøå" # forces replacement
      ~ id                         = "944677071884272" -> (known after apply)
    }
```

alexott:
This fixes #1034

Also, added printing of response headers when debugging is enabled

### DIFF
--- a/common/http.go
+++ b/common/http.go
@@ -343,7 +343,7 @@ func (c *DatabricksClient) completeUrl(r *http.Request) error {
 // Scim sets SCIM headers
 func (c *DatabricksClient) Scim(ctx context.Context, method, path string, request interface{}, response interface{}) error {
 	body, err := c.authenticatedQuery(ctx, method, path, request, c.completeUrl, func(r *http.Request) error {
-		r.Header.Set("Content-Type", "application/scim+json")
+		r.Header.Set("Content-Type", "application/scim+json; charset=utf-8")
 		if c.isAccountsClient() && c.AccountID != "" {
 			// until `/preview` is there for workspace scim
 			r.URL.Path = strings.ReplaceAll(path, "/preview", fmt.Sprintf("/api/2.0/accounts/%s", c.AccountID))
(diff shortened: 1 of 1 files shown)


common/http.go
```diff
@@ -486,7 +487,16 @@ func (c *DatabricksClient) genericQuery(ctx context.Context, method, requestURL
...
-	log.Printf("[DEBUG] %s %v <- %s %s", resp.Status, c.redactedDump(body), method, request.URL.Path)
+	headers = ""
+	if c.DebugHeaders {
+		for k, v := range resp.Header {
+			headers += fmt.Sprintf("\n * %s: %s", k, onlyNBytes(strings.Join(v, ""), c.DebugTruncateBytes))
+		}
+		if len(headers) > 0 {
+			headers += "\n"
+		}
+	}
```
nfx, line 498: can you make it into a function, as lines 455-464 are exactly the same.

nfx (changes requested):
please fix code duplication

common/http.go
alexott: ok

nfx (approved):

PR #1067, opened 2022-01-28
Implicitly force import manually added group

### LINKED ISSUE #1066: force ignore existing groups - like for users
Federation using SCIM can pre create  groups that will cause TF to fail if they exist.
The provider should have the `force` option for **groups** to ignore `cannot create group: ... already exists` errors and implicitly import the specific group into Terraform state.

Same problem, accepted and solved via #1048 for **users**, should be solved same way for **groups** as well.

Using `data "databricks_group"` data provider is not possible if the existence of the group is not known upfront. The data provider should also gracefully fail.

### Expected Behavior

allowing to ignore that case and implicitly import the specific group into Terraform state.

### Actual Behavior
erroring out

### Steps to Reproduce
1. `terraform apply`

v0.4.2

nkvuong:
Add `force` parameter to `databricks_group`, so that SCIM sync corner cases might be simplified

image

Closes #1066

### DIFF
--- a/scim/resource_group.go
+++ b/scim/resource_group.go
@@ -3,6 +3,7 @@ package scim
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/databrickslabs/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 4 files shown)


scim/resource_group.go
```diff
@@ -61,3 +68,23 @@ func ResourceGroup() *schema.Resource {
 		Schema: groupSchema,
 	}.ToResource()
 }
+
+func createForceOverridesManuallyAddedGroup(err error, d *schema.ResourceData, groupsAPI GroupsAPI, g Group) error {
+	forceCreate := d.Get("force").(bool)
+	if !forceCreate {
+		return err
+	}
+	// corner-case for overriding manually provisioned users
```
nfx, line 77: not _users_ ;)

scim/resource_group.go
```diff
@@ -61,3 +68,23 @@ func ResourceGroup() *schema.Resource {
...
+	groupName := g.DisplayName
+	force := fmt.Sprintf("Group with name %s already exists.", groupName)
+	if err.Error() != force {
+		return err
+	}
+	groupList, err := groupsAPI.Filter(fmt.Sprintf("displayName eq '%s'", groupName))
```
nfx, line 83: use `ReadByDisplayName` method instead: https://github.com/databrickslabs/terraform-provider-databricks/blob/2d138fb6b7822f67a13ccbfee7c1c2ad32f54b36/scim/groups.go#L52-L63

nfx (changes requested):
+ change CHANGELOG.md and docs/resources/group.md - make sure to add "experimental" label to it, as in user resource.
+ cover missing line with test (probably test will be simpler if you use ReadByDisplayName)

nfx (approved):

PR #1086, opened 2022-02-03
Added documentation for MLflow Registry Webhooks resource

alexott:


docs/resources/mlflow_webhook.md
```diff
@@ -0,0 +1,116 @@
...
+    spark_version = data.databricks_spark_version.latest.id
+    node_type_id  = data.databricks_node_type.smallest.id
+  }
+
+  notebook_task {
+    notebook_path = databricks_notebook.this.path
+  }
+}
+
+resource databricks_mlflow_webhook "job" {
```
nfx, line 53: add quotes around resource type. try running `make fmt-docs` - maybe it'll do that for you automatically.

docs/resources/mlflow_webhook.md
```diff
@@ -0,0 +1,116 @@
...
+  }
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `model_name` - (Optional) Name of MLflow model for which webhook will be created. If model name is not specified, a registry-wide webhook is created that listens for the specified events across all versions of all registered models
```
nfx, line 82: add dot in the end of the line

nfx (changes requested):
formatting changes required

docs/resources/mlflow_webhook.md
alexott: done, also additional files were reformatted

nfx: hm, it did not add quotes around resource type. definition must be `resource "databricks_mlflow_webhook" "job" {` ...

nfx (changes requested):

docs/resources/mlflow_webhook.md
alexott: I need to write 1000 times "don't multitask" :-)

nfx (approved):

PR #1090, opened 2022-02-04
[FEATURE]: Adds support for git source in preview to job resource

edmondo1984:
This introduce the new git_source support for job API in preview to Terraform Provider.

It's a really powerful feature for deployment automation and CI/CD. To work around the absence of this feature we had developed internally a mix of python scripts and terraform code. The main hurdle for us was that pulling from Github on the workspace would fail whenever history has been rewritten, so you would have to check out the repository either to a new folder or destroy the previous repo and job and try again.

I have no idea of what level of testing is needed for this change. Would appreciate if a maintainer could provide feedback on the topic so I can add tests and resubmit the PR


common/version.go
```diff
@@ -3,7 +3,7 @@ package common
 import "context"
 
 var (
-	version = "0.4.8"
+	version = "0.4.9"
```
alexott, line 6: leave this out - it's usually done separately. For development you can use this: https://www.terraform.io/cli/config/config-file#development-overrides-for-provider-developers

jobs/resource_job.go
```diff
@@ -74,6 +74,17 @@ type TaskDependency struct {
 	TaskKey string `json:"task_key,omitempty"`
 }
 
+// BEGIN Jobs + Repo integration preview
+type GitSource struct {
+	GitUrl      string `json:"git_url"`
```
alexott, line 79: Add validator for this field, like [this](https://github.com/databrickslabs/terraform-provider-databricks/blob/master/mlflow/resource_webhook.go#L99)

jobs/resource_job.go
```diff
@@ -74,6 +74,17 @@ type TaskDependency struct {
...
+	GitProvider string `json:"git_provider"`
```
alexott, line 80: Try to guess provider, like this: https://github.com/databrickslabs/terraform-provider-databricks/blob/master/workspace/resource_repo.go#L49 - reuse `getProviderFromUrl` function (it needs to be exported)

jobs/resource_job.go
```diff
@@ -74,6 +74,17 @@ type TaskDependency struct {
...
+	GitBranch   string `json:"git_branch,omitempty"`
+	GitTag      string `json:"git_tag,omitempty"`
+	GitCommit   string `json:"git_commit,omitempty"`
```
alexott, line 83: Add conflicts for each other, as you can't use `branch` with `tag` or `commit`

jobs/resource_job.go
```diff
@@ -74,6 +74,17 @@ type TaskDependency struct {
 	TaskKey string `json:"task_key,omitempty"`
 }
 
+// BEGIN Jobs + Repo integration preview
+type GitSource struct {
```
alexott, line 78: simplify naming - we already inside `git_source`, why we need to repeat `git_` everywhere? we can just use `url`, `branch`, ...

alexott (changes requested):
also, we'll need tests after adding all these changes

nfx (changes requested):
In addition to previous comments, please add a unit test that creates a job with git source

jobs/resource_job.go
edmondo1984: I am not sure I understood, do you think the field in HCL should be named 'url' instead of git_url?

```go
Url      string `json:"git_url" tf:"alias:url"`
```

edmondo1984: Very helpful, I agree. I imagine tf:"alias:url" sets the name of the property in HCL. Is there somewhere I can read about the struct tags supported by Terraform? I can't find googling it

alexott: Yes, alias will be used as a name in HCL, while json will name the field in JSON object. Really, these things aren’t standard in terraform (although should be, imho) - they were added by Serge in Databricks provider only. Source code is in common/reflect_reaource

edmondo1984: Thanks, would you mind giving a look the structure? I am now writing unit tests for the PR

edmondo1984:
> In addition to previous comments, please add a unit test that creates a job with git source

Will do, what do you test? Correct serialization of the rest API? conflicts handling in the schema?

jobs/resource_job.go
```diff
@@ -120,6 +132,10 @@ type JobSettings struct {
 	Format string            `json:"format,omitempty" tf:"computed"`
 	// END Jobs API 2.1
 
+	// BEGIN Jobs + Repo integration preview
+	GitSource *GitSource `json:"git_source,omitempty" tf:"group:git_source"`
```
nfx, line 136: tf:group not needed here - it's meant for "at least one of these fields" groups.

jobs/resource_job.go
```diff
@@ -321,6 +337,9 @@ func (a JobsAPI) Restart(id string, timeout time.Duration) error {
 func (a JobsAPI) Create(jobSettings JobSettings) (Job, error) {
 	var job Job
 	jobSettings.sortTasksByKey()
+	if jobSettings.GitSource != nil && jobSettings.GitSource.Provider == "" {
+		jobSettings.GitSource.Provider = workspace.GetProviderFromUrl(jobSettings.GitSource.Url)
```
nfx, line 341: rename `GetProviderFromUrl` to `GetGitProviderFromUrl` to have clearer semantics of exported functions

jobs/resource_job.go
```diff
@@ -406,10 +425,20 @@ func jobSettingsSchema(s *map[string]*schema.Schema, prefix string) {
 	}
 }
 
+func gitSourceSchema(s *map[string]*schema.Schema, prefix string) {
```
nfx, line 428: just take `*schema.Resource` :)

jobs/resource_job.go
```diff
@@ -406,10 +425,20 @@ func jobSettingsSchema(s *map[string]*schema.Schema, prefix string) {
...
+	(*s)["tag"].ConflictsWith = []string{"branch", "commit"}
+	(*s)["branch"].ConflictsWith = []string{"commit", "tag"}
+	(*s)["commit"].ConflictsWith = []string{"branch", "tag"}
+}
+
 var jobSchema = common.StructToSchema(JobSettings{},
 	func(s map[string]*schema.Schema) map[string]*schema.Schema {
 		jobSettingsSchema(&s, "")
 		jobSettingsSchema(&s["task"].Elem.(*schema.Resource).Schema, "task.0.")
+		gitSourceSchema(&s["git_source"].Elem.(*schema.Resource).Schema, "")
```
nfx, line 441: i haven't looked at the api spec yet, but why not calling `gitSourceSchema` from `jobSettingsSchema`, so that multi-task jobs are supported as well?

jobs/resource_job.go
```diff
@@ -74,6 +75,17 @@ type TaskDependency struct {
 	TaskKey string `json:"task_key,omitempty"`
 }
 
+// BEGIN Jobs + Repo integration preview
+type GitSource struct {
+	Url      string `json:"git_url" tf:"alias:url"`
+	Provider string `json:"git_provider" tf:"alias:provider"`
+	Branch   string `json:"git_branch,omitempty" tf:"alias:provider"`
```
nfx, line 82: alias:branch

nfx (changes requested):

jobs/resource_job.go
edmondo1984: It's a property of the job and not of the task, you can't have a mixed job git/not-git I think

jobs/resource_job.go
```diff
@@ -74,6 +75,17 @@ type TaskDependency struct {
 	TaskKey string `json:"task_key,omitempty"`
 }
 
+// BEGIN Jobs + Repo integration preview
+type GitSource struct {
+	Url      string `json:"git_url" tf:"alias:url"`
+	Provider string `json:"git_provider" tf:"alias:provider"`
```
nfx, line 81: `provider` is an illegal property name for terraform, because it's a reserved keyword...

@alexott maybe we should keep the non-aliased properties after all, to simplify the support in the long run?

nfx (changes requested):
@edmondo1984 please add resource tests to make sure we have correct serialization. Also add the "acceptance test", where you checkout [this repository](https://github.com/databricks/delta-live-tables-notebooks) as git source, for example.

jobs/resource_job.go
alexott: not sure. I really don't like repetition of the names, because we already in the `git` block

nfx: think about difficulties of our tech support troubleshooting the configs ;)

nfx:
@edmondo1984 okay, spoke to Vaibhav (PM of this feature) - as long as it's private preview, it doesn't need an acceptance test, because preview acceptance infrastructure is a bit difficult right now. Just make a unit test to verify serialization is all right.

edmondo1984:
> @edmondo1984 okay, spoke to Vaibhav (PM of this feature) - as long as it's private preview, it doesn't need an acceptance test, because preview acceptance infrastructure is a bit difficult right now. Just make a unit test to verify serialization is all right.
> 
> And make sure to manually test it before pushing last changes ;)

I am a junior Go developer, how about writing an acceptance test that is explicitly tagged "private-preview" so we can filter it out?

nfx:
@edmondo1984 The way to tag it is to prefix with TestPreviewAcc. You can see examples how it's done in order places. See contributing.md and scripts/readme.md for references

edmondo1984:
> @edmondo1984 The way to tag it is to prefix with TestPreviewAcc. You can see examples how it's done in order places. See contributing.md and scripts/readme.md for references

This is useful. I notice that since I added validation my code doesn't work anymore and as I wrote unit tests the code is failing like so:

"ConflictsWith: branch references unknown attribute (commit) at part (commit)"

Do you see anything obviously wrong here https://github.com/edmondo1984/terraform-provider-databricks/blob/support-repojobapi/jobs/resource_job.go#L432 or here https://github.com/edmondo1984/terraform-provider-databricks/blob/support-repojobapi/jobs/resource_job.go#L441  ?

nfx:
@edmondo1984 You have to specify full path - git_source.0.commit - in the list of conflicts or simply can defer validation to backend

edmondo1984:
> @edmondo1984 You have to specify full path - git_source.0.commit - in the list of conflicts or simply can defer validation to backend

Thank you. This works but it sounds to me like a leap of faith, why wasn't it required here?
https://github.com/databrickslabs/terraform-provider-databricks/blob/master/clusters/resource_cluster.go#L82

Also, if you have time, could you explain me what deferring validation to the backend would have meant? Just let the API call fail?

jobs/resource_job.go
```diff
@@ -74,6 +75,17 @@ type TaskDependency struct {
 	TaskKey string `json:"task_key,omitempty"`
 }
 
+// BEGIN Jobs + Repo integration preview
+type GitSource struct {
+	Url      string `json:"git_url" tf:"alias:url"`
+	Provider string `json:"git_provider,omitempty" tf:"alias:provider"`
```
nfx, line 81: remove alias:provider, because `provider` is a reserved TF keyword...

jobs/resource_job.go
```diff
@@ -321,6 +337,16 @@ func (a JobsAPI) Restart(id string, timeout time.Duration) error {
 func (a JobsAPI) Create(jobSettings JobSettings) (Job, error) {
 	var job Job
 	jobSettings.sortTasksByKey()
+	var gitSource *GitSource = jobSettings.GitSource
+	if gitSource != nil && gitSource.Provider == "" {
+		gitSource.Provider = workspace.GetGitProviderFromUrl(gitSource.Url)
+		if gitSource.Provider == "" {
+			return job, fmt.Errorf("git Source is not empty but Git Provider is not specified and cannot be guessed by url %+v", gitSource)
```
nfx, line 344: %s instead of %+v

jobs/resource_job_test.go
```diff
@@ -429,6 +430,161 @@ func TestResourceJobCreateNWorkers(t *testing.T) {
...
+			task_key = "b"
+
+			notebook_task {
+				notebook_path = "/GitSourcedNotebook"
+			}
+		}
+	`,
+	}.Apply(t)
+	assert.Error(t, err, err)
+	require.Equal(t, true, strings.Contains(err.Error(), "git Source is not empty but Git Provider is not specified and cannot be guessed by url &{Url:https://custom.git.hosting.com/databrickslabs/terraform-provider-databricks Provider: Branch: Tag:0.4.8 Commit:}"))
```
nfx, line 585: imagine an average devops looking at this kind of long error ;) please make the message more concise. also there's inconsistent use of Title Case - e.g. `git Source` should be `git source`. code style conventions of golang 🤷‍♂️

jobs/resource_job_test.go
```diff
@@ -429,6 +430,161 @@ func TestResourceJobCreateNWorkers(t *testing.T) {
...
+	`
+	var hcl = fmt.Sprintf(hclTemplate, gitSource)
+	_, err := qa.ResourceFixture{
+		Create:   true,
+		Resource: ResourceJob(),
+		HCL:      hcl,
+	}.Apply(t)
+	assert.Error(t, err, err)
+
+	var expectedErr = "invalid config supplied."
```
nfx, line 526: looks a bit strange with this concatenation. and a bit fragile to changing error message. generally i tend to create `var found bool` and have assertions in the loop, then `assert.True(t, found)` after the loop. it'll be more readable than concatenation.

jobs/resource_job_test.go
```diff
@@ -429,6 +430,161 @@ func TestResourceJobCreateNWorkers(t *testing.T) {
...
+		}
+
+		task {
+			task_key = "b"
+
+			notebook_task {
+				notebook_path = "/GitSourcedNotebook"
+			}
+		}`,
+	}.Apply(t)
```
nfx, line 498: use either `.ApplyNoError` or `.ApplyAndExpectData`. two lines in the end look a bit ugly :)

jobs/resource_job.go
```diff
@@ -406,10 +432,20 @@ func jobSettingsSchema(s *map[string]*schema.Schema, prefix string) {
 	}
 }
 
+func gitSourceSchema(s *schema.Resource, prefix string) {
+	if p, err := common.SchemaPath(s.Schema, prefix, "url"); err == nil {
```
nfx, line 436: prefix is always an empty string, so why wouldn't you just `s["url"].ValidateFunc = validation.IsURLWithHTTPS`?...

jobs/resource_job.go
```diff
@@ -406,10 +432,20 @@ func jobSettingsSchema(s *map[string]*schema.Schema, prefix string) {
 	}
 }
 
+func gitSourceSchema(s *schema.Resource, prefix string) {
```
nfx, line 435: one-letter variable names should be the first letters of types they refer to - at least that's how most of people read it in the middle of the night ;) so `r *schema.Resource`. Otherwise a longer name :)

jobs/resource_job.go
```diff
@@ -321,6 +337,16 @@ func (a JobsAPI) Restart(id string, timeout time.Duration) error {
...
 	var job Job
 	jobSettings.sortTasksByKey()
+	var gitSource *GitSource = jobSettings.GitSource
+	if gitSource != nil && gitSource.Provider == "" {
+		gitSource.Provider = workspace.GetGitProviderFromUrl(gitSource.Url)
+		if gitSource.Provider == "" {
+			return job, fmt.Errorf("git Source is not empty but Git Provider is not specified and cannot be guessed by url %+v", gitSource)
+		}
+		if gitSource.Branch == "" && gitSource.Tag == "" && gitSource.Commit == "" {
+			return job, fmt.Errorf("git Source is not empty but none of branch, commit and tag is specified")
```
nfx, line 347: `git Source` -> `git source` for consistency

nfx (changes requested):
almost there. and don't forget to run `make fmt` in the end.

jobs/resource_job.go
edmondo1984: Would you please explain me why is this preferrable?

jobs/resource_job_test.go
```diff
@@ -429,6 +430,157 @@ func TestResourceJobCreateNWorkers(t *testing.T) {
...
+
+		task {
+			task_key = "b"
+
+			notebook_task {
+				notebook_path = "/GitSourcedNotebook"
+			}
+		}
+	`,
+	}.Apply(t)
```
nfx, line 579: there's a method `.ExpectError(t, "error string")` for that :)

jobs/resource_job.go
nfx: @edmondo1984 did you test creating real resource with locally built binary? here's how to do it https://github.com/databrickslabs/terraform-provider-databricks/blob/master/CONTRIBUTING.md#developing-provider

i just want to make sure `provider` attribute name is not clashing with a reserved tf keyword

nfx (changes requested):
on the finish line =) two comments remaining

nfx:
@edmondo1984 Will do somewhere next week. We have this whole week of company event meetings, so availability is quite limited for those couple of days.

I'd also need to enable a dynamic feature flag on our tf nightly testing workspace.

nfx (approved):
lgtm so far, let's test it on master

PR #1091, opened 2022-02-04
Prevent creation of`users` and `admins` groups

### LINKED ISSUE #1089: [ISSUE] The databricks_group resource shouldn't allow you to specify system groups
### Configuration

In v0.4.7 of the provider, the `databricks_group` resource has the `force` argument. This allows you to do, for example:

```hcl
resource "databricks_group" "system" {
  for_each = toset(["admins", "user"])

  display_name = each.value
  force        = true
}
```

### Expected Behavior

The provider should forbid this.

### Actual Behavior

This works fine and it will even reflect new argument values to the group (e.g., `allow_cluster_create`, etc.). The problem comes when you try to destroy this resource: The Databricks API forbids this and so `terraform destroy` will fail.

### Steps to Reproduce

1. `terraform apply`
2. `terraform destroy`

```
Terraform v1.0.11
on linux_amd64
+ provider registry.terraform.io/databrickslabs/databricks v0.4.7
```

nfx:
image

Fix #1089

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Version changelog
 
+## 0.4.9
+
+* Prevent creation of `databricks_group` with `users` and `admins` reserved names ([#1089](https://github.com/databrickslabs/terraform-provider-databricks/issues/1089)).
+
 ## 0.4.8
 
 * Added support for `tf:suppress_diff` on primitive types ([#984](https://github.com/databrickslabs/terraform-provider-databricks/issues/984)).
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm, maybe only requires `make fmt`

nfx:
@alexott did run fmt, no changes

ebarault:
@nfx hi!
arfff, i just discovered this while upgrading the databricks terraform provider version to latest

this has one main drawback/side effect:  consider you want to deal with admins group memberships in terraform, and for this you imported the admins group within your modules.

something like
image

now, this piece of code won't work since it raises this error
> Error: expected display_name to not be any of [users admins], got admins

ebarault:
i guess, I need now to refactor this to use the group data source
https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/group#example-usage

PR #1092, opened 2022-02-04
Rename sqlanalytics -> sql

pietern:
This is a mechanical change:

```shell
git mv sqlanalytics sql
perl -pi -e 's/sqlanalytics/sql/g' **/*
```

After this change `git grep sqlanalytics` returns nothing.


nfx (approved):
Lgtm

PR #1100, opened 2022-02-11
Suppress diff for `databricks_group`.`external_id`

### LINKED ISSUE #1099: [ISSUE] external_id Databricks Group and SCIM provisioning status
### Configuration
```hcl
resource "databricks_group" "MY_GROUP" {
  display_name = "MY_GROUP"
}
```

### Expected Behavior
After the first create of the group in Databricks, users are populated through SCIM provisioning correctly. Performing an apply with the same terraform code should produce no changes.

### Actual Behavior
Further plans and apply lead to changes to the group object. This leads to destroy and then create replacement, resulting in an empty group on Databricks side. At this point, SCIM provisioning should be forced again to populate the group, leading to an outage for users which were provisioned inside the workspace through SCIM, because they loose access and permissions.

```hcl
resource "databricks_group" "MY_GROUP" {
      - external_id                = "84e4d0c3-xxxx-xxxx-xxxx-cf5ab19d6e35" -> null # forces replacement
      ~ id                         = "YYYYY" -> (known after apply)
      ~ url                        = "[https://adb-XXXX.azuredatabricks.net/#setting/accounts/groups/YYYYY"](https://adb-XXXX.azuredatabricks.net/#setting/accounts/groups/YYYYY%22) -> (known after apply)
    }
```

### Steps to Reproduce
1. Add Databricks group and perform an apply
2. Execute SCIM provisioning for the new group
3. Perform another `terraform apply` without editing terraform code

Version 0.4.8

nfx:
Fixes #1099

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## 0.4.9
 
 * Prevent creation of `databricks_group` with `users` and `admins` reserved names ([#1089](https://github.com/databrickslabs/terraform-provider-databricks/issues/1089)).
-* Shared job cluster functionality added. ([#1082](https://github.com/databrickslabs/terraform-provider-databricks/issues/1082))
+* Added support for shared clusters in multi-task `databricks_job` ([#1082](https://github.com/databrickslabs/terraform-provider-databricks/issues/1082)).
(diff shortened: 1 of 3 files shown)


alexott (approved):
lgtm

PR #1102, opened 2022-02-11
Suppress diff for `databricks_user`.`external_id`

### LINKED ISSUE #1097: [ISSUE] external_id for databricks_user with force=true isn't set
### Configuration
```hcl
resource "databricks_user" "foo" {
  user_name = "foo@bar.com"
  force     = true
}
```

### Expected Behavior

Upon apply, the pre-existing user is imported into TF state, and running `apply` again doesn't show state skew.

### Actual Behavior

The next `apply` shows state skew showing that the `external_id` property should be set to `null`.

When applied, this doesn't happen. I suspect `external_id` must be set as `computed`, or must be imported if is isn't set.

0.4.7

nfx:
Fixes #1097


alexott (approved):

pietern:
Nice, thanks @nfx !

PR #1124, opened 2022-02-16
Fixed a doc for SQL Endpoint + allow to disable Photon

alexott:
It was impossible to disable Photon if necessary, because it was declared as `omitempty`, so `false` value wasn't sent, and then default `true` was used instead


nfx (approved):
Thank you

PR #1125, opened 2022-02-16
Added exporter for IP Access Lists

alexott:


exporter/importables.go
```diff
@@ -845,4 +847,70 @@ var resourcesMap map[string]importable = map[string]importable{
...
 		},
 	},
+	"databricks_workspace_conf": {
+		Service: "workspace",
+		Name: func(d *schema.ResourceData) string {
+			return globalWorkspaceConfName
+		},
+		Import: func(ic *importContext, r *resource) error {
+			wsConfAPI := workspace.NewWorkspaceConfAPI(ic.Context, ic.Client)
+			keys := map[string]interface{}{"enableIpAccessLists": false}
```
nfx, line 857: this is very limited wsconf exporter. add a way to add more fields there.

nfx (changes requested):
please remove custom body for ip acls - we don't need `depends_on` attribute.

exporter/importables.go
alexott: I didn't want to go with more than officially documented

exporter/importables.go
```diff
@@ -845,4 +847,70 @@ var resourcesMap map[string]importable = map[string]importable{
...
+					Data: ic.Resources["databricks_workspace_conf"].Data(
+						&terraform.InstanceState{
+							ID:         globalWorkspaceConfName,
+							Attributes: map[string]string{},
+						}),
+				})
+			}
+			return nil
+		},
+		Body: func(ic *importContext, body *hclwrite.Body, r *resource) error {
```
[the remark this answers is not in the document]
alexott: ah, then it makes even simpler

exporter/importables.go
nfx: https://registry.terraform.io/providers/databrickslabs/databricks/latest/docs/resources/workspace_conf there are few more ;)

nfx (approved):

exporter/importables.go
alexott: added

PR #1145, opened 2022-03-01
Update databricks_mws_networks Resource doc

yessawab:
Updated the documentation databricks_mws_networks resource in order to clarify that the public subnet can be limited to few IP addresses by adding a secondary cidr_block to the managed vpc.


nfx (approved):
lgtm

PR #1178, opened 2022-03-11
Treat auto-purged `databricks_cluster` as removed

### LINKED ISSUE #1177: [ISSUE] Project with cluster terminated 30+ days ago made unusable.
Hi there,

It appears there might be an issue with terraform projects containing unpinned clusters that have been terminated 30+ days ago. Indeed, because the cluster has been permanently removed from the workspace, the provider, upon calling the Cluster API, will get an error message and won't progress further.

Such a resource seemingly can't be deleted by being removed from the project declaration, and the provider won't try to recreate it either, it will simply return the following error message:

> │ Error: cannot read cluster: Cannot access cluster that was terminated or unpinned more than 30 days ago.
> │ Error: Cannot access cluster <cluster_id> that was terminated or unpinned more than 30 days ago.

This, as far as I can tell, essentially bricks the project.

### Configuration

though probably not that relevant, here is my cluster configuration:

```hcl
resource "databricks_cluster" "project_etl_cluster" {
  cluster_name            = "project_ETL"
  spark_version           = "9.1.x-scala2.12"
  node_type_id            = "i3.4xlarge"
  driver_node_type_id     = "i3.xlarge"
  autotermination_minutes = 20
  enable_local_disk_encryption = true

  spark_env_vars          = {
    PYSPARK_PYTHON = "/databricks/python3/bin/python3"
  }
  spark_conf              = {
    "spark.databricks.hive.metastore.glueCatalog.enabled": true
    "spark.databricks.adaptive.autoOptimizeShuffle.enabled": true
    "spark.sql.adaptive.optimizeSkewedJoin.enabled": true
    "spark.databricks.io.cache.enabled": true
    "spark.databricks.io.cache.maxDiskUsage": "50g"
    "spark.databricks.delta.schema.autoMerge.enabled": true    
    "spark.databricks.io.cache.maxMetaDataCache": "2g"    
  }
  autoscale {
    min_workers = 1
    max_workers = 8
  }
  aws_attributes {
    instance_profile_arn    = local.instance_profile_arn
    availability            = "SPOT_WITH_FALLBACK"
    first_on_demand         = 1
    spot_bid_price_percent  = 100
    zone_id                 = "eu-west-1"
  }

  library {
    pypi {
      package = "Shapely>=1.7.0"
    }
  }

   library {
    pypi {
      package = "libgeohash>=0.1.1"
    }
  }
```

### Expected Behavior
Ideally, the provider should probably treat that specific error message as a "create" when the command `terraform plan` is ran, and  recreate a new cluster with the same configurations and name on `terraform apply`. It shouldn't generate any conflict since the previous one is long destroyed.

### Actual Behavior
Seemingly permanently bricked pipeline?

### Steps to Reproduce
1. Have a terraform project with an unpinned cluster terminated 30+ days ago and thus deleted.
2. `terraform plan`

Terraform v1.0.1
Provider V0.5.2

### Factoids

If I would have to guess, I would say it's related on how [resourceClusterRead manages error on clusterAPI.Get() 
](https://github.com/databrickslabs/terraform-provider-databricks/blob/43b4996851ed5df75f97c4eb2bb4174432532b10/clusters/resource_cluster.go#L187).

Thanks for your time and help!

nfx:
Added special case for handling `Cannot access cluster that was terminated or unpinned more than 30 days ago` error in `databricks_cluster` as an indication of resource removed on the platform side.

Fix #1177

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -600,6 +600,7 @@ func (a ClustersAPI) StartAndGetInfo(clusterID string) (ClusterInfo, error) {
 	return a.waitForClusterStatus(clusterID, ClusterStateRunning)
 }
 
+// make common/resource.go#ToResource read behavior consistent with "normal" resources
 func wrapMissingClusterError(err error, id string) error {
 	if err == nil {
 		return nil
@@ -610,6 +611,14 @@ func wrapMissingClusterError(err error, id string) error {
 	}
 	if apiErr.IsMissing() {
 		return err
+	}
+	// https://github.com/databrickslabs/terraform-provider-databricks/issues/1177
+	// Aligned with Clusters Core team to keep behavior of these workarounds 
+	// as is in the longer term, so that this keeps working.
+	if apiErr.ErrorCode == "INVALID_STATE" {
+		log.Printf("[WARN] assuming that cluster is removed on backend: %s", apiErr)
+		apiErr.StatusCode = 404
+		return apiErr
 	}
 	// fix non-compliant error code
 	if strings.Contains(apiErr.Message,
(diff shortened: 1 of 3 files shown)


clusters/clusters_api.go
```diff
@@ -611,6 +612,14 @@ func wrapMissingClusterError(err error, id string) error {
 	if apiErr.IsMissing() {
 		return err
 	}
+	// https://github.com/databrickslabs/terraform-provider-databricks/issues/1177
+	// Aligned with Clusters Core team to keep behavior of these workarounds 
+	// as is in the longer term, so that this keeps working.
+	if apiErr.ErrorCode == "INVALID_STATE" {
```
pietern, line 618: Add the expectation that the return code is 400?

pietern (approved):

clusters/clusters_api.go
nfx: yep, that's our `INVALID_STATE` mapping. i'd rather have it as `NOT_FOUND` (imho, this is more semantically correct, as the cluster is **_no longer found_**), so that we don't have to have this if. you're on that slack thread.

pietern: So any response with status 400 gets `ErrorCode = "INVALID_STATE"`?

PR #1196, opened 2022-03-15
fix read MLFlow model with versions

nicolai86:
as described in https://github.com/databrickslabs/terraform-provider-databricks/issues/1195
the terraform provider makes incorrect assumptions about MLFlow API responses for models with versions.

This PR adjusts the schema to what the databricks MLFlow registry actually responds with.

Because this provider doesn't have support for mlflow model versions I can't write tests for this schema change just yet. Adding MLFlow model version support needs some leg work because there seems to be no go sdk for it today.

### DIFF
--- a/mlflow/resource_model.go
+++ b/mlflow/resource_model.go
@@ -14,14 +14,14 @@ type Tag struct {
 
 // ModelVersion defines a MLFlow model version as returned by the API
 type ModelVersion struct {
-	Name                 string `json:"name" tf:"computed"`
-	Version              string `json:"version" tf:"computed"`
-	CreationTimestamp    int64  `json:"creation_timestamp,omitempty" tf:"computed"`
-	LastUpdatedTimestamp int64  `json:"last_updated_timestamp,omitempty" tf:"computed"`
-	UserID               string `json:"user_id,omitempty" tf:"computed"`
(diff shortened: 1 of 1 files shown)


mlflow/resource_model.go
```diff
@@ -12,16 +12,28 @@ type Tag struct {
...
-	UserID               string   `json:"user_id,omitempty" tf:"computed"`
-	LatestVersions       []string `json:"latest_versions,omitempty" tf:"computed"`
-	Description          string   `json:"description,omitempty"`
-	Tags                 []Tag    `json:"tags,omitempty"`
-	RegisteredModelID    string   `json:"id,omitempty" tf:"computed,alias:registered_model_id"`
+	Name                 string         `json:"name" tf:"force_new"`
+	CreationTimestamp    int64          `json:"creation_timestamp,omitempty" tf:"computed"`
+	LastUpdatedTimestamp int64          `json:"last_updated_timestamp,omitempty" tf:"computed"`
+	UserID               string         `json:"user_id,omitempty" tf:"computed"`
+	LatestVersions       []ModelVersion `json:"latest_versions,omitempty" tf:"computed"`
```
nfx, line 33: let me check if this won't proliferate into tfstate, otherwise i'll add ignore for `latest_versions` in schema customization function.

PR #1218, opened 2022-03-24
Feature/add reuse key for cluster volumes to cmk

johnc1996:
- Added new field to Terraform resource `mws_customer_managed_key` to stop the key from being used to encrypt cluster volumes
- The new field defaults to `true` to preserve the existing behaviour of this resource so people don't have to update their code to include the new field if they don't want to use it
- Added test to check that setting the field to `false` returns the correct response


mws/resource_customer_managed_key.go
```diff
@@ -19,11 +19,12 @@ type AwsKeyInfo struct {
...
-	AwsKeyInfo           *AwsKeyInfo `json:"aws_key_info" tf:"force_new"`
-	AccountID            string      `json:"account_id" tf:"force_new"`
-	CreationTime         int64       `json:"creation_time,omitempty" tf:"computed"`
-	UseCases             []string    `json:"use_cases"`
+	CustomerManagedKeyID      string      `json:"customer_managed_key_id,omitempty" tf:"computed"`
+	AwsKeyInfo                *AwsKeyInfo `json:"aws_key_info" tf:"force_new"`
+	AccountID                 string      `json:"account_id" tf:"force_new"`
+	CreationTime              int64       `json:"creation_time,omitempty" tf:"computed"`
+	UseCases                  []string    `json:"use_cases"`
+	ReuseKeyForClusterVolumes bool        `json:"reuse_key_for_cluster_volumes,omitempty" tf:"default:true"`
```
nfx, line 27: please bump the schema version to 2 and do schema version upgrader:

https://github.com/databrickslabs/terraform-provider-databricks/blob/bb43dd35d51870538e6670982d31f73ffb7d6e17/mws/resource_customer_managed_key.go#L103-L157

johnc1996: Version 1 or 2 as current one looks to be 0? Are there docs on what the schema version upgrader is and how I can do that please?

johnc1996: Ah was looking at the version within which is at 0. Will update schema version to 2

johnc1996: Hi, still wanting to check how I do a schema version upgrader?

nfx: @johnc1996 Please see the link in the first comment =)

johnc1996: Have pushed a fix now and have passing tests

johnc1996:
When I have tested this in a Terraform project when it is set to `true` it is causing the field to be added to the resource which causes the resource to be rebuilt and in turn it requires the Databricks workspace to be rebuilt. Can you offer any assistance on how to achieve this addition of not having the customer managed key be used to encrypt cluster volumes without introducing breaking changes?

nfx:
@johnc1996 Please add state upgrader, so that workspace is not rebuilt.

nfx:
@johnc1996 Please add state upgrader, so that workspace is not rebuilt.

nfx:
PR is out of date. please create a new one.

PR #1253, opened 2022-04-21
Refactor `databricks_permissions` to common.Resource

nfx:
This guarantees panic recovery and common error messages across all resources

### DIFF
--- a/access/resource_ipaccesslist_test.go
+++ b/access/resource_ipaccesslist_test.go
@@ -313,13 +313,13 @@ func TestIPACLDelete(t *testing.T) {
 }
 
 func TestIPACLDelete_Error(t *testing.T) {
-	d, err := qa.ResourceFixture{
+	qa.ResourceFixture{
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   http.MethodDelete,
 				Resource: "/api/2.0/ip-access-lists/" + TestingID,
 				Response: common.APIErrorBody{
-					ErrorCode: "FEATURE_DISABLE",
+					ErrorCode: "FEATURE_DISABLED",
 					Message:   "IP access list is not available in the pricing tier of this workspace",
 				},
 				Status: 404,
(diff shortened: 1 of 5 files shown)


access/resource_ipaccesslist_test.go
```diff
@@ -313,25 +313,23 @@ func TestIPACLDelete(t *testing.T) {
...
 		},
 		Resource: ResourceIPAccessList(),
 		Delete:   true,
+		Removed:  true,
 		ID:       TestingID,
-	}.Apply(t)
-	assert.Error(t, err)
-	qa.AssertErrorStartsWith(t, err, "IP access list is not available in ")
-	assert.Equal(t, TestingID, d.Id())
+	}.ApplyNoError(t)
```
pietern, line 332: Test name indicates it expects an error but here you expect no error.

permissions/resource_permissions.go
```diff
@@ -336,37 +336,7 @@ func ResourcePermissions() *schema.Resource {
...
 		return s
 	})
-	readContext := func(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-		id := d.Id()
-		objectACL, err := NewPermissionsAPI(ctx, m).Read(id)
-		if common.IsMissing(err) {
-			log.Printf("[INFO] %s is removed on backend", d.Id())
-			d.SetId("")
-			return nil
-		}
```
pietern, line 346: This is missing from the new `Read` implementation. Could it regress?

nfx: this is part of `common.Resource`, which already consistently handles it across all resources

nfx: covered by `TestResourcePermissionsRead_NotFound`

access/resource_ipaccesslist_test.go
nfx: fixed in the next commit

pietern (approved):

PR #1256, opened 2022-04-22
Update devcontainer to support ARM

fjakobs:
this will enable M1 Mac users to develop in dev containers.

Also had to update the version go go to 1.18.1 to make it work.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,7 @@
 FROM ubuntu:20.04
 
 ARG TERRAFORM_VERSION=0.14.8
+ARG TARGETARCH
 
 # Configure apt, install packages and tools
 RUN \
@@ -18,7 +19,7 @@ RUN \
 # Install terraform 
 RUN \
     mkdir -p /tmp/docker-downloads \
-    && curl -sSL -o /tmp/docker-downloads/terraform.zip https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip \
+    && curl -sSL -o /tmp/docker-downloads/terraform.zip https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_${TARGETARCH}.zip \
     && unzip /tmp/docker-downloads/terraform.zip \
     && mv terraform /usr/local/bin \
     && terraform -install-autocomplete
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:20.04
 
 ARG TERRAFORM_VERSION=0.14.8
```
nfx, line 3: Could you also change this to 1.0.0? :)

I never use dev containers 🤷🏻‍♂️

.devcontainer/Dockerfile
```diff
@@ -1,6 +1,7 @@
...
+ARG TARGETARCH
```
nfx, line 4: Can you also include default arch?

.devcontainer/Dockerfile
```diff
@@ -33,7 +34,7 @@ RUN \
 
 # GoLang
 RUN \
-    wget -c https://dl.google.com/go/go1.15.2.linux-amd64.tar.gz -O - | tar -xz -C /usr/local \
+    wget -c https://dl.google.com/go/go1.18.1.linux-${TARGETARCH}.tar.gz -O - | tar -xz -C /usr/local \
```
nfx, line 37: We didn't switch the codebase to Go 1.18 yet - we're still at 1.16 here, I've seen some build failures related to static checking tooling we use. 

Change to 1.18 should be a more global one - CI, releasing, go.mod

.devcontainer/Dockerfile
```diff
@@ -43,16 +44,15 @@ ENV GO111MODULE=on
...
-USER developer
\ No newline at end of file
+    && go install -v golang.org/x/tools/cmd/goimports@latest \
+    # # --> Gotestsum
+    && go install -v gotest.tools/gotestsum@latest \
+    # # --> Go symbols and outline for go to symbol support and test support 
+    && go install github.com/acroca/go-symbols@v0.1.1 \
+    && go install github.com/ramya-rao-a/go-outline@7182a932836a71948db4a81991a494751eccfe77 \
+    # # --> Static checker
+    && go install honnef.co/go/tools/cmd/staticcheck@latest
```
nfx, line 58: And staticcheck that works with 1.16 is v2.2.1 (I think - check push.yml to be sure).

nfx (changes requested):
Thank you for picking up dev container, it needs some occasional love :)

.devcontainer/Dockerfile
fjakobs: Will do - and you should use containers, they are great ;-)

.devcontainer/Dockerfile
fjakobs: Arch is provided by docker depending on which architecture you are building for. 
See https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

.devcontainer/Dockerfile
fjakobs: it's 0.2.2

.devcontainer/Dockerfile
fjakobs: 0.15 didn't work so I upgraded to the latest one.

I'll try 0.16.

nfx (approved):

PR #1257, opened 2022-04-22
Allow focused dashboard selection for DBSQL exporter (WIP)

nfx:


exporter/importables.go
```diff
@@ -946,55 +939,8 @@ var resourcesMap map[string]importable = map[string]importable{
 			{Path: "source", File: true},
 		},
 	},
-	"databricks_sql_query": {
```
alexott, line 949: for visibility: as discussed, I believe that we still need a functionality to export all queries, even those that aren't referenced from the dashboards - it would be useful for migrations.

nfx: yep. one for following commits will bring back listing and introduce it into it's own "service". We need to start working on PRD for this :D

nfx:
as one of the action items of making DBSQL exporter more mature, we have to:

1) Make proper SDK for every DBSQL entity. `json.RawMessage` and `map[string]any` are not good at all to work in a scalable way.
2) figure out the "import a use-case" path, not only "code generate the entire manual workspace"

pietern:
@nfx Some of the fields need those types, because they are arbitrary JSON blobs for which we don't have a stable schema.

exporter/importables.go
```diff
@@ -45,39 +46,31 @@ type dbsqlListResponse struct {
 }
 
 // Generic function to list objects related to the DBSQL
-func dbsqlListObjects(ic *importContext, path string) ([]map[string]interface{}, error) {
+func dbsqlListObjects(ic *importContext, path string) (events []map[string]interface{}, err error) {
```
nfx, line 49: @alexott please fix the pagination to return concrete slices, not `[]map[string]interface{}`

PR #1292, opened 2022-05-03
Ignore new lint error & update formatting

alexott:
Lint error on Go 1.18:

```
provider/generate_test.go:250:18: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (SA1019)
qa/testing.go:458:17: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (SA1019)
```

### DIFF
--- a/catalog/data_views_test.go
+++ b/catalog/data_views_test.go
@@ -16,15 +16,15 @@ func TestViewsData(t *testing.T) {
 					Tables: []TableInfo{
 						{
 							CatalogName: "a",
-							SchemaName: "b",
-							Name: "c",
-							TableType: "MANAGED",
+							SchemaName:  "b",
+							Name:        "c",
+							TableType:   "MANAGED",
 						},
 						{
 							CatalogName: "a",
-							SchemaName: "b",
-							Name: "d",
-							TableType: "VIEW",
+							SchemaName:  "b",
+							Name:        "d",
+							TableType:   "VIEW",
 						},
 					},
 				},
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/zclconf/go-cty v1.10.0
 	golang.org/x/mod v0.5.1
 	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5
+	golang.org/x/text v0.3.7
```
nfx, line 22: why do we need it? What license does it have?

alexott: It's needed to fix error reported by linter, as `strings.Title` may not handle unicode correctly.
License is the same as for Go (https://cs.opensource.google/go/x/text/+/master:LICENSE, https://go.dev/LICENSE )

nfx (approved):

go.mod
nfx: can we add ignore to that rule in those two places and not add more third-party dependencies? =)

I'd park this till more holistic go1.18 update.

alexott: ok. 
it's really not a 3rd party, just "extension pack" from Golang devs...

alexott: @nfx added ignore instead of using new library

PR #1309, opened 2022-05-12
Restrict KMS policy to caller AWS account

amine250:
Current KMS policy is too large allowing all actions by any AWS account. Restrict it to the caller AWS account.


nkvuong (changes requested):
Could you also update this for CMK for workspace storage (line 97)

nkvuong (approved):
LGTM

PR #1330, opened 2022-05-24
Skip validation for unity external location

sridharitv:


sridharitv:
> @sridharitv you can also add `SkipValidation` to the storage credential resource, as that is also validated during creation

@nkvuong Sorry I couldn't find API documentation for unity resources [here](https://docs.databricks.com/dev-tools/api/latest/index.html) & I don't see any validation errors or force save button in the UI as well. So I will refrain from modifying storage credential resource in this PR

nfx (approved):
👍

PR #1333, opened 2022-05-25
Re-create purged cluster for `databricks_mount` for Google Storage

### LINKED ISSUE #1331: [ISSUE] GS mount provider cannot handle mounting cluster deletion
### Configuration
```hcl
 resource "databricks_mount" "foobucket" {
      name       = "foobucket"
      gs {
          bucket_name     = "foobucket"
          service_account = "fooserviceaccount@fooproject.iam.gserviceaccount.com"
        }
    }
```

### Expected Behavior
After this configuration is applied and the mounting cluster manually (accidentally or by expiration) deleted, the template can still be applied without errors.

### Actual Behavior
cluster_id of the mounting cluster is stored in tf state such as `"cluster_id": "0524-131054-9zs18rj2"`, which makes tf apply fail with
`Error: cannot get mounting cluster: Cluster 0524-131054-9zs18rj2 does not exist` because the code goes into [this branch of code](https://github.com/databrickslabs/terraform-provider-databricks/blob/bb43dd35d51870538e6670982d31f73ffb7d6e17/storage/gs.go#L64).

The "fix" is to manually edit tf state and set `"cluster_id": "". Then the cluster is created by tf again.

### Steps to Reproduce
1. `terraform apply`
2. Delete the mounting cluster manually in Databricks workspace.
3. `terraform apply` 

Terraform Version: 0.12
Provider "gcp_databricks" with version "~> 0.5.4"

marekbrysa:
This way if the cluster is deleted, it will be re-created.

This still leaves an issue where if the service account name changes, new cluster won't be created. You'd have to manually delete the mounting cluster so that new cluster is re-created with updated service account.

Hopefully resolves #1331.

### DIFF
--- a/storage/gs_test.go
+++ b/storage/gs_test.go
@@ -94,3 +94,22 @@ func TestCreateOrValidateClusterForGoogleStorage_WorksOnDeletedCluster(t *testin
 		assert.Equal(t, "new-cluster", d.Get("cluster_id"))
 	})
 }
+
+func TestCreateOrValidateClusterForGoogleStorage_FailsOnErrorGettingCluster(t *testing.T) {
+	qa.HTTPFixturesApply(t, []qa.HTTPFixture{
+		{
+			Method:   "GET",
+			Resource: "/api/2.0/clusters/get?cluster_id=my-cluster",
+			Status:   500,
+			Response: common.APIError{
+				ErrorCode:  "SERVER_ERROR",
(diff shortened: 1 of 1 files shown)


alexott:
@marekbrysa please fix test

nfx:
@marekbrysa take a look at implementation in #1345

nfx (approved):

PR #1337, opened 2022-05-26
Add ability to specify tags for jobs

alexott:
Jobs API now allows to specify tags that will be used as cluster tags for created jobs clusters


nfx (approved):
Lgtm

PR #1338, opened 2022-05-26
Apply ownership for UC objects during creation

nkvuong:
This PR makes the ownership behaviour for catalogs, schemas and tables the same as metastores

The current behaviour of Unity Catalog API is as follows:

- `databricks_catalog`, `databricks_metastore`, `databricks_external_location`, `databricks_schema`, `databricks_storage_credentials`, `databricks_table` have an owner field
- on Create, the new object’s owner field is set to the user performing the operation
- to ensure this field is set correctly in Terraform - we need to patch the resource after creating it

Changes applied:
* set owner during creation for UC securables
* update docs
* refactor into update function factory
* add support for locations & credentials
* remove update APIs
* add unit test when cannot delete default schema
* add MarkNewResource to testing fixture


catalog/resource_catalog.go
```diff
@@ -57,6 +57,25 @@ func ResourceCatalog() *schema.Resource {
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			return m
 		})
+	update := func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
```
nfx, line 60: Makes sense. I suggest you create an unexported package level utility "update factory" function of this. This way there's less repetition.

catalog/resource_catalog_test.go
```diff
@@ -50,6 +50,7 @@ func TestCatalogCreateAlsoDeletesDefaultSchema(t *testing.T) {
 		properties = {
 			c = "d"
 		}
+		owner = "f"
```
nfx, line 53: Strange.

catalog/resource_table.go
```diff
@@ -84,6 +84,25 @@ func ResourceTable() *schema.Resource {
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			return m
 		})
+	update := func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+		// other fields to come later
+		updatable := []string{"owner", "name"}
```
nfx, line 89: Are you sure about table having only these fields updatable?

nfx (changes requested):
Makes sense. Please introduce "update function factory" to have less repetition.

Also are you sure about tables?

catalog/resource_catalog.go
```diff
@@ -18,7 +18,7 @@ func NewCatalogsAPI(ctx context.Context, m interface{}) CatalogsAPI {
 }
 
 type CatalogInfo struct {
-	Name        string            `json:"name" tf:"force_new"`
+	Name        string            `json:"name"`
```
nfx, line 21: can we update the name, though?.. why force_new is removed?

catalog/resource_catalog_test.go
```diff
@@ -38,7 +47,14 @@ func TestCatalogCreateAlsoDeletesDefaultSchema(t *testing.T) {
 						"c": "d",
 					},
 					MetastoreID: "e",
-					Owner:       "f",
+					Owner:       "administrators",
+				},
+			},
+			{
```
nfx, line 53: can you also move this fixture a bit up, so it represents the real sequence of requests? sometimes i refrain to these tests as documentation for support, who probably could easier read API request flows, than go code.

the real sequence for create is
1. POST
2. DELETE
3. PATCH
4. GET

correct?

catalog/resource_catalog_test.go
```diff
@@ -50,6 +66,7 @@ func TestCatalogCreateAlsoDeletesDefaultSchema(t *testing.T) {
 		properties = {
 			c = "d"
 		}
+		owner = "administrators"
```
nfx, line 69: could you create new test, where you set the owner? this will give us more confidence that existing behavior is not affected and this PR is only incrementing features, not causing regressions.

catalog/resource_external_location.go
```diff
@@ -18,7 +18,7 @@ func NewExternalLocationsAPI(ctx context.Context, m interface{}) ExternalLocatio
 }
 
 type ExternalLocationInfo struct {
-	Name           string `json:"name" tf:"force_new"`
```
nfx, line 21: so we can change name for this as well?

catalog/resource_external_location.go
```diff
@@ -49,6 +49,7 @@ func ResourceExternalLocation() *schema.Resource {
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			return m
 		})
+	update := updateFunctionFactory("external-location", []string{"owner", "name", "comment", "url", "credential_name", "skip_validation"})
```
nfx, line 52: with these many fields - I'll eventually move this up to common layer. thanks for making the first step! :)

catalog/resource_external_location_test.go
```diff
@@ -31,13 +39,21 @@ func TestCreateExternalLocation(t *testing.T) {
...
+				},
+			},
 		},
 		Resource: ResourceExternalLocation(),
 		Create:   true,
 		HCL: `
 		name = "abc"
 		url = "s3://foo/bar"
 		credential_name = "bcd"
+		owner = "administrators"
```
nfx, line 56: same comment about not changing tests

catalog/resource_schema_test.go
```diff
@@ -40,6 +51,7 @@ func TestCreateSchema(t *testing.T) {
 		name = "a"
 		catalog_name = "b"
 		comment = "c"
+		owner = "administrators"
```
nfx, line 54: same comment about not changing tests and introducing new ones

catalog/resource_storage_credentials_test.go
```diff
@@ -47,6 +57,7 @@ func TestCreateStorageCredentials(t *testing.T) {
 			role_arn = "def"
 		}
 		comment = "c"
+		owner = "administrators"
```
nfx, line 60: same comment about tests :)

catalog/resource_table_test.go
```diff
@@ -45,10 +89,10 @@ func TestTableCreate(t *testing.T) {
 		catalog_name = "main"
 		schema_name = "foo"
 		name = "bar"
+		owner = "administrators"
```
nfx, line 92: same comment about tests :)

catalog/update.go
```diff
@@ -0,0 +1,40 @@
...
+		case "catalog":
+			return NewCatalogsAPI(ctx, c).updateCatalog(d.Id(), patch)
+		case "schema":
+			return NewSchemasAPI(ctx, c).updateSchema(d.Id(), patch)
+		case "table":
+			return NewTablesAPI(ctx, c).updateTable(d.Id(), patch)
+		case "external-location":
+			return NewExternalLocationsAPI(ctx, c).update(d.Id(), patch)
+		case "storage-credential":
+			return NewStorageCredentialsAPI(ctx, c).update(d.Id(), patch)
```
nfx, line 35: i don't like this hardcoding and tight coupling. at the same time this makes it more readable.

there are also two other options: 

very explicit:
```go
return updateFactory(func (ctx context.Context) func(string, map[string]interface{}) error {
  return NewCatalogsAPI(ctx, c).updateCatalog
}, resourceData, client, "owner", "name", "comment", "properties")
```

layer leak:
```go
update := updateFactory("/unity-catalog/catalogs", "owner", "name", "comment", "properties")
```

which would entail implementation like
```go
func updateFunctionFactory(pathPrefix string, updatable string...) func(context.Context, *schema.ResourceData, *common.DatabricksClient) error {
  return func(... {
    ...
    return c.Patch(ctx, path.Join(pathPrefix, d.Id()), patch)
  }
}
```

with this layer leak we technically don't need update methods for all securables, right?

catalog/update.go
```diff
@@ -0,0 +1,40 @@
...
+		default:
+			return nil
```
nfx, line 37: must be an error here ;) like "securable is not implemented for updates" or something. with "layer leak" this is not really a problem

nfx (changes requested):
resources start to look neat! still some comments.

nfx:
@nkvuong please reintegrate this branch because of #1334

catalog/resource_external_location.go
nkvuong: yes 😢

catalog/resource_catalog_test.go
nkvuong: created new tests

catalog/resource_schema_test.go
nkvuong: new tests

catalog/update.go
nkvuong: moved to layer leak - looking relatively clean

catalog/resource_catalog.go
nfx: yep.. that's why i ask... did you extensively test it?

catalog/resource_catalog.go
```diff
@@ -69,7 +67,7 @@ func ResourceCatalog() *schema.Resource {
 				return fmt.Errorf("cannot remove new catalog default schema: %w", err)
 			}
 			d.SetId(ci.Name)
-			return nil
+			return updateOwner(ctx, d, c)
```
nfx, line 70:
```suggestion
			return updateFunctionFactory("/unity-catalog/catalogs/", []string{"owner"})(ctx, d, c)
```

don't introduce single-use variables that make readability harder :)

catalog/resource_catalog.go
```diff
@@ -78,11 +76,7 @@ func ResourceCatalog() *schema.Resource {
 			}
 			return common.StructToData(ci, catalogSchema, d)
 		},
-		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			var ci CatalogInfo
-			common.DataToStructPointer(d, catalogSchema, &ci)
-			return NewCatalogsAPI(ctx, c).updateCatalog(ci)
-		},
+		Update: update,
```
nfx, line 79:
```suggestion
		Update: updateFunctionFactory("/unity-catalog/catalogs/", []string{"owner", "name", "comment", "properties"}),
```

catalog/resource_external_location.go
nfx: 😢

catalog/resource_external_location.go
```diff
@@ -49,6 +45,8 @@ func ResourceExternalLocation() *schema.Resource {
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			return m
 		})
+	updateOwner := updateFunctionFactory("/unity-catalog/external-locations/", []string{"owner"})
+	update := updateFunctionFactory("/unity-catalog/external-locations/", []string{"owner", "name", "comment", "url", "credential_name", "skip_validation"})
```
nfx, line 49: ouch, this looks ugly. there's `d.IsNewResource()` that can probably help to diff between create and update. you can add a special if statement in update factory just for `owner` field (if it's present in the list of fields).

i don't remember if it's easy unit testable or not, so please launch provider in a debug mode to see how to use it and then work on tests :)

catalog/resource_metastore.go
```diff
@@ -87,34 +87,11 @@ func ResourceMetastore() *schema.Resource {
...
-				continue
-			}
-			// delta sharing enabled and new is true must always be accompanied by a value for
-			// delta_sharing_recipient_token_lifetime_in_seconds
-			if field == "delta_sharing_enabled" && old != new && new == true &&
-				!d.HasChange("delta_sharing_recipient_token_lifetime_in_seconds") {
-				patch["delta_sharing_recipient_token_lifetime_in_seconds"] =
-					d.Get("delta_sharing_recipient_token_lifetime_in_seconds")
-			}
+	updateOwner := updateFunctionFactory("/unity-catalog/metastores/", []string{"owner", "delta_sharing_enabled",
```
nfx, line 90:
```suggestion
	updateOwner := updateFunctionFactory("/unity-catalog/metastores", []string{"owner", "delta_sharing_enabled",
```

nit: can we have it wihtout a trailing slash please?... just looks ugly :(

nfx (changes requested):
beef up the logic of update factory

catalog/update.go
```diff
@@ -0,0 +1,32 @@
...
+		for _, field := range updatable {
+			old, new := d.GetChange(field)
+			if !d.HasChange(field) {
+				continue
+			}
+			if field == "delta_sharing_enabled" && old != new && new == true &&
+				!d.HasChange("delta_sharing_recipient_token_lifetime_in_seconds") {
+				patch["delta_sharing_recipient_token_lifetime_in_seconds"] =
+					d.Get("delta_sharing_recipient_token_lifetime_in_seconds")
+			}
```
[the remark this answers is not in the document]
nkvuong: this code is from #1334 - already have a test

nfx (changes requested):
Please build locally and try to create some resources with owners with `d.MarkNewResource()` removed from creates, as it's supposed to be done already implicitly by terraform SDK on creates. 

I've pointed where to "fix" the ResourceFixture, so that it's unit-testable.

catalog/resource_metastore.go
```diff
@@ -87,34 +87,9 @@ func ResourceMetastore() *schema.Resource {
...
-				continue
-			}
-			// delta sharing enabled and new is true must always be accompanied by a value for
-			// delta_sharing_recipient_token_lifetime_in_seconds
-			if field == "delta_sharing_enabled" && old != new && new == true &&
-				!d.HasChange("delta_sharing_recipient_token_lifetime_in_seconds") {
-				patch["delta_sharing_recipient_token_lifetime_in_seconds"] =
-					d.Get("delta_sharing_recipient_token_lifetime_in_seconds")
-			}
+	update := updateFunctionFactory("/unity-catalog/metastores", []string{"owner", "name", "delta_sharing_enabled",
```
nfx, line 90: so, as you told over slack - does it still make sense to have `name` updatable?

catalog/resource_schema.go
```diff
@@ -17,7 +17,7 @@ func NewSchemasAPI(ctx context.Context, m interface{}) SchemasAPI {
 }
 
 type SchemaInfo struct {
-	Name        string            `json:"name" tf:"force_new"`
```
nfx, line 20: so, as you told over slack - does it still make sense to have `name` updatable?

catalog/resource_schema_test.go
```diff
@@ -23,14 +23,15 @@ func TestCreateSchema(t *testing.T) {
...
 					FullName: "b.a",
+					Comment:  "c",
 				},
 			},
 			{
 				Method:   "GET",
 				Resource: "/api/2.0/unity-catalog/schemas/b.a",
 				Response: SchemaInfo{
 					MetastoreID: "d",
-					Owner:       "e",
```
nfx, line 33: does it error out with owner?

catalog/resource_table.go
```diff
@@ -32,7 +32,7 @@ type ColumnInfo struct {
 }
 
 type TableInfo struct {
-	Name                  string            `json:"name" tf:"force_new"`
+	Name                  string            `json:"name"`
```
nfx, line 35: okay, for table names it makes total sense to rename.

docs/resources/catalog.md
```diff
@@ -27,7 +27,7 @@ resource "databricks_catalog" "sandbox" {
 The following arguments are required:
 
 * `name` - Name of Catalog relative to parent metastore. Change forces creation of a new resource.
-* `owner` - (Optional) Username/groupname of catalog owner. Currently this field can only be changed after the resource is created.
+* `owner` - (Optional) Username/groupname/sp id of catalog owner.
```
nfx, line 30: make it a bit more explicit, is it `application_id` or `id` from `databricks_service_principal`? :) please be more explicit

catalog/resource_schema_test.go
nkvuong: forgot to put it back

nfx (approved):
👍

PR #1354, opened 2022-06-04
Add support for managed identity in UC

nkvuong:
On Azure, Unity Catalog now supports system assigned managed identity as credentials for `databricks_metastore_data_access` and `databricks_storage_credential` in addition to service principals.

This PR adds support for this as an optional block

```tf 
azure_managed_identity {
    access_connector_id   = var.access_connector_id
  }
```

### DIFF
--- a/docs/resources/metastore_data_access.md
+++ b/docs/resources/metastore_data_access.md
@@ -29,7 +29,7 @@ resource "databricks_metastore_data_access" "this" {
 }
 ```
 
-For Azure using service principal
+For Azure using service principal as credential
 
 ```hcl
 resource "databricks_metastore" "this" {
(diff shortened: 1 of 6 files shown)


docs/resources/metastore_data_access.md
```diff
@@ -53,6 +53,28 @@ resource "databricks_metastore_data_access" "this" {
 }
 ```
 
+For Azure using managed identity as credential (Private Preview)
```
nfx, line 56: Did you get a green light from @annashres to document it publicly for private preview? 

If not - please remove this section from registry docs just yet, create a google doc in GTM private preview folder and move it there. And also link it on internal wiki for tf provider and azmi preview.

nfx (changes requested):
Very fast! :-) left some comments.

docs/resources/metastore_data_access.md
nkvuong: Anna said it should be fine since the UC is gated as well and we'll ungate MIs when the UC is ungated

nfx: Then it goes to this release :)

PR #1360, opened 2022-06-07
Added `network_id` to `network` block in `databricks_mws_workspaces` for GCP

marekbrysa:
In GCP to use customer-manager VPC, the `network_id` is specified inside the `network` object, not at the top-level workspace object.

This allows the `network_id` to be specified at the correct place.


nfx (commented):
👍

nfx (approved):
👍

PR #1373, opened 2022-06-15
Update resources diagram

alexott:
Incorporate all new resources implemented since last update


nfx (changes requested):
can you move azure and gcp a bit below?

nfx (approved):

PR #1385, opened 2022-06-22
Sanitize host field prior to auth flow

pietern:
I was getting 401s when trying to use the accounts API on GCP, until I realized the host field in the provider configuration had a trailing slash. Removing the trailing slash fixed the issue. To improve the UX here, this PR removes a trailing slash prior to the authentication flow.

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"reflect"
 	"strings"
@@ -237,6 +238,11 @@ func (c *DatabricksClient) Authenticate(ctx context.Context) error {
 	if c.authVisitor != nil {
 		return nil
 	}
+	// Fix host prior to auth, because it may be used in the OIDC flow as "audience" field.
+	// If necessary, this function adds a scheme and strips a trailing slash.
+	if err := c.fixHost(); err != nil {
+		return err
+	}
 	type auth struct {
 		configure func(context.Context) (func(*http.Request) error, error)
 		name      string
@@ -271,7 +277,6 @@ func (c *DatabricksClient) Authenticate(ctx context.Context) error {
 		log.Printf("[INFO] Configured %s auth: %s", auth.name, c.configDebugString()) // lgtm[go/clear-text-logging]
 		c.authVisitor = authorizer
 		c.AuthType = auth.name
-		c.fixHost()
 		return nil
 	}
 	if c.AuthType != "" {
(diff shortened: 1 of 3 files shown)


nfx (changes requested):

common/client.go
```diff
@@ -326,12 +331,32 @@ func (c *DatabricksClient) niceAuthError(message string) error {
 	return fmt.Errorf("%s%s. Please check %s for details", message, info, docUrl)
 }
 
-func (c *DatabricksClient) fixHost() {
-	if c.Host != "" && !(strings.HasPrefix(c.Host, "https://") || strings.HasPrefix(c.Host, "http://")) {
-		// azurerm_databricks_workspace.*.workspace_url is giving URL without scheme
-		// so that is why this line is here
-		c.Host = "https://" + c.Host
+func (c *DatabricksClient) fixHost() error {
+	if c.Host != "" {
```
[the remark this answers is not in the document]
pietern: Done.

nfx (approved):

common/client.go
```diff
@@ -271,7 +277,6 @@ func (c *DatabricksClient) Authenticate(ctx context.Context) error {
 		log.Printf("[INFO] Configured %s auth: %s", auth.name, c.configDebugString()) // lgtm[go/clear-text-logging]
 		c.authVisitor = authorizer
 		c.AuthType = auth.name
-		c.fixHost()
```
nfx, line 274: can we fix it before and after the auth resolution? e.g. misconfigured databrickscfg

nfx (changes requested):

common/client.go
```diff
@@ -271,7 +277,6 @@ func (c *DatabricksClient) Authenticate(ctx context.Context) error {
 		log.Printf("[INFO] Configured %s auth: %s", auth.name, c.configDebugString()) // lgtm[go/clear-text-logging]
 		c.authVisitor = authorizer
 		c.AuthType = auth.name
```
nfx, line 279:
```suggestion
		c.AuthType = auth.name
		c.fixHost()
```

can we fix it before and after the auth resolution? e.g. misconfigured databrickscfg

nfx (changes requested):

common/client.go
```diff
@@ -326,12 +331,34 @@ func (c *DatabricksClient) niceAuthError(message string) error {
...
+	if err != nil {
+		return err
+	}
+
+	// If the host is empty, assume the scheme wasn't included.
+	if u.Host == "" {
+		u, err = url.Parse("https://" + c.Host)
+		if err != nil {
+			return err
+		}
```
nfx, line 350: this is strange, as invocation will always error out because of `url.Parse("https://")`

common/client.go
```diff
@@ -326,12 +331,34 @@ func (c *DatabricksClient) niceAuthError(message string) error {
...
 	}
+
+	// Create new instance to ensure other fields are initialized as empty.
+	u = &url.URL{
+		Scheme: u.Scheme,
+		Host:   u.Host,
+	}
+
+	// Store sanitized version of c.Host.
+	c.Host = u.String()
```
nfx, line 360: why not just `fmt.Sprintf("https://%s", c.Host)`?

common/client.go
pietern: No, u is the parsed URL, c is the input URL.

nfx: ah, it's `u`, i've read it as `c`.

common/client.go
nfx: `TestConfig_PatFromDatabricksCfg` is failing. as long as we're doing https://github.com/databricks/terraform-provider-databricks/blob/master/common/http.go#L338-L343, it's safe to update the test

PR #1399, opened 2022-06-24
Fix Azure authentication for staging workspaces

pietern:


common/client.go
```diff
@@ -62,6 +62,10 @@ type DatabricksClient struct {
 	AzureTenantID      string `name:"azure_tenant_id" env:"ARM_TENANT_ID" auth:"azure"`
 	AzurermEnvironment string `name:"azure_environment" env:"ARM_ENVIRONMENT"`
 
+	// Resource ID of the Azure Databricks application.
+	// This is configured automatically when an Azure authentication method is used.
+	armDatabricksResourceID string
```
nfx, line 67: should it rather be an env variable instead?.. not to publish the uuid in the wild. this will also be easier for sdks.

nfx (changes requested):

pietern:
Turns out that this change alone is not enough. I'm going to use PATs to unblock myself and bother with this later.

PR #1414, opened 2022-06-29
Use SQL warehouses API instead of SQL endpoints API

alexott:
It's only internal change, resource/attributes renaming, etc. could be done later.

New API names: https://docs.databricks.com/sql/api/sql-endpoints.html#create


nfx (approved):
👍

PR #1416, opened 2022-07-01
Fixed `databricks_pipeline` incorrect generation of `cluster` blocks

### LINKED ISSUE #1401: [ISSUE] databricks_pipeline does not support updating the resource
### Affected Resource(s)
https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/pipeline

### Expected Details
The field `cluster.custom_tags.cluster_type` is not listed as a required field (implied to be therefore optional), yet creating a DLT pipeline without at least one `cluster` block containing the following configuration will allow the pipeline to be created but NOT subsequently updated.
```
cluster {
    custom_tags {
        cluster_type = 'default'
    }
}
```

Attempting to update a pipeline created without the above block will return the error:
`Error: cannot update pipeline: clusters must have exactly 1 default cluster`

It seems that updates to the pipeline are trying to reference this tag which could be non-existent, not the `label` field.

**EDIT:** It seems that this is not a documentation issue, requiring the custom tag still will not allow the resource to be updated in subsequent `terraform apply`

alexott:
This fixes #1401 to unblock the production workloads.

But it should be investigated why we generated empty structs when using reflection

### DIFF
--- a/pipelines/acceptance/pipeline_test.go
+++ b/pipelines/acceptance/pipeline_test.go
@@ -94,7 +94,7 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
 }
 
 func TestAccAwsPipelineResource_CreatePipeline(t *testing.T) {
-	if cloud, ok := os.LookupEnv("CLOUD_ENV"); !ok || cloud != "AWS" {
+	if cloud, ok := os.LookupEnv("CLOUD_ENV"); !ok || cloud != "aws" {
 		t.Skip("Test is only for CLOUD_ENV=AWS")
 	}
 	acceptance.Test(t, []acceptance.Step{
--- a/pipelines/acceptance/pipeline_test.go
+++ b/pipelines/acceptance/pipeline_test.go
@@ -1,9 +1,49 @@
 package acceptance
 
 import (
+	"os"
 	"testing"
 
 	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+)
+
+var (
+	dltNotebookResource = `
+	resource "databricks_notebook" "this" {
+		content_base64 = base64encode(<<-EOT
+			CREATE LIVE TABLE clickstream_raw AS 
+			SELECT * FROM json.` + "`/databricks-datasets/wikipedia-datasets/data-001/clickstream/raw-uncompressed-json/2015_2_clickstream.json`" + `
+			
+			-- COMMAND ----------
+			
+			CREATE LIVE TABLE clickstream_clean(
+			  CONSTRAINT valid_current_page EXPECT (current_page_id IS NOT NULL and current_page_title IS NOT NULL),
+			  CONSTRAINT valid_count EXPECT (click_count > 0) ON VIOLATION FAIL UPDATE
+			) TBLPROPERTIES ("quality" = "silver")
+			AS SELECT
+			  CAST (curr_id AS INT) AS current_page_id,
+			  curr_title AS current_page_title,
+			  CAST(n AS INT) AS click_count,
+			  CAST (prev_id AS INT) AS previous_page_id,
+			  prev_title AS previous_page_title
+			FROM live.clickstream_raw
+			
+			-- COMMAND ----------
+			
+			CREATE LIVE TABLE top_spark_referers TBLPROPERTIES ("quality" = "gold")
+			AS SELECT
+			  previous_page_title as referrer,
+			  click_count
+			FROM live.clickstream_clean
+			WHERE current_page_title = 'Apache_Spark'
+			ORDER BY click_count DESC
+			LIMIT 10					
+		  EOT
+		)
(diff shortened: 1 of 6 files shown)


pipelines/resource_pipeline.go
```diff
@@ -22,6 +22,7 @@ const DefaultTimeout = 20 * time.Minute
 // 1. Pipeline clusters include a `Label` field.
 // 2. Spark version is not required (and shouldn't be specified) for pipeline clusters.
 // 3. num_workers is optional, and there is no single-node support for pipelines clusters.
+// When adding new attributes, also modify the isEmptyClusterStruct function below
```
nfx, line 25: have you tried `reflect.IsZero` way?

https://www.geeksforgeeks.org/reflect-iszero-function-in-golang-with-examples/

nfx (changes requested):

pipelines/resource_pipeline.go
alexott: I looked to reflect, but read that it may not handle pointers, etc. correctly

alexott: I’ll look Monday

nfx: i think it's more of an issue in `slice_set`, btw. e.g. diff is not properly picked up

alexott: agree, but this requires more time to investigation. I had a hope to have quick & dirty fix, and then fix correctly when get time

alexott:
It didn't have `suppress_diff` before: https://github.com/databricks/terraform-provider-databricks/commit/8260d907c58b97c4fa2b719c8a8267cb8cd81f05#diff-95aaa2c785925088721c159cdd5b07357cdc3b1899bf4e4d4e236460acf1b737L33

alexott:
Test is added

pipelines/acceptance/pipeline_test.go
```diff
@@ -81,14 +86,99 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
...
-					exclude = ["com.databricks.exclude"]
+				continuous = false
+			}
+			` + dltNotebookResource,
+		},
+	})
+}
+
+func TestAccAwsPipelineResource_CreatePipeline(t *testing.T) {
+	if cloud, ok := os.LookupEnv("CLOUD_ENV"); !ok || cloud != "AWS" {
```
nfx, line 97:
```suggestion
	if cloud, ok := os.LookupEnv("CLOUD_ENV"); !ok || cloud != "aws" {
```

pipelines/acceptance/pipeline_test.go
```diff
@@ -81,14 +86,99 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
...
+					key2 = "value2"
+				}
+				library {
+					notebook {
+						path = databricks_notebook.this.path
+					}
 				}
 
+				cluster {
+					instance_pool_id = "{var.COMMON_INSTANCE_POOL_ID}"
```
nfx, line 120: we don't have this instance pool defined anywhere...

pipelines/acceptance/pipeline_test.go
```diff
@@ -81,14 +86,99 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
...
+					num_workers = 2
+					custom_tags = {
+						cluster_type = "default"
+					}
+					aws_attributes {
+						first_on_demand = 1
+					}
+				}
+				cluster {
+					instance_pool_id = "{var.COMMON_INSTANCE_POOL_ID}"
```
nfx, line 131: here as well

pipelines/acceptance/pipeline_test.go
```diff
@@ -81,14 +86,99 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
...
+					key2 = "value2"
+				}
+				library {
+					notebook {
+						path = databricks_notebook.this.path
+					}
+				}
+
+				cluster {
+					instance_pool_id = "{var.COMMON_INSTANCE_POOL_ID}"
```
nfx, line 161: here as well

pipelines/acceptance/pipeline_test.go
```diff
@@ -81,14 +86,99 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
...
+					num_workers = 3
+					custom_tags = {
+						cluster_type = "default"
+					}
+					aws_attributes {
+						first_on_demand = 2
+					}
+				}
+				cluster {
+					instance_pool_id = "{var.COMMON_INSTANCE_POOL_ID}"
```
nfx, line 172: here as well

nfx (changes requested):

nfx (approved):
image

PR #1417, opened 2022-07-04
Better handling of missing objects during import

alexott:
Right now, when importing we're reading data, and ignoring the missing resources.  But it
doesn't work well, and users get a cryptic error, like this:

```
>terraform import databricks_cluster.this 12313231232
databricks_cluster.this: Importing from ID "12313231232"...
╷
│ Error: The provider returned a resource missing an identifier during
ImportResourceState. This is generally a bug in the resource implementation for
import. Resource import code should not call d.SetId("") or create an empty
ResourceData. If the resource is missing, instead return an error. Please report this to
the provider developers.
```

With this PR, error message is clear:

```
Error: Cannot import non-existent remote object

While attempting to import an existing object to "databricks_cluster.this",
the provider detected that no object exists with the given id. Only
pre-existing objects can be imported; check that the id is correct and that
it is associated with the provider's configured region or endpoint, or use
"terraform apply" to create a new remote object for this resource.
```

### DIFF
--- a/common/resource.go
+++ b/common/resource.go
@@ -92,20 +92,23 @@ func (r Resource) ToResource() *schema.Resource {
 			}
 		}
 	}
-	read := func(ctx context.Context, d *schema.ResourceData,
+	generateReadFunc := func(ignoreMissing bool) func(ctx context.Context, d *schema.ResourceData,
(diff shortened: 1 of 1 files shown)


nfx (approved):
👍

PR #1437, opened 2022-07-08
Fixed `databricks_job` recreation on changed `docker_image` URL

### LINKED ISSUE #1434: [ISSUE] `databricks_job` is replaced every time the Docker Image url is updated
Each time the docker image url changes, the job gets deleted and recreated with a new job id, thus removing all run history of the job. This is a regression introduced by https://github.com/databricks/terraform-provider-databricks/commit/735ea5ce60d2bb0532294bd8fc599e286954b765. Proper fix should be propagating `force_new` type tag down to child entities - in this case, for instance pools.

```
docker_image {
  ~ url = "abc/bcd:2.1.0" -> "abc/bcd:2.2.0" # forces replacement
}
```

# Expected behaviour

When the docker url is updated for an existing job cluster, the Job is not deleted and re-created with a new Job Id. We should be able to update the cluster config without re-creating the job via terraform. 

# Steps to reproduce 

- Create a job with a cluster via terraform that installs a Docker Image from a url
- Change that docker image url via an update of any sort
- Attempt to deploy the job via terraform provider. Job will get deleted and re-created.

alexott:
Added use `tf:force_new` on Docker config only for `databricks_instance_pool`

this fixes #1434

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -241,14 +241,14 @@ type LogSyncStatus struct {
 
 // DockerBasicAuth contains the auth information when fetching containers
 type DockerBasicAuth struct {
-	Username string `json:"username" tf:"force_new"`
-	Password string `json:"password" tf:"force_new"`
+	Username string `json:"username"`
+	Password string `json:"password" tf:"sensitive"`
 }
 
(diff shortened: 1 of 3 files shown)


nfx (approved):
👍

PR #1454, opened 2022-07-12
[DRAFT] Retry storage credential create/update to allow for IAM propagation

### LINKED ISSUE #1424: [ISSUE] Cannot create databricks_mws_credentials because aws_iam not ready
Hi there,

Thank you for opening an issue. Please note that we try to keep the Databricks Provider issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Configuration
```hcl
terraform {
  required_providers {
    databricks = {
      version = "1.0.0"
    }
    aws = {
      source  = "hashicorp/aws"
      version = "~> 3.0"
    }
  }
}

provider "databricks" {
  alias = "mws"
  host  = "https://accounts.cloud.databricks.com"
  username = "foo"
  password = "bar"
}

# Configure the AWS Provider
provider "aws" {
  region = "us-west-2"
}

variable "databricks_account_id" {
  description = "Account Id that could be found in the bottom left corner of https://accounts.cloud.databricks.com/"
  default = "foo"
}

data "databricks_aws_assume_role_policy" "this" {
  external_id = var.databricks_account_id
}

resource "aws_iam_role" "cross_account_role" {
  name               = "test-crossaccount"
  assume_role_policy = data.databricks_aws_assume_role_policy.this.json
}

data "databricks_aws_crossaccount_policy" "this" {
}

resource "aws_iam_role_policy" "this" {
  name   = "test-policy"
  role   = aws_iam_role.cross_account_role.id
  policy = data.databricks_aws_crossaccount_policy.this.json
}

resource "databricks_mws_credentials" "this" {
  provider         = databricks.mws
  account_id       = var.databricks_account_id
  credentials_name = "test-creds"
  role_arn         = aws_iam_role.cross_account_role.arn
}

```

### Expected Behavior
Terraform apply should have just worked.

### Actual Behavior
Terraform applied failed with
```
│ Error: cannot create mws credentials: MALFORMED_REQUEST: Failed credential validation checks: please use a valid cross account IAM role with permissions setup correctly
│
│   with databricks_mws_credentials.this,
│   on mws-workspace.tf line 55, in resource "databricks_mws_credentials" "this":
│   55: resource "databricks_mws_credentials" "this" {
```

### Steps to Reproduce
1. `terraform apply`

```
Terraform v1.1.7
on darwin_amd64
+ provider registry.terraform.io/databricks/databricks v1.0.0
+ provider registry.terraform.io/hashicorp/aws v3.75.2
```

### Important Factoids
You can do a workaround with doing `terraform apply`. The 2nd time the apply will be successful and mws_credentials will be created

neinkeinkaffee:
Retry the create/update operation for storage credentials if the error indicates that we may have to allow more time for IAM propagation. The timeout of 2 minutes could be further decreased, the current number is taken from corresponding timeouts for the IAM propagation case in the AWS Terraform provider.

### DIFF
--- a/catalog/resource_storage_credential.go
+++ b/catalog/resource_storage_credential.go
@@ -63,6 +63,10 @@ func ResourceStorageCredential() *schema.Resource {
 		return nil
 	}
 	return common.Resource{
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(2 * time.Minute),
+			Update: schema.DefaultTimeout(2 * time.Minute),
+		},
 		Schema: s,
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			var sci StorageCredentialInfo
--- a/catalog/resource_storage_credential.go
+++ b/catalog/resource_storage_credential.go
@@ -2,7 +2,6 @@ package catalog
 
 import (
 	"context"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"strings"
 	"time"
 
@@ -52,15 +51,11 @@ func ResourceStorageCredential() *schema.Resource {
 			return m
 		})
 	update := func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-		err := retryOnIAMError(ctx, d.Timeout(schema.TimeoutUpdate), func() error {
+		return common.RetryOnError(ctx, d.Timeout(schema.TimeoutUpdate), isIAMError, func() error {
 			return updateFunctionFactory("/unity-catalog/storage-credentials", []string{
 				"owner", "comment", "aws_iam_role", "azure_service_principal", "azure_managed_identity"})(
 				ctx, d, c)
 		})
-		if err != nil {
-			return err
-		}
-		return nil
 	}
 	return common.Resource{
 		Timeouts: &schema.ResourceTimeout{
@@ -72,7 +67,7 @@ func ResourceStorageCredential() *schema.Resource {
 			var sci StorageCredentialInfo
 			common.DataToStructPointer(d, s, &sci)
 			sci.Owner = ""
-			if err := retryOnIAMError(ctx, d.Timeout(schema.TimeoutCreate), func() error {
+			if err := common.RetryOnError(ctx, d.Timeout(schema.TimeoutCreate), isIAMError, func() error {
 				return NewStorageCredentialsAPI(ctx, c).create(&sci)
 			}); err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


catalog/resource_storage_credential.go
```diff
@@ -76,3 +92,17 @@ func ResourceStorageCredential() *schema.Resource {
 		},
 	}.ToResource()
 }
+
+func waitIAMPropagation(ctx context.Context, c *common.DatabricksClient, f func() error) error {
+	return resource.RetryContext(ctx, IAMPropagationTimeout,
+		func() *resource.RetryError {
+			cerr := f()
+			if e, ok := cerr.(common.APIError); ok && e.StatusCode == 403 && strings.Contains(cerr.Error(), "IAM role") {
```
nfx, line 100: Please make error message matching a bit more concrete

catalog/resource_storage_credential.go
```diff
@@ -76,3 +92,17 @@ func ResourceStorageCredential() *schema.Resource {
 		},
 	}.ToResource()
 }
+
+func waitIAMPropagation(ctx context.Context, c *common.DatabricksClient, f func() error) error {
+	return resource.RetryContext(ctx, IAMPropagationTimeout,
```
nfx, line 97: There’s a concept of “resource timeouts” that are configurable on instance levels. Take a look at sql endpoints, library, cluster, workspace resources to see how to add them. You’ll have to change method signature to accept that.

catalog/resource_storage_credential.go
```diff
@@ -76,3 +92,17 @@ func ResourceStorageCredential() *schema.Resource {
 		},
 	}.ToResource()
 }
+
+func waitIAMPropagation(ctx context.Context, c *common.DatabricksClient, f func() error) error {
```
nfx, line 96: Rename method to include retry

catalog/resource_storage_credential.go
neinkeinkaffee: I'm now matching on a larger portion of the error message, maybe still not ideal, but hopefully a bit better than before

catalog/resource_storage_credential.go
neinkeinkaffee: Now using the resource timeouts which I've set to 2 minutes. That should be fine for storage credential, I'm just wondering whether in general the IAM propagation timeout might potentially sometimes have to be more conservative than the general resource timeouts. If the IAM role takes more than 2 minutes than maybe something's really wrong, and the role isn't there or not properly configured. Whereas some resources may want to have longer timeouts for creation and update.

catalog/resource_storage_credential.go
neinkeinkaffee: Renamed :)

neinkeinkaffee:
I have moved the retryOnError method into http.go in the common package in order to reuse it for the MWS credential. I'm not sure if that's the best location. I'm also wondering whether it might be better to keep the value of the timeout duration for IAM propagation separate from the resource timeouts for creation and update, given that one might not want to retry on IAM error until the overall resource creation/update timeout limit is reached. It would have the additional benefit of not having to pass down the resource timeout.

nfx:
@neinkeinkaffee I’m not sure if common package is the right place for it. Cost of refactoring/supporting it in the common package is higher than copy paste of approximately 10 lines. So far i see it used only two times (for IAM roles). This is really retry-and-wait for error that might be a valid error after one or two minutes. Common package does another kind of error retries, which are immediate. More than that, common package is being refactored into standalone SDK as we speak, so I don’t want more things added over there.

bottom line: these retries are great, but I think they belong to mws and catalog packages.

neinkeinkaffee:
Makes sense! I've moved the retry code back into the catalog and mws.

oleksandr-gubchenko:
any updates on this?

hectorcast-db:
Retries and other client side patches for non consistent resources have been proven to be problematic. The solution needs to come from the server side with the exposure of a method or field which notifies/confirms when a the resource is ready (e.g., by using Long Running Operations). Closing this PR.

PR #1459, opened 2022-07-14
Add more attributes to `databricks_cluster`

alexott:


clusters/resource_cluster.go
```diff
@@ -83,6 +83,19 @@ func resourceClusterSchema() map[string]*schema.Schema {
 		s["driver_node_type_id"].ConflictsWith = []string{"driver_instance_pool_id", "instance_pool_id"}
 		s["node_type_id"].ConflictsWith = []string{"driver_instance_pool_id", "instance_pool_id"}
 
+		p, err := common.SchemaPath(s, "workload_type", "clients", "notebooks")
+		if err == nil {
+			p.Optional = true
+			p.Required = false
+			p.Default = true
+		}
+		p, err = common.SchemaPath(s, "workload_type", "clients", "jobs")
```
nfx, line 92: why not `tf:"default:true"`?

nfx (changes requested):

clusters/resource_cluster.go
nfx: so tf:optional is merged, can we remove what's not needed? :)

nfx (changes requested):

clusters/resource_cluster.go
alexott: I’ll update this tomorrow morning

nfx (approved):

PR #1463, opened 2022-07-15
Add more attributes to `databricks_instance_pool`

alexott:
P.S. Also added reporting of actual error when unmarshall JSON, otherwise it's hard to debug issues

### DIFF
--- a/common/http.go
+++ b/common/http.go
@@ -308,8 +308,7 @@ func (c *DatabricksClient) unmarshall(path string, body []byte, response any) er
 		ErrorCode:  "UNKNOWN",
 		StatusCode: 200,
 		Resource:   "..." + path,
-		Message: fmt.Sprintf("Invalid JSON received (%d bytes): %v",
-			len(body), string(body)),
+		Message:    fmt.Sprintf("%s\n%s", err.Error(), onlyNBytes(string(body), 32)),
 	}
 }
 
--- a/common/http_test.go
+++ b/common/http_test.go
@@ -269,7 +269,7 @@ func TestPost_Error(t *testing.T) {
 		ScimDetail: "some",
 	}, &resp)
 	require.Error(t, err)
-	assert.True(t, strings.HasPrefix(err.Error(), "Invalid JSON received (16 bytes)"),
+	assert.True(t, strings.HasPrefix(err.Error(), "invalid character 'c'"),
 		"Actual message: %s", err.Error())
 }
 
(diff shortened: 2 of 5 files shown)


nfx (changes requested):

common/http.go
```diff
@@ -308,8 +308,8 @@ func (c *DatabricksClient) unmarshall(path string, body []byte, response interfa
 		ErrorCode:  "UNKNOWN",
 		StatusCode: 200,
 		Resource:   "..." + path,
-		Message: fmt.Sprintf("Invalid JSON received (%d bytes): %v",
-			len(body), string(body)),
+		Message: fmt.Sprintf("Invalid JSON received (%d bytes): %v, error: %v",
```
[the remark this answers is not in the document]
alexott: added test

common/http_test.go
```diff
@@ -311,6 +311,18 @@ func TestUnmarshall(t *testing.T) {
...
+func TestUnmarshallError(t *testing.T) {
+	v := struct {
+		Correct   string `json:"c"`
+		Incorrect int    `json:"i"`
+	}{}
+	ws := DatabricksClient{}
+	err := ws.unmarshall("/a/b/c", []byte(`{"c":"val", "i":"123"}`), &v)
+	require.Error(t, err)
+	require.EqualError(t, err,
+		`Invalid JSON received (22 bytes): {"c":"val", "i":"123"}, error: json: cannot unmarshal string into Go struct field .i of type int`)
```
nfx, line 323: can you rather do it so that the error is `error: json: cannot unmarshal string into Go struct field .i of type int - '{"c":"val", "i":"12 (128 more bytes)'` (by using the `onlyNBytes` function)

nfx (changes requested):

common/http_test.go
alexott: done

common/http_test.go
```diff
@@ -311,6 +311,18 @@ func TestUnmarshall(t *testing.T) {
...
+func TestUnmarshallError(t *testing.T) {
+	v := struct {
+		Correct   string `json:"c"`
+		Incorrect int    `json:"i"`
+	}{}
+	ws := DatabricksClient{}
+	err := ws.unmarshall("/a/b/c", []byte(`{"c":"val", "i":"123"}`), &v)
+	require.Error(t, err)
+	require.EqualError(t, err,
+		`Invalid JSON: json: cannot unmarshal string into Go struct field .i of type int - {"c":"val", "i":"123"}`)
```
nfx, line 323: in 99.999% of cases there'll be `Invalid JSON: json: cannot`, so that's why i suggest to make the format as simple as `%s\n%s`

nfx (changes requested):

nfx (approved):
👍

PR #1468, opened 2022-07-15
Fixed `databricks_table` update for `column` block

### LINKED ISSUE #1455: [ISSUE] Updates to Unity table columns don't seem to take effect
Currently, changes we make to the fields of a `column` of an existing `databricks_table` don't seem to get reflected in the actual resource. The Terrraform plan reflects the expected change, but the apply step doesn't make a patch request to effect the change.

### Configuration
An example would be when we change the type of a column from string to integer.

For example when we start out with a table like this
```hcl
resource "databricks_table" "this" {
  name               = "foo"
  catalog_name       = "main"
  schema_name        = "bar"
  storage_location   = "s3://some-bucket/location"

  column {
    name      = "id"
    position  = 0
    type_name = "STRING"
    type_text = "varchar(64)"
  }
}
```
and then change it to this
```hcl
resource "databricks_table" "this" {
  name               = "foo"
  catalog_name       = "main"
  schema_name        = "bar"
  storage_location   = "s3://some-bucket/location"

  column {
    name      = "id"
    position  = 0
    type_name = "INT"
    type_text = "int"
  }
}
```

### Expected Behavior
We'd expect the Unity table to now have an "id" column of integer type.

### Actual Behavior
The "id" column is still of type string/varchar(64).

### Steps to Reproduce
1. `terraform plan` and `terraform apply` the first HCL snippet to create the table with an integer type "id" column
2. Exchange the first for the the second HCL snippet and `terraform plan` to create the table with an integer type "id" column
We see the expected diff 
```
  # databricks_table.this will be updated in-place
  ~ resource "databricks_table" "this" {
        id                 = "main.bar.foo"
        name               = "foo"
        # (7 unchanged attributes hidden)

      ~ column {
            name            = "id"
          ~ type_name       = "STRING" -> "INT"
          ~ type_text       = "varchar(64)" -> "int"
        }
        # (5 unchanged blocks hidden)
    }

```
3.  `terraform apply` the second HCL snippet to create the table with an "id" column of type string/varchar(64)

4. Run `terraform plan` again
The output suggests that the change got applied (`Apply complete! Resources: 0 added, 1 changed, 0 destroyed.`) However, no PATCH request is sent to the Databricks API, the schema change therefore doesn't reflect in the Unity table, and when we run terraform plan, we get the same diff as before the terraform apply.

nkvuong:
Close #1455

There must be a more programmatic way to do this...

### DIFF
--- a/catalog/update.go
+++ b/catalog/update.go
@@ -29,9 +29,18 @@ func updateFunctionFactory(pathPrefix string, updatable []string) func(context.C
 			}, field) {
 				continue
 			}
-			old, new := d.GetChange(field)
-			if !d.HasChange(field) {
-				continue
+
+			var old, new interface{}
+			if field == "columns" {
+				old, new = d.GetChange("column")
+				if !d.HasChange("column") {
+					continue
+				}
+			} else {
+				old, new = d.GetChange(field)
+				if !d.HasChange(field) {
+					continue
+				}
 			}
 
 			if contains([]string{
--- a/catalog/update.go
+++ b/catalog/update.go
@@ -21,6 +21,8 @@ func updateFunctionFactory(pathPrefix string, updatable []string) func(context.C
 	return func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 		patch := map[string]any{}
 		for _, field := range updatable {
+
+			// these fields cannot be set during creation
 			if d.IsNewResource() && !contains([]string{
 				"owner",
 				"delta_sharing_scope",
@@ -30,7 +32,7 @@ func updateFunctionFactory(pathPrefix string, updatable []string) func(context.C
 				continue
 			}
 
-			var old, new interface{}
+			var old, new any
 			if field == "columns" {
 				old, new = d.GetChange("column")
 				if !d.HasChange("column") {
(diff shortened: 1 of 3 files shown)


catalog/update.go
```diff
@@ -29,8 +29,22 @@ func updateFunctionFactory(pathPrefix string, updatable []string) func(context.C
 			}, field) {
 				continue
 			}
-			old, new := d.GetChange(field)
-			if !d.HasChange(field) {
+
+			var old, new interface{}
+			if field == "columns" {
```
alexott, line 34: wouldn't it be simpler to do:

```
if field == "columns" {
  field = "column"
}
old, new = d.GetChange(field)
if !d.HasChange(field) {
  continue
}

File: catalog/update.go [hunk 2]
```diff
@@ -29,8 +29,22 @@ func updateFunctionFactory(pathPrefix string, updatable []string) func(context.C
...
+				}
+			} else {
+				old, new = d.GetChange(field)
+				if !d.HasChange(field) {
+					continue
+				}
+			}
+
+			//ignore storage_location for MANAGED tables
+			if field == "storage_location" && d.Get("table_type") != "EXTERNAL" {
```
alexott, line 47: shouldn't it be handled as suppress diff?
Also, should we handle the case when people will change from managed to external?

catalog/update.go
nkvuong: the API expect `columns` in the payload however

catalog/update.go
nkvuong: `table_type` cannot be updated

PR #1474, opened 2022-07-18
add `PHOTON` runtime engine to cluster configuration

rondefreitas:
The PHOTON runtime engine feature may still be flagged as preview, but it is available for selection through both the UI and API, at least on AWS and the latest DBR releases.

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -96,6 +96,14 @@ const (
 	ClusterStateUnknown = "UNKNOWN"
 )
 
+// RuntimeEngine is for describing whether the Photon runtime engine was used
+type RuntimeEngine string
+
+const (
+	// Photon indicates that the Photon runtime engine is being used
+	Photon = "PHOTON"
+)
+
 var stateMachine = map[ClusterState][]ClusterState{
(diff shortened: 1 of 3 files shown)


alexott:
Photon usually requires both node type and Spark version. That’s why we have corresponding selector in both data sources…

rondefreitas:
That's correct, @alexott , however without this feature flag on the cluster configuration, you cannot use terraform to set clusters (or job clusters) to leverage Photon.

alexott:
```hcl
data "databricks_node_type" "photon" {
  photon_worker_capable = true
}

data "databricks_spark_version" "photon_lts" {
  long_term_support = true
  photon            = true
}

resource "databricks_cluster" "photon" {
  cluster_name            = "Photon Cluster"
  spark_version           = data.databricks_spark_version.photon_lts.id
  node_type_id            = data.databricks_node_type.photon.id
  autotermination_minutes = 20
  num_workers             = 3
}
```

creates Photon-enabled cluster:

Screenshot 2022-07-19 at 08 04 02

rondefreitas:
Closing for now while I investigate further on my end... My terraform logic is the same as yours and yet I end up with no photon runtime

PR #1479, opened 2022-07-19
add support for gcp network info block

### LINKED ISSUE #1393: add support for byo vpc based workspace creation on gcp
please add GCP support to https://github.com/databricks/terraform-provider-databricks/blob/master/docs/resources/mws_networks.md , byo vpc on GCP feature - https://docs.gcp.databricks.com/administration-guide/cloud-configurations/gcp/customer-managed-vpc.html

stikkireddy:
- [ ] create doc updates/guide updates
- [ ] run acceptance tests for aws
- [ ] test creating mws for aws with 1.0.2 and then upgrade and see impact on previous state
- [x] unit tests

Attempts to resolve #1393 

byovpc example:

```hcl
resource "google_compute_network" "dbx_private_vpc" {
  project                 = var.google_project
  name                    = "sri-tf-network"
  auto_create_subnetworks = false
#  mtu                     = 1460
}

resource "google_compute_subnetwork" "network-with-private-secondary-ip-ranges" {
  name          = "test-dbx"
  ip_cidr_range = "10.0.0.0/16"
  region        = "us-central1"
  network       = google_compute_network.dbx_private_vpc.id
  secondary_ip_range {
    range_name    = "pods"
    ip_cidr_range = "10.1.0.0/16"
  }
  secondary_ip_range {
    range_name    = "svc"
    ip_cidr_range = "10.2.0.0/20"
  }
}

resource "databricks_mws_networks" "this" {
  provider       = databricks.accounts
  account_id     = var.databricks_account_id
  network_name = "sri-test-fs-demo"
  gcp_network_info {
    network_project_id = var.google_project
    vpc_id = google_compute_network.dbx_private_vpc.name
    subnet_id = google_compute_subnetwork.network-with-private-secondary-ip-ranges.name
    subnet_region = google_compute_subnetwork.network-with-private-secondary-ip-ranges.region
    pod_ip_range_name = "pods"
    service_ip_range_name = "svc"
  }
}
```


mws/mws.go
```diff
@@ -46,19 +46,30 @@ type NetworkVPCEndpoints struct {
...
+	VPCID            string               `json:"vpc_id,omitempty"`
+	SubnetIds        []string             `json:"subnet_ids,omitempty" tf:"slice_set"`
 	VPCEndpoints     *NetworkVPCEndpoints `json:"vpc_endpoints,omitempty" tf:"computed,force_new"`
-	SecurityGroupIds []string             `json:"security_group_ids" tf:"slice_set"`
+	SecurityGroupIds []string             `json:"security_group_ids,omitempty" tf:"slice_set"`
 	VPCStatus        string               `json:"vpc_status,omitempty" tf:"computed"`
 	ErrorMessages    []NetworkHealth      `json:"error_messages,omitempty" tf:"computed"`
 	WorkspaceID      int64                `json:"workspace_id,omitempty" tf:"computed"`
 	CreationTime     int64                `json:"creation_time,omitempty" tf:"computed"`
+	GcpNetworkInfo   *GcpNetworkInfo      `json:"gcp_network_info,omitempty"`
```
alexott, line 72: If you don’t specify this on gcp, what read operation returns? Maybe we need suppress diff?

alexott (commented):
From code perspective looks good. But I haven’t seen api docs. It also makes sense to validate against api spec in the source code

mws/mws.go
stikkireddy: afaik you cannot create a network object in gcp without this field being configured. I will be testing it on aws which shouldnt return an object for this struct. @bhavink-db can correct me if i am wrong.

alexott: I meant that sometimes get api returns less information, or instead of null returns empty object

common/version.go
```diff
@@ -3,7 +3,7 @@ package common
 import "context"
 
 var (
-	version = "1.0.2" // TODO: check if automatable via https://pkg.go.dev/runtime/debug@master#ReadBuildInfo
+	version = "1.0.3" // TODO: check if automatable via https://pkg.go.dev/runtime/debug@master#ReadBuildInfo
```
nfx, line 6:
```suggestion
```

this change is done in release prs

nfx (changes requested):
Please add an acceptance test in order for this to qualify for merge. You can expect `DATABRICKS_ACCOUNT_ID`, `GOOGLE_PROJECT`, `TEST_PREFIX`,  and `GOOGLE_REGION` environment variables to be there.

nfx:
Thank you for the example, will also add the BYOVPC to integration testing infra.

mws/mws.go
stikkireddy: i validated this via drift to see if anything changes; after creating it i ran apply again to avoid any drift

but i can do two additional things:

1. confirm with eng on this api to confirm that the input is exactly what will be returned by the get and not be omitted or returned as a null. 
2. I think i can also confirm this by trying an import for an existing network and see if there is any drift.

or if you think we should add a suppress diff

@alexott let me know if this is sufficient.

nfx:
@stikkireddy could you also remove `TEST_PREFIX` env variable in the other gcp test? =)

redcape:
@stikkireddy / @nfx is this still active?

nfx:
okay, we can live without full "read" implementation, so that we can release this faster

stikkireddy:
I can pick this up next week and ignore the changes

nfx:
@stikkireddy i think you can close the PR yourself if needed ;)

PR #1507, opened 2022-07-29
Fixed `databricks_permissions` for calling user to `CAN_MANAGE` on `databricks_mlflow_model`

### LINKED ISSUE #1504: [ISSUE] Granting permissions to an MLflow model removes manage permissions from MLflow model creator by default
Hi there,

Thank you for opening an issue. Please note that we try to keep the Databricks Provider issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Configuration
I'm currently applying the following Terraform config, authenticating to Databricks using an Azure service principal.

```hcl
resource "databricks_mlflow_model" "registered_model" {
  name = "staging-mlops-azure-demo-model"
  description = "My MLflow model description"
}

resource "databricks_permissions" "mlflow_model_permissions" {
  registered_model_id = databricks_mlflow_model.registered_model.registered_model_id

  access_control {
    group_name       = "${local.read_user_group}"
    permission_level = "CAN_READ"
  }
}
```

### Expected Behavior
What should have happened?

I would expect that a) the registered model is created and b) that the "users" group is granted read access to the model in addition to c) preserving `CAN_MANAGE` permissions on the model for the service principal that created the model.

### Actual Behavior
a) and b) do happen, but c) does not. The configuration removes `CAN_MANAGE` permissions from the Databricks service principal applying Terraform changes, resulting in the following error:

```
╷
│ Error: cannot create permissions: *** does not have Owner permissions on 4e4cd6a8cafe4a14b942d236dc804f7c. Please contact the owner or an administrator for access. Using pat auth: host=https://adb-1173035118216611.11.azuredatabricks.net/, token=***REDACTED***
│ 
│   with databricks_permissions.mlflow_model_permissions,
│   on permissions.tf line 28, in resource "databricks_permissions" "mlflow_model_permissions":
│   28: resource "databricks_permissions" "mlflow_model_permissions" {
│ 
╵
```

Users are successfully granted read permissions, but the model is shown to be managed only by admins in the UI:
image

For MLflow models (seemingly also Databricks clusters), you can remove `CAN_MANAGE` permissions on the model owner (creator), without specifying another owner. This means that if you declare permissions on a model using infra as code, it’s pretty easy to create a model and accidentally remove the owner’s permissions, as shown in the above example. There is a user-code fix which is to explicitly grant the current service principal manage permissions on the model, but this is counterintuitive & a little brittle (requires knowing whether the user applying the Terraform config is a normal user vs service principal):

```
data "databricks_current_user" "current_sp" {}

resource "databricks_permissions" "mlflow_model_permissions" {
  registered_model_id = databricks_mlflow_model.registered_model.registered_model_id

  access_control {
    group_name       = "${local.read_user_group}"
    permission_level = "CAN_READ"
  }
  // New: preserve CAN_MANAGE permissions on the model for the current user
  access_control {
    service_principal_name       = data.databricks_current_user.current_sp.user_name
  }
}
```

I'm curious if this issue also affects clusters which have a similar permission model to MLflow models. If no, I wonder if we can reuse the fix for clusters for MLflow models. Either way, it seems we should also consider updating the underlying permissions REST APIs to ensure that there is always an owner of the MLflow model with `CAN_MANAGE` permissions, similar to [Databricks job permissions](https://docs.databricks.com/security/access-control/jobs-acl.html), but that may be further away

### Steps to Reproduce
1. `terraform apply`

Terraform version:

```
Terraform v1.2.2
on darwin_amd64
```

Provider version: 1.1.0

### Debug Output
Please add turn on logging, e.g. `TF_LOG=DEBUG terraform apply` and run command again, paste it to gist & provide the link to gist. If you're still willing to paste in log output, make sure you provide only relevant log lines with requests.

It would make it more readable, if you pipe the log through `| grep databricks | sed -E 's/^.* plugin[^:]+: (.*)$/\1/'`, e.g.:

```
TF_LOG=DEBUG terraform plan 2>&1 | grep databricks | sed -E 's/^.* plugin[^:]+: (.*)$/\1/'
```

If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Important Factoids
Are there anything atypical about your accounts that we should know?

smurching:
Ensure that calling user always has CAN_MANAGE permission on MLflow models.

Fixes #1504


permissions/resource_permissions.go
```diff
@@ -124,18 +124,40 @@ func urlPathForObjectID(objectID string) string {
...
+// Helper function for applying permissions changes. Ensures that
+// we select the correct HTTP method based on the object type and preserve the calling
+// user's ability to manage the specified object when applying permissions changes.
 func (a PermissionsAPI) put(objectID string, objectACL AccessControlChangeList) error {
+	objectACL, err := a.ensureCurrentUserCanManageObject(objectID, objectACL)
+	if err != nil {
+		return err
+	}
 	if isDbsqlPermissionsWorkaroundNecessary(objectID) {
-		// SQLA entities always have `CAN_MANAGE` permission for the calling user.
```
nfx, line 130: Probably this line is no longer necessary.

nfx (changes requested):
Good start. Make sure you have 100% unit test coverage for the changes

smurching:
Thanks @nfx ! I just added & pushed some tests if you're able to take another pass :)

permissions/resource_permissions_test.go
```diff
@@ -121,6 +121,167 @@ func TestResourcePermissionsRead_RemovedCluster(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		Resource: ResourcePermissions(),
+		Read:     true,
+		New:      true,
+		ID:       "/registered-models/fakeuuid123",
+	}.Apply(t)
+	assert.NoError(t, err, err)
```
nfx, line 153: Please rewrite with ApplyAndEqual method of resource fixture. It provides more readable tests

permissions/resource_permissions_test.go
```diff
@@ -121,6 +121,167 @@ func TestResourcePermissionsRead_RemovedCluster(t *testing.T) {
...
+			"registered_model_id": "fakeuuid123",
+			"access_control": []any{
+				map[string]any{
+					"user_name":        TestingUser,
+					"permission_level": "CAN_READ",
+				},
+			},
+		},
+		Create: true,
+	}.Apply(t)
```
nfx, line 212: Rewrite with ApplyAndEqual

permissions/resource_permissions_test.go
```diff
@@ -121,6 +121,167 @@ func TestResourcePermissionsRead_RemovedCluster(t *testing.T) {
...
+	assert.NoError(t, err, err)
+	ac := d.Get("access_control").(*schema.Set)
+	require.Equal(t, 1, len(ac.List()))
+	firstElem := ac.List()[0].(map[string]any)
+	assert.Equal(t, TestingUser, firstElem["user_name"])
+	assert.Equal(t, "CAN_READ", firstElem["permission_level"])
+}
+
+func TestResourcePermissionsUpdate_Mlflow_Model(t *testing.T) {
```
nfx, line 221: Where’s delete? =)

permissions/resource_permissions_test.go
```diff
@@ -121,6 +121,167 @@ func TestResourcePermissionsRead_RemovedCluster(t *testing.T) {
...
+		access_control {
+			user_name = "ben"
+			permission_level = "CAN_READ"
+		}
+		`,
+		Resource: ResourcePermissions(),
+		Update:   true,
+		// Removed:  true,
+		ID: "/registered-models/fakeuuid123",
+	}.Apply(t)
```
nfx, line 275: Please rewrite with more readable and newer version - ApplyAndEqual

nfx (changes requested):
Tests have to be made more readable and add a test for deleting

permissions/resource_permissions_test.go
smurching: Thanks for the pointer - I can't find an `ApplyAndEqual` method, do you mean `ApplyAndExpectData`? I'm trying that out but hitting issues with [an assumption in that method](https://github.com/databricks/terraform-provider-databricks/blob/c72250b58ca18883fd62745902cbdd33eb881527/qa/testing.go#L273) that the key of the provided data object is a string array. Thus when I try to pass a `data` argument (representing expected resource data) like

```
	expectedData := map[string]any{
		"registered_model_id": "fakeuuid123",
		"access_control": []any{
			map[string]any{
				"user_name":        TestingUser,
				"permission_level": "CAN_READ",
			},
		},
	}
```

I get a runtime type conversion error:

```
--- FAIL: TestResourcePermissionsRead_Mlflow_Model (0.27s)
panic: interface conversion: interface {} is []interface {}, not []string [recovered]
	panic: interface conversion: interface {} is []interface {}, not []string

goroutine 8 [running]:
testing.tRunner.func1.2({0x18d1900, 0xc00014c390})
	/usr/local/Cellar/go/1.18.4/libexec/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/usr/local/Cellar/go/1.18.4/libexec/src/testing/testing.go:1392 +0x39f
panic({0x18d1900, 0xc00014c390})
	/usr/local/Cellar/go/1.18.4/libexec/src/runtime/panic.go:838 +0x207
github.com/databricks/terraform-provider-databricks/qa.ResourceFixture.ApplyAndExpectData({{0xc000538aa0, 0x2, 0x2}, 0xc00071a2a0, 0x0, 0x0, 0x0, {0x0, 0x0}, 0x0, ...}, ...)
	/Users/sid.murching/go/src/terraform-provider-databricks/qa/testing.go:274 +0x53d
github.com/databricks/terraform-provider-databricks/permissions.TestResourcePermissionsRead_Mlflow_Model(0x0?)
	/Users/sid.murching/go/src/terraform-provider-databricks/permissions/resource_permissions_test.go:161 +0x528
testing.tRunner(0xc000702b60, 0x1a21d88)
	/usr/local/Cellar/go/1.18.4/libexec/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/usr/local/Cellar/go/1.18.4/libexec/src/testing/testing.go:1486 +0x35f
```

I naively tried the solution in [this StackOverflow post](https://stackoverflow.com/a/18211675/12111347) for asserting deep equality, but found that `access_control` blocks seem to have a different type from what's expected & also contain more fields than I care to assert on:

```
2022/08/01 21:52:11 [DEBUG] GET /api/2.0/preview/scim/v2/Me
2022/08/01 21:52:11 [DEBUG] 200 OK  {
  "userName": "admin"
} <- GET /api/2.0/preview/scim/v2/Me
    /Users/sid.murching/go/src/terraform-provider-databricks/permissions/testing.go:275:
        	Error Trace:	/Users/sid.murching/go/src/terraform-provider-databricks/permissions/testing.go:275
        	            				/Users/sid.murching/go/src/terraform-provider-databricks/permissions/resource_permissions_test.go:161
        	Error:      	Should be true
        	Test:       	TestResourcePermissionsRead_Mlflow_Model
        	Messages:   	&{%!s(schema.SchemaSetFunc=0x1722300) map[1040111181:map[group_name: permission_level:CAN_READ service_principal_name: user_name:ben]] {%!s(uint32=1) {%!s(int32=0) %!s(uint32=0)}}} did not have expected value [map[permission_level:CAN_READ user_name:ben]]
--- FAIL: TestResourcePermissionsRead_Mlflow_Model (0.27s)
FAIL
FAIL	github.com/databricks/terraform-provider-databricks/permissions	1.788s
```

Any suggestions here? I'd prefer not to make major changes to the lower level testing library in this PR as a Golang noob :) but happy to send a follow-up PR to use `ApplyAndExpectData` if it's enhanced to support my use case

permissions/resource_permissions_test.go
smurching: Happy to add a test for delete!

permissions/resource_permissions_test.go
smurching: For good measure I also did some searching and found that `ApplyAndExpectData` is always called on a map from `string -> string[]` or map from `string -> string`, but that seems insufficient for testing the `access_control` blocks in my case

nfx: @smurching i've realised that i wrote `ApplyAndExpectData` to work properly with sets of strings only. i'm relaxing the requirement on `ApplyAndExpectData` for this pr :)

nfx (approved):

PR #1510, opened 2022-08-01
Adds PolicyId and ApplyPolicyDefaultId parameters for resource_pipeline.

divyavanmahajan:
Databricks DLT pipelines provide cluster configuration parameters. Two parameters Policy_id and apply_policy_default_id are available in the Databricks Pipeline API - cluster configuration parameters. This adds them to the provider.


divyavanmahajan:
This may not be needed anymore since pull request 1554 - is already merged.

PR #1534, opened 2022-08-15
Fix node_type data source by skipping node types that aren't available from cloud provider

### LINKED ISSUE #1533: [ISSUE] Issue with data `databricks_node_type` (Azure) while setting local_disk = true
When we set local_disk=true in data “databricks_node_type” to fetch the node, it picks up unsupported node(Standard_L8as_v3) in the region (Central India) as well. 

While the [list-node-types](https://docs.databricks.com/dev-tools/api/latest/clusters.html#list-node-types) API is called in the process and receives the below status for that unsupported node in the response **(for Azure)**- 

"node_info": {
        "status": [
          "NotEnabledOnSubscription",
          "NotAvailableInRegion"
        ]
}

The data “databricks_node_type” doesn’t honor the "NotAvailableInRegion" status for that node_type_id and give that as a response. 

_Went through the [provider code](https://github.com/databricks/terraform-provider-databricks/blob/master/clusters/data_node_type.go) and doesn’t look like the exception is handled._

### Configuration
```hcl
  local_disk = true
}

  long_term_support = true
}

  autotermination_minutes = 20
  autoscale {
    min_workers = 1
    max_workers = 50
  }
}
```

### Expected Behavior
The node_type_id returned via data "databricks_node_type" should be a supported one in that region. (Azure)

### Actual Behavior
It is returning unsupported nodes too, for e.g Standard_L8as_v3 in Central India (Azure)

Error thrown is - 
Could not launch cluster due to cloud provider failures. azure_error_code: InvalidParameter, azure_error_message: The requested VM size Standard_L8as_v3 is not available in the current region. The ..., Termination info: code: CLOUD_PROVIDER_LAUNCH_FAILURE, type: CLOUD_FAILURE, parameters: map[azure_error_code:InvalidParameter azure_error_message:The requested VM size Standard_L8as_v3 is not available in the current region. 

### Steps to Reproduce
You can run the above code in a workspace region where the Standard_L8as_v3 node is not supported.

### Debug Output
https://gist.github.com/pratik-databricks/d92103543eeb7f21b6ef6358233687cb

alexott:
Also, fixed tests that use azure-cli - Go stopped to execute programs when PATH has relative paths

This fixes #1533

### DIFF
--- a/clusters/data_node_type.go
+++ b/clusters/data_node_type.go
@@ -50,6 +50,11 @@ func (l *NodeTypeList) Sort() {
 	})
 }
 
+const (
+	CloudProviderNodeStatusNotEnabled           = "NotEnabledOnSubscription"
+	CloudProviderNodeStatusNotAvailableInRegion = "NotAvailableInRegion"
+)
+
 // ClusterCloudProviderNodeInfo encapsulates the existing quota available from the cloud service provider.
 type ClusterCloudProviderNodeInfo struct {
 	Status             []string `json:"status,omitempty"`
@@ -89,6 +94,19 @@ type NodeType struct {
 	Graviton              bool                          `json:"is_graviton,omitempty"`
 }
 
+func (nt NodeType) shouldBeSkipped() bool {
+	if nt.NodeInfo == nil {
+		return false
+	}
+	for _, st := range nt.NodeInfo.Status {
+		switch st {
+		case CloudProviderNodeStatusNotAvailableInRegion, CloudProviderNodeStatusNotEnabled:
+			return true
+		}
+	}
+	return false
+}
+
 func (a ClustersAPI) defaultSmallestNodeType() string {
 	if a.client.IsAzure() {
 		return "Standard_D3_v2"
@@ -117,6 +135,9 @@ func (a ClustersAPI) GetSmallestNodeType(r NodeTypeRequest) string {
 	}
 	list.Sort()
 	for _, nt := range list.NodeTypes {
+		if nt.shouldBeSkipped() {
+			continue
+		}
 		gbs := (nt.MemoryMB / 1024)
 		if r.VCPU && !strings.HasPrefix(nt.NodeTypeID, "vcpu") {
 			continue
--- a/clusters/data_node_type_test.go
+++ b/clusters/data_node_type_test.go
@@ -224,3 +224,83 @@ func TestSmallestNodeTypeClouds(t *testing.T) {
 		},
 	}.defaultSmallestNodeType())
 }
+
+func TestNodeTypeCategoryNotAvailable(t *testing.T) {
+	d, err := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:       "GET",
+				ReuseRequest: true,
+				Resource:     "/api/2.0/clusters/list-node-types",
+				Response: NodeTypeList{
+					[]NodeType{
+						{
+							NodeTypeID:     "Random_05",
+							InstanceTypeID: "Random_05",
+							MemoryMB:       1024,
+							NumCores:       32,
+							NodeInstanceType: &NodeInstanceType{
+								LocalDisks:      3,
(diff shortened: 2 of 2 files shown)


clusters/data_node_type.go
```diff
@@ -117,6 +122,18 @@ func (a ClustersAPI) GetSmallestNodeType(r NodeTypeRequest) string {
 	}
 	list.Sort()
 	for _, nt := range list.NodeTypes {
+		if nt.NodeInfo != nil {
+			shouldBeSkipped := false
+			for _, st := range nt.NodeInfo.Status {
+				if st == CloudProviderNodeStatusNotAvailableInRegion || st == CloudProviderNodeStatusNotEnabled {
+					shouldBeSkipped = true
```
nfx, line 129: please create a method on NodeInfo and reduce the level of nesting to one

nfx (changes requested):

clusters/data_node_type.go
```diff
@@ -89,6 +94,19 @@ type NodeType struct {
...
+func (nt NodeType) shouldBeSkipped() bool {
+	shouldBeSkipped := false
+	if nt.NodeInfo != nil {
+		for _, st := range nt.NodeInfo.Status {
+			if st == CloudProviderNodeStatusNotAvailableInRegion || st == CloudProviderNodeStatusNotEnabled {
+				shouldBeSkipped = true
+				break
+			}
+		}
+	}
```
redcape, line 106: should we also check `node_info.available_core_quota >= num_cores`

alexott: not sure if it will help a lot, as we're working on the individual node level...

redcape: I suppose it depends on the use-case. For mounts, it will help since we only start one node always and use it for all mounts. When the use-case is to for user-defined clusters, it's less clear how much it would help since the number of cluster and number of executors is unknown at this stage.

clusters/data_node_type.go
```diff
@@ -89,6 +94,19 @@ type NodeType struct {
 	Graviton              bool                          `json:"is_graviton,omitempty"`
 }
 
+func (nt NodeType) shouldBeSkipped() bool {
+	shouldBeSkipped := false
+	if nt.NodeInfo != nil {
```
nfx, line 99: i'll quote my favorite chapter from https://livebook.manning.com/book/100-go-mistakes-how-to-avoid-them/chapter-2/v-10/30
image

clusters/data_node_type.go
```diff
@@ -89,6 +94,19 @@ type NodeType struct {
...
+		for _, st := range nt.NodeInfo.Status {
+			if st == CloudProviderNodeStatusNotAvailableInRegion || st == CloudProviderNodeStatusNotEnabled {
```
nfx, line 101: nit: i think switch statement would look more elegant here

clusters/data_node_type.go
```diff
@@ -89,6 +94,19 @@ type NodeType struct {
...
+				shouldBeSkipped = true
```
nfx, line 102: why not immediately return here? would be more robust.

nfx (changes requested):

clusters/data_node_type.go
alexott: addressed all comments

clusters/data_node_type.go
```diff
@@ -89,6 +94,18 @@ type NodeType struct {
 	Graviton              bool                          `json:"is_graviton,omitempty"`
 }
 
+func (nt NodeType) shouldBeSkipped() bool {
+	if nt.NodeInfo != nil {
```
nfx, line 98: can we further simplify it to `if nt.NodeInfo == nil return false`?

nfx (changes requested):

clusters/data_node_type.go
redcape: comment doesn't block merge

nfx (approved):

PR #1538, opened 2022-08-15
Fix Azure CLI tests on Go 1.19

alexott:
In Go 1.19 the Command and LookPath no longer allow results from a PATH search to be found relative to the current directory. https://go.dev/doc/go1.19#os-exec-path


nfx (approved):
it's great that tests are passing on go1.18

PR #1548, opened 2022-08-17
[BUG FIX] Cluster edit fails if cluster transitions into TERMINATED s…

shreyas-goenka:
…tate from PENDING / RESIZING / RESTARTING states

To reproduce issue:
1. start clus
2. edit terraform config and terraform apply (while cluster is pending)
3. terminate cluster

You would get a error message like:
```
databricks_cluster.shared_autoscaling: Modifying... [id=0815-200440-bsovpk1b]
databricks_cluster.shared_autoscaling: Still modifying... [id=0815-200440-bsovpk1b, 10s elapsed]
databricks_cluster.shared_autoscaling: Still modifying... [id=0815-200440-bsovpk1b, 20s elapsed]
╷
│ Error: cannot update cluster: 0815-200440-bsovpk1b is not able to transition from TERMINATED to RUNNING: Termination requested by shreyas.goenka@databricks.com, Termination info: code: USER_REQUEST, type: SUCCESS, parameters: map[username:shreyas.goenka@databricks.com]. Please see https://docs.databricks.com/dev-tools/api/latest/clusters.html#clusterclusterstate for more details
│ 
│   with databricks_cluster.shared_autoscaling,
│   on main.tf line 49, in resource "databricks_cluster" "shared_autoscaling":
│   49: resource "databricks_cluster" "shared_autoscaling" {
```

This issue happens because the Edit API method incorrectly assumes that PENDING / RESIZING / RESTARTING states will be followed by RUNNING state, and since RUNNING state can't be reached from TERMINATED state we see this error

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -542,8 +542,8 @@ func (a ClustersAPI) Edit(cluster Cluster) (info ClusterInfo, err error) {
 		// it's already running or terminated, so we're safe to edit
 		break
 	case ClusterStatePending, ClusterStateResizing, ClusterStateRestarting:
-		// let's wait tiny bit, so we return RUNNING cluster info
-		info, err = a.waitForClusterStatus(info.ClusterID, ClusterStateRunning)
+		// let's wait tiny bit, so we return RUNNING or TERMINATED cluster info
+		info, err = a.waitForClusterStatusRunningOrTerminated(info.ClusterID)
 		if err != nil {
 			return info, err
 		}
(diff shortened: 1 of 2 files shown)


clusters/clusters_api.go
```diff
@@ -542,8 +542,8 @@ func (a ClustersAPI) Edit(cluster Cluster) (info ClusterInfo, err error) {
 		// it's already running or terminated, so we're safe to edit
 		break
 	case ClusterStatePending, ClusterStateResizing, ClusterStateRestarting:
-		// let's wait tiny bit, so we return RUNNING cluster info
```
nfx, line 545: Can you have two branches: one where you wait for terminated, the other where you wait for running.

creating a method with duplicate logic is not a good idea

clusters/clusters_api_test.go
```diff
@@ -166,6 +166,137 @@ func TestGetOrCreateRunningCluster_Existing_AzureAuth(t *testing.T) {
...
+	assert.NotNil(t, clusterInfo)
+}
+
+func TestWaitForClusterStatusRunningOrTerminated_StopRetryingEarly(t *testing.T) {
+	client, server, err := qa.HttpFixtureClient(t, []qa.HTTPFixture{
+		{
+			Method:   "GET",
+			Resource: "/api/2.0/clusters/get?cluster_id=abc",
+			Response: common.APIErrorBody{
+				Message: "I am a teapot. The cutest little teapot.",
```
nfx, line 205: 😂

HTTP 418 is real. https://www.rfc-editor.org/rfc/rfc2324#section-2.3.2

nfx (changes requested):

shreyas-goenka:
On further discussion, we decided to not merge this bug fix since resource_cluster is super critical, users are not expected to manually edit terraformed resources and retry by user is a very easy and valid workaround

PR #1554, opened 2022-08-19
Add cluster policy support for `databricks_pipeline`

### LINKED ISSUE #1551: [FEATURE] Support policies in DLT pipelines
We use policies to manage the configurations/cost of clusters in our environments (including dlt pipelines), so that users can create their own pipelines within the guards we have set.

I can see currently that the provider doesn't support using policies with dlt pipelines, when will it be added?

alexott:
Also added `apply_policy_default_values` attribute to automatically add missing attributes from the policy.

This fixes #1551


alexott:
ah, we have #1510 already

nfx (approved):

PR #1560, opened 2022-08-24
Make debugging work again

nfx:
TODO: test


nfx:
image
quick test confirms it working as expected

PR #1567, opened 2022-08-26
Remove support for `ANY` access level in `mws_private_access_settings`

alexott:
Support for `ANY` value for `private_access_level` in `mws_private_access_settings`
resource was removed. `ACCOUNT` is the new default now


nfx (approved):
👍

PR #1570, opened 2022-08-30
Remove "Public Preview" marker from Unity Catalog resources docs

alexott:
Unity Catalog (UC) moved to General Availability last week, so we need to remove Public Preview marker from documentation for UC resources

### DIFF
--- a/docs/resources/catalog.md
+++ b/docs/resources/catalog.md
@@ -3,8 +3,6 @@ subcategory: "Unity Catalog"
 ---
 # databricks_catalog Resource
 
--> **Public Preview** This feature is in [Public Preview](https://docs.databricks.com/release-notes/release-types.html). Contact your Databricks representative to request access. 
-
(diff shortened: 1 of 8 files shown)


docs/resources/table.md
```diff
@@ -3,8 +3,6 @@ subcategory: "Unity Catalog"
 ---
 # databricks_table (Resource)
 
--> **Public Preview** This feature is in [Public Preview](https://docs.databricks.com/release-notes/release-types.html). Contact your Databricks representative to request access.
```
nfx, line 6: @alexott can we keep preview just for table? we might deprecate this one...

nfx (changes requested):

docs/resources/table.md
alexott: done

nfx (approved):

PR #1577, opened 2022-09-01
Fix `databricks_user` creation with `force` on account

### LINKED ISSUE #1549: [ISSUE] `databricks_user` with `force` on account does not always work
### Configuration
```hcl
resource "databricks_user" "uc_admin" {
  for_each  = toset(module.defaults.admins)
  user_name = each.key
  force     = true
}
```

### Expected Behavior
`terraform apply` goes on

### Actual Behavior
`terraform apply` tells that users are in the account already

v1.0.1

nkvuong:
Backend API changes error messages, updating code to match

Closes #1549

### DIFF
--- a/scim/resource_user.go
+++ b/scim/resource_user.go
@@ -10,6 +10,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func userExistsErrorMessage(userName string, isAccount bool) string {
+	if isAccount {
+		return fmt.Sprintf("User with email %s already exists in this account", userName)
+	} else {
+		return fmt.Sprintf("User with username %s already exists.", userName)
+	}
+}
+
 // ResourceUser manages users within workspace
 func ResourceUser() *schema.Resource {
 	type entity struct {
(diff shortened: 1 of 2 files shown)


scim/resource_user.go
```diff
@@ -85,7 +85,7 @@ func createForceOverridesManuallyAddedUser(err error, d *schema.ResourceData, us
 	}
 	// corner-case for overriding manually provisioned users
 	userName := strings.ReplaceAll(u.UserName, "'", "")
-	force := fmt.Sprintf("User with username %s already exists.", userName)
+	force := fmt.Sprintf("User with email %s already exists in this account", userName)
```
alexott, line 88: Will it continue to work with the workspace level identities?

scim/resource_user_test.go
```diff
@@ -459,7 +459,7 @@ func TestCreateForceOverwriteCannotListUsers(t *testing.T) {
 		d := ResourceUser().TestResourceData()
 		d.Set("force", true)
 		err := createForceOverridesManuallyAddedUser(
-			fmt.Errorf("User with username me@example.com already exists."),
+			fmt.Errorf("User with email me@example.com already exists in this account"),
```
alexott, line 462: Maybe move this format string into a constant, or create a function that will accept email & return string?

nkvuong: refactored the error messages into a function

scim/resource_user.go
nkvuong: you are correct - updated the correct error message

alexott (approved):

nfx (approved):

PR #1578, opened 2022-09-02
add `force_destroy` flag to `databricks_schema` & `databricks_catalog`

### LINKED ISSUE #1576: [FEATURE] Provide purge flag for databricks_schema
Currently when doing a terraform destroy if there are existing tables the schema can't be destroyed, it would be ideal if there was a purge flag so if when set to true it will go through and drop all tables in schema when its being destroyed. See databricks-cli for purge functionality https://docs.databricks.com/dev-tools/cli/unity-catalog-cli.html#delete-a-schema

nkvuong:
Add `force_destroy` flag to UC schema & catalog. This achieves the same functionality as the `purge` flag in the cli, which loop through all child schemas & tables and delete them before delete the parent securable

Closes #1576

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -51,7 +51,9 @@ func (a CatalogsAPI) deleteCatalog(name string, force bool) error {
 			return err
 		}
 		for _, s := range schemas.Schemas {
-			schemasAPI.deleteSchema(s.Name, true)
+			if s.Name != name+".information_schema" {
+				schemasAPI.deleteSchema(s.FullName, true)
+			}
 		}
 	}
 	return a.client.Delete(a.context, "/unity-catalog/catalogs/"+name, nil)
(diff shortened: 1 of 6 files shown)


catalog/resource_catalog.go
```diff
@@ -43,14 +43,30 @@ func (a CatalogsAPI) getCatalog(name string) (ci CatalogInfo, err error) {
 	return
 }
 
-func (a CatalogsAPI) deleteCatalog(name string) error {
-	// TODO: force_destroy attribute
-	return a.client.Delete(a.context, "/unity-catalog/catalogs/"+name+"?force=true", nil)
+func (a CatalogsAPI) deleteCatalog(name string, force bool) error {
```
nfx, line 46: @nkvuong can you create a separate method and call it `forceDelete`? this will align with future SDK migration

nfx (changes requested):

nfx (approved):

PR #1584, opened 2022-09-07
Fix TestAccAwsOboFlow

pietern:
The test creates an SP to test the on-behalf-of token creation flow. It failed because the SP wasn't active.


nfx (approved):
Thanks for fixing! 

hm... really wonder why service principals are not created active now.

PR #1606, opened 2022-09-16
clean up hcl parse errors in `/docs`

crankswagon:
general linting for /docs so new contributors should not longer see hcl parse errors when running `make fmt-docs`


nfx (approved):
👍

PR #1607, opened 2022-09-16
Fix Azure authentication for dev and staging workspaces

mehdikit-db:


common/azure_auth.go
```diff
@@ -17,9 +17,21 @@ import (
 )
 
 // List of management information
-const armDatabricksResourceID string = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
+const armDatabricksProdResourceID string = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
+const armDatabricksStagingResourceID string = "4a67d088-db5c-48f1-9ff2-0aace800ae68"
+const armDatabricksDevResourceID string = "62a912ac-b58e-4c1d-89ea-b2dbfc7358fc"
+
+func (aa *DatabricksClient) GetArmDatabricksResourceID() string {
```
nfx, line 24: @mehdikit-db i'd suggest not to leak our internal resource UUIDs and rather supply this as configurable property through `DatabricksClient`, that gets to provider config (if mapped the right way)

nfx (changes requested):

common/azure_auth.go
```diff
@@ -17,9 +17,18 @@ import (
 )
 
 // List of management information
-const armDatabricksResourceID string = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
+const azureDatabricksProdLoginAppID string = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
+
+func (aa *DatabricksClient) GetAzureDatabricksLoginAppId() string {
+	res := azureDatabricksProdLoginAppID
```
nfx, line 23: 1) make this method "unexported" - make it start with lowecase letter (`getAzure...`)
2) `strings.Contains(aa.Host, ".staging.")` is unnecessary and it leaks internal infra references. simplify it to

```
if aa.AzureDatabricksLoginAppId != "" {
			return aa.AzureDatabricksLoginAppId
		}
return azureDatabricksProdLoginAppID
```

nfx (changes requested):

common/azure_auth.go
mehdikit-db: 1 - This method is being used in `common/azure_cli_auth.go`.

2 - Good point, Fixed it.

nfx (approved):

PR #1618, opened 2022-09-20
Fix dashboard acceptance test

pietern:
DBSQL resources are organized in a workspace structure as of recently. As such it is no longer possible to use the same name for different assets. This acceptance test used to use the same name for both the dashboard and the query and started failing. The solution here is to use a different suffix for the different resources.


alexott (approved):
lgtm

nfx (approved):

PR #1620, opened 2022-09-21
Fix unity catalog docs to use correct metastore

npvinhphat:
aws_iam_policy.unity_metastore is referencing
aws_s3_bucket.unity_metastore.arn, which is non-existent. Fix to use aws_s3_bucket.metastore.arn instead. This also matches with the online tutorial.

https://docs.databricks.com/data-governance/unity-catalog/automate.html


nfx (approved):
👍

Issue #1615, opened 2022-09-20
[ISSUE] Issue with `databricks_sql_visualization` resource - changes to `option` from `inline` to `file()` strips permissions from referenced `sql_query` resource


crankswagon:
when attempting to move `databricks_sql_visualization.options` from inline `json` to a `file` reference:
- `terraform plan` succeeds
- `terraform apply` fails to apply `databricks_sql_visualization` but also strips `permissions` from `databricks_sql_query` and breaks terraform state (subsequent `terraform plan` is not able to accss the referenced query)

### Configuration
note that this bug was seen when moving from `current state` to `desired state` (both configurations are shown below)

`current_state`
```hcl
### current_state
## creates a query for every .sql file in databricks/jobs/redacted/analytics/dashboards
resource "databricks_sql_query" "redacted_analytics_queries" {
  for_each       = fileset("${path.module}/jobs/redacted/analytics/dashboards", "*.sql")
  data_source_id = data.terraform_remote_state.dataeng-deltalake-platform.outputs.redacted-analytics-warehouse-data-source-id
  name           = split(".", "${each.value}")[0]
  query          = file("${path.module}/jobs/redacted/analytics/dashboards/${each.value}")
  run_as_role    = "viewer"

  parameter {
    name  = "Vertical"
    title = "Vertical"
    enum {
      options = ["a", "b", "c"]
      values   = ["a"]

      multiple {
        prefix    = "\""
        suffix    = "\""
        separator = ","
      }
    }
  }

}

resource "databricks_permissions" "redacted_analytics_queries-permissions" {
  for_each     = databricks_sql_query.redacted_analytics_queries
  sql_query_id = each.value.id

  access_control {
    group_name       = data.terraform_remote_state.dataeng-deltalake-platform.outputs.dataeng-group
    permission_level = "CAN_RUN"
  }
}

resource "databricks_sql_visualization" "test_pie" {
  query_id    = databricks_sql_query.redacted_analytics_queries["rpt_test.sql"].id
  type        = "CHART"
  name        = "Test Pie"
  description = "Derp"

  # The options encoded in this field are passed verbatim to the SQLA API.

  # in sql.queries.create_new_visualization context, open the network tab in your browser (F12) and look for `post` calls with `visualizations` file
  # then copy out the `options` object from request payload and chuck it in here verbatim
  options = jsonencode(
    {
      "alignYAxesAtZero" : false,
      "coefficient" : 1,
[...]

PR #1644, opened 2022-09-29
prevent stripping of `CAN_MANAGE` permission from caller of `sql_query` resource

crankswagon:
relates to https://github.com/databricks/terraform-provider-databricks/issues/1615

this PR uses the fix from https://github.com/databricks/terraform-provider-databricks/issues/1504 to append `CAN_MANAGE` perms to the caller explicitly when modifying permissions on `databricks sql queries`

### DIFF
--- a/permissions/resource_permissions_test.go
+++ b/permissions/resource_permissions_test.go
@@ -1401,7 +1401,7 @@ func TestResourcePermissionsCreate_RepoPath(t *testing.T) {
 	assert.Equal(t, "CAN_READ", firstElem["permission_level"])
 }
 
-//when caller does not specify CAN_MANAGE permission during create, it should be explictly added
+// when caller does not specify CAN_MANAGE permission during create, it should be explictly added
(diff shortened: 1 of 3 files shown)


permissions/resource_permissions_test.go
```diff
@@ -1400,3 +1400,125 @@ func TestResourcePermissionsCreate_RepoPath(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		InstanceState: map[string]string{
+			"sql_query_id": "id111",
+		},
+		HCL: `
+		"sql_query_id" = "id111",
+		"access_control" = {
```
nfx, line 1509: nit: please format this with spaces

nfx (changes requested):
great PR, please add documentation to make it explicit for users

permissions/resource_permissions_test.go
crankswagon: thank you 👍  ive linted it into `blocks` with the new commit as well

nfx (approved):

Issue #1693, opened 2022-10-25
[ISSUE] Issue bucket policy for root bucket exposes confused deputy problem


edmondo1984:
A recent email from Databricks has been sent to improve the security of DBFS Root S3 bucket and KMS key(s) for E2 workspaces and help address the confused deputy problem.

Probably the fix is in a modification of the iam policy statement https://github.com/databricks/terraform-provider-databricks/blob/master/aws/data_aws_bucket_policy.go#L37

PR #1694, opened 2022-10-25
Restricting access to S3 bucket by custom tag on the IAM identity according to security email in `databricks_aws_bucket_policy` data resource

edmondo1984:
This pull request add a restriction on the root bucket s3 policy, using the PrincipalTag of the request. An equivalent change should be implemented by customers who use SS3-KMS on their own IAM policies that are not managed by the Terraform plugin

Fixes https://github.com/databricks/terraform-provider-databricks/issues/1693


aws/data_aws_bucket_policy.go
```diff
@@ -36,6 +36,11 @@ func DataAwsBucketPolicy() *schema.Resource {
 						Principal: map[string]string{
 							"AWS": fmt.Sprintf("arn:aws:iam::%s:root", d.Get("databricks_account_id").(string)),
 						},
+						Condition: map[string]map[string]string{
```
nfx, line 39: @edmondo1984 please make this field optional.

aws/data_aws_bucket_policy.go
```diff
@@ -36,6 +36,11 @@ func DataAwsBucketPolicy() *schema.Resource {
...
+							"StringEquals": {
+								"aws:PrincipalTag/DatabricksAccountId": d.Get("databricks_account_id").(string),
```
nfx, line 41: > Get your databricks-account-id. Go to the [account console](https://protect-us.mimecast.com/s/GK-NCOY0xwuNywrCER2IH?domain=docs.databricks.com) and click the in the bottom left corner. Under Account ID, select and copy the ID.

documentation meant the `E2 account id`, but in this data resource the `databricks_account_id` is AWS Account ID of Databricks

nfx (changes requested):

edmondo1984:
> @edmondo1984 could you also add documentation? :)

Sure, is the only thing that's missing?

nfx:
@edmondo1984 yes, otherwise how users will discover this change? :)

nfx (approved):

Issue #1688, opened 2022-10-24
Issue with exporter emitting`databricks_user` resource - unable to handle multidomain


rohitnijhawan:
### Expected Behavior
A user `databricksu@acmecorp.com` and another user `databricksu@acmecorp-biz.com` inside a workspace should create two different user blocks

- `resource "databricks_user" "databricksu_acmecorp_com" {}` 
- `resource "databricks_user" "databricksu_acmecorp_biz_com" {}` 

### Actual Behavior
Instead we get two identical resource blocks

- `resource "databricks_user" "databricksu" {}` 
- `resource "databricks_user" "databricksu" {}` 

That are semantically distinguishable to terraform

### Steps to Reproduce
Add 2 or more users with the same username prior to the `@domain` expression
Then run terraform export and observe the users.tf file

### Terraform and provider versions
Any terraform version
Provider version up to and including 1.6.1

PR #1697, opened 2022-10-26
Handling databricks_user block to include multiple domains

rohitkn:
Resolves issue #1688
Issue with exporter emitting `databricks_user` resource

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -648,8 +648,10 @@ var resourcesMap map[string]importable = map[string]importable{
 		Name: func(d *schema.ResourceData) string {
 			// TODO: if I have 2 users from different domains: test@domain1.com & test@domain2.com - then I'll generate the same name
 			// use another algorithm for name generation, like, just replace non-word characters with '_'
-			s := strings.Split(d.Get("user_name").(string), "@")
-			return s[0]
+			//s := strings.Split(d.Get("user_name").(string), "@")
+			//return s[0]
+			s := d.Get("user_name").(string)
+			return regexp.MustCompile(`[^a-zA-Z0-9]+`).ReplaceAllString(s, "_")
 		},
 		Search: func(ic *importContext, r *resource) error {
 			u, err := ic.findUserByName(r.Value)


exporter/importables.go
```diff
@@ -648,8 +648,10 @@ var resourcesMap map[string]importable = map[string]importable{
 		Name: func(d *schema.ResourceData) string {
 			// TODO: if I have 2 users from different domains: test@domain1.com & test@domain2.com - then I'll generate the same name
 			// use another algorithm for name generation, like, just replace non-word characters with '_'
-			s := strings.Split(d.Get("user_name").(string), "@")
-			return s[0]
+			//s := strings.Split(d.Get("user_name").(string), "@")
+			//return s[0]
+			s := d.Get("user_name").(string)
+			return regexp.MustCompile(`[^a-zA-Z0-9]+`).ReplaceAllString(s, "_")
```
alexott, line 654: for performance, it's better to make it a variable that is declared with `var` at the beginning of importables.
Also remove commented out code.
Although, the best case would be to rely on the name fixes declared here: https://github.com/databricks/terraform-provider-databricks/blob/master/exporter/context.go#L87, although we may need to fix this problematic line: https://github.com/databricks/terraform-provider-databricks/blob/master/exporter/context.go#L91 - it will fix other issues as well

alexott (changes requested):
we can simplify the stuff - remove the comments, and rely on `NameFixes`

exporter/importables.go
rohitnijhawan: NameFixes are applied to **every** name.
That can have other effects across all names. 
Despite the NameFixes array of substitutions we split the databricks_user hcl name then return the user portion.

alexott: What I meant is just return `d.Get("user_name").(string)` and allow `NameFixes` to handle it...

exporter/importables.go
```diff
@@ -648,8 +648,10 @@ var resourcesMap map[string]importable = map[string]importable{
 		Name: func(d *schema.ResourceData) string {
 			// TODO: if I have 2 users from different domains: test@domain1.com & test@domain2.com - then I'll generate the same name
 			// use another algorithm for name generation, like, just replace non-word characters with '_'
-			s := strings.Split(d.Get("user_name").(string), "@")
-			return s[0]
+			//s := strings.Split(d.Get("user_name").(string), "@")
+			//return s[0]
+			s := d.Get("user_name").(string)
```
nfx, line 653: please add the command-line flag to include email domain in the naming callback, because it changes existing  behavior and thus, must be an opt-in feature.

https://github.com/databricks/terraform-provider-databricks/blob/master/exporter/command.go#L89-L95

to have that flag accessible, add import context as the first argument to a callback interface.

nfx (changes requested):

exporter/importables.go
nfx: @rohitnijhawan @rohitkn okay, `NameFixes` is a less invasive approach, but still make the full-email-resource-name as opt-in feature, as you're changing the behavior.

Issue #1654, opened 2022-10-05
[ISSUE] TF exporter creating`databricks_group` resource, not data block for admins and users


rohitnijhawan:
### Configuration [Generated from the exporter]
```hcl
resource "databricks_group" "admins" {
  display_name               = "admins"
  ...
}

resource "databricks_group" "users" {
  display_name               = "users"
  ...
}
```

### Expected Behavior
* Databricks TF Provider should export a data {} block reference to groups with names "admins" and "users"
* Using a resource {} block this way is against the validation Rules
* Databricks TF Provider should have a display_name attribute off the data so it can be referenced in other blocks.

### Actual Behavior
Error:
**Terraform databricks_group resource does not support group names “admins” and “users”**

For Admins
> Error: expected display_name to not be any of [users admins], got admins
>   with databricks_group.admins_4425219924063031,
>   on [groups.tf](http://groups.tf/) line 3, in resource "databricks_group" "admins_4425219924063031":
>    3:   display_name               = "admins"

For users
>Error: expected display_name to not be any of [users admins], got users
>   with databricks_group.users_54623195191167,
>   on [groups.tf](http://groups.tf/) line 532, in resource "databricks_group" "users_54623195191167":
>  532:   display_name          = "users"

### Steps to Reproduce
* Run the Terraform Databricks exporter on a workspace with user principals in users and admins groups
* Run `terraform validate`

### Terraform and provider versions
TF > 0.15
DB Provider > 1.1.0

PR #1698, opened 2022-10-26
Generating HCL data block for users and admins, omit entitlements

rohitnijhawan:
Fix for issue #1654 
Fix for Issue #1688 - databricks_user hcl name includes cleansed domain

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -527,9 +527,15 @@ var resourcesMap map[string]importable = map[string]importable{
 					log.Printf("[INFO] Group %s doesn't match %s filter", g.DisplayName, ic.match)
 					continue
 				}
+				hclblockMode := "resource"
+				genDataBlock := map[string]bool{"admins": true, "users": true}
+				if genDataBlock[g.DisplayName] {
+					hclblockMode = "data"
+				}
 				ic.Emit(&resource{
 					Resource: "databricks_group",
 					ID:       g.ID,
+					Mode:     hclblockMode,
 				})
 			}
 			return nil
@@ -547,10 +553,12 @@ var resourcesMap map[string]importable = map[string]importable{
 			return nil
 		},
 		Import: func(ic *importContext, r *resource) error {
-			if r.Name == "admins" || r.Name == "users" {
+			if r.Mode == "data" {
 				// admins & users are to be imported through "data block"
-				// TODO: it doesn't work, fix it...
-				r.Mode = "data"
+				r.Data.Set("workspace_access", false)
+				r.Data.Set("databricks_sql_access", false)
+				r.Data.Set("allow_instance_pool_create", false)
+				r.Data.Set("allow_cluster_create", false)
 				r.Data.State().Set(&terraform.InstanceState{
 					ID: r.ID,
 					Attributes: map[string]string{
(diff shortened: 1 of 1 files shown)


exporter/importables.go
```diff
@@ -671,10 +677,6 @@ var resourcesMap map[string]importable = map[string]importable{
 					log.Printf("Skipping non-direct group %s/%s for user %s", g.Value, g.Display, username)
 					continue
 				}
-				ic.Emit(&resource{
-					Resource: "databricks_group",
-					ID:       g.Value,
-				})
```
alexott, line 677: This should be restored back, as we may have a situation when user is discovered from thing like email notification, etc., but we also request to extract groups - without this code groups won't be exported

exporter/importables.go
```diff
@@ -527,9 +527,15 @@ var resourcesMap map[string]importable = map[string]importable{
 					log.Printf("[INFO] Group %s doesn't match %s filter", g.DisplayName, ic.match)
 					continue
 				}
+				hclblockMode := "resource"
+				genDataBlock := map[string]bool{"admins": true, "users": true}
+				if genDataBlock[g.DisplayName] {
+					hclblockMode = "data"
+				}
```
alexott, line 534: This splits logic into two pieces that a harder to handle. It's better to do everything in the `Import` function (see comment there)

exporter/importables.go
```diff
@@ -547,10 +553,12 @@ var resourcesMap map[string]importable = map[string]importable{
 			return nil
 		},
 		Import: func(ic *importContext, r *resource) error {
-			if r.Name == "admins" || r.Name == "users" {
+			if r.Mode == "data" {
```
alexott, line 556: it's easier to handle with something like this (or using map as in your code, but this map needs to be initialized as a constant to avoid creating it each time):

```
groupName := r.Data.Get("display_name").(string)
if groupName == "admins" || groupName == "users" {
```

exporter/importables.go
```diff
@@ -527,9 +527,15 @@ var resourcesMap map[string]importable = map[string]importable{
 					log.Printf("[INFO] Group %s doesn't match %s filter", g.DisplayName, ic.match)
 					continue
 				}
+				hclblockMode := "resource"
+				genDataBlock := map[string]bool{"admins": true, "users": true}
```
alexott, line 531: technically, `map[string]interface{}` is better because it could be used as a `set`, so we can use it as:

```
if _, exists := genDataBlock[g.DisplayName]; exists {
...
}
```

alexott (changes requested):

exporter/importables.go
rohitnijhawan: Yes. The logic I feel - should be split
List() is for discovery. You mark all your special groups there. Right now it's admins and users.. in the future it can be others.
By the time you get to import you're not creating exceptions - just handling data blocks differently. 
Perhaps in the future we should create config files with maps in them so that instead of rolling out code changes we roll out new config files

exporter/importables.go
rohitnijhawan: Can you please elaborate how a new user can be discovered that isn't already in the path to be discovered through a group membership? 
There is only 2 references to databricks_user emission in importables.go and that's in the databricks_group Import implementation and one in the databricks_permissions' import

exporter/importables.go
alexott: The problem is that emit of the `databricks_group` not necessary happens via `List` - if you check the code, it happens from inside group's Import, from `databricks_user`, from `databricks_permission`, and maybe some other places.  

This is necessary because emit could be triggered by another object, for example, if you do a selective export of a single cluster, then you emit a cluster, Import of a cluster emits a permission, and Import of permission emits group that is then handled in Import

exporter/importables.go
alexott: user will be discovered via List only if you have `-listing users`, otherwise users will be discovered only by transitive dependencies

alexott (changes requested):

nfx:
this is a duplicate of #1722 and #1721

PR #1721, opened 2022-11-02
Export `admin` and `users` groups as `databricks_group` data, not resource

rohitnijhawan:
Fix for issue #1654

TF exporter creatingdatabricks_group resource, not data block for admins and users

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -543,10 +543,14 @@ var resourcesMap map[string]importable = map[string]importable{
 			return nil
 		},
 		Import: func(ic *importContext, r *resource) error {
-			if r.Name == "admins" || r.Name == "users" {
+			groupName := r.Data.Get("display_name").(string)
+			if groupName == "admins" || groupName == "users" {
 				// admins & users are to be imported through "data block"
-				// TODO: it doesn't work, fix it...
 				r.Mode = "data"
+				r.Data.Set("workspace_access", false)
+				r.Data.Set("databricks_sql_access", false)
+				r.Data.Set("allow_instance_pool_create", false)
+				r.Data.Set("allow_cluster_create", false)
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm, although we need to fix the broken test

alexott:
Replace `r` in line 118 of `importables_test.go` with:

```
	d := scim.ResourceGroup().TestResourceData()
	d.Set("display_name", "foo")
	r := &resource{
		Value:     "foo",
		Attribute: "display_name",
		Data:      d,
	}
```

nfx (changes requested):
looks good. please make sure that build is passing

exporter/importables.go
```diff
@@ -560,18 +560,24 @@ var resourcesMap map[string]importable = map[string]importable{
...
-				r.Data.State().Set(&terraform.InstanceState{
-					ID: r.ID,
-					Attributes: map[string]string{
-						"display_name": r.Name,
-					},
-				})
-			} else if r.Data != nil {
-				r.Data.Set("force", true)
+			if r.Data != nil {
+				groupName := r.Data.Get("display_name").(string)
```
nfx, line 564: could you please add unit tests for this change, so that the following commits won't break it?

nfx (changes requested):

exporter/importables.go
rohitnijhawan: I can do that @nfx 

What should the unit test be looking for? An r.Data that **is nil**?
Before we decided to fix this bug, that was the original if condition.

Thanks

alexott:
Really, the correct fix was not adding `if r.Data != nil`, but just fixing a test as I've shown here: https://github.com/databricks/terraform-provider-databricks/pull/1721#issuecomment-1301742079

nfx (approved):

Issue #1720, opened 2022-11-02
Issue with `databricks_group_member` resource


rohitnijhawan:
This is observed from the exporter

### **Part of Generated** Configuration
```hcl
resource "databricks_group" "grp_prod_data_read_5498777571163099" {
  force        = true
  display_name = "grp-prod-data-read"
}
resource "databricks_group_member" "cluster_managers_5498777571163099_grp_prod_data_read" {
  member_id = "5498777571163099"
  group_id  = databricks_group.cluster_managers_7696875338796019.id
}
resource "databricks_group_member" "grp_prod_data_read_575982911243271_trade_capture_support" {
  member_id = "575982911243271"
  group_id  = databricks_group.grp_prod_data_read_5498777571163099.id
}
resource "databricks_group" "trade_capture_support_575982911243271" {
  force        = true
  display_name = "Trade Capture Support"
}
resource "databricks_group_member" "trade_capture_support_575982911243271_databricks_misterx" {
  member_id = databricks_user.databricks_misterx.id
  group_id  = databricks_group.trade_capture_support_575982911243271.id
}

```

### Expected Behavior
What is expected is this

1. The **member_id** of a nested group - in this case _grp-prod-data-read_ and _Trade Capture Support_ in the resource block: **databricks_group_member** should **not** be a literal. It should be (for example) `databricks_group.trade_capture_support_575982911243271.id`. This is done correctly when the member is a databricks_user object.
2. When there is a `databricks_group_member` block for **a databricks_user** the name follows this pattern `<parent_group_name>_parentGroupID_username` and this is correct. Thus, for a **nested group** (we'll call it _childGroup_) `databricks_group_member` resource should follow this naming pattern: `<parent_group_name>_parentGroupID_childGroupName`

### Actual Behavior
1. The **member_id** of a nested group - in this case _grp-prod-data-read_ and _Trade Capture Support_ in the resource block: **databricks_group_member** is a literal like "575982911243271"
2. However for a nested group (we'll call it _childGroup_) `databricks_group_member` naming follows this pattern: `<parent_group_name>_childGroupID_childGroupName`

### Steps to Reproduce
- Create group "bricksters"
- Create group "bricksters_nested" inside bricksters
- Add users to bricksters and "bricksters_nested"
- Run the exporter with -services 'groups,users' and -listing 'groups,users'

### Terraform and provider versions
TF version 1.1 or greater
Provider version 1.6.2 or lower

nfx:
[...]

PR #1723, opened 2022-11-02
Fixed `member_id` reference of nested groups and `databricks_group_member` for exporter

rohitnijhawan:
Fix for issue #1720 
Problem with old code shown below is that, it checks **only one** class of resources before quitting. The fixed code checks all references from Depends:
```
if traversal != nil {
   break
}
```


alexott (approved):
lgtm

nfx (approved):
lgtm

nfx:
The only thing remaining in this PR is to rebase/update with master

Issue #1725, opened 2022-11-04
Issue with `databricks_workspace_conf` resource


rohitnijhawan:
There is no path to generating `databricks_workspace_conf` resource block unless the IP Access List setting is enabled with at least one item in the list.

### Expected Behavior
Generate this block with only the custom config keys that are **non-null**

```
resource "databricks_workspace_conf" "global_workspace_conf" {
  custom_config = { 
    enableTokensConfig = "true"
  }
}
```

### Actual Behavior
databricks_workspace_conf resource block is not generated unless ip access list is enabled **and** default values are cause errors on `terraform apply` from the exporter as a result of null values from the workspace API

### Steps to Reproduce
Run the Terraform exporter with -services "workspace" and -listing "workspace"

### Terraform and provider versions
Terraform - any version
Databricks provider - Version 1.6.2 and lower

PR #1727, opened 2022-11-04
First commit to fix databricks_workspace_conf generation 

rohitkn:
Fix for issue #1725 
Generate workspace conf with workspace in -services

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -966,16 +966,24 @@ var resourcesMap map[string]importable = map[string]importable{
 		Name: func(d *schema.ResourceData) string {
 			return globalWorkspaceConfName
 		},
+		List: func(ic *importContext) error {
+			ic.Emit(&resource{
+				Resource: "databricks_workspace_conf",
+				ID:       globalWorkspaceConfName,
+			})
+			return nil
+		},
 		Import: func(ic *importContext, r *resource) error {
 			wsConfAPI := workspace.NewWorkspaceConfAPI(ic.Context, ic.Client)
-			keys := map[string]any{
-				"enableIpAccessLists":  false,
-				"maxTokenLifetimeDays": 0,
-				"enableTokensConfig":   false,
-			}
+			keys := ic.workspaceConfKeys
 			err := wsConfAPI.Read(&keys)
 			if err != nil {
 				return err
+			}
+			for k, v := range keys {
+				if v == nil {
+					delete(keys, k)
+				}
 			}
 			r.Data.Set("custom_config", keys)
 			return nil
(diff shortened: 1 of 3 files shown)


exporter/importables.go
```diff
@@ -953,17 +953,36 @@ var resourcesMap map[string]importable = map[string]importable{
...
+				"maxTokenLifetimeDays":                             0,
+				"enableTokensConfig":                               false,
+				"storeInteractiveNotebookResultsInCustomerAccount": false,
+				"maxUserInactiveDays":                              0,
 			}
 			err := wsConfAPI.Read(&keys)
 			if err != nil {
 				return err
 			}
+			for k, _ := range keys {
```
alexott, line 981: it could be simplified to 
```
for k, v := range keys {
 if v == nil {
	delete(keys, k)
 }
}
```

exporter/importables.go
```diff
@@ -953,17 +953,36 @@ var resourcesMap map[string]importable = map[string]importable{
...
 		},
+		List: func(ic *importContext) error {
+			ic.Emit(&resource{
+				Resource: "databricks_workspace_conf",
+				ID:       globalWorkspaceConfName,
+				Data: ic.Resources["databricks_workspace_conf"].Data(
+					&terraform.InstanceState{
+						ID:         globalWorkspaceConfName,
+						Attributes: map[string]string{},
+					}),
```
alexott, line 964: It's not strictly required to explicitly create empty `r.Data` - it will be created automatically: https://github.com/databricks/terraform-provider-databricks/blob/master/exporter/context.go#L441. You may need to provide `InstanceState` only when you're initializing it

alexott (changes requested):

exporter/importables.go
nfx: won't ti be a concurrent modification? :)

alexott: According to this SO, no: https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop, especially this answer: https://stackoverflow.com/a/23231539/18627

exporter/importables.go
rohitnijhawan: @alexott In the implementation of the List: func
I've removed the unnecessary Data: block and pushed the commit after testing it.

alexott (approved):
lgtm

nfx (changes requested):
please make sure to run `make test` locally before pushing

alexott:
To fix test you need to update URL here: https://github.com/databricks/terraform-provider-databricks/blob/master/exporter/exporter_test.go#L1331, and update response as well: https://github.com/databricks/terraform-provider-databricks/blob/master/exporter/exporter_test.go#L1332

nfx (approved):

PR #1730, opened 2022-11-04
Escape database name in sql_permissions resource

### LINKED ISSUE #1729: [ISSUE] Issue with `databricks_sql_permissions` resource - Filters out backticks
### Configuration
```hcl
```

### Expected Behavior

Able to use `<hyphenated-database>`

### Actual Behavior
Fails on Operation not allowed: SHOW GRANT(line 1, pos 0) with hyphenated database. When trying to pass a database with a hyphen and without backticks, show grant fails. When adding backticks, it looks like it strips them out from SHOW GRANT command which also causes it to fail.

### Steps to Reproduce
 Create databricks_sql_permissions resource with a hyphenated database.

0x91:
Fixes #1729

We have an issue using the sql_permissions with database names that require escaping. Putting backticks in the database name seems to get filtered out somewhere in Terraform.

This change explicitly escapes every database name to avoid that issue, similar to the escaping we do for Tables and Views.

I would like to add a test case for this but I'll need some pointers on how the CommandMock works.

### DIFF
--- a/access/resource_sql_permissions.go
+++ b/access/resource_sql_permissions.go
@@ -50,7 +50,7 @@ func (ta *SqlPermissions) typeAndKey() (string, string) {
 		return "VIEW", fmt.Sprintf("`%s`.`%s`", ta.actualDatabase(), ta.View)
 	}
 	if ta.Database != "" {
-		return "DATABASE", ta.Database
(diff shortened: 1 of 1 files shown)


nfx (approved):

Lironal:
Hey @nfx - how can I see when/which provider version is this going to be out with?

Issue #1737, opened 2022-11-09
[ISSUE] Issue with `databricks_library` resource - Terraform workspace resource creation failed  at library installation if one of the clusters is missing


yugireddy:
### Configuration
```hcl
# Copy-paste your Terraform configuration here
```

### Expected Behavior
Terraform Databricks provider should have moved forward with the resource creation for the available clusters when one of the clusters doesn't exist while installing the libraries, instead the resource creation process is halted and no deployment is done.

### Actual Behavior
Resource creation process is halted and no deployment is done even when other clusters exist while libraries installation process is in progress.

### Steps to Reproduce

1.6.2

### Debug Output

2022-11-08T22:15:52.212-0600 [DEBUG] provider.terraform-provider-databricks: 400 Bad Request {
  "error_code": "INVALID_PARAMETER_VALUE",
  "message": "Cluster 1006-208273-a3js222ew does not exist"
}: timestamp=2022-11-08T22:15:52.212-0600
2022-11-08T22:15:52.213-0600 [WARN]  provider.terraform-provider-databricks: /api/2.0/clusters/get:400 - Cluster 1006-208273-a3js222ew does not exist https://docs.databricks.com/dev-tools/api/latest/clusters.html#get: timestamp=2022-11-08T22:15:52.213-0600

### Important Factoids

The above error doesn't arise if the cluster is Active Running or in Stopped state, it'll only arise if the cluster doesn't exist in the workspace

nfx:
@yugireddy it'll be great if you provide the error example for `/libraries/get-status` path as well. filing internal issue to fix it in Go SDK as well

yugireddy:
@nfx Here's the error example:

2022-11-08T22:15:52.231-0600 [DEBUG] provider.terraform-provider-databricks: GET **/api/2.0/libraries/cluster-status**: timestamp=2022-11-08T22:15:52.231-0600
2022-11-08T22:15:52.607-0600 [DEBUG] provider.terraform-provider-databricks: 400 Bad Request {
  "error_code": "INVALID_PARAMETER_VALUE",
  "message": "Cluster 1102-152045-85l25cgn does not exist"
}: timestamp=2022-11-08T22:15:52.607-0600
[...]

PR #1745, opened 2022-11-09
Fixed auto-purged cluster behaviour for `databricks_library`

yugireddy:
Changes to libraries module to address the Issue #1737

The changes enable:
1. Move forward with the resource creation on available clusters when the library installation error occurs on the non-existent cluster, previously the flow used to stop and failed with `Error: cannot read library: Cluster 1234-123456-u2ab23hh does not exist
`

### DIFF
--- a/libraries/libraries_api.go
+++ b/libraries/libraries_api.go
@@ -83,7 +83,7 @@ func (a LibrariesAPI) WaitForLibrariesInstalled(wait Wait) (result *ClusterLibra
 			return resource.RetryableError(err)
 		}
 		if err != nil {
-			return resource.NonRetryableError(err)
+			return resource.NonRetryableError(wrapClusterStatusAPIError(err, wait.ClusterID))
 		}
 		if !wait.IsRunning {
 			log.Printf("[INFO] Cluster %s is currently not running, so just returning list of %d libraries",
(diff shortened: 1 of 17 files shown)


libraries/libraries_api.go
```diff
@@ -83,7 +83,7 @@ func (a LibrariesAPI) WaitForLibrariesInstalled(wait Wait) (result *ClusterLibra
 			return resource.RetryableError(err)
 		}
 		if err != nil {
-			return resource.NonRetryableError(err)
+			return resource.NonRetryableError(wrapClusterStatusAPIError(err, wait.ClusterID))
```
nfx, line 86: okay, so please simplify the logic here between lines 81 and 87. the new logic should be something like: if we get error from `a.ClusterStatus`, return it as non-retriable.

you probably need to directly paste only these lines:

```go
        apiErr, ok := err.(common.APIError)
	if !ok {
		return resource.NonRetryableError(err)
	}
	// fix non-compliant error code
	if apiErr.StatusCode != 404 && strings.Contains(apiErr.Message, 
                fmt.Sprintf("Cluster %s does not exist", id)) {
		apiErr.StatusCode = 404
	}
        return resource.NonRetryableError(apiErr)
```

then amount of unit tests will be smaller ;)

nfx (changes requested):

libraries/libraries_api.go
yugireddy: @nfx Updated per your suggestion, please review and confirm if we are good.

libraries/libraries_api_test.go
```diff
@@ -114,6 +123,12 @@ func TestWaitForLibrariesInstalled(t *testing.T) {
 			"failed-wheel", 50 * time.Millisecond, true, true,
 		})
 		assert.NoError(t, err, "library should have been uninstalled and work proceeded")
+
+		// Cluster not available or doesn't exist
+		_, err = libs.WaitForLibrariesInstalled(Wait{
+			"1005-abcd", 50 * time.Millisecond, false, false,
+		})
+		assert.EqualError(t, err, "Cluster 1005-abcd does not exist")
```
nfx, line 131: can you also check for 404 StatusCode of APIError here? this is the condition that triggers resource deletion.

nfx (changes requested):

nfx (approved):

PR #1788, opened 2022-11-21
Improve export of DBSQL objects

alexott:
This includes:

* Fixing problems with listing of objects
* Correct export of queries with date/time range parameters
* Introduced separate categories for queries, dashboards & SQL endpoints to allow granular export

### DIFF
--- a/sql/api/query.go
+++ b/sql/api/query.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"strings"
 )
 
 // Query ...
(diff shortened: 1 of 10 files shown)


sql/api/query.go
```diff
@@ -398,17 +400,45 @@ func (p *QueryParameterDateTimeSec) UnmarshalJSON(b []byte) error {
...
+		p.Value = p.StringValue
+	}
+}
+
+func (p *QueryParameterRangeBase) decodeQueryParameter() {
+	if v, ok := p.Value.(map[string]any); ok {
+		p.RangeValue = &DateTimeRange{Start: v["start"].(string), End: v["end"].(string)}
+	} else {
+		p.StringValue = fmt.Sprintf("%v", p.Value)
+	}
```
pietern, line 429: You could set value to nil from here instead of right after this call.

pietern (commented):
LGTM on the changes under `sql/`.

While listing will be much much much easier with the SDK, the type of polymorphism you need to do to make these query parameters work unfortunately remains.

sql/api/query.go
alexott: oh, yes - it was an artifact where I did first it as a separate function, not function on the type

nfx (approved):
LGTM

alexott:
I need to run the full scale test again, and then mark it as non-draft

alexott:
@nfx @pietern Ok, it survives export from DF :-)

nfx:
@alexott could you rebase?

alexott:
@nfx rebased

PR #1791, opened 2022-11-22
Added `home` & `repos` attributes to `databricks_user` and `databricks_service_principal`

alexott:
They will be used in Exporter to simplify references in paths, etc.


pietern (commented):
Looks reasonable. Re: naming, in isolation `user.repos` looks ambiguous. I think something like `repos_[path,location,...]` would be more descriptive. Home is less ambiguous but for symmetry you could argue to use the same suffix.

alexott:
@pietern that was the names that were originally selected for corresponding data sources, so I've used it for compatibility

nfx (approved):
lgtm

PR #1792, opened 2022-11-23
add `databricks_cluster_policy` data source

### LINKED ISSUE #1785: [ISSUE] Issue with `databricks_cluster_policy` resource. Data block needed
When a user authors a terraform descriptor file (.tf file) to create a policy with one of these names:

1. Personal Compute
2. Power User Compute
3. Shared Compute
4. Job Compute

Errors are thrown

### Expected Behavior
For terraform resource like

```
resource "databricks_cluster_policy" {
  name = "Job Compute"
  definition =.....
}
```

The policy should either be created or ignored.

### Actual Behavior
The back end of the API throws an error informing the user that, "This policy already exists"

### Steps to Reproduce
Attempt to create a databricks cluster policy named Job Compute (or one of those listed above) using Terraform

### Exporter Notes
This isn't an exporter related ticket but the exporter will fetch all policies and an attempt to re-create them in a target workspace will fail.

These policies should be handled in a special way - ie. by using a a data block - and this ticket is about implementing that block.

alexott:
initial approach to fix #1785

### DIFF
--- a/docs/data-sources/cluster_policy.md
+++ b/docs/data-sources/cluster_policy.md
@@ -1,0 +1,37 @@
+---
+subcategory: "Compute"
+---
+
+# databricks_cluster_policy Data Source
+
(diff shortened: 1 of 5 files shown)


nkvuong (changes requested):

alexott:
Thank you @nkvuong - copy paste errors :-)

nkvuong (approved):
lgtm

PR #1798, opened 2022-11-25
Fix missing attributes from `databricks_job` data source

### LINKED ISSUE #1797: [ISSUE] Issue with databricks_job Data Source
### Configuration
```hcl
terraform {
    required_providers {
      databricks = {
        version = "1.6.5"
      }
    }
}

data "databricks_jobs" "this" {
  job_name = "Vivian-auto-test"
}

output "raw_out" {
  value     = data.databricks_job.this
}
```

### Expected Behavior

The output has to fetch all the job configurations. The field under "job_settings" is not getting populated.
### Actual Behavior

The output of databricks_job does not populate data in the "job_settings" list in its output JSON.
### Steps to Reproduce

You can easily reproduce this using the above hcl code that I have attached.

Terraform v1.3.5
on darwin_amd64
+ provider registry.terraform.io/databricks/databricks v1.6.5
### Debug Output

Debug logs: https://gist.github.com/vivianwilfred/bbbdfd4277fde352dbc3d3a4f0ecc467

nkvuong:
This happens because we did not set expandTasks to the API when listing the jobs. For search by id, list without expandTasks and then re-search when job name is found 

This closes #1797

### DIFF
--- a/jobs/data_job.go
+++ b/jobs/data_job.go
@@ -20,29 +20,17 @@ func DataSourceJob() *schema.Resource {
 		var list []Job
 		var err error
 		if data.Name != "" {
+			// if name is provided, need to list all jobs ny name
 			list, err = jobsAPI.ListByName(data.Name, true)
 		} else {
-			// if name is not provided, need to list all the jobs and search by id
-			// does not expand tasks to limit the results size
-			list, err = jobsAPI.List()
+			// otherwise, just read the job
+			var job Job
+			job, err = jobsAPI.Read(data.Id)
 			if err != nil {
 				return err
 			}
-			jobName := ""
-			for _, job := range list {
-				currentJob := job // De-referencing the temp variable used by the loop
-				currentJobId := currentJob.ID()
-				if currentJobId == data.Id {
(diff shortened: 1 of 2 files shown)


jobs/data_job.go
```diff
@@ -20,9 +20,29 @@ func DataSourceJob() *schema.Resource {
 		var list []Job
 		var err error
 		if data.Name != "" {
-			list, err = jobsAPI.ListByName(data.Name, false)
+			list, err = jobsAPI.ListByName(data.Name, true)
 		} else {
+			// if name is not provided, need to list all the jobs and search by id
+			// does not expand tasks to limit the results size
 			list, err = jobsAPI.List()
```
alexott, line 27: If we have a job ID, then why not simply call `Read` using that ID? It will be more efficient :-)

alexott (changes requested):

jobs/data_job.go
nkvuong: this is a lot simpler now

alexott (approved):
lgtm. Thank you for simplifying it

Issue #1754, opened 2022-11-11
[FEATURE] Allow data databricks_group on account


BostjanBozic:
### Use-cases
It would be great to be able to fetch user group information on Databricks account level.

Use case behind it is:
* enable SSO with SCIM enabled (e.g. using Okta)
* groups are synced via SCIM, which means groups are available on Databricks account level
* those groups would be assigned to workspace using `databricks_mws_permission_assignment` resource

### Attempted Solutions

I tried to use data source object `databricks_group` with provider configured pointing toward account, but this object requires `host` parameter in provider configuration, which indicates that it is working on workspace level.

Code snippet includes only related resources.
```
provider "databricks" {
  alias    = "mws"
  host     = "https://accounts.cloud.databricks.com"
  username = var.databricks_account.username
  password = var.databricks_account.password
}

provider "databricks" {
  alias = "ws"
  host  = databricks_mws_workspaces.test.workspace_url
  token = databricks_mws_workspaces.test.token[0].token_value
}

data "databricks_group" "test_group" {
  provider     = databricks.mws
  display_name = "test-group"
}

resource "databricks_mws_permission_assignment" "test_group" {
  provider     = databricks.mws
  workspace_id = databricks_mws_workspaces.test.workspace_id
  principal_id = data.databricks_group.test_group.id
  permissions  = ["USER"]
}
```

This would fail with following error, which makes me believe it is only available on workspace scope:
```
│ Error: Databricks API (/api/2.0/preview/scim/v2/Groups) requires you to set `host` property (or DATABRICKS_HOST env variable) to result of `databricks_mws_workspaces.this.workspace_url`. This error may happen if you're using provider in both normal and multiworkspace mode. Please refactor your code into different modules. Runnable example that we use for integration testing can be found in this repository at https://registry.terraform.io/providers/databricks/databricks/latest/docs/guides/aws-workspace
```

### Proposal

Add additional resource/data source, which would work on Databricks account level, similar to how `databricks_mws_permission_assignment` is used. Maybe call it `databricks_mws_group` or something similar.

### References

- https://github.com/databricks/terraform-provider-databricks/issues/1687

BostjanBozic:
It seems that this is actually a bug in error message output. What happened that once I added `account_id` parameter to provider, it did the trick. I will open separate bug report.

PR #1816, opened 2022-12-05
Improve `databricks_group` documentation

fuselessmatt:
I found the previous documentation very confusing and this documentation would have [prevent this issue](https://github.com/databricks/terraform-provider-databricks/issues/1754) that I also had.

In summary 
* The documentation now use the official terminology account groups and workspace-local groups. This is better than the imprecise "creating a group in the account" as the workspace is also in the account. 
* The instructions states that you need to specify the host to create account groups, but in you also need account_id.


fuselessmatt:
Once we can agree on the phrasing of the instructions, we should add it to https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/group_member which as the same problem

daniel-anova:
Please make it clearer that account_id is supposed to contain an actual value and where to get it from.

fuselessmatt:
Do we have any update on the review status of  this @PaulCornellDB ? An alternative solution would be to centralise the account vs workspace provider to a separate document somewhere, but that is a bigger change.

alexott (approved):
lgtm, thank you

@nfx fyi

PR #1845, opened 2022-12-13
Allow init scripts on ADLS in `databricks_cluster` and related resources

### LINKED ISSUE #1786: [FEATURE] New feature request: abfss:// cluster-scoped `init_scripts`
I read #1372 and it says that there is no support for cluster-scoped init_scripts in ADLS, but the web UI now gives me the option of adding an init_script in ADLS via abfss in Azure Databricks:

image

- If abfss for init script is now supported in Databricks it would be nice to have it also in the Terraform provider and to add an example to the documentation as there is for other cloud storages
- If it's still not technically possible to implement in terraform it would be nice to add this info in the documentation and make a note that abfss is not supported by terraform but you can do it in the web UI

alexott:
This allows to specify path to init script as `abfss://container@storage.dfs.core.windows.net/path`

this fixes #1786

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -191,6 +191,11 @@ type GcsStorageInfo struct {
 	Destination string `json:"destination,omitempty"`
 }
 
+// AbfssStorageInfo contains the struct for when storing files in ADLS
+type AbfssStorageInfo struct {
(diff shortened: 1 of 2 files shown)


nfx (approved):
lgtm

PR #1856, opened 2022-12-15
Use NVMe disk size for selection in the `databricks_node_type` data source

alexott:
Due not completely correct sorting, instead of real smallest node, other node type was returned as smallest. For example, on Azure it was `L8` series.

Also added the `local_disk_min_size` selector to allow to select nodes with local disks not smaller than specified.


alexott:
pr-manual check is running for all 3 clouds

alexott:
corresponding acceptance tests passed, failures were in not related tests

nfx (approved):

PR #1861, opened 2022-12-19
Purge home dir and repos on user delete #1557

### LINKED ISSUE #1557: [FEATURE] add `purge_homedir_on_delete` for `databricks_user` resource
### Use-cases
After creating temporary users on the workspace, we need to give them ability to experiment within their home folders. Once the experiment is done, we want to remove the user trace completely, including the `/Users/user@example.com` directory. Current user deletion does not remove user notebooks created by the user as part of standard Databricks REST API behavior

mberk06:
TODO
* Get flags = false to fail
* Add the same logic to service principles
* Test manually via TF (for fun)
* Get review

Fixes #1557


mberk06:
Hey @nfx QQ. I'm looking to remove home dir and repo on user delete. For instance, for repos, the intended functionality is to delete the repo if the `delete_repos` flag is `nil` or `true`. If it's `false`, don't delete the repo. 

I got the above tests to pass but I want to test `delete_repos = false`, however I don't get a test failure when changing the HCL in the fixture to `delete_repos = false`. Am I testing this correctly? `d.Get("Id")` exists, so I'm not testing empty `ResourceData`. 

Thanks!

qa/testing.go
```diff
@@ -91,7 +91,7 @@ type ResourceFixture struct {
 	New bool
 }
 
-// wrapper type for calling resource methords
+// wrapper type for calling resource methods
```
nfx, line 94: please send a separate PR for this, it's unrelated

scim/resource_user.go
```diff
@@ -44,6 +43,14 @@ func ResourceUser() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			}
+			m["delete_repos"] = &schema.Schema{
```
nfx, line 46: perhaps doing `force_delete_content` is a better setting name for both repos and notebooks.

scim/resource_user.go
```diff
@@ -93,7 +100,18 @@ func ResourceUser() *schema.Resource {
 			return NewUsersAPI(ctx, c).Update(d.Id(), u)
 		},
 		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			return NewUsersAPI(ctx, c).Delete(d.Id())
+			usersAPI := NewUsersAPI(ctx, c)
+
+			if d.Get("delete_repos").(bool) {
+				d.Set("cheese", "cheese")
```
nfx, line 106: what is this doing here?

scim/resource_user.go
```diff
@@ -93,7 +100,18 @@ func ResourceUser() *schema.Resource {
...
+				_ = usersAPI.DeleteRepos(d.Id())
```
nfx, line 107: 1. scim can be account-level and workspace-level. how do you plan to check for that here?
2. do not swallow the error on deletion and if these properties are set, return error back to user. please be very careful for misinterpretation of non-found resources on workspace/repos to delete user from terraform state, but not deleting it from actual workspace.

scim/resource_user_test.go
```diff
@@ -440,6 +442,62 @@ func TestResourceUserDelete_Error(t *testing.T) {
 	require.Error(t, err, err)
 }
 
+func TestResourceUserDeleteReposAndDirs(t *testing.T) {
```
nfx, line 445: please add separate test invariants:
* where deleting repos fail
* where repos to be deleted is not found (assert actual user deletion is called).
* where deleting notebook dirs fail
* where notebook dirs to be deleted is not found (assert actual user deletion is called).

scim/resource_user_test.go
```diff
@@ -440,6 +442,62 @@ func TestResourceUserDelete_Error(t *testing.T) {
...
+		},
+		Resource: ResourceUser(),
+		Delete:   true,
+		ID:       "abc",
+		HCL: `
+			user_name    = "abc"
+			delete_repos = true 
+			delete_dirs = true 
+		`,
+	}.Apply(t)
```
nfx, line 495: please use `.ApplyNoError` or data-checking equivalent

scim/users.go
```diff
@@ -88,6 +89,18 @@ func (a UsersAPI) Delete(userID string) error {
 	return a.client.Scim(a.context, http.MethodDelete, userPath, nil, nil)
 }
 
+func (a UsersAPI) DeleteRepos(userID string) error {
```
nfx, line 92: This does not belong here, as we're migrating the provider to Go SDK https://github.com/databricks/databricks-sdk-go/blob/main/service/scim/api.go#L827-L985

move this code to delete callback body

scim/users.go
```diff
@@ -88,6 +89,18 @@ func (a UsersAPI) Delete(userID string) error {
...
+	repoPath := fmt.Sprintf("/Repos/%v", userID)
+	recursive := true
```
nfx, line 94: why do you extract variables if they're used only once and don't aid readability?

nfx (changes requested):

scim/users.go
mberk06: I thought they do aid readability. Will remove

scim/resource_user.go
mberk06: And only include one flag for deleting both repos and dirs?

mberk06: Changed so that there's one `force_delete_content ` flag and downstream if statement.

scim/resource_user.go
mberk06: 1. I don't see anything relating to workspace/account anywhere in the users go/resource/testing files. Would love a hint.
2. Noted very interesting point. Makes sense. My current delete function has 3 calls that could produce errors (image). Should I just return the first not `nil` error, else return `NewUsersAPI(ctx, c).Delete(d.Id())`? That seems like bad design but I'm not sure how to handle 3 potential errors. 
image

mberk06: Note to self. Once each error is accessible via the return, it will be easy to implement the below test.

mberk06:
@nfx @alexott we ran out of time during office hours. I have a couple outstanding questions that are blocking next steps. Any comments would be much appreciated 🙂

scim/resource_user.go
```diff
@@ -44,6 +43,14 @@ func ResourceUser() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			}
+			m["delete_repos"] = &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			}
+			m["delete_dirs"] = &schema.Schema{
```
alexott, line 50: I would call it `delete_home_dir` or something like that

scim/resource_user.go
alexott: Wrap original error into another error.

scim/users.go
```diff
@@ -88,6 +89,18 @@ func (a UsersAPI) Delete(userID string) error {
...
+	return workspace.NewNotebooksAPI(a.context, a.client).Delete(repoPath, recursive)
+}
+
+func (a UsersAPI) DeleteDirs(userID string) error {
```
alexott, line 98: Better name is `DeleteHomeDirectory`

scim/resource_user.go
```diff
@@ -93,7 +100,18 @@ func ResourceUser() *schema.Resource {
...
+			if d.Get("delete_repos").(bool) {
+				d.Set("cheese", "cheese")
+				_ = usersAPI.DeleteRepos(d.Id())
+			}
+			if d.Get("delete_dirs").(bool) {
+				d.Set("crackers", "crackers")
+				_ = usersAPI.DeleteDirs(d.Id())
+			}
+
+			return usersAPI.Delete(d.Id())
```
alexott, line 114: I'm not sure that you can delete home folder & repos if user isn't deleted yet.

alexott (changes requested):
There are few things that needs to be done:

* Check resource isn't used at account level as pointed by Serge
* The same functionality should be implemented for `dataricks_service_principal`
* Documentation needs to be updated

scim/resource_user.go
mberk06: Chatted during office hours and we'll just return error if not nil - no wrapping.

docs/resources/service_principal.md
```diff
@@ -96,6 +96,8 @@ The following arguments are available:
 * `workspace_access` - (Optional) This is a field to allow the group to have access to Databricks Workspace.
 * `active` - (Optional) Either service principal is active or not. True by default, but can be set to false in case of service principal deactivation with preserving service principal assets.
 * `force` - (Optional) Ignore `cannot create service principal: Service principal with application ID X already exists` errors and implicitly import the specified service principal into Terraform state, enforcing entitlements defined in the instance of resource. _This functionality is experimental_ and is designed to simplify corner cases, like Azure Active Directory synchronisation.
+* `delete_repos` - (Optional) This flag determines whether the service principal's repo directory is deleted when the user is deleted. True by default.
```
nfx, line 99: default must be false, otherwise it's a behavior change

nfx (changes requested):

scim/resource_service_principal.go
```diff
@@ -83,6 +84,16 @@ func (a ServicePrincipalsAPI) Delete(servicePrincipalID string) error {
 	return a.client.Scim(a.context, "DELETE", servicePrincipalPath, nil, nil)
 }
 
+func (a ServicePrincipalsAPI) DeleteRepos(userID string) error {
```
nfx, line 87: This does not belong do `ServicePrincipalsAPI`, but rather to a resource code. with SDK migration, `ServicePrincipalsAPI` won't exist in this codebase anymore

scim/resource_service_principal.go
```diff
@@ -109,6 +120,16 @@ func ResourceServicePrincipal() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			}
+			m["delete_repos"] = &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
nfx, line 126:
```suggestion
```
no way.

scim/resource_service_principal.go
```diff
@@ -109,6 +120,16 @@ func ResourceServicePrincipal() *schema.Resource {
...
+			}
+			m["delete_home_dir"] = &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
```
nfx, line 131:
```suggestion
```
again, no way.

scim/resource_service_principal.go
```diff
@@ -161,7 +182,27 @@ func ResourceServicePrincipal() *schema.Resource {
 			})
 		},
 		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			return NewServicePrincipalsAPI(ctx, c).Delete(d.Id())
+			spAPI := NewServicePrincipalsAPI(ctx, c)
+			if c.AccountID != "" {
```
nfx, line 186: logic is broken. this change means "user won't be deleted AT ALL if we are in the account context". 

i think we need so add some validation/explanation/documentation that deleting home dirs has no effect on the account level.

scim/resource_service_principal.go
```diff
@@ -161,7 +182,27 @@ func ResourceServicePrincipal() *schema.Resource {
...
+				return nil
+			}
+			var err = spAPI.Delete(d.Id())
```
nfx, line 189:
```suggestion
			err := spAPI.Delete(d.Id())
```

scim/resource_service_principal.go
```diff
@@ -161,7 +182,27 @@ func ResourceServicePrincipal() *schema.Resource {
...
+				return err
+			}
+			if d.Get("delete_repos").(bool) {
+				err = spAPI.DeleteRepos(d.Id())
+				if err != nil {
+					return err
+				}
+			}
+			if d.Get("delete_home_dir").(bool) {
+				err = spAPI.DeleteHomeDirectory(d.Id())
```
nfx, line 200: please add `acceptance.Test` (the one talking to the real workspace) to verify that homedirs are not deleted if this flag is not set and another test to verify that home dir is deleted, when it's set

scim/resource_service_principal.go
```diff
@@ -161,7 +182,27 @@ func ResourceServicePrincipal() *schema.Resource {
...
+				if err != nil {
+					return err
```
nfx, line 202:
```suggestion
					return fmt.Errorf("delete_home_dir: %w", err)
```

scim/resource_service_principal.go
```diff
@@ -161,7 +182,27 @@ func ResourceServicePrincipal() *schema.Resource {
...
+				return nil
+			}
+			var err = spAPI.Delete(d.Id())
+			if err != nil {
+				return err
+			}
+			if d.Get("delete_repos").(bool) {
+				err = spAPI.DeleteRepos(d.Id())
+				if err != nil {
+					return err
```
nfx, line 196:
```suggestion
					return fmt.Errorf("delete_repos: %w", err)
```

nfx (changes requested):

scim/resource_service_principal.go
mberk06: Added path delete code into resource `Delete` callback

alexott (changes requested):
@nfx @mberk06 we need to discuss error handling when they occur during deletion of home_dir and/or repos dir

scim/acceptance/service_principal_test.go
```diff
@@ -42,3 +47,78 @@ func TestAccServicePrincipalResourceOnAws(t *testing.T) {
...
+	appId := "12345a67-8b9c-0d1e-23fa-4567b89cde04"
+	os.Setenv("appId", appId)
+	ctx := context.Background()
+	client := common.CommonEnvironmentClient()
+	notebooksAPI := workspace.NewNotebooksAPI(ctx, client)
+	acceptance.Test(t, []acceptance.Step{
+		{
+			Template: `
+			resource "databricks_service_principal" "a" {
+				application_id = "{env.appId}"
```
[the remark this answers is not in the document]
mberk06: Weird I tested on AWS, GCP, and Azure and it works. Since the `application_id` is used to create the home directory, how should I reference it without passing it in?

scim/acceptance/service_principal_test.go
```diff
@@ -42,3 +47,78 @@ func TestAccServicePrincipalResourceOnAws(t *testing.T) {
...
+				delete_home_dir = true
+			}`,
+			Check: func(s *terraform.State) error {
+				return nil
+			},
+		},
+		{
+			Template: `
+			resource "databricks_service_principal" "b" {
+				application_id = "12345a67-8b9c-0d1e-23fa-4567b89cde10"
```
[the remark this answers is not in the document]
mberk06: I'm running a check on the prior step's destroy. If I create a SP with the same display name, it will recreate the dir. Should I pass in a variable instead of hard-coding?

scim/resource_service_principal.go
```diff
@@ -161,7 +170,28 @@ func ResourceServicePrincipal() *schema.Resource {
...
+				return err
+			}
+			if c.AccountID != "" {
+				return nil
+			}
+			if d.Get("delete_repos").(bool) {
+				err = workspace.NewNotebooksAPI(ctx, c).Delete(fmt.Sprintf("/Repos/%v", appId), true)
+				if err != nil {
+					return fmt.Errorf("delete_repos: %w", err)
+				}
```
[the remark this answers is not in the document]
mberk06: Agreed. When we first talked we scoped that the errors should be logged. I'll defer to you and @nfx on how to proceed.

scim/acceptance/service_principal_test.go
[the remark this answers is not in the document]
mberk06: TODO: only set to run on azure.

scim/resource_service_principal.go
```diff
@@ -161,7 +170,28 @@ func ResourceServicePrincipal() *schema.Resource {
...
+			appId := d.Get("application_id").(string)
+			err := spAPI.Delete(d.Id())
+			if err != nil {
+				return err
+			}
+			if c.AccountID != "" {
+				return nil
+			}
+			if d.Get("delete_repos").(bool) {
+				err = workspace.NewNotebooksAPI(ctx, c).Delete(fmt.Sprintf("/Repos/%v", appId), true)
```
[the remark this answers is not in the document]
mberk06: I didn't hear back from the repos team, but I tested it on a demo workspace and it works.

mberk06:
Moving to PR above - couldn't resolve SDK conflicts

PR #1862, opened 2022-12-19
Delete empty host odbc parameter

fuselessmatt:
As the comment says, `host` never gets populated, but `Hostname` does. It the UI it is called "Server hostname". Let's finally remove it in the Terraform module as it looks broken when it is simply not set.


nfx:
@fuselessmatt there's an upcoming Go SDK migration, i'd keep any struct changes out of scope until then. I'll close this PR until then.

Issue #1867, opened 2022-12-21
[DOC] Argument `num_workers` is missing


SatoruItaya:
### Affected Resource(s)

[databricks_cluster](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/cluster)

### List of things to potentially add/remove:
This is a list of things to manipulate in the docs:
- [ ] num_workers

### Important Factoids

https://github.com/databricks/terraform-provider-databricks/blob/8b3c2c907db4d726169f127a0992b42dc9c8ce23/clusters/resource_cluster.go#L107-L112

iandexter:
Thanks for your report, @SatoruItaya -- the change has been merged, and will be available in the public documentation in the next available deployment

PR #1868, opened 2022-12-21
Add num_workers optional argument

iandexter:
Ref: https://github.com/databricks/terraform-provider-databricks/issues/1867


nfx (approved):
wow, i really wonder why this was not yet documented.

Issue #1764, opened 2022-11-15
[ISSUE] Issue with `databricks_sql_permissions` resource


TheoPth:
### Configuration
```hcl
# Copy-paste your Terraform configuration here

resource "databricks_sql_permissions" "databases_groups_sql_permissions" {
  cluster_id = "<my-cluster-id>"
  database   = "mydatabase"

  privilege_assignments {
    principal  = "Grouptest"
    privileges = ["USAGE", "SELECT", "READ_METADATA"]
  }
}
```

### Expected Behavior
the resource is applied and should no longer appear as a change in the next plan 

### Actual Behavior
the block "privilege_assignments" appears in every plan and is applied every time
![image](https://user-images.githubusercontent.com/32432470/201950441-fb698d4d-4870-4e11-bdf9-1747a28f4a5c.png)

### Steps to Reproduce

1. `terraform plan`-->
Create  a databricks_sql_permissions with privilege_assignments
Plan and apply two times.
you'll see that it's planned and applied every time.

### Terraform and provider versions
terraform version : 1.2.2
databricks provider version : 1.6.3 and after

### Important Factoids
The bug was introduced in the version 1.6.3. this feature was working in version 1.6.2 (I did not manage to find the root cause by myself)

nfx:
I think this might be the regression caused by #1730 https://github.com/databricks/terraform-provider-databricks/compare/v1.6.2...v1.6.3 

could you share the logs?

TheoPth:
Hi @nfx 
Thanks for your response, here LOG_TRACE of the plan
[run-5Mxk5yxFoDMbZLYs-plan-log.txt](https://github.com/databricks/terraform-provider-databricks/files/10019761/run-5Mxk5yxFoDMbZLYs-plan-log.txt)

TheoPth:
I created a exemple with only one ressource, it's easier to understand the issue.

It's a plan / apply / plan , you'll see that the first and second plan are same. So the apply takes place every time (big issue when you have a lot of sql permissions)

Removed the commit 976b793fd947e3a5a7c52ddbfcaf992cccf0e488 fixed the issue.

This is my first contribution, if you have any comments I will be glad to hear them


[logs.txt](https://github.com/databricks/terraform-provider-databricks/files/10288455/logs.txt)

nfx:
@adamcain-db - could you take a look at this table-acl permissions issue, when you have time?

TheoPth:
```hcl
terraform {
[...]

PR #1872, opened 2022-12-22
Fix bug in `privilege_assignment` of `databricks_sql_permissions`

TheoPth:
…ed code to correctly assign permissions without causing infinite loop

### DIFF
--- a/access/resource_sql_permissions.go
+++ b/access/resource_sql_permissions.go
@@ -148,9 +148,12 @@ func (ta *SqlPermissions) read() error {
 		if !strings.EqualFold(currentType, thisType) {
 			continue
 		}
-		if !strings.EqualFold(currentKey, thisKey) {
+
+		noBackticks := strings.ReplaceAll(thisKey, "`", "")
+		if !strings.EqualFold(currentKey, thisKey) && !strings.EqualFold(currentKey, noBackticks) {
 			continue
 		}
+
 		if strings.HasPrefix(currentAction, "DENIED_") {
 			// DENY statements are intentionally not supported.
 			continue
--- a/access/resource_sql_permissions_test.go
+++ b/access/resource_sql_permissions_test.go
@@ -89,6 +89,22 @@ func TestTableACLGrants(t *testing.T) {
 	assert.Len(t, ta.PrivilegeAssignments[0].Privileges, 2)
 }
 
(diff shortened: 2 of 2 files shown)


TheoPth:
fix issue https://github.com/databricks/terraform-provider-databricks/issues/1764

access/resource_sql_permissions.go
```diff
@@ -50,7 +50,7 @@ func (ta *SqlPermissions) typeAndKey() (string, string) {
 		return "VIEW", fmt.Sprintf("`%s`.`%s`", ta.actualDatabase(), ta.View)
 	}
 	if ta.Database != "" {
-		return "DATABASE", fmt.Sprintf("`%s`", ta.Database)
+		return "DATABASE", ta.Database
```
nfx, line 53: backticks are valid quoting characters. why do you remove them?

nfx (changes requested):

access/resource_sql_permissions.go
TheoPth: Hi !  I explain it there https://github.com/databricks/terraform-provider-databricks/issues/1764#issuecomment-1363029378 with logs and an example 
privilege_assignments are planned and applied every times because of this line 
I didn't find the root cause

nkvuong:
@TheoPth 
the root cause is around line 150, you need to update the key match to
```go
noBackticks := strings.ReplaceAll(thisKey, "`", "")
if !strings.EqualFold(currentKey, thisKey) && !strings.EqualFold(currentKey, noBackticks) {
    continue
}
```

TheoPth:
Thanks a lot @nkvuong, I tested your solution and it's working like a charm

@nfx I think this is a better solution than the one I initially proposed, what do you think?

TheoPth:
Thanks a lot @nkvuong, I tested your solution and it works fine

@nfx The fix is much better now, what do you think ?

nkvuong:
@TheoPth could you also add a test to access/resource_sql_permissions_test.go, similar to `TestTableACLGrants`?

tommy-watts-depop:
Hi all, is there an ETA on this? thanks !

TheoPth:
Hi @nkvuong, I provided the test. is it enough ?

nkvuong (approved):
lgtm

nfx (approved):

PR #1873, opened 2022-12-22
replace the incorrect 'id' attribute with the correct 'sp_id' attribute

liamperritt:
replace the incorrect 'id' attribute with the correct 'sp_id' attribute in service principal data source docs example

### DIFF
--- a/docs/data-sources/service_principal.md
+++ b/docs/data-sources/service_principal.md
@@ -23,7 +23,7 @@ data "databricks_service_principal" "spn" {
 
 resource "databricks_group_member" "my_member_a" {
   group_id  = data.databricks_group.admins.id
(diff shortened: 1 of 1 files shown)


docs/data-sources/service_principal.md
```diff
@@ -23,7 +23,7 @@ data "databricks_service_principal" "spn" {
 
 resource "databricks_group_member" "my_member_a" {
   group_id  = data.databricks_group.admins.id
-  member_id = data.databricks_service_principal.spn.id
```
alexott, line 26: really, it would be easier to fix service principal data source to expose `id` instead of `sp_id` - this will match attributes of other resources.

alexott (commented):
we need to discuss if it's better to fix data source or not

PR #1871, opened 2022-12-22
State upgraders for `databricks_mws_workspaces`

nfx:

PR #1879, opened 2022-12-28
Added support for `databricks_mws_workspaces` on GCP (Public Preview)

nkvuong:
Docs for #1871 

- Update `databricks_mws_networks` and `databricks_mws_workspaces` doc
- Rename `AWS` subcategory to `Deployment`
- Add GCP auth information to index


docs/resources/mws_networks_gcp.md
```diff
@@ -0,0 +1,108 @@
+---
+subcategory: "GCP"
```
nfx, line 2: this creates a strange precedent. do we want it?

nkvuong: `databricks_mws_workspaces` and `databricks_mws_networks` would apply to both GCP & AWS, so we will need a new subcategory for both resources?

grayluck:
Expected change log: Add Terrafrom schema support for GCP account API public preview workspace management.

nfx (changes requested):
change suffices

PR #1881, opened 2022-12-28
Small doc fixes

nkvuong:


docs/resources/grants.md
```diff
@@ -6,6 +6,9 @@ subcategory: "Unity Catalog"
 -> **Note**
   This article refers to the privileges and inheritance model in Privilege Model version 1.0. If you created your metastore during the public preview (before August 25, 2022), you can upgrade to Privilege Model version 1.0 following [Upgrade to privilege inheritance](https://docs.databricks.com/data-governance/unity-catalog/hive-metastore.html)
 
+-> **Notes**
+  Unity Catalog APIs are accessible via **workspace-level APIs**. This design may change in the future.
```
fuselessmatt, line 10:
```suggestion
  Unity Catalog APIs are accessible via **workspace-level APIs**. This design may change in the future. Account-level principal grants can be assigned with any valid workspace as the Unity Catalog is decoupled from specific workspaces.
```

I think we need to write something like this because otherwise the implications are not clear unless you already know the deal

nfx (approved):

PR #1883, opened 2022-12-28
removed unneeded brace from integration test that causes the plan to fail for int test

stikkireddy:
Can you confirm that the acceptance test run on acceptance env.


stikkireddy:
looks like #1878 resolves this

PR #1884, opened 2022-12-28
Fix UC secret update and correct syntax of a gcp test

shreyas-goenka:
Makes `recipient_profile_str` a write only field thus fixing the ACC test

### DIFF
--- a/catalog/acceptance/provider_test.go
+++ b/catalog/acceptance/provider_test.go
@@ -19,7 +19,7 @@ func TestUcAccCreateProviderDb2Open(t *testing.T) {
 			  recipient_profile_str = jsonencode({
 					"shareCredentialsVersion":1,
 					"bearerToken":"dapiabcdefghijklmonpqrstuvwxyz",
-					"endpoint":"https://sharing.delta.io/delta-sharing/"}
+					"endpoint":"https://sharing.delta.io/delta-sharing/"
 				}
 			  )
 			}`,
--- a/catalog/resource_provider.go
+++ b/catalog/resource_provider.go
@@ -21,7 +21,13 @@ type ProviderInfo struct {
 	Name                string `json:"name" tf:"force_new"`
 	Comment             string `json:"comment,omitempty"`
 	AuthenticationType  string `json:"authentication_type"`
-	RecipientProfileStr string `json:"recipient_profile_str" tf:"sensitive,suppress_diff"`
+	RecipientProfileStr string `json:"recipient_profile_str" tf:"sensitive"`
+}
+
+type ProviderInfoMetadata struct {
+	Name               string `json:"name" tf:"force_new"`
+	Comment            string `json:"comment,omitempty"`
+	AuthenticationType string `json:"authentication_type"`
 }
 
 type Providers struct {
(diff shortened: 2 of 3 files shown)


catalog/acceptance/provider_test.go
```diff
@@ -19,7 +19,7 @@ func TestUcAccCreateProviderDb2Open(t *testing.T) {
 			  recipient_profile_str = jsonencode({
 					"shareCredentialsVersion":1,
 					"bearerToken":"dapiabcdefghijklmonpqrstuvwxyz",
-					"endpoint":"https://sharing.delta.io/delta-sharing/"}
+					"endpoint":"https://sharing.delta.io/delta-sharing/"
```
shreyas-goenka, line 22: Fixes the syntax issue, but the gcp test is still broken

catalog/resource_provider.go
```diff
@@ -51,15 +57,19 @@ func (a ProvidersAPI) updateProvider(ci *ProviderInfo) error {
 }
 
 func ResourceProvider() *schema.Resource {
-	providerSchema := common.StructToSchema(ProviderInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+	providerInfoSchema := common.StructToSchema(ProviderInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		m["authentication_type"].ValidateFunc = validation.StringInSlice([]string{"TOKEN"}, false)
```
nfx, line 61: and where `recipient_profile_str` will be coming from?...

nfx (changes requested):

catalog/resource_provider.go
nfx: you need to write `m["recipient_profile_str"].DiffSuppressFunc`, which will suppress the diff when platform returns zero value and we have some in the config. this is "inverted" behavior from `tf:"suppress_diff"` and has to be hand-coded

catalog/resource_provider.go
```diff
@@ -51,15 +57,19 @@ func (a ProvidersAPI) updateProvider(ci *ProviderInfo) error {
 }
 
 func ResourceProvider() *schema.Resource {
-	providerSchema := common.StructToSchema(ProviderInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
```
nfx, line 54: here's how we can make this more readable and clear:

1. revert the `providerSchema` rename. It diverts the attention from the intention of this commit by making more changes than necessary.
2. remove `ProviderInfoMetadata` as it's not necessary anymore.
3. create the copy of `providerSchema`, but without `recipient_profile_str`:

```go
providerSchemaForRead := map[string]*schema.Schema{
  "name": providerSchema["name"],
  "comment": providerSchema["comment"],
  "authentication_type": providerSchema["authentication_type"],
}
```

you can also just inline that in the `Read` callback.

nfx (changes requested):

catalog/resource_provider.go
shreyas-goenka: Done

shreyas-goenka:
test run for this PR: https://github.com/databricks/eng-dev-ecosystem/actions/runs/3801210820

nfx (approved):

PR #1885, opened 2022-12-29
Allow to override resource ID for specific data source

alexott:
The `DataResource` function now checks if the provided schema has the field with name `id` and set its value as resource ID.  This will allow to have uniform references between normal resources and data sources. If this change is ok, then I'll proceed with other data sources. (See https://github.com/hashicorp/terraform-plugin-sdk/issues/607 for reference about definition of `id` attribute).

This change was already applied to the `databricks_service_principal` data source where old attribute (`sp_id`) was left for backward compatibility.

### DIFF
--- a/common/resource.go
+++ b/common/resource.go
@@ -199,6 +199,8 @@ func DataResource(sc any, read func(context.Context, any, *DatabricksClient) err
 				diags = diag.FromErr(err)
 			}
 			StructToData(ptr.Elem().Interface(), s, d)
+			// check if the resource schema has the `id` attribute (marked with `json:"id"` in the provided structure).
+			// and if yes, then use it as resource ID. If not, then use default value for resource ID (`_`)
 			if _, ok := s["id"]; ok {
 				d.SetId(d.Get("id").(string))
 			} else {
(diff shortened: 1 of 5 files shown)


common/resource.go
```diff
@@ -199,7 +199,11 @@ func DataResource(sc any, read func(context.Context, any, *DatabricksClient) err
 				diags = diag.FromErr(err)
 			}
 			StructToData(ptr.Elem().Interface(), s, d)
-			d.SetId("_")
+			if _, ok := s["id"]; ok {
```
nfx, line 202: please add more extended comments - it's clear to me and you, but possibly not to others :)

docs/data-sources/service_principal.md
```diff
@@ -37,7 +37,8 @@ Data source allows you to pick service principals by the following attributes
 
 Data source exposes the following attributes:
 
-- `sp_id` - The id of the service principal.
+- `id` - The id of the service principal.
+- `sp_id` - The id of the service principal (deprecated, kept for compatibility with earlier versions).
```
nfx, line 41: probably just remove it from docs. Also please mention this in the PR description.

nfx (approved):

nkvuong (approved):

nfx (approved):

PR #1887, opened 2022-12-30
Extend `databricks_catalog` resource to support Delta Sharing Catalog

### LINKED ISSUE #1858: [FEATURE] Extend the `databricks_catalog` resource to allow for the creation of catalogs from Delta Shares
### Use-cases

When creating a Databricks-to-Databricks Delta Share via Terraform a user can create a Recipient, a Share, and add tables to a share.  However, there is no way to create a catalog on the recipient Metastore for the share via the databricks_catalog resource.  This step is necessary as the catalog is then used for a recipient to provide access to other users/groups in their workspace.

### Attempted Solutions

Only current workaround is to do this step manually via the UI.  Steps are detailed here via Data Explorer, SQL, or CLI.

https://docs.databricks.com/data-sharing/read-data-databricks.html

### Proposal

Extend the TF databricks_catalog resource to allow for the creation of catalogs from Delta Shares.

### References

https://api-docs.databricks.com/rest/latest/unity-catalog-api-specification-2-1.html?_ga=2.256550600.2105607269.1671031203-450893202.1659449140#:~:text=the%20External%20Location.-,Catalog%20CRUD%20API,-Object%20Models

nkvuong:
Closes #1858

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -18,12 +18,14 @@ func NewCatalogsAPI(ctx context.Context, m any) CatalogsAPI {
 }
 
 type CatalogInfo struct {
-	Name        string            `json:"name" tf:"force_new"`
-	Comment     string            `json:"comment,omitempty"`
-	StorageRoot string            `json:"storage_root,omitempty" tf:"force_new"`
-	Properties  map[string]string `json:"properties,omitempty"`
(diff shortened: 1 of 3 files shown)


nkvuong:
Running acceptance tests for ucws

nkvuong:
Acceptance tests passed

alexott (approved):
lgtm

nfx (approved):

PR #1889, opened 2022-12-30
Fixed the issue of simple named objects not appearing in .tf output files

rohitkn:
Changes were made to the `importables.go`, and `util.go` such that the `ResourceName(r)` function in context.go would return a normalized but **changed** name
Changes to the supplied names were very basic: 1) either changed an underscore `_` to a dash `-` or 2) `upper`cased the name.

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -174,7 +174,7 @@ var resourcesMap map[string]importable = map[string]importable{
 				ic.Emit(&resource{
 					Resource: "databricks_permissions",
 					ID:       fmt.Sprintf("/instance-pools/%s", r.ID),
-					Name:     "inst_pool_" + ic.Importables["databricks_instance_pool"].Name(ic, r.Data),
+					Name:     "inst-pool_" + ic.Importables["databricks_instance_pool"].Name(ic, r.Data),
 				})
 			}
 			return nil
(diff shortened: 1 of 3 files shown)


rohitkn:
This is a fix for ISSUE number 1838

exporter/importables.go
```diff
@@ -174,7 +174,7 @@ var resourcesMap map[string]importable = map[string]importable{
 				ic.Emit(&resource{
 					Resource: "databricks_permissions",
 					ID:       fmt.Sprintf("/instance-pools/%s", r.ID),
-					Name:     "inst_pool_" + ic.Importables["databricks_instance_pool"].Name(ic, r.Data),
+					Name:     "inst-pool_" + ic.Importables["databricks_instance_pool"].Name(ic, r.Data),
```
alexott, line 177: although `-` is allowed in resources identifiers, they aren't considered idiomatic: https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#identifiers

exporter/importables.go
```diff
@@ -790,7 +790,7 @@ var resourcesMap map[string]importable = map[string]importable{
 					ic.Emit(&resource{
 						Resource: "databricks_secret_scope",
 						ID:       scope.Name,
-						Name:     scope.Name,
+						Name:     strings.ToUpper(scope.Name), //strings.ToUpper fixes 1838
```
alexott, line 793: the lower-cased names are recommended.

alexott (changes requested):
it's not a real fix, but more a workaround. I was planning to fix this issue during this week as I'm back to office

alexott:
#1891 fixes it with smaller amount of changes

PR #1890, opened 2023-01-02
Exporter: generate references that are matching only to a prefix

### LINKED ISSUE #1777: [ISSUE] Issue with Users folder and `databricks_notebook` resource on `databricks_user` object
### Configuration
```hcl
resource "databricks_notebook" "users_<REDACTED>_ag_com_test_1566736564850376" {
  source = "${path.module}/notebooks/Users/<REDACTED>@ag.com/test/yodlee_demographics_2.py"
  path   = "/Users/<REDACTED>@ag.com/test/yodlee_demographics_2"
}

```

### Expected Behavior
Notebook should be created in databricks workspace under the user's directory and subdirectory tree

### Actual Behavior
Databricks provider's user (engineer) reports that they see errors:

1. 
>  Error: cannot create notebook: Folder Users is protected
>    with databricks_notebook.users_<REDACTED>_ag_com_test_1566736564850376,
>    on notebooks.tf line 8233, in resource "databricks_notebook" "users_<REDACTED>_ag_com_test_1566736564850376":       
>  8233: resource "databricks_notebook" "users_<REDACTED>_ag_com_test_1566736564850376" {
> 

2.
Problems occur when a notebook is *not* in the user's root path 
**this works** ie: /Users/johnsmith@databricks.com/MyExporterNotebook.py
**this does not** ie: /Users/johnsmith@databricks.com/exporter/MyExporterNotebook.py

### Steps to Reproduce
1. Run exporter on a workspace with user directories, subdirectories and notebooks
2. Run import (terraform init, terraform plan, terraform apply) off exporter's output into a new or empty workspace

### Important notes
Adding a 
```
depends_on = [databricks_user.johnsmith_ag_com] 
```
to the notebook resource fixes the problem

alexott:
There are cases when reference isn't a full match of a given value but a prefix of it. For example, user's notebook has a path of `/Users/user@domain/notebook`.  To make it working correctly we need to emit user or service principal, and then make a correct reference with dependency on emitted user or notebook, like, `${databricks_user.user_domain.home}/notebook`.

This PR adds a new field to dependency specification: `MatchType` that may have following values:

- `prefix` - value of attribute of resource is a prefix of a given attribute
- `exact` (or empty string) - must have exact matching (previous behaviour)

this fixes #1777

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -301,7 +301,7 @@ func (ic *importContext) MatchesName(n string) bool {
 	return strings.Contains(strings.ToLower(n), strings.ToLower(ic.match))
 }
 
-func (ic *importContext) Find(r *resource, pick string) hcl.Traversal {
+func (ic *importContext) Find(r *resource, pick, matchType string) (string, hcl.Traversal) {
 	for _, sr := range ic.State.Resources {
 		if sr.Type != r.Resource {
 			continue
@@ -313,24 +313,28 @@ func (ic *importContext) Find(r *resource, pick string) hcl.Traversal {
 					r.Attribute, r.Resource, r.Name, r.ID)
 				continue
 			}
-			if v.(string) == r.Value {
-				if sr.Mode == "data" {
-					return hcl.Traversal{
-						hcl.TraverseRoot{Name: "data"},
-						hcl.TraverseAttr{Name: sr.Type},
-						hcl.TraverseAttr{Name: sr.Name},
-						hcl.TraverseAttr{Name: pick},
-					}
-				}
-				return hcl.Traversal{
-					hcl.TraverseRoot{Name: sr.Type},
+			strValue := v.(string)
+			if ((matchType == "" || matchType == "exact") && strValue != r.Value) ||
+				(matchType == "prefix" && !strings.HasPrefix(r.Value, strValue)) {
+				continue
+			}
+			if sr.Mode == "data" {
+				return strValue, hcl.Traversal{
+					hcl.TraverseRoot{Name: "data"},
+					hcl.TraverseAttr{Name: sr.Type},
 					hcl.TraverseAttr{Name: sr.Name},
 					hcl.TraverseAttr{Name: pick},
 				}
 			}
+			return strValue, hcl.Traversal{
+				hcl.TraverseRoot{Name: sr.Type},
+				hcl.TraverseAttr{Name: sr.Name},
+				hcl.TraverseAttr{Name: pick},
+			}
+
 		}
 	}
-	return nil
+	return "", nil
 }
 
 func (ic *importContext) Has(r *resource) bool {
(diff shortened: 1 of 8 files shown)


exporter/context.go
```diff
@@ -509,13 +513,22 @@ func (ic *importContext) reference(i importable, path []string, value string) hc
...
+		attrValue, traversal := ic.Find(&resource{
 			Resource:  d.Resource,
 			Attribute: attr,
 			Value:     value,
-		}, attr)
-		//at least one invocation of ic.Find will assign Nil to traversal if resource with value is not found
+		}, attr, d.MatchType)
+		// at least one invocation of ic.Find will assign Nil to traversal if resource with value is not found
 		if traversal != nil {
+			if d.MatchType == "prefix" { // we're replacing the found prefix with the HCL reference using the "${reference}rest" syntax
```
nfx, line 523: is it more complex if `ic.Find` returns prefixed traversal?

exporter/model.go
```diff
@@ -57,11 +57,12 @@ type importable struct {
...
 type reference struct {
-	Path     string
-	Resource string
-	Match    string
-	Variable bool
-	File     bool
+	Path      string
+	Resource  string
+	Match     string
+	MatchType string // type of match, `prefix` - reference is embedded into string, `` (or `exact`) - full match
```
nfx, line 63:
```suggestion
	MatchPrefix bool
```
Wouldn't a simple flag here be simpler? is there a third match type we have in mind to justify a new category?

nfx (changes requested):

exporter/model.go
alexott: I'm thinking right now about additional  match types:
* `substring` - secret scope references
* `suffix` - for workspace support in SQL queries/dashboards.

exporter/context.go
```diff
@@ -312,24 +312,28 @@ func (ic *importContext) Find(r *resource, pick string) hcl.Traversal {
...
-						hcl.TraverseRoot{Name: "data"},
-						hcl.TraverseAttr{Name: sr.Type},
-						hcl.TraverseAttr{Name: sr.Name},
-						hcl.TraverseAttr{Name: pick},
-					}
-				}
-				return hcl.Traversal{
-					hcl.TraverseRoot{Name: sr.Type},
+			strValue := v.(string)
+			if ((matchType == "" || matchType == "exact") && strValue != r.Value) ||
```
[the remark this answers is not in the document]
alexott: good idea, will do

exporter/context.go
alexott: you mean to generate tokens directly in the `ic.Find` ? yes, maybe it's good idea to isolate it there

exporter/context.go
```diff
@@ -509,13 +513,22 @@ func (ic *importContext) reference(i importable, path []string, value string) hc
 		if d.Match != "" {
 			attr = d.Match
 		}
-		traversal := ic.Find(&resource{
+		attrValue, traversal := ic.Find(&resource{
```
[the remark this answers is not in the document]
alexott: https://go.dev/doc/faq#Does_Go_have_a_ternary_form

nfx (approved):

PR #1891, opened 2023-01-02
Exporter: fix issue with exporting of resources with simple name

### LINKED ISSUE #1838: [ISSUE] Issue with all `databricks_*` resources. Resource export is missed entirely for simple names
This is the implementation of the `string()` implementation on a resource

```
// String doubles its use as "hashcode"
func (r *resource) String() string {
	n := r.Name
	if n == "" {
		n = "<unknown>"
	}
	k, v := r.MatchPair()
	return fmt.Sprintf("%s[%s] (%s: %s)", r.Resource, n, k, v)
}
```

The `Emit(r *resource)` function in context.go sets an entry **to mark the resources to be set to import** like this

1. `ic.importing[r.String()] = true`

Subsequently the resource name - `r.Name` variable is expected to change (normalized) with this:

2. `r.Name = ic.ResourceName(r)`

Lastly, the function `Add(r)` is called to put the resource in `Scope` to be exported

3. Without this step, the resource *is never exported*. If it's **not in scope, it won't be exported.**

- `Add(r)` in the function `Emit(r *resource)` first checks `Has(r)` that is - have we already added this to `ic.importing` map?
-  When the `r.Name` does not change in step 2 from what it was in step 1, then `Has(r)` returns **true** and the resource is never added to the scope.

### Expected Behavior
All resources should be exported regardless of their customer provided or computer generated name or display name

### Actual Behavior

- A name for a secret scope like `lab39`

never gets normalized because it has no special characters and thus... it is never exported.

### Steps to Reproduce

1. Create a secret scope called `lab39` and `lab-41`
2. Run the exporter with -services secrets and -listing secrets
3. Inspect secrets.tf

All terraform versions
All Provider versions up to 1.6.5

alexott:
The `Add` function is called only from the `Emit` so it's safe to remove the check if resource was already imported or not.

this fixes #1838

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -333,8 +333,14 @@ func (ic *importContext) Find(r *resource, pick string) hcl.Traversal {
 	return nil
 }
 
+// This function checks if resource exist in any state (already added or in process of addition)
 func (ic *importContext) Has(r *resource) bool {
-	if _, visiting := ic.importing[r.String()]; visiting {
+	return ic.HasInState(r, false)
+}
+
+// This function checks if resource exist. onlyAdded flag enforces that true is returned only if it was added with Add()
+func (ic *importContext) HasInState(r *resource, onlyAdded bool) bool {
+	if v, visiting := ic.importing[r.String()]; visiting && (v || !onlyAdded) {
 		return true
 	}
 	k, v := r.MatchPair()
@@ -352,6 +358,10 @@ func (ic *importContext) Has(r *resource) bool {
 }
 
 func (ic *importContext) Add(r *resource) {
+	if ic.HasInState(r, true) { // resource must exist and already marked as added
+		return
+	}
+	ic.importing[r.String()] = true // mark resource as added
 	state := r.Data.State()
 	if state == nil {
 		log.Printf("[ERROR] state is nil for %s", r)
@@ -422,7 +432,7 @@ func (ic *importContext) Emit(r *resource) {
 		ic.testEmits[r.String()] = true
 		return
 	}
-	ic.importing[r.String()] = true
+	ic.importing[r.String()] = false // we're starting to add a new resource
 	pr, ok := ic.Resources[r.Resource]
 	if !ok {
 		log.Printf("[ERROR] %s is not available in provider", r)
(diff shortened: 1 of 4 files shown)


exporter/context.go
```diff
@@ -351,9 +351,6 @@ func (ic *importContext) Has(r *resource) bool {
 }
 
 func (ic *importContext) Add(r *resource) {
-	if ic.Has(r) {
```
nfx, line 354: this is a short-circuit for not adding duplicate resources. cannot be removed.

exporter/importables.go
```diff
@@ -776,7 +776,8 @@ var resourcesMap map[string]importable = map[string]importable{
 	"databricks_secret_scope": {
 		Service: "secrets",
 		Name: func(ic *importContext, d *schema.ResourceData) string {
-			return d.Get("name").(string)
+			name := d.Get("name").(string)
+			return name + "_" + fmt.Sprintf("%x", md5.Sum([]byte(name)))
```
nfx, line 780: please use either concatenation or sprintf, but not both in the same statement

exporter/importables_test.go
```diff
@@ -194,7 +194,7 @@ func TestSecretScope(t *testing.T) {
 	d.Set("name", "abc")
 	ic := importContextForTest()
 	name := ic.Importables["databricks_secret_scope"].Name(ic, d)
-	assert.Equal(t, "abc", name)
+	assert.Equal(t, "abc_900150983cd24fb0d6963f7d28e17f72", name)
```
nfx, line 197: not too critical for this PR, but what if we use sha1 and get just first 9 characters from it, like git does - e.g https://github.com/databricks/terraform-provider-databricks/commit/8c16069abaf9843a3955230ab82f1a194199a23f for `8c16069abaf9843a3955230ab82f1a194199a23f`

then the generated conf would look less ugly.

nfx (changes requested):

exporter/context.go
alexott: ah, ok - I see, there could be a case where were nested import will re-add the same resource. I'll re-implement it a bit differently

alexott: @nfx what about updated implementation?

nfx (approved):

PR #1897, opened 2023-01-03
Exporter: emit secret scopes from Spark Conf and Env Vars

alexott:
Exporter now detects Spark conf and environment variables values that are references to secrets (the `{{secrets/scope/secret}}` syntax) and emit a secret scope referenced there.

P.S. It doesn't put explicit reference into the value, but it could be added after #1890 is merged (it will require addition of `substring` match type, but it's doable).


nfx (approved):
lgtm

PR #1898, opened 2023-01-03
Add `config_reference` attribute to `databricks_secret` for easier reference from Spark configuration

alexott:
The `config_reference` attribute exposes string value in form of `{{secrets/scope/key}}` for easier reference from Spark configuration or environment variables.

It will be also used later in exporter for handling dependencies

### DIFF
--- a/docs/resources/secret.md
+++ b/docs/resources/secret.md
@@ -33,6 +33,7 @@ In addition to all arguments above, the following attributes are exported:
 
 * `id` - Canonical unique identifier for the secret.
 * `last_updated_timestamp` - (Integer) time secret was updated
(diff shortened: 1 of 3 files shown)


docs/resources/secret.md
```diff
@@ -33,6 +33,7 @@ In addition to all arguments above, the following attributes are exported:
 
 * `id` - Canonical unique identifier for the secret.
 * `last_updated_timestamp` - (Integer) time secret was updated
+* `config_reference` - (String) value to use as a secret reference in [Spark configuration and environment variables](https://docs.databricks.com/security/secrets/secrets.html#use-a-secret-in-a-spark-configuration-property-or-environment-variable): `{{secrets/scope/key}}`.
```
nfx, line 36: an extended TF example could be good to add in a follow-up PR, so that users know about it in context

nfx (approved):

docs/resources/secret.md
alexott: good. I'll add it when I work on the exporter pieces

Issue #1768, opened 2022-11-16
[ISSUE] Issue with `databricks_mws_permission_assignment` resource


BostjanBozic:
### Configuration
```hcl
provider "databricks" {
  alias      = "mws"
  host       = "https://accounts.cloud.databricks.com"
  account_id = var.databricks_account.account_id
  username   = var.databricks_account.username
  password   = var.databricks_account.password
}

provider "databricks" {
  alias = "ws"
  host  = databricks_mws_workspaces.test.workspace_url
  token = databricks_mws_workspaces.test.token[0].token_value
}

data "databricks_group" "test_group" {
  provider     = databricks.mws
  display_name = "test-group"
}

resource "databricks_mws_permission_assignment" "test_group" {
  provider     = databricks.mws
  workspace_id = databricks_mws_workspaces.test.workspace_id
  principal_id = data.databricks_group.test_group.id
  permissions  = ["USER"]
}
```

### Expected Behavior

Permissions are assigned without any problem.

### Actual Behavior

At initial `terraform apply`, apply fails with error below:
```
│ Error: cannot create mws permission assignment: Databricks API (/api/2.0/preview/accounts/<account_id>/workspaces/<workspace_id>/permissionassignments/principals/<principal_id>) requires you to set `host` property (or DATABRICKS_HOST env variable) to result of `databricks_mws_workspaces.this.workspace_url`. This error may happen if you're using provider in both normal and multiworkspace mode. Please refactor your code into different modules. Runnable example that we use for integration testing can be found in this repository at https://registry.terraform.io/providers/databricks/databricks/latest/docs/guides/aws-workspace
│ 
│   with databricks_mws.databricks_mws_permission_assignment.test_group,
│   on permissions.tf line 11, in resource "databricks_mws_permission_assignment" "test_group":
│   11: resource "databricks_mws_permission_assignment" "test_group" {
│ 
```
This always fails on initial apply, but after that it occurs sporadically. It seems that for some reason it takes the logic from workspace-level provider, eventhough that account-level provider is specified (`databricks.mws`).

### Steps to Reproduce

1. `terraform apply`

### Terraform and provider versions

```
Terraform v1.3.4
on darwin_arm64
+ provider registry.terraform.io/databricks/databricks v1.6.5
```

### Important Factoids

Nothing atypical about the Databricks account. As mentioned though, this occurs only at initial apply and later on sporadically.

BostjanBozic:
[...]

Issue #1832, opened 2022-12-09
[ISSUE] Issue with `databricks_metastore_assignment` resource - No metastore assigned for the current workspace


alias-santi:
#### providers.tf
```hcl
terraform {
  required_providers {
    databricks = {
      source = "databricks/databricks"
    }
    aws = {
      source  = "hashicorp/aws"
      version = ">= 3.34"
    }
  }
}

provider "databricks" {
  alias      = "mws"
  host       = "https://accounts.cloud.databricks.com"
  account_id = "<account_id>"
  username   = "<username>"
  password   = "<password>"
}

# https://discuss.hashicorp.com/t/databricks-unity-catalog-account-vs-workspace-level-understanding/42570#:~:text=Unity%20Catalog%20API,as%20the%20host.
# Per the above link, workspace level provider is currently needed for creating unity resources
# this will most likely be fixed in api2.1 at which point refactors will be required (to use the mws account level provider only)
provider "databricks" {
  alias    = "workspace"
  host       = "<workspace_url>"
  username   = "<username>"
  password   = "<password>"
}
```
#### data.tf
```hcl
#-------------------------------------------
# data.tf
# retrieve common data sources in AWS account
#--------------------------------------------

# current AWS account id
data "aws_caller_identity" "current" {}

#  current AWS account_alias
#data "aws_iam_account_alias" "current" {}

# Get AWS AZ zones
data "aws_availability_zones" "available" {}

# get current AWS provider region
data "aws_region" "current" {}

# Groups are automatically added by SCIM at an account level so we need to pull
# them in as data blocks here.
data "databricks_group" "data_engineer" {
  provider = databricks.mws
  display_name = "data-engineer"
}

data "databricks_group" "devops_admin" {
  provider = databricks.mws
  display_name = "devops-admin"
}

data "databricks_group" "data_scientist" {
  provider = databricks.mws
  display_name = "data-scientist"
}
```
#### locals.tf
```hcl
#-----------------------------------
# locals.tf
#-----------------------------------

locals {

    create_layer = var.create_layer
    common_tags = {
        terraform = "true"
    }

    region_short  = join("",[for r in split("-", data.aws_region.current.name) : substr(r,0,1)])
    account_alias_number = "sso001"

    bucket_name           = "<bucket_name>"
    bucket_acl            = "private"
    bucket_grant          = []
    bucket_lifecycle_rule = []
    
    uc_role_name = "<aws_role_name>"
[...]

PR #1900, opened 2023-01-03
fix metastore assignment

guillaumestevens:
Link to issue #1768
Fix the object type return by metastore_summary API call.


nkvuong:
@guillaumestevens this fix makes sense to me, but how is it linked to #1768?

guillaumestevens:
oups my bad. 
I think it's link to #1832
Sorry for that.

nkvuong (changes requested):
@guillaumestevens could you also fix the return type of the `metastore_summary` API call in the metastore_assignment tests?

guillaumestevens:
I've just updated the test file.

nkvuong (approved):
looks good to me

alexott:
@nfx one more approved PR

PR #1901, opened 2023-01-04
add search by name for `databricks_cluster` data source

alexott:
Also expose `id` for `databricks_cluster` & `databricks_job` data sources so it's easier to use results.

This is a follow up for #1885


nkvuong (approved):
lgtm

nfx (changes requested):

nfx (approved):

PR #1905, opened 2023-01-05
Exporter: use `databricks_group_role` instead of deprecated `databricks_group_instance_profile`

alexott:


exporter/importables.go
```diff
@@ -190,11 +190,11 @@ var resourcesMap map[string]importable = map[string]importable{
 			return splits[len(splits)-1]
 		},
 	},
-	"databricks_group_instance_profile": {
+	"databricks_group_role": {
 		Service: "access",
 		Depends: []reference{
 			{Path: "group_id", Resource: "databricks_group"},
-			{Path: "instance_profile_id", Resource: "databricks_instance_profile"},
+			{Path: "role", Resource: "databricks_instance_profile"},
```
nfx, line 197: should it match the `role` attribute of instance profile, instead of the `id`?..

alexott: added explicit attribute, although it's the same as `id`

nfx (approved):

PR #1908, opened 2023-01-05
Exporter: export permissions for `databricks_notebook` & `databricks_directory`

### LINKED ISSUE #1850: [ISSUE] Issue with `databricks_permissions` resource. Permissions are missing on notebooks & directories
When `access` is added to the list of services, the user of the exporter reports that permission on notebook and directories are not created as a `resource databricks_permission` terraform resource block.

### Expected Behavior
Permissions granted to users other than the owner / author of the notebook should be reflected in the exported output

### Actual Behavior
Permission resource blocks are missing (at least for notebook and directories) and for other objects, also.

qili86:
Tested 
For notebook, 
it generates the permission block in the `access.tf`, an example is:
`resource "databricks_permissions" "notebook_users_qi_li_mckesson_com_tf_idf_test_3534645884600778" {
  notebook_id = "3534645884600778"
  access_control {
    user_name        = "XXXXXXXX"            (replace the real email)
    permission_level = "CAN_READ"
  }
}`

in `import.sh`, it generates a new line:
`terraform import databricks_permissions.notebook_users_qi_li_mckesson_com_tf_idf_test_3534645884600778 "/notebooks/3534645884600778"`

For directory:
it generates the permission block in the `access.tf`, an example is:
`resource "databricks_permissions" "directory_users_qi_li_mckesson_com_directory_test_2864623526846789" {
  directory_id = "2864623526846789"
  access_control {
    user_name        = "XXXXXXXX". (replace the real email)
    permission_level = "CAN_READ"
  }
}`

in `import.sh`, it generates a new line:
`terraform import databricks_permissions.directory_users_qi_li_mckesson_com_directory_test_2864623526846789 "/directories/2864623526846789"`

This fixes #1850


qili86:
the test fails, will debug more on that

exporter/importables.go
```diff
@@ -1133,6 +1133,17 @@ var resourcesMap map[string]importable = map[string]importable{
...
 			fileName, err := ic.createFileIn("notebooks", name, []byte(content))
+			splits := strings.Split(r.Name, "_")
+			notebookId := splits[len(splits)-1]
+
+			if ic.meAdmin {
+				ic.Emit(&resource{
+					Resource: "databricks_permissions",
+					ID:       fmt.Sprintf("/notebooks/%s", notebookId),
+					Name:     "notebook_" + ic.Importables["databricks_notebook"].Name(ic, r.Data),
+				})
```
alexott, line 1144: Main problem here is that it will generate too many Terraform resources that aren't necessary, and this may lead to very slow execution. Otherwise it would be implemented long timeago

exporter/importables.go
```diff
@@ -1480,4 +1491,52 @@ var resourcesMap map[string]importable = map[string]importable{
...
+					Resource: "databricks_directory",
+					ID:       directory.Path,
+				})
+				if offset%50 == 0 {
+					log.Printf("[INFO] Scanned %d of %d directories",
+						offset+1, len(directoryList))
+				}
+			}
+			return nil
+		},
```
alexott, line 1526: This is not necessary. It's just enough to emit `databricks_directory` from `databricks_notebook` import

exporter/importables.go
```diff
@@ -1480,4 +1491,52 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			splits := strings.Split(r.Name, "_")
+			directoryId := splits[len(splits)-1]
+
+			if ic.meAdmin {
+				ic.Emit(&resource{
+					Resource: "databricks_permissions",
+					ID:       fmt.Sprintf("/directories/%s", directoryId),
+					Name:     "directory_" + ic.Importables["databricks_directory"].Name(ic, r.Data),
+				})
+			}
```
alexott, line 1538: Same as for notebooks - too many not necessary resources.

workspace/resource_notebook.go
```diff
@@ -161,6 +161,35 @@ func (a NotebooksAPI) recursiveAddPaths(path string, pathList *[]ObjectStatus) e
 	return err
 }
 
+func (a NotebooksAPI) ListDirectories(path string, recursive bool) ([]ObjectStatus, error) {
```
alexott, line 164: This may not be necessary if you emit directory from notebooks

exporter/importables.go
```diff
@@ -1480,4 +1491,52 @@ var resourcesMap map[string]importable = map[string]importable{
 			{Path: "library.whl", Resource: "databricks_dbfs_file", Match: "dbfs_path"},
 		},
 	},
+	"databricks_directory": {
+		Service: "directories",
```
alexott, line 1495: directories belong to workspace service, no need to introduce new service type.

alexott (changes requested):
see comments. Also, tests are completely missing for new functionality

exporter/importables.go
```diff
@@ -1480,4 +1491,52 @@ var resourcesMap map[string]importable = map[string]importable{
...
+					ID:       directory.Path,
+				})
+				if offset%50 == 0 {
+					log.Printf("[INFO] Scanned %d of %d directories",
+						offset+1, len(directoryList))
+				}
+			}
+			return nil
+		},
+		Import: func(ic *importContext, r *resource) error {
```
alexott, line 1527: Import won't be called if you don't do emit on directory (not listing)

alexott (changes requested):

alexott:
Another issue - you can't expose user's directories or something like `/Users` as resources - they already exist

exporter/importables.go
qili86: if we don't separate them out, users won't have the option to only export notebook or directory?

exporter/importables.go
qili86: got it, do we have solutions for this?

qili86:
> Another issue - you can't expose user's directories or something like `/Users` as resources - they already exist

Can you explain more on this? thanks!

qili86:
> see comments. Also, tests are completely missing for new functionality

will add tests after we finalize the right solution, thanks!

exporter/importables.go
Tagar: my 2 cents, perhaps we can have an option to export permissions and this wouldn't be on by default?

alexott: Not necessary to have a separate option, just need to do incremental export

exporter/importables.go
alexott: Emitting of only directories could be strange requirement - who will need a directory hierarchy without any content?

qili86: maybe for auditing purpose, especially when the parent directory permission ACLs are different from children notebooks'?

alexott: it should be either covered by the difference of directory permissions from its parent, or permissions on individual notebooks.

P.S. You may see that not all resources have a `List` method, for example, `databricks_user`, or `databricks_dbfs_file` - in most cases they will be covered by emitting from other resources

qili86:
@alexott , apologies that we don't know that you started working on this issue, if you think it is easier for you to implement this, feel free to close this PR, and change the owner in the following spreadsheet, thank you!

https://docs.google.com/spreadsheets/d/1IA8oBNczSPrt2NmATbY8X0V7JhIpm2oKPTWY_4HfxsM/edit#gid=0



> In general, it would be less work if the intended design would be first coordinated/discussed, as I was already starting working on this issue.
> 
> To the rest of the comments:
> 
> * Some directories, like, `/Users`, `/`, and user's directories need to be either not exported as they are always exist, or should be exported as references to user's home, etc. - like it's already done with reference handling in notebooks, ...
> * Permissions on notebooks & directories need to be exported only if they are different from the parent's resource permissions. In this case we won't generate too many resources and this will solve main issue with the current approach
> * Directories need to be exported only their permissions are different from parent (this could be done differently - either by emitting all directories and then filtering them out later, or by emitting them only if permissions are different)
> * Specific directories could be exported by using `-match` argument for notebooks.

exporter/importables.go
qili86: what's the issue here?

exporter/importables.go
qili86: actually, for notebook permission, only permissions granted to users other than the owner/author of the notebook are exported (might need your advice on how this works! )

qili86:
> In general, it would be less work if the intended design would be first coordinated/discussed, as I was already starting working on this issue.
> 
> To the rest of the comments:
> 
> * Some directories, like, `/Users`, `/`, and user's directories need to be either not exported as they are always exist, or should be exported as references to user's home, etc. - like it's already done with reference handling in notebooks, ...
> * Permissions on notebooks & directories need to be exported only if they are different from the parent's resource permissions. In this case we won't generate too many resources and this will solve main issue with the current approach
> * Directories need to be exported only their permissions are different from parent (this could be done differently - either by emitting all directories and then filtering them out later, or by emitting them only if permissions are different)
> * Specific directories could be exported by using `-match` argument for notebooks.

exporter/exporter_test.go
```diff
@@ -460,6 +460,11 @@ func TestImportingUsersGroupsSecretScopes(t *testing.T) {
 				Resource: "/api/2.0/secrets/acls/get?principal=users&scope=a",
 				Response: secrets.ACLItem{Permission: "READ", Principal: "users"},
 			},
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/workspace/list?path=%2F",
+				Response: workspace.ObjectList{},
+			},
```
alexott, line 467: You can make a variable out of this - similar to `emptyRepos` and other - there will be less code

nfx:
@qili86 any updates? :)

qili86:
> @qili86 any updates? :)

@alexott, could you do a final review, let me know what else we need to improve, thank you!

alexott (approved):
lgtm

nfx (approved):

PR #1925, opened 2023-01-14
Fixed `databricks_cluster` resource missing library definitions issues

ukannikanti:


libraries/libraries_api.go
```diff
@@ -231,6 +231,10 @@ func (library Library) String() string {
 	return "unknown"
 }
 
+func (library Library) GetID(clusterId string) string {
+	return fmt.Sprintf("%s/%s", clusterId, library.String())
+}
+
```
alexott, line 237: We need a test for it in the `libraries` package, otherwise the code coverage will be decreasing (Go's code coverage tool calculates code coverage per each package, so even if there is a usage in another package it's not taken into account)

alexott (changes requested):
Minor changes are required to prevent decrease in code coverage...

Also, check that all code in `importables.go` and `utils.go` is covered by tests - not only "happy path" but also error handling

alexott (approved):
Code wise it looks good...

nfx (approved):

Issue #1743, opened 2022-11-09
[ISSUE] Issue with `databricks_sql_endpoint` resource


eczerwin:
### Configuration
resource "databricks_sql_endpoint" "this" {
  name             = "test"
  cluster_size     = "2X-Small"
  max_num_clusters = 1
  min_num_clusters = 1
  num_clusters     = 1
  auto_stop_mins   = 30

### Expected Behavior
Terraform should apply successfully and then on new plan not find any changes

### Actual Behavior
 # databricks_sql_endpoint.this will be updated in-place
  ~ resource "databricks_sql_endpoint" "this" {
        id                        = "21fa58321fa1c6a7"
        name                      = "azadbengcloud01"
      ~ num_clusters              = 0 -> 1
        # (10 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

### Steps to Reproduce
run a terraform plan and apply and the behavior repeats

### Terraform and provider versions
Terraform v1.2.5
on windows_amd64
+ provider registry.terraform.io/databricks/databricks v1.6.3
+ provider registry.terraform.io/hashicorp/azuread v2.30.0
+ provider registry.terraform.io/hashicorp/azurerm v3.30.0
+ provider registry.terraform.io/hashicorp/time v0.9.1

fava9410:
Totally agree with this one, for me the `num_clusters` shouldnt be pass because this is like the active cluster in that moment, with the min and max should be enough

eczerwin:
I see this was considered config drift...  However, I have noted it comes down simply to if the cluster is running or not.  At time of creation it is for sure running but since it is configurable timeout..  It cannot always be running during a plan and terraform does not start it on interaction

Israphel:
I'd recommend ignoring that line in your module (at least for now)

```terraform
# Databricks SQL Endpoint
resource "databricks_sql_endpoint" "sql_endpoint" {
  name                      = var.name
  cluster_size              = var.cluster_size
  min_num_clusters          = var.min_num_clusters
  max_num_clusters          = var.max_num_clusters
  auto_stop_mins            = var.auto_stop_mins
  spot_instance_policy      = var.spot_instance_policy
  enable_serverless_compute = var.enable_serverless_compute

  dynamic "tags" {
    for_each = var.tags != null ? [var.tags] : []
    content {
      dynamic "custom_tags" {
        for_each = var.tags
        content {
          key   = custom_tags.key
          value = custom_tags.value
        }
      }
[...]

PR #1927, opened 2023-01-16
Fixed incorrect diff for `num_clusters` and `min_num_clusters` for stopped `databricks_sql_endpoint`

### LINKED ISSUE #1204: [ISSUE] Incorrect diff for num_clusters in SQL Endpoint when endpoint is not running
Hi,

### Configuration
```hcl
   name             = "Terraform-testing Endpoint"
   cluster_size     = "Large"
   min_num_clusters = 1
   max_num_clusters = 2
   auto_stop_mins   = 10
 }
```

### Expected Behavior
No change on subsequent applies when endpoint is not running

### Actual Behavior
image

### Steps to reproduce
create the endpoint
**turn it off**
`terraform apply` again

provider 0.5.4

### Notes
Already reported in https://github.com/databrickslabs/terraform-provider-databricks/issues/1116 with attached fix: https://github.com/databrickslabs/terraform-provider-databricks/pull/1117

however the fix does not seem to work

nfx:
Correctly suppress diff for `num_clusters` on warehouses that are not running, in case `num_clusters` is not specified.

Note for contributors: `tf:"suppress_diff"` and `tf:"default"` don't work well together.

Fixes #1204 #1743


sql/resource_sql_endpoint.go
```diff
@@ -27,7 +27,7 @@ type SQLEndpoint struct {
 	AutoStopMinutes         int             `json:"auto_stop_mins" tf:"optional,default:120"`
 	MinNumClusters          int             `json:"min_num_clusters,omitempty" tf:"default:1,suppress_diff"`
```
alexott, line 28: Should we do the same for `MinNumClusters` ?

alexott (changes requested):
tests are failing :-)
plus we need to do the same for `MinNumClusers`.

Plus, maybe it makes sense to add `.AtLeastOneOf`  to `num_clusers` and `min_num_clusters`?

nfx:
@alexott having a sound CI system makes me sometimes forget running tests locally. 

Added the same fix for `min_num_clusters`. not really willing to `. AtLeastOneOf` because the validation should happen on the backend.

alexott (approved):
lgtm

PR #1933, opened 2023-01-17
Remove "computed" annotation for "aws_attributes"

pietern:
This annotation means the field shows up in plans for Azure and GCP, where it doesn't make sense.

It was added in #1463. Not clear if it was intentional at that time.


pietern:
I confirmed that after removing the annotation, `aws_attributes` no longer shows up in the plans for Azure and GCP.

nfx (approved):
lgtm, if pr-manual checks pass on azure-prod and gcp-prod

alexott (approved):
lgtm

pietern:
Integration tests passed.

PR #1935, opened 2023-01-18
Remove superfluous err argument in NoError test function

pietern:
I noticed the duplicate message in an assertion failure. This is caused by passing `err` as message to `NoError`.

```
    resource_token_test.go:272: 
        	Error Trace:	<...>/tokens/resource_token_test.go:272
        	Error:      	Received unexpected error:
        	            	Post "***/api/2.0/token/create": context deadline exceeded [...]
        	Test:       	TestAccCreateToken
        	Messages:   	Post "***/api/2.0/token/create": context deadline exceeded [...]
```


storage/azure_blob_mount_test.go
```diff
@@ -52,7 +52,7 @@ func TestResourceAzureBlobMountCreate(t *testing.T) {
 		},
 		Create: true,
 	}.Apply(t)
-	require.NoError(t, err, err) // TODO: global search-replace for NoError
```
pietern, line 55: Note: someone before me may have meant to change this before.

alexott (approved):
lgtm

access/resource_ip_access_list_test.go
```diff
@@ -109,7 +109,7 @@ func TestIPACLCreate(t *testing.T) {
 		},
 		Create: true,
 	}.Apply(t)
```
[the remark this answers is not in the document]
pietern: Modified everything where applicable to use `ApplyNoError`.

Had to update the return type to `*schema.ResourceData` to accommodate.

qa/testing.go
```diff
@@ -256,16 +256,17 @@ func (f ResourceFixture) requiresNew(diff *terraform.InstanceDiff) error {
 	return nil
 }
 
-// ApplyNoError is a convenience method for no-data tests
-func (f ResourceFixture) ApplyNoError(t *testing.T) {
-	_, err := f.Apply(t)
-	assert.NoError(t, err, err)
+// ApplyNoError is a convenience method that asserts no error occurred.
+func (f ResourceFixture) ApplyNoError(t *testing.T) *schema.ResourceData {
```
nfx, line 260: Can we not add the `*schema.ResourceData` as the return type?.. we have `ApplyAndExpectData` to assert for data in a more readable way.

The main intent of `ApplyNoError` is to make tests more readable. I want to prevent unreadable tests proliferate with copy-paste

nfx (changes requested):
please don't add `*schema.ResourceData` to `ApplyNoError`. We have `ApplyAndExpectData` for resource data assertions

qa/testing.go
pietern: * `Apply` returns `d, err`
* `ApplyNoError` returns `d`
* `ApplyAndExpectData` returns nothing

Seems reasonable to me.

What about this approach is not / less readable?

pietern: I reverted back to the original intent and approach of this PR; only replace `Error(t, err, err)` and `NoError(t, err, err)`.

pietern:
@nfx Output of `git grep "err, err"` doesn't contain these anymore.

nfx (approved):

PR #1943, opened 2023-01-20
Added support for `iam_role_arn` in `databricks_instance_profile`

### LINKED ISSUE #1710: [FEATURE] Add iam_role_arn for databricks_group_instance_profile resource
### Use-cases

In noticeable amount of cases instance profile name and associated IAM role have different names. Currently `databricks_group_instance_profile` resource supports only `instance_profile_id` input and based on the profile name it generates associated role with the same name. 

### Attempted Solutions

Set up correct association within Admin UI on Databricks instance profiles page. But it's not acceptable in IaC paradigm.

### Proposal

Add associated IAM role ARN input into the `databricks_group_instance_profile` resource.

### LINKED ISSUE #1911: [FEATURE] New feature request :
IHAC who wants to use IAM role ARN argument also while creating the instance profile through Terraform. 

They are making the IAM role through command line with --path option in AWS CLI and attaching Instance profile with the role. The IP and Role ARN are different in names. So they wanted to create instance profile with these values. They are able to do it in the UI but not through Terraform.

Jorge-Rodriguez:
This PR adds support to manage an instance profile's IAM role ARN when it does not match the instance profile name. This feature is needed when managing instance profiles for use with Databricks SQL serverless.

See [Databricks REST API documentation for instance profiles](https://docs.databricks.com/dev-tools/api/latest/instance-profiles.html#add)

Fixes #1911 
Fixes #1710

### DIFF
--- a/aws/resource_instance_profile.go
+++ b/aws/resource_instance_profile.go
@@ -18,7 +18,7 @@ import (
 
 // InstanceProfileInfo contains the ARN for aws instance profiles
 type InstanceProfileInfo struct {
-	InstanceProfileArn    string `json:"instance_profile_arn,omitempty"`
+	InstanceProfileArn    string `json:"instance_profile_arn"`
 	IamRoleArn            string `json:"iam_role_arn,omitempty"`
 	IsMetaInstanceProfile bool   `json:"is_meta_instance_profile,omitempty"`
 	SkipValidation        bool   `json:"skip_validation,omitempty" tf:"computed"`
--- a/aws/resource_instance_profile.go
+++ b/aws/resource_instance_profile.go
@@ -86,16 +86,14 @@ func (a InstanceProfilesAPI) Delete(instanceProfileARN string) error {
 
 // Update updates the IAM role ARN of an existing instance profile
 func (a InstanceProfilesAPI) Update(ipi InstanceProfileInfo) error {
-	return a.client.Post(a.context, "/instance-profiles/edit", map[string]any{
-		"instance_profile_arn": ipi.InstanceProfileArn,
-		"iam_role_arn": func(ipi InstanceProfileInfo) string {
-			if ipi.IamRoleArn == "" {
-				return ipi.InstanceProfileArn
-			} else {
-				return ipi.IamRoleArn
-			}
-		}(ipi),
-	}, nil)
+	data := map[string]any{
+		"inatance_profile_arn": ipi.InstanceProfileArn,
+		"iam_role_arn":         ipi.InstanceProfileArn,
+	}
+	if ipi.IamRoleArn != "" {
+		data["iam_role_arn"] = ipi.IamRoleArn
+	}
+	return a.client.Post(a.context, "/instance-profiles/edit", data, nil)
 }
 
 // IsRegistered checks if instance profile exists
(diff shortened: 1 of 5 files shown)


Jorge-Rodriguez:
Please note that this is my first stab at writing anything in go, so feel free to consider this PR just as a PoC for the intended support feature if there is anything fundamentally wrong otherwise.

Jorge-Rodriguez:
ping @alexott

aws/resource_instance_profile.go
```diff
@@ -19,6 +19,7 @@ import (
 // InstanceProfileInfo contains the ARN for aws instance profiles
 type InstanceProfileInfo struct {
 	InstanceProfileArn    string `json:"instance_profile_arn,omitempty"`
+	IamRoleArn            string `json:"iam_role_arn,omitempty" tf:"optional"`
```
alexott, line 22: `tf:"optional` is not needed if you have `omitempty` in `json`

aws/resource_instance_profile.go
```diff
@@ -138,6 +147,7 @@ func ResourceInstanceProfile() *schema.Resource {
 	instanceProfileSchema := common.StructToSchema(InstanceProfileInfo{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			m["instance_profile_arn"].ValidateDiagFunc = ValidInstanceProfile
+			m["iam_role_arn"].ValidateDiagFunc = ValidIamRole
```
alexott, line 150: it should allow empty string as well if it's optional

alexott (changes requested):

alexott:
Also, add documentation

aws/resource_instance_profile.go
Jorge-Rodriguez: Do you mean the `ValidIamRole` validation function? And allowing empty string as opposed to `null` or missing parameter?

alexott: when string initialized it's getting initialized to empty string, not to `null`

aws/resource_instance_profile.go
```diff
@@ -83,6 +84,20 @@ func (a InstanceProfilesAPI) Delete(instanceProfileARN string) error {
...
+	return a.client.Post(a.context, "/instance-profiles/edit", map[string]any{
+		"instance_profile_arn": ipi.InstanceProfileArn,
+		"iam_role_arn": func(ipi InstanceProfileInfo) string {
+			if ipi.IamRoleArn == "" {
+				return ipi.InstanceProfileArn
+			} else {
+				return ipi.IamRoleArn
+			}
+		}(ipi),
+	}, nil)
```
alexott, line 98: This could be made much simpler:

```golang
data := map[string]any{"instance_profile_arn": ipi.InstanceProfileArn,}
if ipi.IamRoleArn != "" {
  data["iam_role_arn"]=ipi.IamRoleArn
}
return a.client.Post(a.context, "/instance-profiles/edit", data, nil)
```

alexott (changes requested):

aws/resource_instance_profile.go
Jorge-Rodriguez: Yeah, the assignment to data is clearer, but it still needs the `else` clause.

Jorge-Rodriguez: Modified, but left the initial assignment to `data["iam_role_arn"]` since the argument is required for the `edit` endpoint.

alexott (approved):
lgtm. Let check with the PR runner

Jorge-Rodriguez:
> lgtm. Let check with the PR runner

There seems to be a typo

Jorge-Rodriguez:
Sure. That'll handle the DRY issue as well

alexott:
One thing that is different between instance profile & role validation is that we're allowing empty role, while instance profile shouldn't allow it

Jorge-Rodriguez:
> One thing that is different between instance profile & role validation is that we're allowing empty role, while instance profile shouldn't allow it

Oh, I got confused by the fact that InstanceProfileArn is marked as `omitempty`

In other matters, I don't seem to be able to write a test for `if !ok {`, I can't think of a configuration value that wouldn't get marshalled into a string without throwing a panic before even starting the test.

alexott:
Hmmm, interesting - `omitempty` is wrong here - [doc](https://docs.databricks.com/dev-tools/api/latest/instance-profiles.html) says that this field is required so `omitempty` should be removed

aws/resource_instance_profile.go
```diff
@@ -182,6 +201,18 @@ func ValidInstanceProfile(v any, c cty.Path) diag.Diagnostics {
...
 	}
+	var arnType string
+	switch c[0].(cty.GetAttrStep).Name {
+	case "instance_profile_arn":
+		arnType = "instance-profile"
+	case "instance_profile_id":
+		arnType = "instance-profile"
+	case "iam_role_arn":
+		arnType = "role"
+	}
```
alexott, line 212: should we have `default` here?

Jorge-Rodriguez: Yeah, Otherwise the check down the line is against an empty string and whatever failure message comes out of it would be quite useless.

alexott (approved):
lgtm

@nfx - please look & merge

nfx (approved):

PR #1956, opened 2023-01-25
Fix `max_clusters_per_user` to `databricks_cluster_policy` resource

alexott:
:facepalm: we need to get rid from the non-struct schemas


iandexter (approved):
lgtm

nfx (approved):

PR #1959, opened 2023-01-26
Add description to `max_clusters_per_user` in policies schema. Re-format docs.

iandexter:


docs/resources/cluster_policy.md
```diff
@@ -99,7 +99,7 @@ The following arguments are required:
 
 * `name` - (Required) Cluster policy name. This must be unique. Length must be between 1 and 100 characters.
 * `definition` - (Required) Policy definition: JSON document expressed in [Databricks Policy Definition Language](https://docs.databricks.com/administration-guide/clusters/policies.html#cluster-policy-definition).
-* `max_clusters_per_user` - (Optional, integer) Maximum number of clusters allowed per user. When omitted, there is no limit.
+* `max_clusters_per_user` - (Optional, integer) Maximum number of clusters allowed per user. When omitted, there is no limit. If specified, value must be non-zero.
```
alexott, line 102: it's better to say "greater than zero" :-)

policies/resource_cluster_policy.go
```diff
@@ -125,6 +125,9 @@ func ResourceClusterPolicy() *schema.Resource {
 			"max_clusters_per_user": {
 				Type:     schema.TypeInt,
 				Optional: true,
+				Description: "Max number of clusters per user that can be active\n" +
+				    "using this policy. If not present, there is no max limit.",
+				ValidateFunc: validation.IntAtLeast(1),
```
alexott, line 130: does it work when omitted?

alexott (changes requested):
exclude `go.mod` & `go.sum` from PR. We need to test that it allows 0 value (when it's omitted).

You can easily test by building it and specifying dev override like described [here](https://developer.hashicorp.com/terraform/cli/config/config-file#development-overrides-for-provider-developers) or [here](https://nicovibert.com/2021/03/29/dev-overrides-terraform/)

docs/resources/cluster_policy.md
iandexter: Changed.

policies/resource_cluster_policy.go
iandexter: Yes, I've tested it with and without the parameter.

alexott (approved):
lgtm

PR #1965, opened 2023-01-27
Removed check for creation of Azure Key Vault-based `databricks_secret_scope` using Service Principals

mattiazenidb:
This PR removes the check that was forbidding to call this from a Service Principal since it was not working. Now that MSFT fixed it we can remove the check and allow customers use it.


secrets/resource_secret_scope.go
```diff
@@ -122,10 +118,6 @@ func kvDiffFunc(ctx context.Context, diff *schema.ResourceDiff, v any) error {
 	if len(kvLst) == 0 {
 		return nil
 	}
-	client := v.(*common.DatabricksClient)
```
alexott, line 125: We can remove the whole `kvDiffFunc` function - it was necessary only to check if service principal is used or not

alexott (changes requested):
we need to remove `kvDiffFunc` completely

alexott (approved):
lgtm. Integration tests are also passing

fyi @nfx

Issue #1951, opened 2023-01-23
[DOC] Clarify support for using dedicated SQL Warehouses for SQL jobs


fuselessmatt:
According to the [official documentation](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/job#sql_task-configuration-block), it is only possible to define a SQL warehouse using serverless warehouses.

> [warehouse_id](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/job#warehouse_id) - (Required) ID of the (the [databricks_sql_endpoint](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/sql_endpoint)) that will be used to execute the task. Only serverless warehouses are supported right now.

In the UI, it is however possible to use any SQL Pro warehouse when you create a new SQL job. Why not add support for that in the provider?

> Classic warehouses are unavailable for this schedule
> Only serverless or SQL Pro warehouses are supported. [Learn more](https://docs.databricks.com/administration-guide/account-settings/workload-types.html).

### Use-cases

Have terraformed SQL jobs
### Attempted Solutions

I haven't actually tried to ignore the warning and simply give it a SQL warehouse pro, but I suppose the warning is there for a reason. In the meanwhile I tried to add my SQL jobs manually and that works great.

alexott:
It's more documentation issue

fuselessmatt:
@alexott : I thought I would give it a try before sending a PR with this correction, but I now fail on defining an sql_task

```hcl
resource "databricks_sql_query" "firebase_ingestion" {
  data_source_id = databricks_sql_endpoint.ingestion_sql_warehouse.id
  name           = "Firebase Ingestion SQL Task"
  query          = file("job_files/firebase.sql")
}

resource "databricks_job" "load_firebase" {
  name = "Firebase Ingestion Job"
  sql_task {
    warehouse_id = databricks_sql_endpoint.ingestion_sql_warehouse.id
    query = databricks_sql_query.firebase_ingestion.id
  }
}
```

>Error: Unsupported block type
> 
>   on custom_jobs.tf line 11, in resource "databricks_job" "load_firebase":
>   11:   sql_task {
> 
> Blocks of type "sql_task" are not expected here.

alexott:
[...]

PR #1967, opened 2023-01-29
Added SQL example for `databricks_job` documentation

fuselessmatt:
I found this advanced Terraform syntax a bit confusion, so I think it would be nice to give an example would it easier. 

We could alternatively also add it as a task in the main example? Is it possible to combine it with other types? I've only tried it as a stand separate thing.

Another option would be to include how to add in an job it in the [databricks_sql_query docs](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/sql_query).

https://github.com/databricks/terraform-provider-databricks/issues/1951


alexott (approved):
lgtm

PR #1969, opened 2023-01-30
Move tags to default_tags in AWS provider

KneeShard:
Default tags on AWS resources can be specified on the provider as of v3.38.0, see https://www.hashicorp.com/blog/default-tags-in-the-terraform-aws-provider

Tag specifications on individual resources can then either be removed or simplified (e.g. no need to merge tags).


alexott (approved):
lgtm

PR #1970, opened 2023-01-30
Remove deprecated params from bucket

KneeShard:
The `acl` and `versioning` parameters are deprecated on the `aws_s3_bucket` resource. The docs recommend using `aws_s3_bucket_acl` and `aws_s3_bucket_versioning` resources instead. From what I can gather, this helps with drift detection in Terraform.


alexott (approved):
lgtm

Issue #1986, opened 2023-02-07
[DOC] Correct identifier of data source "databricks job" is "job_id" and not "id"


guillesd:
### Affected Resource(s)
In https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/job it is specified in the Attribute reference that: "[id](https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/job#id) - the id of [databricks_job](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/job) if the resource was matched by name."

### Expected Details
This is actually not true, since `id` just returns "-". The actual id of the job is `job_id`.

### List of things to potentially add/remove:
Just change `id` for `job_id` in the docs.

iandexter:
I think what this means is the ID to reference when **importing** a `databricks_job` resource? Can you elaborate what you mean with an example, @guillesd ? Thanks!

guillesd:
sure, what I mean is, when you have a databricks job resource, i.e.:
```hcl
data "databricks_job" "this" {
  job_name = "My job"
}
```
and then let's say you want to explore what is the actual content of this data source, so you create an output to explore easier:
```hcl
output "my_databricks_job" {
  value = data.databricks_job.this
}
```
This is what `data.databricks_job.this` provides as an output:
```
      + my_job  = {
          + id           = "_"
          + job_id       = "764062641151921"
          + job_name     = "My job"
          + job_settings = [
              + {
                  + created_time      = 1673259719254
                  + creator_user_name = "9f4d197a-a483-4659-adbf-51f52d693f5d"
                  + job_id            = 764062641151921
                  + settings          = [
                      + {
                          + dbt_task                  = []
                          + email_notifications       = [
                              + {
                                  + alert_on_last_attempt     = false
                                  + no_alert_for_skipped_runs = false
                                  + on_failure                = []
                                  + on_start                  = []
                                  + on_success                = []
                                },
                            ]
                          + existing_cluster_id       = ""
                          + format                    = "MULTI_TASK"
                          + git_source                = []
                          + job_cluster               = []
                          + library                   = []
                          + max_concurrent_runs       = 1
                          + max_retries               = 0
                          + min_retry_interval_millis = 0
                          + name                      = "my_job"
                          + new_cluster               = []
                          + notebook_task             = []
                          + pipeline_task             = []
                          + python_wheel_task         = []
                          + retry_on_timeout          = false
                          + schedule                  = []
[...]

PR #2016, opened 2023-02-17
[DOC] Fix job datasource docs

guillesd:
Addresses #1986

- As explained in the issue, in the output of the data source `databricks_job`, the attribute containing the job ID is `job_id` and not `id`. This PR fixes that.
- Additionally, I saw that in the Databricks resource `databricks_job`, in the list of "additionally" exported values, only `url` is listed. Since the job ID is the most important exported attribute (it is needed for any permission that you may want to apply later) I also added `id` to the list of exported attributes. 

**Note:** There is an inconsistency in the job ID naming, since in the data source response the ID is under the attribute `job_id` and in the resource it is under `id`. This is quite confusing.

### DIFF
--- a/docs/data-sources/job.md
+++ b/docs/data-sources/job.md
@@ -27,7 +27,7 @@ output "job_num_workers" {
 This data source exports the following attributes:
 
 
-* `id` - the id of [databricks_job](../resources/job.md) if the resource was matched by name.
+* `job_id` - the id of [databricks_job](../resources/job.md) if the resource was matched by name.
 * `name` - the job name of [databricks_job](../resources/job.md) if the resource was matched by id.
 * `job_settings` - the same fields as in [databricks_job](../resources/job.md).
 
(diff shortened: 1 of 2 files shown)


alexott:
what version have you tested? at least for job by name, the `id` should be filled correctly

guillesd:
Hi @alexott indeed after running it with both 1.5 and 1.9, I saw that in more recent versions this discrepancy has been fixed. Moreover, in the 1.5 docs the attribute for job ID is referenced as `job_id`, so everything is OK. I just looked at the wrong provider version docs. I'm closing this.

Anyhow, my second point regarding `id` not being mentioned in "exported Attributes" of the resource "databricks_job" is still valid I feel like

alexott:
all data sources & resources always have `id`, although maybe it makes sense to explicitly state it - please rework PR to mention only it & I'll approve it

guillesd:
done @alexott :)

alexott (approved):
thank you

Issue #1651, opened 2022-10-03
[FEATURE] add databricks_metastore data source to retrieve an existing metastore


nicolasbailly1:
### Use-cases
Hello,

I need to retrieve an existing metastore to attach it to a new workspace.
Currently, we only can create a new metatstore.

### Proposal

```terraform
data "databricks_metastore" "this" {
  name = "primary"
}

output "metastore_id" {
  value = data.databricks_metastore.this.id
}
```

guillesd:
Hi there! This is much needed for my client... If you want to share metastores among workspace deployments I don't wanna be hardcoding metastore id values!

@nkvuong why do you label this issue as `needs-api`? I checked the Databricks CLI implementation and there's seems to be an endpoint to list metastores https://github.com/databricks/databricks-cli/blob/701959037eac5fa6f168f47af546545609db116e/databricks_cli/unity_catalog/uc_service.py#L44

nkvuong:
@guillesd that endpoint in the CLI implementation is a workspace-specific one, what we need is an account-level endpoint (it should be available some time this month is what we last heard from eng)

guillesd:
@nkvuong but even the provider being used right now to create the metastore is at the workspace level, so I don't see any problem on extending this with the functionality to fetch a metastore as a data source. I'm aware that the idea is to migrate the metastore implementation to the account level APIs but in the meantime this could be quite a quick win in my view

alyssons-db:
This has become an issue for one of my customers as well. It would be great to have a data source for Metastores even if it means using the workspace APIs as @guillesd mentioned

waskelton4:
This feature would really help us out with our TF deployments as well!

scheleaap:
This is an issue for us as well. It would be very helpful to be able to retrieve the metastore instead of having to hard code the ID.

FantomKun:
+1 on behalf of my customer, we have to use http to retrieve the metastore id which is very heavy to operate.

nkvuong:
Closes by #2492

PR #2017, opened 2023-02-17
Feature/add metastore datasource

guillesd:
Addresses #1651

This PR would enable users to retrieve existing metastores, which will allow them to get the needed ID of a Metastore by name. This ID can then be used to assign a metastore to a workspace, without the need of hardcoding an ID for different workspace deployments.

### DIFF
--- a/catalog/data_metastores.go
+++ b/catalog/data_metastores.go
@@ -1,0 +1,37 @@
+package catalog
+
+import (
+	"context"
+
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceMetastores() *schema.Resource {
+	return common.DataResource(MetastoreInfo{}, func(ctx context.Context, e any, c *common.DatabricksClient) error {
+		data := e.(*MetastoreInfo)
+		metastoresAPI := NewMetastoresAPI(ctx, c)
+		metastores, err := metastoresAPI.listMetastores()
+		if err != nil {
+			return err
+		}
+		for _, v := range metastores.Metastores {
+			data.Name = append(data.Name, v.name)
+			data.StorageRoot = append(data.StorageRoot, v.storage_root)
+			data.DefaultDacID = append(data.DefaultDacID, v.default_data_access_config_id)
+			data.Owner = append(data.Owner, v.owner)
+			data.MetastoreID = append(data.MetastoreID, v.metastore_id)
+			data.Region = append(data.Region, v.region)
+			data.Cloud = append(data.Cloud, v.cloud)
+			data.GlobalMetastoreId = append(data.GlobalMetastoreId, v.global_metastore_id)
+			data.CreatedAt = append(data.CreatedAt, v.created_at)
+			data.CreatedBy = append(data.CreatedBy, v.created_by)
+			data.UpdatedAt = append(data.UpdatedAt, v.updated_at)
+			data.UpdatedBy = append(data.UpdatedBy, v.updated_by)
+			data.DeltaSharingScope = append(data.DeltaSharingScope, v.delta_sharing_scope)
+			data.DeltaSharingRecipientTokenLifetimeInSeconds = append(data.DeltaSharingRecipientTokenLifetimeInSeconds, v.delta_sharing_recipient_token_lifetime_in_seconds)
+			data.DeltaSharingOrganizationName = (data.DeltaSharingOrganizationName, v.delta_sharing_organization_name)
+		}
+		return nil
+	})
+}
--- a/catalog/data_metastores_test.go
+++ b/catalog/data_metastores_test.go
@@ -1,0 +1,53 @@
+package catalog
+
+import (
+	"testing"
+
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestMetastoresData(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "GET",
+				Resource: "/api/2.1/unity-catalog/metastores",
+				Response: Metastores{
+					Metastores: []MetastoreInfo{
+						{
+							Name: "a",
+							StorageRoot: "",
+							DefaultDacID: "",
+							Owner: "",
+							MetastoreID: "",
+							Region: "",
+							Cloud: "",
+							GlobalMetastoreId: "",
+							CreatedAt: 0,
+							CreatedBy: "",
+							UpdatedAt: 0,
+							UpdatedBy: "",
+							DeltaSharingScope: "",
+							DeltaSharingRecipientTokenLifetimeInSeconds: 0,
+							DeltaSharingOrganizationName: "",
+						},
+					},
+				},
+			},
+		},
+		Resource:    DataSourceMetastores(),
+		Read:        true,
+		NonWritable: true,
+		ID:          "_",
+	}.ApplyNoError(t)
+}
+
+func TestMetastoresData_Error(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures:    qa.HTTPFailures,
+		Resource:    DataSourceMetastores(),
+		Read:        true,
+		NonWritable: true,
+		ID:          "_",
+	}.ExpectError(t, "I'm a teapot")
+}
--- a/catalog/resource_metastore.go
+++ b/catalog/resource_metastore.go
@@ -43,10 +43,10 @@ type CreateMetastore struct {
 	StorageRoot string `json:"storage_root"`
 }
 
-// func (a MetastoresAPI) listMetastores() (mis []MetastoreInfo, err error) {
-// 	err = a.client.Get(a.context, "/unity-catalog/metastores", nil, &mis)
-// 	return
-// }
+func (a MetastoresAPI) listMetastores() (mis []MetastoreInfo, err error) {
+	err = a.client.Get(a.context, "/unity-catalog/metastores", nil, &mis)
+	return
+}
 
 func (a MetastoresAPI) createMetastore(cm CreateMetastore) (mi MetastoreInfo, err error) {
 	err = a.client.Post(a.context, "/unity-catalog/metastores", cm, &mi)
--- a/docs/data-sources/metastores.md
+++ b/docs/data-sources/metastores.md
@@ -1,0 +1,33 @@
+---
+subcategory: "Unity Catalog"
+---
+# databricks_metastores Data Source
+
+-> **Note** If you have a fully automated setup with workspaces created by [databricks_mws_workspaces](../resources/mws_workspaces.md) or [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace), please make sure to add [depends_on attribute](../index.md#data-resources-and-authentication-is-not-configured-errors) in order to prevent _authentication is not configured for provider_ errors.
+
+Retrieves a list of [databricks_metastore](../resources/metastore.md) objects, that were created by Terraform or manually, so that special handling could be applied.
+
+## Example Usage
+
+Listing all catalogs:
+
+```hcl
+data "databricks_metastores" "all" {}
+
+output "all_metastores" {
+  value = data.databricks_metastores.all
+}
+```
+
+## Attribute Reference
+
+This data source exports the following attributes:
+
+* `metastores` - list of [databricks_metastore](../resources/share.md)
+
+## Related Resources
+
+The following resources are used in the same context:
+
+* [databricks_metastore](../resources/metastore.md) to manage Metastores within Unity Catalog.
+* [databricks_catalog](../resources/catalog.md) to manage catalogs within Unity Catalog.


nkvuong:
@guillesd we should wait until the account API for Unity Catalog is available before implementing this data source.

for this PR, you also need to add the mapping to [provider.go]https://github.com/databricks/terraform-provider-databricks/tree/master/provider/provider.go

guillesd:
Hi @nkvuong! As discussed in the issue #1651, every unity endpoint uses the Workspace API, so i don't see why we shouldn't make this data source available. Even the `databricks_metastore` resource uses the Workspace API

guillesd:
@nkvuong I added the mapping to provider.go

docs/data-sources/metastores.md
```diff
@@ -0,0 +1,33 @@
...
+output "all_metastores" {
+  value = data.databricks_metastores.all
+}
+```
+
+## Attribute Reference
+
+This data source exports the following attributes:
+
+* `metastores` - list of [databricks_metastore](../resources/share.md)
```
alexott, line 26: it's pointing to the wrong file

catalog/data_metastores.go
```diff
@@ -0,0 +1,37 @@
...
+			data.Region = append(data.Region, v.region)
+			data.Cloud = append(data.Cloud, v.cloud)
+			data.GlobalMetastoreId = append(data.GlobalMetastoreId, v.global_metastore_id)
+			data.CreatedAt = append(data.CreatedAt, v.created_at)
+			data.CreatedBy = append(data.CreatedBy, v.created_by)
+			data.UpdatedAt = append(data.UpdatedAt, v.updated_at)
+			data.UpdatedBy = append(data.UpdatedBy, v.updated_by)
+			data.DeltaSharingScope = append(data.DeltaSharingScope, v.delta_sharing_scope)
+			data.DeltaSharingRecipientTokenLifetimeInSeconds = append(data.DeltaSharingRecipientTokenLifetimeInSeconds, v.delta_sharing_recipient_token_lifetime_in_seconds)
+			data.DeltaSharingOrganizationName = (data.DeltaSharingOrganizationName, v.delta_sharing_organization_name)
```
alexott, line 33: How the data is look like in use? I don't see where do you expose the `metastores` attribute...

catalog/data_metastores_test.go
```diff
@@ -0,0 +1,53 @@
...
+						},
+					},
+				},
+			},
+		},
+		Resource:    DataSourceMetastores(),
+		Read:        true,
+		NonWritable: true,
+		ID:          "_",
+	}.ApplyNoError(t)
```
alexott, line 42: Also, add a check for the data returned

alexott (changes requested):

nfx:
SDK is merged in #1848 , please update code accordingly

guillesd:
@alexott I fixed:
- test/linting errors (sorry for this, my bad)
- wrong reference in docs
I also added:
- A specific data test

@nfx not sure what you mean by "SDK is merged update code accordingly"

nfx:
@guillesd it means that you have to use https://pkg.go.dev/github.com/databricks/databricks-sdk-go@v0.3.2/service/unitycatalog#MetastoresAPI 

e.g.

```go
w, err := c.WorkspaceClient()
..
w.Metastores....
```

guillesd:
ok @nfx, so if I understand correctly then all the API functions with this type of structure:
```go
func (a MetastoresAPI) listMetastores() (mis Metastores, err error) {
	err = a.client.Get(a.context, "/unity-catalog/metastores", nil, &mis)
	return
}
```
will be replaced with their SDK counterpart?

nfx:
> will be replaced with their SDK counterpart?

@guillesd that's correct

nfx:
@guillesd all tests must pass

catalog/data_metastores.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/unitycatalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceMetastores() *schema.Resource {
+	type MetastoresData struct {
+		Metastores []unitycatalog.MetastoreInfo `json:"metastores,omitempty" tf:"computed"`
```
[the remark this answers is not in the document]
guillesd: I've seen indeed that this is something that is done with other data sources... but for me this is not really nice for the user, since they would need two resource blocks to get the information of my metastores (first the `data_metastores`, then the `data_metastore` passing the specific ID of the metastore).

guillesd:
@nfx yes, all test should pass. But the specific tests that are failing have nothing to do with my addition, there's just some test that fail for weird reasons:
```
=== FAIL: provider TestConfig_ConfigFile (0.00s)
2023/03/17 14:25:22 [INFO] Explicit and implicit attributes: 
2023/03/17 14:25:22 [TRACE] Loading config via environment
2023/03/17 14:25:22 [TRACE] Loading config via config-file
2023/03/17 14:25:22 [DEBUG] Loading DEFAULT profile from /Users/guillermosanchez/.databrickscfg
2023/03/17 14:25:22 [TRACE] Attempting to configure auth: pat
```
This is trying to connect to my default databricks CLI profile, but why?

guillesd:
@nfx anything blocking this?

nkvuong:
@guillesd we now have account-level API to list metastores in the SDK, so it would make sense to use that instead
https://github.com/databricks/databricks-sdk-go/blob/main/service/unitycatalog/api.go#L189

guillesd:
@nkvuong this is not released yet for what I see, I should wait for that I guess!

nfx:
@guillesd it is released :)

alexott:
@guillesd would you able to finish this PR?

guillesd:
I'm taking a look at it now @alexott. Trying to find an example of another resource that uses the account API and the SDK

catalog/data_metastores.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceMetastores() *schema.Resource {
+	return common.WorkspaceData(func(ctx context.Context, data *MetastoresData, w *databricks.WorkspaceClient) error {
```
guillesd, line 12: So @alexott here I'm switching to `*databricks.AccountClient` but I'm having a weird issue now. Can I use `common.WorkspaceData` with the `AccountClient` context?

tanmay-db:
Hi @guillesd, thanks for creating a PR on this. If it's okay with you, I will add changes to your PR and address the comments / integration tests etc...

internal/acceptance/data_metastores_test.go
```diff
@@ -0,0 +1,16 @@
...
+func TestAccDataSourceMetastore(t *testing.T) {
+	workspaceLevel(t, step{
+		Template: `
+		data "databricks_metastores" "this" {}
+		
+		output "all_metastores" {
+			value = data.databricks_metastores.this
+		}`,
+	})
+}
```
tanmay-db, line 16: Using output should be fine, used cluster test for reference: https://github.com/databricks/terraform-provider-databricks/blob/master/internal/acceptance/data_cluster_test.go

tanmay-db: Tested locally, working fine

provider/provider.go
```diff
@@ -62,6 +62,7 @@ func DatabricksProvider() *schema.Provider {
 			"databricks_instance_pool":           pools.DataSourceInstancePool(),
 			"databricks_jobs":                    jobs.DataSourceJobs(),
 			"databricks_job":                     jobs.DataSourceJob(),
+			"databricks_metastores":              catalog.DataSourceMetastores(),
```
tanmay-db, line 65: should we have `databricks_metastores` or `databricks_metastore`?

catalog/data_metastores.go
tanmay-db: Is this outdated? What was the context / discussion behind this

internal/acceptance/data_metastores_test.go
tanmay-db: Also created a run on CI

provider/provider.go
tanmay-db: most likely metastore`s` since there can be multiple of them but highlighting for visibility

alexott: It should be `metastores` because it returns a list. A standard pattern usually is to have two data sources - one returning a list, and another searching for a specific resource and throwing an error if it's not found

catalog/data_metastores_test.go
```diff
@@ -0,0 +1,93 @@
...
+			},
+		},
+		Resource:    DataSourceMetastores(),
+		Read:        true,
+		NonWritable: true,
+		ID:          "_",
+		HCL: `metastores {
+			name = "a"
+			storage_root = "abc"
+		}`,
```
alexott, line 79: why it's required?

alexott (approved):
code wise looks good, but really maybe we need to add some filtering capabilities? Like, search by substring in the name, etc.?

tanmay-db:
@alexott makes sense, I will add databricks_metastore resource as well 
@nkvuong thanks, will move it to be a account level resource.

tanmay-db:
https://github.com/databricks/terraform-provider-databricks/pull/2429/files

provider/provider.go
tanmay-db: I will add that in another PR as it will require docs and tests and we can merge this before that.

catalog/data_metastores_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestMetastoresData(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/accounts/testaccount/metastores",
```
tanmay-db, line 15: We use 2.0 https://github.com/databricks/databricks-sdk-go/blob/main/service/catalog/impl.go#L78

catalog/data_metastores.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceMetastores() *schema.Resource {
+	return common.AccountData(func(ctx context.Context, data *MetastoresData, acc *databricks.AccountClient) error {
```
tanmay-db, line 12: This is added in https://github.com/databricks/terraform-provider-databricks/pull/2429/files

catalog/data_metastores.go
```diff
@@ -0,0 +1,20 @@
...
+		metastores, err := acc.Metastores.List(ctx)
+		if err != nil {
+			return err
+		}
+		data.Metastores = metastores.Metastores
+		return nil
+	})
```
tanmay-db, line 19: cc: @mgyucht

catalog/resource_metastore.go
```diff
@@ -38,16 +38,15 @@ type MetastoreInfo struct {
...
 type CreateMetastore struct {
 	Name        string `json:"name"`
 	StorageRoot string `json:"storage_root"`
 }
 
-// func (a MetastoresAPI) listMetastores() (mis []MetastoreInfo, err error) {
-// 	err = a.client.Get(a.context, "/unity-catalog/metastores", nil, &mis)
-// 	return
-// }
-
```
tanmay-db, line 50: remove dead code, we already have crud

mgyucht (commented):
Let's double check with UC team how they feel about data sources implemented with accounts client only before going forward with this.

adamcain-db:
> Let's double check with UC team how they feel about data sources implemented with accounts client only before going forward with this.

On behalf of the UC team, we are quite in favor of the terraform provider using the account-level UC Metastore CRUD/assign APIs instead of the workspce-level UC Metastore APIs.  Thanks!

internal/acceptance/data_metastores_test.go
```diff
@@ -0,0 +1,16 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+func TestUcAccDataSourceMetastore(t *testing.T) {
+	accountLevel(t, step{
+		Template: `
+		data "databricks_metastores" "this" {}
```
mgyucht, line 10: This is a reasonable integration test for the "no metastores" case. Can we add an integration test for the "one or more metastores case"?

mgyucht (commented):
Almost good to go, but let's improve the integration test.

internal/acceptance/data_metastores_test.go
nkvuong: @mgyucht not sure how easy it is to test both scenarios in a single account, unless we add a filter option to this data source

nkvuong:
Account-level APIs for UC now works without the additional header 🎉 

Acceptance tests passed

nkvuong (changes requested):
it would make more sense to return a list of name -> id mapping, otherwise the id on itself is not very useful

catalog/data_metastores.go
```diff
@@ -0,0 +1,32 @@
...
+		metastores, err := acc.Metastores.List(ctx)
+		if err != nil {
+			return err
+		}
+		data.Ids = map[string]string{}
+		for _, v := range metastores.Metastores {
+			name := v.Name
+			_, duplicateName := data.Ids[name]
+			if duplicateName {
+				return fmt.Errorf("duplicate metastore name detected: %s", name)
```
tanmay-db, line 26: cc: @nkvuong on this for visiblity

nkvuong (approved):

internal/acceptance/data_metastores_test.go
tanmay-db: Check has been added in the test

Issue #1802, opened 2022-11-29
[ISSUE] Issue with `databricks_workspace` resource (destroy) 


abij:
Destroy a workspace with non-boolean value in the config settings.

### Configuration
```hcl
resource "databricks_workspace_conf" "this" {
  custom_config = {
    "enableIpAccessLists" : false
    "maxTokenLifetimeDays": 1  # Non boolean, BUT can be string parsed AS a boolean!
  }
}
```

### Expected Behavior
- Workspace config settings are destroyed

### Actual Behavior
`Error: cannot delete workspace conf: Some values are not allowed: {"maxTokenLifetimeDays":"false"}`

### Steps to Reproduce
1. `terraform detroy`

### Terraform and provider versions
- TF: 1.3.4
- Databricks: 1.3.0

### Important Factoids
Bug is here: https://github.com/databricks/terraform-provider-databricks/blob/4a7ca5374fbcf454b8bce8180c653ee79f2bb260/workspace/resource_workspace_conf.go#L111
- Parsing 1 becomes false -> no error -> value: "false"

Related tickets: #823 #1229

abij:
Workaround is to use `2` as a minimum, until fixed.

abij:
The documentation states, but cannot find this in the code.

[Argument Reference](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/workspace_conf#argument-reference)
The following arguments are available:

[custom_config](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/workspace_conf#custom_config) - (Required) Key-value map of strings that represent workspace configuration. Upon resource deletion, properties that start with **enable** or **enforce** will be reset to false value, regardless of initial default one.

abij:
Our workaround is remove the resource `databricks_workspace_conf` from the state before a `terraform destroy` to remove the whole Databricks Workspace.

Is the solution to **only** reset if the key's are starting with _enable_ or _enforce_ like is described in the docs?
If so, I can make a PR.

alexott:
Fixed by #3737

PR #2025, opened 2023-02-20
resource_workspace_conf Delete: only reset enable and enforce properties

abij:
Make the code aligned with the documentation: https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/workspace_conf#argument-reference

PR for Issue: #1802 (remove a workspace with a non-enable/enforce setting)

### DIFF
--- a/workspace/resource_workspace_conf.go
+++ b/workspace/resource_workspace_conf.go
@@ -106,6 +106,7 @@ func ResourceWorkspaceConf() *schema.Resource {
 			for k, v := range config {
 				if !strings.HasPrefix(k, "enable") && !strings.HasPrefix(k, "enforce") {
 					log.Printf("[DEBUG] Skip erasing configuration of %s", k)
+					delete(config, k)
 					continue
 				}
 				switch r := v.(type) {
--- a/workspace/resource_workspace_conf.go
+++ b/workspace/resource_workspace_conf.go
@@ -104,6 +104,10 @@ func ResourceWorkspaceConf() *schema.Resource {
 		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			config := d.Get("custom_config").(map[string]any)
 			for k, v := range config {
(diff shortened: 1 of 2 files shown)


workspace/resource_workspace_conf.go
```diff
@@ -104,6 +104,11 @@ func ResourceWorkspaceConf() *schema.Resource {
 		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			config := d.Get("custom_config").(map[string]any)
 			for k, v := range config {
+				if !strings.HasPrefix(k, "enable") && !strings.HasPrefix(k, "enforce") {
```
alexott, line 107: Really it's not ideal situation as the workspace isn't returned to the original state. One approach would be to collect original values on create/update, and then reset to that values on delete.

abij: I get your point, for example the `projectsAllowList` (allowed git url's) is something you would like to reset (`""`), with the companion setting: `projectsAllowListPermissions
: "ALL_REPOS"` (default). 

Is there a way how to fetch or store the default values?

alexott: yes. something similar. 

You can retrieve original values when creating/updating a resource and store original values in a computed property (for example as JSON string, or something like)

Nestor10: a simpler solution could be to do nothing. Delete just removes the resource from backend / state. (This follows others work arounds now) The logic here would be that we are asking for the lack of management over the resource. As we know, we can't actually delete the workspace conf. All workspaces start with a configuration, even if its default. To get a default configuration, a simple inclusion of a workspace_conf with default settings. That to me IS a configuration. Its all grey but i fall into the camp that https://github.com/databricks/terraform-provider-databricks/blob/f29e655c5e71d00f09044dce323e16b6ab463b7b/workspace/resource_workspace_conf.go#L104
is asking for no more management not default setup.

nkvuong:
Resolved by #3737

PR #2032, opened 2023-02-20
Purge home dir and repos on user delete (SDK-compatible)

mberk06:
I was unable to resolve conflicts with the original PR so I created a PR

### DIFF
--- a/internal/acceptance/service_principal_test.go
+++ b/internal/acceptance/service_principal_test.go
@@ -1,7 +1,14 @@
 package acceptance
 
 import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
 	"testing"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
 const azureSpn = `resource "databricks_service_principal" "this" {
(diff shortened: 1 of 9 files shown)


internal/acceptance/service_principal_test.go
```diff
@@ -14,6 +21,76 @@ const awsSpn = `resource "databricks_service_principal" "this" {
 	display_name = "SPN {var.RANDOM}"
 }`
 
+func TestAccServicePrincipalHomeDeleteSuccess(t *testing.T) {
+	appId := "12345a67-8b9c-0d1e-23fa-4567b89cde04"
+	appId2 := "22345a67-8b9c-0d1e-23fa-4567b89cde04"
+	os.Setenv("appId", appId)
```
nfx, line 27: in the case of this particular test, use string concatenation instead of `env.appId`. integration tests are run in parallel and environment variables are not isolated per test.

also, environment variables are usually ALL_CAPS.

internal/acceptance/service_principal_test.go
```diff
@@ -14,6 +21,76 @@ const awsSpn = `resource "databricks_service_principal" "this" {
...
+	os.Setenv("appId2", appId2)
+	workspaceLevel(t, step{
+		Template: `
+			resource "databricks_service_principal" "a" {
+				application_id = "{env.appId}"
```
nfx, line 32:
```suggestion
				application_id = "{var.RANDOM_UUID}"
```

you can use the random UUID here

internal/acceptance/service_principal_test.go
```diff
@@ -14,6 +21,76 @@ const awsSpn = `resource "databricks_service_principal" "this" {
...
+				force_delete_home_dir = true
+			}`,
+		Check: func(s *terraform.State) error {
```
nfx, line 35: you can iterate over the `*terraform.State` here and pick up the application ID and set it to a test-level variable, that you'll use in the check in the next step

internal/acceptance/service_principal_test.go
```diff
@@ -14,6 +21,76 @@ const awsSpn = `resource "databricks_service_principal" "this" {
...
+			return nil
+		},
+	}, step{
+		Template: `
+			resource "databricks_service_principal" "b" {
+				application_id = "{env.appId2}"
+			}
+			`,
+		Check: func(s *terraform.State) error {
```
nfx, line 82: same comments as for previous test

internal/acceptance/service_principal_test.go
```diff
@@ -14,6 +21,76 @@ const awsSpn = `resource "databricks_service_principal" "this" {
...
+			w, err := databricks.NewWorkspaceClient()
+			if err != nil {
+				return err
+			}
+			ctx := context.Background()
+			_, err = w.Workspace.GetByPath(ctx, fmt.Sprintf("/Users/%v", appId))
+			if err != nil {
+				if strings.Contains(err.Error(), "doesn't exist") {
```
nfx, line 52: there's a helper - `apperr.IsMissing(err)`

scim/resource_user.go
```diff
@@ -93,7 +100,28 @@ func ResourceUser() *schema.Resource {
...
+			userName := d.Get("user_name").(string)
+			err := user.Delete(d.Id())
+			if err != nil {
+				return err
+			}
+			if c.Config.IsAccountClient() && c.Config.AccountID != "" {
+				return nil
+			}
+			if d.Get("force_delete_repos").(bool) {
+				err = NewUsersAPI(ctx, c).DeletePath(fmt.Sprintf("/Repos/%v", userName))
```
nfx, line 113: initialize workpsace api, don't add new method in the users API. it doesn't exist in the SDK.

nfx (changes requested):
1. don't add `DeletePath` to UsersAPI, because it doesn't exist in the SDK
2. adjust the tests

internal/acceptance/service_principal_test.go
mberk06: I think the new implementation is much cleaner and aligns with your comments. Lmk if there are issues

mberk06:
I was unable to run the acceptance tests and slacked about it. It would be helpful to know the needed configs and I'll update the docs I'm working on. Thanks.

alexott (approved):
lgtm

mberk06:
@nfx and @alexott it should be good to go. Thanks for all the help!

nfx (approved):

PR #2038, opened 2023-02-22
Migrate `databricks_catalogs` data to Go SDK

nfx:
This PR also introduces `common.WorkspaceData` to reduce code boilerplate. The minimal data resource is now to be defined as:

```go
return common.WorkspaceData(func(ctx context.Context, data *struct {
         // now even possible with inline structs!
	Ids []string `json:"ids,omitempty" tf:"computed,slice_set"`
}, w *databricks.WorkspaceClient) error {
	catalogs, err := w.Catalogs.ListAll(ctx)
	if err != nil {
		return err
	}
	for _, v := range catalogs {
		data.Ids = append(data.Ids, v.Name)
	}
	return nil
})
```

### DIFF
--- a/catalog/data_catalogs.go
+++ b/catalog/data_catalogs.go
@@ -3,6 +3,7 @@ package catalog
 import (
 	"context"
 
+	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/terraform-provider-databricks/common"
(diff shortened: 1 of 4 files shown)


alexott (approved):
great!

PR #2039, opened 2023-02-22
Update ip_access_list.md

watsonjm:
This example is wrong, it is also wrong on the https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/workspace_conf page. Terraform v1.3.7 will attempt to apply the resource and then error out.


alexott (approved):
lgtm

nfx (approved):

Issue #1436, opened 2022-07-07
[ISSUE] `enable_serverless_compute = false` has no effect in `databricks_sql_endpoint`


ravirahul-db:
Have enabled the workspace for serverless SQL warehouse. All the new warehouses are created by default as serverless even when I have "enable_serverless_compute = false" 

### Configuration
```hcl
resource "databricks_sql_endpoint" "this" {
  name                      = "ravi-tf-test"
  cluster_size              = "Small"
  max_num_clusters          = 1
  enable_serverless_compute = false
}
```

### Expected Behavior
Provider should consider the value of "enable_serverless_compute" when creating/updating the warehouse.

### Actual Behavior
"enable_serverless_compute" has no effect.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Terraform and provider versions
1.0.0

### Debug Output
Could not see the enable_serverless_compute passed to REST api. Same applicable for edit endponit.
During warehouse creation:
```
2022-07-08T01:03:34.759+0530 [INFO]  provider.terraform-provider-databricks: Configured pat auth: host=https://kavya.cloud.databricks.com/, token=***REDACTED***: timestamp=2022-07-08T01:03:34.759+0530
2022-07-08T01:03:34.759+0530 [DEBUG] provider.terraform-provider-databricks: POST /api/2.0/sql/endpoints {
  "auto_stop_mins": 120,
  "cluster_size": "Small",
  "enable_photon": true,
  "max_num_clusters": 1,
  "min_num_clusters": 1,
  "name": "ravi-tf-test",
  "num_clusters": 1,
  "spot_instance_policy": "COST_OPTIMIZED"
}: timestamp=2022-07-08T01:03:34.759+0530
2022-07-08T01:03:36.328+0530 [DEBUG] provider.terraform-provider-databricks: 200 OK  {
  "id": "0a8161fe7886e0cf"
} <- POST /api/2.0/sql/endpoints: timestamp=2022-07-08T01:03:36.328+0530
2022-07-08T01:03:36.329+0530 [DEBUG] provider.terraform-provider-databricks: Waiting for state to become: [success]: timestamp=2022-07-08T01:03:36.328+0530
2022-07-08T01:03:36.329+0530 [DEBUG] provider.terraform-provider-databricks: GET /api/2.0/sql/endpoints/0a8161fe7886e0cf: timestamp=2022-07-08T01:03:36.328+0530
[...]

PR #2046, opened 2023-02-27
Allow creation of classic endpoints using terraform

shreyas-goenka:
This PR is a fix for issue: https://github.com/databricks/terraform-provider-databricks/pull/new/classic-endpoint

This issue was caused because the default for enable_serverless_compute on the server side is true but we had provided the "omitempty" tag on the client side.

This has been tested manually and through unit tests

### DIFF
--- a/sql/resource_sql_endpoint.go
+++ b/sql/resource_sql_endpoint.go
@@ -29,7 +29,7 @@ type SQLEndpoint struct {
 	MaxNumClusters          int             `json:"max_num_clusters,omitempty" tf:"default:1"`
 	NumClusters             int             `json:"num_clusters,omitempty" tf:"suppress_diff"`
 	EnablePhoton            bool            `json:"enable_photon" tf:"optional,default:true"`
-	EnableServerlessCompute bool            `json:"enable_serverless_compute,omitempty" tf:"suppress_diff"`
+	EnableServerlessCompute bool            `json:"enable_serverless_compute" tf:"suppress_diff,optional,default:true"`
 	InstanceProfileARN      string          `json:"instance_profile_arn,omitempty"`
 	State                   string          `json:"state,omitempty" tf:"computed"`
 	JdbcURL                 string          `json:"jdbc_url,omitempty" tf:"computed"`
--- a/sql/resource_sql_endpoint_test.go
+++ b/sql/resource_sql_endpoint_test.go
@@ -41,12 +41,13 @@ func TestResourceSQLEndpointCreate(t *testing.T) {
 				Method:   "POST",
 				Resource: "/api/2.0/sql/warehouses",
 				ExpectedRequest: SQLEndpoint{
-					Name:               "foo",
-					ClusterSize:        "Small",
-					MaxNumClusters:     1,
-					AutoStopMinutes:    120,
-					EnablePhoton:       true,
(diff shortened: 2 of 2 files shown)


sql/resource_sql_endpoint.go
```diff
@@ -29,7 +29,7 @@ type SQLEndpoint struct {
 	MaxNumClusters          int             `json:"max_num_clusters,omitempty" tf:"default:1"`
 	NumClusters             int             `json:"num_clusters,omitempty" tf:"suppress_diff"`
 	EnablePhoton            bool            `json:"enable_photon" tf:"optional,default:true"`
-	EnableServerlessCompute bool            `json:"enable_serverless_compute,omitempty" tf:"suppress_diff"`
+	EnableServerlessCompute bool            `json:"enable_serverless_compute" tf:"suppress_diff,optional,default:true"`
```
alexott, line 32: `default:true` - you can't set it to this - it will fail on the workspaces without serverless enabled.

alexott (changes requested):
If it would be some simple, it would be fixed long time ago ;-)

The problem with that fix is that it won't work on the workspaces with no serverless enabled - creation of warehouses will fail. There are many reasons when customers may not adopt serverless - not in region with serverless, compliance restrictions, etc.  And requiring explicit disable of serverless is also not a way forward because it will require to people their existing code.

And warehouses are created by default with serverless (and PRO SKU) if workspace has serverless enabled, and without serverless and (CLASSIC SKU) if it doesn't have serverless enabled.  And although we have a way to check if serverless is enabled or not via SQL Global Config API we can't use it as it's available only for admins, and this will prohibit "normal users" to create warehouses.

This will require a more complex solution:

-  check if serverless is explicitly enabled or disabled
- if it's not explicitly enabled/disabled and warehouse SKU isn't set to CLASSIC:
  - try to create warehouse with serverless enabled
  - if it fails with a specific error saying that serverless isn't enabled, then try to create warehouse with serverless disabled.

nfx:
@shreyas-goenka please link the github issue to this PR

sql/resource_sql_endpoint_test.go
```diff
@@ -80,19 +81,68 @@ func TestResourceSQLEndpointCreate(t *testing.T) {
 	assert.Equal(t, "d7c9d05c-7496-4c69-b089-48823edad40c", d.Get("data_source_id"))
 }
 
+func TestResourceClassicSQLEndpointCreate(t *testing.T) {
+	d, err := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "POST",
+				Resource: "/api/2.0/sql/warehouses",
+				ExpectedRequest: SQLEndpoint{
```
nfx, line 90: can you rewrite the test to work with `map[string]any`, this will show if the fix is exactly there and we're properly serializing the request payload.

nfx (changes requested):

shreyas-goenka:
It seems this issue is not solvable right now due to this limitation: https://github.com/databricks/terraform-provider-databricks/issues/1436#issuecomment-1446608069. Basically, we cannot detect whether an API field is unspecified, and if we cannot detect that, we cannot comply with the API behavior. 

Thus closing this PR. We can revisit this in the future

PR #2047, opened 2023-02-27
Added `databricks_sql_alert` resource

### LINKED ISSUE #1780: [FEATURE] New feature request: databricks_sql_alert resource
Hello!

I am working in a databricks workspace and want to use SQL Alerts to enhance monitoring for data issues.  I would like to create the Alerts via terraform, but there does not seem to be a resource yet to create Alerts.  It would be great if a databricks_sql_alert resource could be created to use in my terraform module.

kartikgupta-db:
Adds `databricks_sql_alert` resource to manage SQL Alerts.

Fixes #1780


CONTRIBUTING.md
```diff
@@ -61,7 +61,7 @@ After installing necessary software for building provider from sources, you shou
 
 Make sure you have `$GOPATH/bin` in your `$PATH`:
 ```
-echo "export PATH=\$PATH:$(go env GOPATH)/bin" >> ~/.bash_profile
```
alexott, line 64: Please revert this as it's incorrect. The goal of this line to generate a line like this in the `.bash_profile`:

```
export PATH=$PATH:/Users/<user>/exp/go/bin
```

but without `\` it will expand content of the `PATH` environment variable that is incorrect.

aws/resource_instance_profile_test.go
```diff
@@ -374,4 +374,3 @@ func TestResourceInstanceProfileUpdate_Error(t *testing.T) {
 	qa.AssertErrorStartsWith(t, err, "Internal error happened")
 	assert.Equal(t, "arn:aws:iam::999999999999:instance-profile/my-fake-instance-profile", d.Id())
 }
-
```
alexott, line 377: not necessary change

sql/resource_sql_alerts.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (a AlertApi) Delete(id string) error {
+	path := fmt.Sprintf("/preview/sql/alerts/%v", id)
+	return a.client.Get(a.context, path, nil, nil)
+}
+
+func (a AlertApi) Update(id string, request *api.CreateAlert) error {
+	path := fmt.Sprintf("/preview/sql/alerts/%v", id)
+	return a.client.Put(a.context, path, request)
+}
```
alexott, line 104: Shouldn't we use Databricks Go SDK instead? https://github.com/databricks/databricks-sdk-go/blob/main/service/sql/api.go#L16

alexott (changes requested):
Besides the comments on specific piece of the code:

* unit test is missing so coverage is low
* documentation is missing

sql/resource_sql_alerts.go
alexott: @nfx ^

kartikgupta-db:
> unit test is missing so coverage is low
documentation is missing

@alexott these are still a TODO. I am working on them and will update the PR soon.

alexott:
I was confused by fact that it's not a draft :-)

kartikgupta-db:
> I was confused by fact that it's not a draft :-)

Yep. Changed that now 😄.

sql/resource_sql_alerts.go
```diff
@@ -0,0 +1,152 @@
...
+func ResourceSqlAlert() *schema.Resource {
+	s := common.StructToSchema(AlertEntity{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		options := m["options"].Elem.(*schema.Resource)
+		options.Schema["op"].ValidateFunc = validation.StringInSlice([]string{">", ">=", "<", "<=", "==", "!="}, true)
+		return m
+	})
+
+	return common.Resource{
+		Create: func(ctx context.Context, data *schema.ResourceData, c *common.DatabricksClient) error {
+			var a AlertEntity
```
nfx, line 115: use Go SDK to write this resource, don't implement the resource API client from scratch

nfx (changes requested):

sql/resource_sql_alerts.go
nfx: we _must use Go SDK_ for all new resources

internal/acceptance/sql_alert_test.go
```diff
@@ -0,0 +1,27 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+func TestAccAlert(t *testing.T) {
+	workspaceLevel(t, step{
```
nfx, line 8: can you add one more step to perform editing of this alert?

sql/resource_sql_alerts.go
```diff
@@ -0,0 +1,143 @@
...
+		Muted:         apiAlert.Options.Muted,
+		CustomBody:    apiAlert.Options.CustomBody,
+		CustomSubject: apiAlert.Options.CustomSubject,
+	}
+
+	return common.StructToData(a, s, data)
+}
+
+func ResourceSqlAlert() *schema.Resource {
+	s := common.StructToSchema(AlertEntity{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
```
nfx, line 86: can you check if we can do it without `to/fromAPIObject`? `ForceNew` you can add via this schema callback for sure. suppress diff - not quite sure if we need it.

nfx (changes requested):
+ this needs docs

nfx (approved):

PR #2067, opened 2023-03-05
Fixed InitScripts Type to work with GCS and ABFS.

martinzuern:
On Azure and GCP, InitScripts can be pulled from their respective Cloud Storage offerings. However due to the wrong type, when refreshing only DBFS and S3 are considered, leading to diffs (where in reality there is no diff). 

This PR updates the type so that also ABFS and GCS InitScripts are considered on refresh.


nkvuong:
@martinzuern do you mind adding a unit test that covers this?

martinzuern:
@nkvuong added a test

martinzuern:
Is there anything else you need from my side to get it merged? Would be thrilled if it makes it into the next release. @nkvuong

alexott (approved):
lgtm. thank you!

nfx (approved):

PR #2114, opened 2023-03-14
Allow rotating `token` block in `databricks_mws_workspaces` resource by only changing `comment` field

shreyas-goenka:
Tested manually for the following cases
1. Without this PR the provider recreates the entire workspace on a token update
2. With changes in this PR only the token is refreshed
3. When both token and `storage_configuration_id` are changed then the entire workspace is recreated

Additional unit tests also added that allow checks that patch workspace calls are not made when only token is changed

Also added an integration test to check tokens are successfully updated


nfx:
@shreyas-goenka do acceptance tests pass on this PR? can you run the PR-manual on it?

shreyas-goenka:
Passes PR Manual: https://github.com/databricks/eng-dev-ecosystem/actions/runs/4574355963/jobs/8075877072

nfx (approved):

PR #2135, opened 2023-03-20
Add example for automatically rotating a token

pietern:


docs/resources/token.md
```diff
@@ -30,6 +30,22 @@ output "databricks_token" {
 }
 ```
 
+A token can be automatically rotated by taking a dependency on the `time_rotating` resource:
+
+```hcl
+resource "time_rotating" "this" {
+  rotation_months = 30
```
alexott, line 37: per common on  PAT this should be `rotation_days` instead of `rotation_months`

alexott (changes requested):

nfx (approved):

docs/resources/token.md
pietern: Whoops, thank you!

alexott (approved):
lgtm

PR #2139, opened 2023-03-21
Deprecate `enable_serverless_compute` on `databricks_sql_global_config` resource

nfx:
cc @jxbell

### DIFF
--- a/sql/resource_sql_endpoint.go
+++ b/sql/resource_sql_endpoint.go
@@ -194,6 +194,8 @@ func (a SQLEndpointsAPI) Delete(endpointID string) error {
 func ResourceSqlEndpoint() *schema.Resource {
 	s := common.StructToSchema(SQLEndpoint{}, func(
 		m map[string]*schema.Schema) map[string]*schema.Schema {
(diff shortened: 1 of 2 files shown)


sql/resource_sql_endpoint.go
```diff
@@ -194,6 +194,8 @@ func (a SQLEndpointsAPI) Delete(endpointID string) error {
 func ResourceSqlEndpoint() *schema.Resource {
 	s := common.StructToSchema(SQLEndpoint{}, func(
 		m map[string]*schema.Schema) map[string]*schema.Schema {
+		m["enable_serverless_compute"].Deprecated = "This field is intended as an internal API " +
+			"and may be removed from the Databricks Terraform provider in the future"
```
alexott, line 198: No, it should be done only for Global SQL config. There are a valid reasons for disabling serverless on the specific warehouses.

alexott (changes requested):
we need to deprecate only global config

sql/resource_sql_endpoint.go
nfx: reverted

alexott (approved):
lgtm!

PR #2147, opened 2023-03-23
Added experimental queue field

andrepibo-db:
Add support for private preview functionality queueing.

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -185,6 +185,9 @@ type ContinuousConf struct {
 	PauseStatus string `json:"pause_status,omitempty" tf:"computed"`
 }
 
+type Queue struct {
+}
+
 // JobSettings contains the information for configuring a job on databricks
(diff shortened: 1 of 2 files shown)


jobs/resource_job.go
```diff
@@ -222,6 +225,7 @@ type JobSettings struct {
 	EmailNotifications   *EmailNotifications   `json:"email_notifications,omitempty" tf:"suppress_diff"`
 	WebhookNotifications *WebhookNotifications `json:"webhook_notifications,omitempty" tf:"suppress_diff"`
 	Tags                 map[string]string     `json:"tags,omitempty"`
+	Queue                *Queue                `json:"queue,omitempty"`
```
KoningJasper, line 228: Not sure omitempty works here because the object is (intentionally) empty.

andrepibo-db:
I added a unit test and tested it manually against a real workspace.

PR #2170, opened 2023-03-29
Add caching of scim/me API to `databricks_current_user` data source and `databricks_permissions` resource.

kartikgupta-db:
## Changes
* cache scim/me API
## Tests
* manual
*  existing acceptance tests
   * gcp-prod
   * aws-prod
   * azure-prod
   * aws-prod-ucws
   * aws-prod-ucacct
   * azure-prod-acct
   * gcp-prod-acct

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -15,13 +15,13 @@ import (
 	"github.com/golang-jwt/jwt/v4"
 )
 
-type cachedCurrentUserService struct {
+type cachedMe struct {
 	internalImpl scim.CurrentUserService
 	cachedUser   *scim.User
 	mu           sync.Mutex
 }
 
-func (a *cachedCurrentUserService) Me(ctx context.Context) (*scim.User, error) {
+func (a *cachedMe) Me(ctx context.Context) (*scim.User, error) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	if a.cachedUser != nil {
@@ -58,16 +58,12 @@ func (c *DatabricksClient) WorkspaceClient() (*databricks.WorkspaceClient, error
 	if err != nil {
 		return nil, err
 	}
-	c.cachedWorkspaceClient = w
-	return c.withCachedCurrentUserApi(w), nil
-}
-
-func (c *DatabricksClient) withCachedCurrentUserApi(w *databricks.WorkspaceClient) *databricks.WorkspaceClient {
 	internalImpl := w.CurrentUser.Impl()
-	w.CurrentUser.WithImpl(&cachedCurrentUserService{
+	w.CurrentUser.WithImpl(&cachedMe{
 		internalImpl: internalImpl,
 	})
-	return w
+	c.cachedWorkspaceClient = w
+	return w, nil
 }
 
 // Get on path
--- a/common/client.go
+++ b/common/client.go
@@ -10,8 +10,27 @@ import (
 	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/client"
 	"github.com/databricks/databricks-sdk-go/config"
+	"github.com/databricks/databricks-sdk-go/service/scim"
 	"github.com/golang-jwt/jwt/v4"
 )
+
+type CachedCurrentUserService struct {
+	client     *client.DatabricksClient
+	oldImpl    scim.CurrentUserService
+	cachedUser *scim.User
+}
+
+func (a *CachedCurrentUserService) Me(ctx context.Context) (*scim.User, error) {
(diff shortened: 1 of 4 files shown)


common/client.go
```diff
@@ -22,10 +41,26 @@ type DatabricksClient struct {
 
 	// callback used to create API1.2 call wrapper, which simplifies unit tessting
 	commandFactory func(context.Context, *DatabricksClient) CommandExecutor
+	cachedUser     scim.User
 }
 
 func (c *DatabricksClient) WorkspaceClient() (*databricks.WorkspaceClient, error) {
-	return databricks.NewWorkspaceClient((*databricks.Config)(c.DatabricksClient.Config))
+	// return databricks.NewWorkspaceClient((*databricks.Config)(c.DatabricksClient.Config))
```
nfx, line 48: remove commented code

common/client.go
```diff
@@ -10,9 +10,28 @@ import (
...
+
+func (a *CachedCurrentUserService) Me(ctx context.Context) (*scim.User, error) {
+	if a.cachedUser.DisplayName != "" {
+		return a.cachedUser, nil
+	}
+	user, err := a.oldImpl.Me(ctx)
+	if err != nil {
+		return user, err
+	}
+	*a.cachedUser = *user
```
nfx, line 31: why not `a.cachedUser = user`?..

common/client.go
```diff
@@ -10,9 +10,28 @@ import (
...
 	"github.com/golang-jwt/jwt/v4"
 )
 
+type CachedCurrentUserService struct {
+	client     *client.DatabricksClient
+	oldImpl    scim.CurrentUserService
+	cachedUser *scim.User
+}
+
+func (a *CachedCurrentUserService) Me(ctx context.Context) (*scim.User, error) {
```
nfx, line 23: add `sync.Mutex`, otherwise it's not threadsafe.

common/client.go
```diff
@@ -22,10 +41,26 @@ type DatabricksClient struct {
 
 	// callback used to create API1.2 call wrapper, which simplifies unit tessting
 	commandFactory func(context.Context, *DatabricksClient) CommandExecutor
+	cachedUser     scim.User
 }
 
 func (c *DatabricksClient) WorkspaceClient() (*databricks.WorkspaceClient, error) {
-	return databricks.NewWorkspaceClient((*databricks.Config)(c.DatabricksClient.Config))
+	// return databricks.NewWorkspaceClient((*databricks.Config)(c.DatabricksClient.Config))
+	w, err := databricks.NewWorkspaceClient((*databricks.Config)(c.DatabricksClient.Config))
```
nfx, line 49: if we're creating a workspace client every time we do `c.WorkspaceClient()`, there's no caching ;)

nfx (changes requested):
can you just create func (c *DatabricksClient) CurrentUser(ctx), that'll lock/unlock properly. just because there's only one instance of DatabricksClient per provider ;)

common/client.go
kartikgupta-db: The pointer points to the cached user in client. This is why we need to copy into the cachedUser value.

common/client.go
```diff
@@ -22,10 +47,27 @@ type DatabricksClient struct {
 
 	// callback used to create API1.2 call wrapper, which simplifies unit tessting
 	commandFactory func(context.Context, *DatabricksClient) CommandExecutor
+	cachedUser     scim.User
```
nfx, line 50:
```suggestion
	cachedUser     *scim.User
```

common/client.go
nfx: just make it `cachedUser     *scim.User` on the client.

common/client.go
nfx: add mutex here

common/client.go
kartikgupta-db: That would not work I think. If we do this, then `a.cachedUser = user` will lead to the cachedUser within `CachedUserService` to point to the new user, but the changes won't be propagated to the client.

common/client.go
kartikgupta-db: This should be fine because the address of cachedUser is not going to change.

common/client.go
kartikgupta-db: This would cause caching to not propagate. The explanation https://github.com/databricks/terraform-provider-databricks/pull/2170#discussion_r1151711693

common/client.go
nfx: i don't understand it

nfx (approved):
lgtm

PR #2181, opened 2023-03-31
Exclude roles in scim API list calls

shreyas-goenka:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/data_service_principal_test.go
+++ b/scim/data_service_principal_test.go
@@ -12,7 +12,7 @@ func TestDataServicePrincipalReadByAppId(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "GET",
-				Resource: "/api/2.0/preview/scim/v2/ServicePrincipals?filter=applicationId%20eq%20%27abc%27",
+				Resource: "/api/2.0/preview/scim/v2/ServicePrincipals?excludedAttributes=roles&filter=applicationId%20eq%20%27abc%27",
 				Response: UserList{
 					Resources: []User{
 						{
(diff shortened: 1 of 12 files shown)


exporter/exporter_test.go
```diff
@@ -340,7 +340,7 @@ func TestImportingUsersGroupsSecretScopes(t *testing.T) {
 			},
 			{
 				Method:   "GET",
-				Resource: "/api/2.0/preview/scim/v2/ServicePrincipals?filter=applicationId%20eq%20%27spn%27",
+				Resource: "/api/2.0/preview/scim/v2/ServicePrincipals?excludedAttributes=roles&filter=applicationId%20eq%20%27spn%27",
```
nfx, line 343: This should stay, it’s called once and cached after

scim/data_service_principal_test.go
```diff
@@ -56,7 +56,7 @@ func TestDataServicePrincipalReadNotFound(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "GET",
-				Resource: "/api/2.0/preview/scim/v2/ServicePrincipals?filter=applicationId%20eq%20%27abc%27",
+				Resource: "/api/2.0/preview/scim/v2/ServicePrincipals?excludedAttributes=roles&filter=applicationId%20eq%20%27abc%27",
```
nfx, line 59: For single service principal/user - keep roles, it’s part of the public APIs

shreyas-goenka: We do not have output attributes that depend read roles for users/service principals. For groups we do, which is the instance_profiles. Doing a second pass on this just to be sure

shreyas-goenka: So we use the list user API in three instances:
1. Create user (resource): We only use the Id here
2. Read user (data source): we use UserName, DisplayName, ExternalID, ApplicationID, ID here
3. Exporter:  UserName, DisplayName, Id, Groups

Does not use roles

exporter/exporter_test.go
shreyas-goenka: It's still not being used, can we add it later on if we need it?

nfx (approved):

PR #2197, opened 2023-04-06
Fix diff detection for `databricks_share` resource

### LINKED ISSUE #2167: [ISSUE] Issue with `databricks_share` update detection
### Configuration
Adding a new `object` to a `databricks_share` resource e.g.
```hcl
databricks_share my_share {
  name = "my_share"
  object {
    data_object_type = "TABLE"
    name = "main.foo.table_A"
  }
  object {
    data_object_type = "TABLE"
    name = "main.foo.table_C"
  }
}
```
->
```hcl
databricks_share my_share {
  name = "my_share"
  object {
    data_object_type = "TABLE"
    name = "main.foo.table_A"
  }
  object {
    data_object_type = "TABLE"
    name = "main.foo.table_B"
  }
  object {
    data_object_type = "TABLE"
    name = "main.foo.table_C"
  }
}
```

### Expected Behavior
Object for `main.foo.table_B` is added to the share

### Actual Behavior
Apply fails with `cannot update share: Shared Table 'foo.table_C' already exists`

When explicitly providing `shared_as` both before and after adding new object, it works.

Most likely https://github.com/databricks/terraform-provider-databricks/blob/0693e4c31f4fa84f2e01d5b0f75271f4c0b46fd8/catalog/resource_share.go#L152 does not work as expected.

1.10.0

nkvuong:
## Changes

Improve diff detection by suppressing diff for `shared_as` field. Close #2167

## Tests

- [x] `make test` run locally

### DIFF
--- a/catalog/resource_share.go
+++ b/catalog/resource_share.go
@@ -128,6 +128,13 @@ func (si ShareInfo) resourceShareMap() map[string]SharedDataObject {
 	return m
 }
 
+func Equal(a, b SharedDataObject) bool {
+	if b.SharedAs == "" {
+		b.SharedAs = a.SharedAs
+	}
+	return reflect.DeepEqual(a, b)
+}
+
 func (si ShareInfo) Diff(other ShareInfo) []ShareDataChange {
 	beforeMap := si.resourceShareMap()
 	afterMap := other.resourceShareMap()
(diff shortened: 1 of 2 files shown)


nkvuong:
acceptance tests passed

nkvuong:
@marekbrysa do you mind validating this manually with your issue as well?

catalog/resource_share.go
```diff
@@ -128,6 +128,13 @@ func (si ShareInfo) resourceShareMap() map[string]SharedDataObject {
 	return m
 }
 
+func Equal(a, b SharedDataObject) bool {
```
alexott, line 131: I would either rename it to be more explicit on what we're comparing, or make it a member function of the corresponding struct.

alexott (commented):
in general looks good, but may need to change name...

catalog/resource_share.go
marekbrysa: Also perhaps rename the parameters? `b` has the meaning of the "after"/goal state?

nfx: not blocking, but this should be the method on `SharedDataObject` in the follow-up PR

nfx (approved):

Issue #2193, opened 2023-04-05
[ISSUE] Issue with `databricks_group` data requiring workspace admin privileges since 1.14.1


mtrampont:
### Configuration
```hcl
terraform {
  required_providers {
    databricks = {
      source = "databricks/databricks"
      version = "1.14.1"
    }
  }
}

provider "databricks" {
  host  = var.db_url
  token = var.databricksPAT
}

data "databricks_group" "somegroup" {
  display_name = "somegroup-display-name"
}

```

### Expected Behavior

No issue.

### Actual Behavior

Terraform plan fails with following error:
```
��� Error: /2.0/preview/scim/v2/Groups/916570868973800 is only accessible by admins

��� 

���   with databricks_group.somegroup,

���   on tstDatabricksInstance.tf line [...], in data "databricks_group" "somegroup":

���    4: data "databricks_group" "somegroup" {
```

### Steps to Reproduce

Terraform version 1.3.9
Databricks provider version: 1.14.1

### Important Factoids

I think the issue come from PR#2181, more specifically [this line](https://github.com/databricks/terraform-provider-databricks/commit/cb2f98b42c4407e2e67a34b31553c9738840cdd4#diff-11e6b6f6d6e6f4e8664a09b309d609b81f368b5dc272a1e565aa1f96e2980082R67).
A call to the Get Group by Id API was added and this is only usable by admin of the workspace from the [doc](https://learn.microsoft.com/en-us/azure/databricks/dev-tools/api/latest/scim/scim-groups#get-group-by-id).

mtrampont:
@shreyas-goenka regarding your changes to [scim/groups.go](https://github.com/databricks/terraform-provider-databricks/commit/cb2f98b42c4407e2e67a34b31553c9738840cdd4#diff-11e6b6f6d6e6f4e8664a09b309d609b81f368b5dc272a1e565aa1f96e2980082), I'm not an expert but was it really necessary to call `Filter` with `includeRoles` parameter to `false` on [this line](https://github.com/databricks/terraform-provider-databricks/blame/cb2f98b42c4407e2e67a34b31553c9738840cdd4/scim/groups.go#L57),
given that `Filter` is called with a matching condition on exactly one group display_name ?

In this case, is fetching the roles an issue since only one group should match ?

nfx:
@mtrampont can you confirm this error didn't happen before 1.14.1?

mtrampont:
[...]

PR #2210, opened 2023-04-11
Fixed regression in `databricks_group` data not to require workspace admin privileges

shreyas-goenka:
## Changes

We had a regression as specified in https://github.com/databricks/terraform-provider-databricks/issues/2193

The fix is to revert back to using the list groups API since that is not gated to admins

## Tests

Tested using existing unit tests and manually with a non admin account with the following hcl

```
terraform {
  required_providers {
    databricks = {
    }
  }
}

provider "databricks" {
  profile = "aws-prod-nonadmin"
}

data databricks_group "this" {
    display_name = "users"
}
```

- [ ] `make test` run locally
- [ ] using Go SDK


shreyas-goenka:
Initiated manual acc tests for account and workspace level

nfx (approved):

PR #2137, opened 2023-03-21
Document `enable_serverless_compute` API changes in `databricks_sql_endpoint` resource

jxbell:
**DEPLOYMENT NOTE: TO BE DEPLOYED WITH AWS GA SERVERLESS SQL ON MARCH 22 AND IN CONJUNCTION WITH THIS DOCS PR: https://github.com/databricks/docs/pull/10197** PROBABLY BEST TO WAIT FOR CONFIRMATION OF GO/NOGO!

Three related PRs:
- Docs repo PR for main docs including release note = https://github.com/databricks/docs/pull/10398 , a followup to https://github.com/databricks/docs/pull/10197
- Universe PR for API Explorer = https://github.com/databricks/universe/pull/282673 a followup to another universe PR https://github.com/databricks/universe/pull/280282
- Terraform PR for Terraform = https://github.com/databricks/terraform-provider-databricks/pull/2137

[DOC-6788] Terraform AWS serverless SQL API changes

IMPORTANT: NO API changes to remove global config field IN TERRAFORM CONTRACT PER @nfx -- HE DID DEPRECATE IT IN https://github.com/databricks/terraform-provider-databricks/pull/2139/files

**Cleanup:**

- Cleanup: fix "sql endpoint" in English descriptions (not code/spec) to "SQL warehouse" per name change.

- Cleanup: fix to lowercase for all this type of case for "serverless" (per Marketing Department dictum)

**API change to regular `enable_serverless_compute` field (not global config) :**

- Update the `enable_serverless_compute` guidance for AWS and Azure enablement: ((Make more accurate enablement info for AWS only (not "enable for workspace" since we don't do that as of GA). Azure doesn't need special terms of use flags at account level, but does have workspace flag) : 

    `If your account was created before October 1, 2021, your organization's owner or account administrator must [accept applicable terms of use](https://docs.databricks.com/sql/admin/serverless.html#accept-terms) before workspaces are enabled for serverless compute. A workspace must meet the [requirements](https://docs.databricks.com/sql/admin/serverless.html#requirements) and might require an update its instance profile role to [add a trust relationship](https://docs.databricks.com/sql/admin/serverless.html#aws-instance-profile-setup). For Azure, you must [enable your workspace for serverless SQL warehouse](https://learn.microsoft.com/azure/databricks/sql/admin/serverless).
     `
 
 - Also update that field per agreement with PM to match API Explorer and main docs... Update AWS specific instructions when we enable "auto enablement at account level" as part of GA for AWS

[...]

PR #2249, opened 2023-04-26
Update `databricks_sql_endpoint` docs for Azure DBSQL Serverless GA

jxbell:
## Deploy requirements

This must be merged and deploy so that the changes are visible on May 3, 2023 but NOT before -- that date is stage 1.5 for the Azure GA -- this matches the corresponding docs repo PR https://github.com/databricks/docs/pull/10837

## Docs-only changes description

[DOC-8669] Required Terraform docs changes for Azure DBSQL Serverless GA

This is in effect a docs-only Azure followup PR to the AWS changes from this PR: https://github.com/databricks/terraform-provider-databricks/pull/2137

Basically Azure now gets the complicated story for the default for the serverless enabled field that has a magic set of dates... although it's DIFFERENT sets of dates for Azure.

[DOC-8669]: https://databricks.atlassian.net/browse/DOC-8669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/resources/sql_endpoint.md
```diff
@@ -36,14 +36,15 @@ The following arguments are supported:
 * `tags` - Databricks tags all endpoint resources with these tags.
 * `spot_instance_policy` - The spot policy to use for allocating instances to clusters: `COST_OPTIMIZED` or `RELIABILITY_OPTIMIZED`. This field is optional. Default is `COST_OPTIMIZED`.
 * `enable_photon` - Whether to enable [Photon](https://databricks.com/product/delta-engine). This field is optional and is enabled by default.
-* `enable_serverless_compute` - Whether this SQL warehouse is a serverless endpoint. If this value is true explicitly or through the default, you **must** also set `warehouse_type` field to `pro`.
+* `enable_serverless_compute` - Whether this SQL warehouse is a serverless endpoint. If this value is true explicitly or through the default, you **must** also set `warehouse_type` field to `pro`. To avoid ambiguity, especially for organizations with many workspaces, Databricks recommends that you always set this field explicitly.
```
alexott, line 39: as I remember it's not necessary to set it to `PRO` as it's default on the backend

docs/resources/sql_endpoint.md
```diff
@@ -36,14 +36,15 @@ The following arguments are supported:
...
 * `channel` block, consisting of following fields:
   * `name` - Name of the Databricks SQL release channel. Possible values are: `CHANNEL_NAME_PREVIEW` and `CHANNEL_NAME_CURRENT`. Default is `CHANNEL_NAME_CURRENT`.
-* `warehouse_type` - SQL warehouse type. See for [AWS](https://docs.databricks.com/sql/admin/sql-endpoints.html#switch-the-sql-warehouse-type-pro-classic-or-serverless) or [Azure](https://docs.databricks.com/sql/admin/sql-endpoints.html#switch-the-sql-warehouse-type-pro-classic-or-serverless). Set to `PRO` or `CLASSIC` (default).  If you want to use serverless compute, you must set to `PRO` and **also** set the field `enable_serverless_compute` to `true`.
+* `warehouse_type` - SQL warehouse type. See for [AWS](https://docs.databricks.com/sql/admin/sql-endpoints.html#switch-the-sql-warehouse-type-pro-classic-or-serverless) or [Azure](https://learn.microsoft.com/en-us/azure/databricks/sql/admin/create-sql-warehouse#--upgrade-a-pro-or-classic-sql-warehouse-to-a-serverless-sql-warehouse). Set to `PRO` or `CLASSIC` (default).  For serverless SQL warehouses, you must set to `PRO` and also set the field `enable_serverless_compute` to `true`.
```
alexott, line 47: We need to remove that `CLASSIC` is default - we removed any default from the code, and now backend selects what is default...

alexott (changes requested):

docs/resources/sql_endpoint.md
jxbell: That is not consistent with what I've been told. My understanding is that the REST API backend has CLASSIC as the default, and thus must be set explicitly.

Moreover, fields are either required or optional, and if optional (as this is), API reference must specify the default if omitted. If there is a more complicated default based on the value of enable_serverless_compute, sure let's doc it.... but that is different from what has been technically reviewed in the past for the REST API which is that the default is classic. 

@mengqiu-teng @superdupershant can you confirm?

docs/resources/sql_endpoint.md
jxbell: See comment thread https://github.com/databricks/terraform-provider-databricks/pull/2249#discussion_r1179487705

docs/resources/sql_endpoint.md
mengqiu-teng: For warehouse API, warehouse_type is an optional field. If enable_serverless_compute is explicitly set to true, then warehouse_type will be set to PRO by backend. If the user does not specify warehouse_type and enable_serverless_compute, then it will be CLASSIC by default

jxbell: When you say "If the user does not specify warehouse_type and enable_serverless_compute, then it will be CLASSIC by default"

I think you mean "If the user does not specify warehouse_type and the field enable_serverless_compute is either false or unspecified, then it will be CLASSIC by default", is that correct?

@mengqiu-teng

mengqiu-teng: Yes. This will be the behavior for users that are not in the special list for API change. Thank you for the precise wording!

gaurav-pm (approved):

docs/resources/sql_endpoint.md
alexott: @mengqiu-teng @jxbell not exactly - if `enable_serverless_compute` isn't specified, then the behavior will depend on if the Serverless is enabled for the workspace or not. If yes, then the warehouse will be created as Serverless, not ClASSIC.

Here is an example for my workspace that has Serverless enabled:

```
  name             = "Channels test"
  cluster_size     = "2X-Small"
  max_num_clusters = 1
}
```

gives:

Screenshot 2023-04-28 at 08 34 37

Unfortunately this implicit behaviour leads to the problem like #1436 - we can't always send the `enable_serverless_compute` as the warehouse creation logic depends on the global configuration - it's very hard to fix this correctly.

docs/resources/sql_endpoint.md
jxbell: Per meng, changing this....

```suggestion
* `enable_serverless_compute` - Whether this SQL warehouse is a serverless endpoint. See below for details about the default values. To avoid ambiguity, especially for organizations with many workspaces, Databricks recommends that you always set this field explicitly.
```

docs/resources/sql_endpoint.md
jxbell: @alexott @mengqiu-teng @gaurav-pm 

(1) I changed the enable_serverless_compute field a few lines up to remove the reference to setting warehouse type since.

(2) Alex is right that the wording needs to change for warehouse type -- it's not always classic, but my understanding is that the correct answer for the default is based on the EFFECTIVE value of enable_serverless_compute... which depends (and Alex does not mention this...) based on whether the workspace is in the list of ones that used the API to create a warehouse in the special date range. ("the default remains the previous behavior which is default to true if the workspace is enabled for serverless and fits the requirements for serverless SQL warehouses" -- note no reference to global workspace enablement switch -- see item 6 below)

(3) proposed warehouse type new version (removing default for classic):

```suggestion
* `warehouse_type` - SQL warehouse type. See for [AWS](https://docs.databricks.com/sql/admin/sql-endpoints.html#switch-the-sql-warehouse-type-pro-classic-or-serverless) or [Azure](https://learn.microsoft.com/en-us/azure/databricks/sql/admin/create-sql-warehouse#--upgrade-a-pro-or-classic-sql-warehouse-to-a-serverless-sql-warehouse). Set to `PRO` or `CLASSIC`.  If the field `enable_serverless_compute` has the value `true` either explicitly or through the default logic (see that field above for details), the default is `PRO`, which is required for serverless SQL warehouses. Otherwise, the default is `CLASSIC`.
```

@mengqiu-teng @alexott does that sound right?

(4) @alexott In your example, I assume it created serverless merely because that workspace DID use the SQL Warehouses API to create a warehouse in the magic date range... and that's why it created serverless. 

(5) As for the issue https://github.com/databricks/terraform-provider-databricks/issues/1436 , where it's impossible to not have serverless, it sounds just like a **temporary bug in the provider** rather than an intended behavior, right? if so, the documentation typically should reflect the intended behavior, not the bug behavior. If this is supposed to follow the rules of the REST API, passing 'false' **explicitly** should disable serverless even if the workspace in the "special date range workspace list" since that change was about the defaults and setting explicitly to false should override, @mengqiu-teng please confirm. 

(6) @alexott you mention the 'depends on the global configuration' -- does that refer to the internal workspace switch? if so, that went away in GA docs though I seem to remember it appears in the terraform provider temporarily as deprecated and do not use for AWS. And as of Azure GA, the same would apply -- a global workspace switch is NOT supported for customer use for any cloud.

cc: @gaurav-pm @amji12

alexott: yes, that flag that is removed (from docs only), but people may still have it.

mengqiu-teng: For (3), that sounds right to me
for (5), I believe the behavior mentioned in https://github.com/databricks/terraform-provider-databricks/issues/1436 buggy. If customer specifies enable_serverless_compute in the request, in the backend, we always respect this and will create non-Serverless warehouse even tho Serverless is enabled in the workspace. Terraform team should fix this.

jxbell:
@alexott @nfx , there's three problems...

(1) This PR wasn't supposed to be merged yet. This is for May 3 release.

(2) The suggestion edit I proposed [in this comment](https://github.com/databricks/terraform-provider-databricks/pull/2249#discussion_r1180698237) was not committ3d -- I was waiting to hear back from Alex about @mengqiu-teng  comment before I committed it (which would make the thread resoved by default). That change still needs to be made.

(3) We just had a last minute change to the date range for the API change, it now ends at May 19, 2023, not the 30th.

I will prep a NEW PR for those changes

cc: @amji12 @gaurav-pm @mengqiu-teng

docs/resources/sql_endpoint.md
jxbell: ^^^ @alexott @nfx  see @mengqiu-teng comments above.

Also, btw my suggested change that Meng approved wasn't committed yet, so it's not in this PR.

I've made a secondary PR as followup with that change... https://github.com/databricks/terraform-provider-databricks/pull/2254 -- it also does the date change

PR #2257, opened 2023-05-02
Migrate `databricks_mlflow_model` to go sdk

mberk06:
## Changes

Migrate MLflow model to Go SDK and update testing file.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


nfx (changes requested):
please address the comments from the previous PR review

mberk06:
@nfx I think this is ready to merge - all comments from the prior PR are addressed

mlflow/resource_mlflow_model.go
```diff
@@ -3,112 +3,72 @@ package mlflow
...
 			}
-			d.SetId(m.Name)
+			var req ml.CreateModelRequest
+			common.DataToStructPointer(d, s, &req)
+			res, err := w.ModelRegistry.CreateModel(ctx, req)
+			if err != nil {
+				return err
+			}
+			d.SetId(res.RegisteredModel.Name)
+			d.Set("registered_model_id", res.RegisteredModel.Name) // alias
```
nfx, line 37: didn't we agree it's supposed to be `res.RegisteredModel.Id`? 

see: 

> 	RegisteredModelID string `json:"id,omitempty" tf:"computed,alias:registered_model_id"`

nfx (changes requested):

mlflow/resource_mlflow_model.go
mberk06: > didn't we agree it's supposed to be `res.RegisteredModel.Id`?
> 
> see:
> 
> > RegisteredModelID string `json:"id,omitempty" tf:"computed,alias:registered_model_id"`

yeah but `w.ModelRegistry.CreateModel` returns an [ml.Model](https://github.com/databricks/databricks-sdk-go/blob/main/service/ml/model.go#L800) which doesn't have an ID field, which changed in the [most recent PR](https://github.com/databricks/databricks-sdk-go/pull/370/files#). Also, model name and d.ID should be identical (and are based on the prior line).

nfx: @mberk06 make an additional call to get model

nfx (changes requested):

mberk06:
@nfx update is done. `d.SetId(res.RegisteredModel.Id)` causes the request for a GET stub with empty name (screenshot), so I used `d.SetId(res.RegisteredModel.Name)`
image

mlflow/resource_mlflow_model.go
```diff
@@ -3,112 +3,78 @@ package mlflow
...
-			return m
+		ml.Model{},
+		func(s map[string]*schema.Schema) map[string]*schema.Schema {
+			delete(s, "latest_versions")
+			s["name"] = &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			}
+			return s
```
nfx, line 21: i think you have to add `registered_model_id` as a TypeString schema

nfx (changes requested):

mberk06:
Waiting on [go sdk fix](https://github.com/databricks/databricks-sdk-go/issues/395). Currently the ml model get request returns a nil Registered model.

mberk06:
@nfx thanks for the fix. Acceptance and unit tests are passing.

mberk06:
@alexott @nfx this should be good to merge.

alexott (approved):
lgtm

mlflow/resource_mlflow_model.go
```diff
@@ -3,112 +3,78 @@ package mlflow
...
-			delete(m, "latest_versions")
-			return m
+		ml.Model{},
+		func(s map[string]*schema.Schema) map[string]*schema.Schema {
+			delete(s, "latest_versions")
+			s["name"] = &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			}
```
mgyucht, line 20: The schema should already be `TypeString` if I'm not mistaken. This will ensure that customizations otherwise propagated to the `Schema` of this field will not be overwritten.
```suggestion
			s["name"].Required = true
			s["name"].ForceNew = true
```

mlflow/resource_mlflow_model.go
```diff
@@ -3,112 +3,78 @@ package mlflow
...
+			model, err := w.ModelRegistry.GetModel(ctx, ml.GetModelRequest{
+				Name: res.RegisteredModel.Name,
+			})
+			if err != nil {
 				return err
 			}
-			d.SetId(m.Name)
+			d.SetId(model.RegisteredModelDatabricks.Name)
+			d.Set("registered_model_id", model.RegisteredModelDatabricks.Id) // alias
```
mgyucht, line 43: Is this a bugfix? Is it not already handled in the `Read` implementation below?

mgyucht (commented):
Thanks for contributing this change! Moving to the Go SDK is a thankless task. Two small nits, otherwise we can get this into the next release!

mlflow/resource_mlflow_model.go
mberk06: This commit throws an error. Also specifying a schema without a type throws an error. Will leave as is unless you suggest otherwise

mlflow/resource_mlflow_model.go
mberk06: The `RegisteredModel` object that is returned by the CREATE request does not have an ID in the payload. To get the model ID we have to run a GET. Not sure why it's designed like this but I'll ask

mlflow/resource_mlflow_model.go
mgyucht: Ok. I maybe misunderstood as I thought the name field would have been filled in by StructToSchema already, so the point was to only change the attributes of the schema you care about rather than replacing the entire schema of this field. However, given that this is a string field and these are its only attributes, I feel ok with this as well.

mlflow/resource_mlflow_model.go
mgyucht: Yeah, I saw this as well. I just missed how this worked in the original implementation :p

mgyucht (approved):

Issue #2256, opened 2023-05-02
[ISSUE] Terraform Exporter notebook listing breaks when it encounters special characters and/or backslash in the workspace path


antoniopinheirofilho:
Hi there,

I'm using the Terraform Exporter to migrate a client to a new E2 environment.

The creation of databricks_notebook fails when a notebook or folder has some special characters in their path and display the messages below:

```
[DEBUG] GET /api/2.0/workspace/list?path=/Users/xxx@xxx.com/xxx/aaa/bbb/ccc
< HTTP/2.0 404 Not Found
[ERROR] **databricks_notebook (notebooks service) listing failed**: Path (/Users/xxx@xxx.com/xxx/aaa/bbb/ccc) doesn't exist.
```

In that case, "aaa/bbb/ccc" is a folder name, but since it has backslash in its name, the API gets lost.
Emojis in the path such as ⭐️ and 🔧 also cause the same disruption

I'm wondering if we can do anything in the Terraform Exporter to avoid this error. 

### Steps to Reproduce
Execute `./terraform-provider-databricks_v1.14.3 exporter -skip-interactive -listing=notebooks,repos -debug 2>&1 |tee tr_exporter_logs` in an environment where there are folders or notebooks with special characters and/or backslash in their names.

alexott:
@antoniopinheirofilho for `/` character, for me it's a platform bug - it shouldn't allow to use `/` character in the path - there is a check for notebooks, but not for the directories.

PR #2258, opened 2023-05-03
Exporter: relax handling of problematic files/directories

alexott:
## Changes

The new UI has a bug that allows creation of directories with `/` character in the name -
it breaks the Workspace List API, and prevents exporting of the workspace objects.   The
temporary solution is explicitly add the `ignoreErrors` flag when using exporter.

It's a workaround for #2256

## Tests

- [x] `make test` run locally
- [x] tested manually 
- [ ] using Go SDK


workspace/resource_notebook.go
```diff
@@ -131,10 +131,10 @@ func (a NotebooksAPI) Mkdirs(path string) error {
 // List will list all objects in a path on the workspace
 // and with the recursive flag it will recursively list
 // all the objects
-func (a NotebooksAPI) List(path string, recursive bool) ([]ObjectStatus, error) {
+func (a NotebooksAPI) List(path string, recursive bool, ignoreErrors bool) ([]ObjectStatus, error) {
```
nfx, line 134: please make a PR to SDK

nfx (changes requested):

workspace/resource_notebook.go
alexott: Will do separately - right now it's blocking some PS work... Really I need to think a bit about that List APIs

nfx (approved):

PR #2288, opened 2023-05-10
Exporter: don't emit `databricks_permissions` for `/Shared` directory

### LINKED ISSUE #2269: [ISSUE] Experimental exporter generates permissions for /shared in the workspace
The Experimental exporter generates permissions for the /shared directory in the databricks workspace.

As a result, the `terraform apply` operation tries to apply permissions - such an operation is not supported.

The following resource definition 
```
resource "databricks_permissions" "directory_shared_430507847510155" {
  directory_id = data.databricks_directory.shared_430507847510155.object_id
  access_control {
    group_name       = data.databricks_group.users_54623195191167.display_name
  }
}
```

leads to this error

> Error: cannot create permissions: Cannot modify permissions of directory 357400256613881
>
>   with databricks_permissions.directory_shared_430507847510155,
>   on [access.tf](http://access.tf/) line 68, in resource "databricks_permissions" "directory_shared_430507847510155":
>   68: resource "databricks_permissions" "directory_shared_430507847510155" {
>

alexott:
## Changes
Current Permissions API doesn't allow change of the permissions for the `/Shared` folder, so when we generate a permission resource for it, we get Terraform error.

fixes #2269

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -1202,6 +1202,7 @@ var resourcesMap map[string]importable = map[string]importable{
 				})
 			}
 
+			// TODO: it's not completely correct condition - we need to make emit smarter - emit only if permissions are different from their parent's permission.
 			if ic.meAdmin {
 				ic.Emit(&resource{
 					Resource: "databricks_directory",
@@ -1677,6 +1678,7 @@ var resourcesMap map[string]importable = map[string]importable{
 				if strings.HasPrefix(directory.Path, "/Repos") {
 					continue
 				}
+				// TODO: don't emit directories for deleted users/SPs (how to identify them?)
 				ic.Emit(&resource{
 					Resource: "databricks_directory",
 					ID:       directory.Path,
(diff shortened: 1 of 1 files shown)


exporter/importables.go
```diff
@@ -1202,6 +1202,7 @@ var resourcesMap map[string]importable = map[string]importable{
 				})
 			}
 
+			// TODO: it's not completely correct condition - we need to make emit smarter - emit only if permissions are different...
```
mgyucht, line 1205: Is this comment meant for the block before this?
What do you mean by "permissions are different"? Different from what exactly?

mgyucht: Ah I understand. Could you just change your comment to this answer? It is much clearer :)

alexott: done

PR #2306, opened 2023-05-16
Migrated `databricks_ip_access_list` resource to Go SDK

tanmay-db:
## Changes

- Migrate IP ACL to Go SDK 
- Update and enable integration tests
- Fix tests to use body for updates
- No relevant change is needed to be updated in docs since argument references are the same.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

Integration test: internal/acceptance/ip_access_list_test.go. Full integration test suite is also running successfully.

### DIFF
--- a/access/resource_ip_access_list.go
+++ b/access/resource_ip_access_list.go
@@ -3,6 +3,7 @@ package access
 import (
 	"context"
 
+	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/terraform-provider-databricks/common"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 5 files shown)


internal/acceptance/ip_access_list_test.go
```diff
@@ -4,21 +4,16 @@ import (
...
 )
 
-func TestIPACLListsResourceFullLifecycle(t *testing.T) {
+func TestAccIPACLListsResourceFullLifecycle(t *testing.T) {
 	workspaceLevel(t, step{
 		Template: `
-		resource "databricks_workspace_conf" "features" {
-			enable_ip_access_lists = "true"
-		}
-
```
alexott, line 13: Why is this removed? Do we have IP Access Lists enabled by default now?

tanmay-db: Hi @alexott, thanks for taking a look but the PR is still WIP and not ready for review (I have removed review from eng-dev-ecosystem for the time being). 

This change was suggested by @nfx when we were discussing the tests weren't enabled (i.e Acc isn't present).

internal/acceptance/ip_access_list_test.go
```diff
@@ -4,21 +4,16 @@ import (
...
 
-func TestIPACLListsResourceFullLifecycle(t *testing.T) {
+func TestAccIPACLListsResourceFullLifecycle(t *testing.T) {
 	workspaceLevel(t, step{
 		Template: `
-		resource "databricks_workspace_conf" "features" {
-			enable_ip_access_lists = "true"
-		}
-		  
 		resource "databricks_ip_access_list" "this" {
```
nfx, line 10: please add one more step, where you update this IP access list - e.g. by adding one more IP address range

internal/acceptance/ip_access_list_test.go
nfx: it's fine to remove it, it didn't run before.

nfx (changes requested):

access/resource_ip_access_list.go
```diff
@@ -105,30 +31,47 @@ func ResourceIPAccessList() *schema.Resource {
...
 		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			var iacl ipAccessListUpdateRequest
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var iacl settings.UpdateIpAccessList
 			common.DataToStructPointer(d, s, &iacl)
-			return NewIPAccessListsAPI(ctx, c).Update(d.Id(), iacl)
+			iacl.IpAccessListId = d.Id()
```
tanmay-db, line 66: We need to do this because Update call doesn't take id so we have to pass it through the resource data.
Any reason why we don't pass id through the Update method call?

access/resource_ip_access_list_test.go
```diff
@@ -9,20 +9,21 @@ import (
...
+	TestingId               = "234567"
 	TestingLabel            = "Naughty"
 	TestingListTypeString   = "BLOCK"
-	TestingListType         = "BLOCK"
+	TestingListType         = settings.ListType("BLOCK")
 	TestingEnabled          = true
-	TestingIPAddresses      = []string{"1.2.3.4", "1.2.4.0/24"}
-	TestingIPAddressesState = []any{"1.2.3.4", "1.2.4.0/24"}
+	TestingIpAddresses      = []string{"1.2.3.4", "1.2.4.0/24"}
+	TestingIpAddressesState = []any{"1.2.3.4", "1.2.4.0/24"}
```
tanmay-db, line 26: We should rename this (IP -> Ip) as well for consistency

access/resource_ip_access_list_test.go
```diff
@@ -117,37 +118,43 @@ func TestIPACLUpdate(t *testing.T) {
...
-						Label:         TestingLabel,
-						ListType:      TestingListType,
-						IPAddresses:   TestingIPAddresses,
-						AddressCount:  2,
-						CreatedAt:     87939234,
-						CreatorUserID: 1234556,
-						UpdatedAt:     87939234,
-						UpdatorUserID: 1234556,
-						Enabled:       TestingEnabled,
+				Method:   http.MethodPatch,
```
tanmay-db, line 138: Put -> Patch because: https://docs.databricks.com/api-explorer/workspace/ipaccesslists/update update uses PATCH.  Can this be breaking for customers? Also why was put using query parameters earlier (wasn't taking expected request)? Should put use body?

access/resource_ip_access_list.go
```diff
@@ -3,92 +3,18 @@ package access
...
-
-type IpAccessListStatusWrapper struct {
-	IPAccessList IpAccessListStatus `json:"ip_access_list,omitempty"`
-}
-
 type ipAccessListUpdateRequest struct {
-	Label       string   `json:"label"`
-	ListType    string   `json:"list_type"`
-	IPAddresses []string `json:"ip_addresses"`
-	Enabled     bool     `json:"enabled,omitempty" tf:"default:true"`
```
tanmay-db, line 43: Since all these are same, I don't think updates is needed in docs.

nfx (changes requested):
lgtm for code, make PR title changelog-friendly

PR #2307, opened 2023-05-17
Fixed updates for `databricks_share` resource

### LINKED ISSUE #2167: [ISSUE] Issue with `databricks_share` update detection

### LINKED ISSUE #2240: [ISSUE] Updating the `shared_as` field for tables is not supported yet for `databricks_share` resource
The [API spec](https://docs.databricks.com/api-explorer/workspace/shares/update) is incorrect, `shared_as` is only allowed for new tables in shares at the moment. Therefore, need to fix diff calculation and patch payload

nkvuong:
## Changes

- Improve diff detection for `databricks_share` by selectively comparing certain fields. Also suppress sending `shared_as` for update calls.
- Migrate Create & Delete to Go SDK. Update & Get calls cannot be migrated yet, due to missing alias support.
- Update doc to call out the ordering of `object`
This closes #2240, which should closes #2167

## Tests

Also manually tested locally.

- [x] `make test` run locally
- [x] using Go SDK


nkvuong:
Acceptance tests passed - failed tests are not related to shares

marekbrysa (approved):
I also verified that it worked for me on a real-world case.

catalog/resource_share.go
```diff
@@ -128,19 +120,23 @@ func (si ShareInfo) resourceShareMap() map[string]SharedDataObject {
...
-func Equal(a, b SharedDataObject) bool {
-	if b.SharedAs == "" {
-		b.SharedAs = a.SharedAs
+func (sdo SharedDataObject) Equal(other SharedDataObject) bool {
+	if other.SharedAs == "" {
+		other.SharedAs = sdo.SharedAs
 	}
-	return reflect.DeepEqual(a, b)
+	//don't compare computed fields
+	other.AddedAt = sdo.AddedAt
```
nfx, line 128: this will pollute the state of `other`, leading to unexpected behavior

nfx (changes requested):

catalog/resource_share.go
nkvuong: I thought `other` is passed by value, so won't be modified outside the function?

junwei-db:
@nkvuong Any updates on this PR?

nfx (approved):

PR #2308, opened 2023-05-17
Added missing `serverless` option to PipelineSpec

andrewnester:
## Changes
Added missing `serverless` option to PipelineSpec

## Tests

- [x] `make test` run locally
- [ ] ~covered with integration tests in `internal/acceptance`~
- [ ] ~relevant acceptance tests are passing~
- [ ] ~using Go SDK~

### DIFF
--- a/pipelines/resource_pipeline.go
+++ b/pipelines/resource_pipeline.go
@@ -102,6 +102,7 @@ type PipelineSpec struct {
 	Edition             string            `json:"edition,omitempty" tf:"suppress_diff,default:ADVANCED"`
 	Channel             string            `json:"channel,omitempty" tf:"suppress_diff,default:CURRENT"`
 	Notifications       []Notification    `json:"notifications,omitempty" tf:"alias:notification"`
+	Serverless          bool              `json:"serverless,omitempty"`
 }
 
 type createPipelineResponse struct {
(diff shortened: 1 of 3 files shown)


pipelines/resource_pipeline.go
```diff
@@ -102,6 +102,7 @@ type PipelineSpec struct {
 	Edition             string            `json:"edition,omitempty" tf:"suppress_diff,default:ADVANCED"`
 	Channel             string            `json:"channel,omitempty" tf:"suppress_diff,default:CURRENT"`
 	Notifications       []Notification    `json:"notifications,omitempty" tf:"alias:notification"`
+	Serverless          bool              `json:"serverless,omitempty"`
```
alexott, line 105: Be careful with `omitempty` because if you create a DLT pipeline with serverless, then you won't be able to disable it because the `false` value will be omitted from the payload. it's better to use `json:"serverless" tf:"optional"` instead

alexott (changes requested):

pipelines/resource_pipeline.go
andrewnester: It looks like currently it works because we don't use direct json unmarshalling but reflection here
We just mark such field optional and handle them accordingly

https://github.com/databricks/terraform-provider-databricks/blob/master/common/reflect_resource.go#L88

Correct me if I'm wrong

alexott: yes, it's relying on reflection. The situation with `serverless` flag could be a bit different from the `development` & `photon`. 

Maybe I incorrectly formulated - we'll omit the field from the payload, and it will be applied correctly if control plane will fill the defaults. But we may get to the situation with serverless in DBSQL (#1436) where default depends on the global setting, and if it's enabled globally, we can't disable serverless

alexott (approved):
lgtm

PR #2315, opened 2023-05-19
Remove support for releasing 32-bit binaries

### LINKED ISSUE #2304: [FEATURE] Deprecate 32-bit builds
The use of 32-bit versions is leading to many issues ([provider repo](https://github.com/databricks/terraform-provider-databricks/issues?q=is%3Aissue++label%3A32bit-problem), [StackOverflow](https://stackoverflow.com/search?q=%5Bterraform-provider-databricks%5D+32+bit)) because 64-bit identifiers that are heavily used by the platform aren't handled correctly. 

To avoid users confusion it would be useful to stop releasing of the 32-bit versions.

840:
## Changes

Removes 32-bit builds from .goreleaser.yml

Closes #2304

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,12 +14,7 @@ builds:
     - darwin
   goarch:
     - amd64
-    - '386'
-    - arm
     - arm64
-  ignore:
-    - goos: darwin
-      goarch: '386'
 archives:
 - format: zip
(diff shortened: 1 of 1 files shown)


nfx:
this may require 2.0 version, which needs a bit more comms and preparations than just a PR

nfx (approved):

PR #2324, opened 2023-05-22
Added support for Unity Catalog `databricks_volume` resource

tanmay-db:
## Changes

Added support for UC Volumes - CRUD APIs. 
 

## Tests

Added unit and integration tests 

- [x] `make test` run locally
- [x] using Go SDK


catalog/resource_volumes.go
```diff
@@ -0,0 +1,104 @@
...
+	err = a.client.Get(a.context, "/unity-catalog/volumes/"+url.PathEscape(name), nil, &v)
+	return
+}
+
+func (a VolumesAPI) delete(name string) error {
+	return a.client.Delete(a.context, "/unity-catalog/volumes/"+url.PathEscape(name), nil)
+}
+
+func ResourceVolumes() *schema.Resource {
+	s := common.StructToSchema(VolumeInfo{},
```
nfx, line 74: use `catalog.UpdateVolumeRequestContent` as the schema for the request: https://pkg.go.dev/github.com/databricks/databricks-sdk-go@v0.8.1/service/catalog#UpdateVolumeRequestContent

catalog/resource_volumes_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"testing"
+
+	"github.com/databricks/terraform-provider-databricks/qa"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	TestingName             = "testName"
```
nfx, line 11: don't introduce variables for HTTP tests - they're harder to debug as HTTP fixtures.

instead, do inline those variables.

nfx (changes requested):
- it's not using SDK
- add integration tests with steps to update volume name, owner and comments.

internal/acceptance/volumes_test.go
```diff
@@ -0,0 +1,42 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+func TestAccVolumesResourceFullLifecycle(t *testing.T) {
+	t.Skip("Not running until we have UC Volumes enabled in our aws test workspace")
```
tanmay-db, line 8: We have made a request for this, will check for full lifecycle once we have this capability in our workspace. Skipping for now.

catalog/resource_volumes.go
```diff
@@ -0,0 +1,83 @@
...
+	Name string `json:"name"`
+	// The identifier of the user who owns the volume
+	Owner      string             `json:"owner,omitempty"`
+	VolumeType catalog.VolumeType `json:"volume_type"`
+	// The storage location on the cloud
+	StorageLocation string `json:"storage_location,omitempty"`
+}
+
+func ResourceVolumes() *schema.Resource {
+	// We cannot use catalog.UpdateVolumeRequestContent because it doesn't contain all the necessary fields
```
tanmay-db, line 29: Hi @nkvuong, @nfx suggested to use `catalog.UpdateVolumeRequestContent` but not sure how we can use that here since catalog_name and schema_name for example are required.

catalog/resource_volumes.go
nkvuong: I don't think it is possible to move a volume to a new catalog & schema, which is why these parameters are not part of the update request

docs/resources/volumes.md
```diff
@@ -0,0 +1,61 @@
...
+* `storage_location` - (Optional) If `EXTERNAL` volume type is used, then location of that volume
+* `comment` - (Optional) User-supplied free-form text.
+
+## Import
+
+This resource can be imported by `name`
+
+```bash
+$ terraform import databricks_volumes.this <name>
+```
```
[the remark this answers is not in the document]
nkvuong: you could write it as <catalog_name>.<schema_name>.<volume_name>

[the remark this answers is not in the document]
tanmay-db: Thanks updated

docs/resources/volumes.md
```diff
@@ -0,0 +1,61 @@
+---
+subcategory: "Unity Catalog"
+---
+# databricks_volumes (Resource)
+Volumes are a new Unity Catalog (UC) capability for accessing, storing, governing, and organizing files. Volumes unlock new processing capabilities for data governed by the Unity Catalog, including support for most machine learning and data science workloads. You can use volumes to store and access files in any format; data can be structured, semi-structured, or unstructured.
+
+With Volumes, files managed centrally in Unity Catalogare are organized under a 3-level namespace: `<catalog>.<schema>.<volume>`.
```
tanmay-db, line 7: Hi @bogdanrdc, I have used these from the doc, please let me know if these need to be changed or if you have any suggestions. Thank you.

nkvuong (approved):
went through this with @tanmay-db  and all suggested changes are made. integration tests are also passing

tanmay-db:
Thanks @nkvuong for the review. Hi @nfx, I cannot merge because although we have approval, it says:
`1 review requesting changes and 1 approving review by reviewers with write access. `

Doing a dismiss review to unblock.

internal/acceptance/volume_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+)
+
+func TestUcAccVolumesResourceWithoutInitialOwnerAWSFullLifecycle(t *testing.T) {
+	GetEnvOrSkipTest(t, "TEST_BUCKET")
```
tanmay-db, line 11: Not sure if there is a better way for filtering specific cloud with UC and UC Volumes enabled.

pietern: You could set a separate var on the env to filter on. Would this in its current form fail in any of our nightly test environments?

catalog/resource_volume.go
```diff
@@ -0,0 +1,98 @@
...
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var updateVolumeRequestContent catalog.UpdateVolumeRequestContent
+			common.DataToStructPointer(d, s, &updateVolumeRequestContent)
+			updateVolumeRequestContent.FullNameArg = d.Id()
+			v, err := w.Volumes.Update(ctx, updateVolumeRequestContent)
+			d.SetId(v.FullName)
```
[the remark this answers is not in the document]
tanmay-db: We need this in update because `name` is updatable and is part of the `full_name` which is used for API calls. If we don't update the FullName then the API calls (post update) will be made to the previous name. Example:

FullName = a.b.c
Name updated from c to d
FullName now is a.b.d but if we don't update it in the Resource Data then the next calls will be made to a.b.c which doesn't exist.

@nkvuong and me discussed this in our meet and one reason why this might not have been required till now is because in other such cases we use `tf:"force_new"` annotation but we shouldn't create and destroy resource when we can just update the ID in resource data. This would need to be done for other resources as well (maybe when they are being migrated).

internal/acceptance/volume_test.go
tanmay-db: > Would this in its current form fail in any of our nightly test environments?
No, this in its current form won't fail any of the nightly test environments because `"TEST_BUCKET"` only exists for one environment currently but this is fragile for short term (in case "TEST_BUCKET" is added to some other environment), in long term I think UC Volumes would be available in UC enabled prod workspaces so we might not required special filtering based on this.

Do you suggest creating a separate variable for this in short term or leave as is? We would need to remove that separate variable once Volumes is default UC enabled so it's an extra overhead.

catalog/resource_volume.go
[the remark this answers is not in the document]
tanmay-db: Added a comment locally for this, will push with other changes.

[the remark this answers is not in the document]
pietern: Understood, thank you. Please add a distilled version of this comment to the call site for future readers.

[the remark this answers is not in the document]
tanmay-db: Added the comment.

catalog/resource_volume.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type UpdateVolumeRequestContent struct {
```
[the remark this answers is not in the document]
tanmay-db: Updated to `VolumeInfo`

pietern (approved):
Please confirm the integration tests ran against the latest version of this PR.

tanmay-db:
> Please confirm the integration tests ran against the latest version of this PR.

Confirmed - integration tests are passing on latest version of PR.

catalog/resource_volume.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// This structure contains the fields of catalog.UpdateVolumeRequestContent and catalog.CreateVolumeRequestContent
+// We need to create this because we need Owner, FullNameArg, SchemaName and CatalogName which aren't present in a single of them.
+// We also need to annotate tf:"computed" for the Owner field.
+type VolumeInfo struct {
+	// The name of the catalog where the schema and the volume are
+	CatalogName string `json:"catalog_name"`
```
nfx, line 16: please run `make fmt`

catalog/resource_volume.go
```diff
@@ -0,0 +1,100 @@
...
+			var createVolumeRequestContent catalog.CreateVolumeRequestContent
+			common.DataToStructPointer(d, s, &createVolumeRequestContent)
+			v, err := w.Volumes.Create(ctx, createVolumeRequestContent)
+			if err != nil {
+				return err
+			}
+			d.SetId(v.FullName)
+
+			// Update owner if it is provided
+			if d.Get("owner") != "" {
```
nfx, line 52: switch the logic a bit for the sake of readability:

```go
if d.Get("owner") == "" {
  return nil
}

var updateVolumeRequestContent catalog.UpdateVolumeRequestContent
...

```

catalog/resource_volume.go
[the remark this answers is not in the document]
nfx: is there a test that renames it?

catalog/resource_volume_test.go
```diff
@@ -0,0 +1,440 @@
...
+					CatalogName: "testCatalogName",
+					SchemaName:  "testSchemaName",
+					Comment:     "This is a test comment.",
+					FullName:    "testCatalogName.testSchemaName.testName",
+					Owner:       "InitialOwner",
+				},
+			},
+			{
+				Method:   http.MethodGet,
+				Resource: "/api/2.1/unity-catalog/volumes/" + "testCatalogName.testSchemaName.testName" + "?",
```
nfx, line 44: `				Resource: "/api/2.1/unity-catalog/volumes/" + "testCatalogName.testSchemaName.testName" + "?",`

remove the `" + "`, hinders the greppability

docs/resources/volume.md
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+resource "databricks_volume" "this" {
+  name = "quickstart_volume"
+  catalog_name = databricks_catalog.sandbox.name
+  schema_name = databricks_schema.things.name 
+  owner = "volume_owner"
+  volume_type = "EXTERNAL"
+  storage_location   = ""
```
nfx, line 38: include how to create `storage_location` in this example

internal/acceptance/volume_test.go
nfx: we don't `GetEnvOrSkipTest` in the beginning of the test.

internal/acceptance/volume_test.go
```diff
@@ -0,0 +1,143 @@
...
+	randomName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	unityWorkspaceLevel(t, step{
+		Template: fmt.Sprintf(`
+		resource "databricks_schema" "this" {
+			name 		 = "schema-%[1]s"
```
nfx, line 16: use `{var.STICKY_RANDOM}` instead of `fmt.Sprintf`.

internal/acceptance/volume_test.go
nfx: it's be skipped by `{env.TEST_BUCKET}`

nfx (commented):
please address the changes

nfx (commented):
please address the changes

catalog/resource_volume.go
tanmay-db: This is not required, it's no-op and fmt test was passed https://github.com/databricks/terraform-provider-databricks/actions/runs/5131170653/jobs/9230902643

tanmay-db:
@nfx we use GetEnvOrSkipTest for other tests: https://github.com/databricks/terraform-provider-databricks/blob/master/internal/acceptance/mws_networks_test.go#L8

what is the reason for:
> we don't GetEnvOrSkipTest in the beginning of the test.

tanmay-db:
> is there a test that renames it?
@nfx yes please take a look at the integration tests that updates it: https://github.com/databricks/terraform-provider-databricks/blob/master/internal/acceptance/volume_test.go#L66

PR #2331, opened 2023-05-23
Added `databricks_model_serving` support to `databricks_permissions`

### LINKED ISSUE #2333: [FEATURE] Add support for model serving endpoints in `databricks_permissions`
Hi there,

Feature Request Proposal: Functionality to Add User , Groups and service principals to Model Serving Points using Databricks Terraform

### Use-cases
The ability to add user groups to model serving point using Databricks Terraform would allow users to more easily manage access to their models. Currently, the only way to add user groups to model serving point is through the Databricks UI. This can be a time-consuming and error-prone process, especially for large organizations with many user groups.

### Proposal

We propose extending the Databricks Terraform provider with a new resource type that allows users to define and manage user groups for model serving points. This new resource will provide the following functionality:

a. Ability to define user groups: Users will be able to define user groups within the Terraform configuration, specifying the desired group name, members, and associated permissions.

b. Association of user groups with model serving points: Users can assign one or more user groups to a model serving point, granting the defined group members access to the serving point. This association will ensure that the specified user groups have the necessary permissions to interact with the models.

c. Support for dynamic updates: The Terraform configuration should support dynamic updates to user groups and their associations with model serving points. This allows for easy modifications and adjustments as user groups evolve over time.

alexott:
## Changes
To support permissions assignment we also need to expose a new attribute with internal ID of the model serving endpoint.

## Tests

- [x] `make test` run locally
- [x] tested manually
- [x] using Go SDK

fixes #2333

### DIFF
--- a/docs/resources/model_serving.md
+++ b/docs/resources/model_serving.md
@@ -72,7 +72,7 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `id` - Serving endpoint identifier (equal to `name` argument).
+* `id` - Equal to the `name` argument and used to identify the serving endpoint.
 * `serving_endpoint_id` - Unique identifier of the serving endpoint primarily used to set permissions and refer to this instance for other operations.
 
 ## Access Control
--- a/docs/resources/model_serving.md
+++ b/docs/resources/model_serving.md
@@ -73,7 +73,7 @@ The following arguments are supported:
 In addition to all arguments above, the following attributes are exported:
 
 * `id` - Serving endpoint identifier (equal to `name` argument).
-* `serving_endpoint_id` - unique identifier of the serving endpoint that could be used to set permissions and other operations.
+* `serving_endpoint_id` - Unique identifier of the serving endpoint primarily used to set permissions and refer to this instance for other operations.
 
 ## Access Control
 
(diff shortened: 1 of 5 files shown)


docs/resources/model_serving.md
```diff
@@ -68,6 +68,17 @@ The following arguments are supported:
 * `served_model_name` - (Required) The name of the served model this route configures traffic for. This needs to match the name of a `served_models` block
 * `traffic_percentage` - (Required) The percentage of endpoint traffic to send to this route. It must be an integer between 0 and 100 inclusive.
 
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - Serving endpoint identifier (equal to `name` argument).
```
arpitjasa-db, line 75:
```suggestion
* `id` - Equal to the `name` argument and used to identify the serving endpoint.
```
Just making it more clear that this attribute is effectively a dupe and to make it easier to distinguish from `serving_endpoint_id`.

docs/resources/model_serving.md
```diff
@@ -68,6 +68,17 @@ The following arguments are supported:
...
+* `serving_endpoint_id` - unique identifier of the serving endpoint that could be used to set permissions and other operations.
```
arpitjasa-db, line 76:
```suggestion
* `serving_endpoint_id` - Unique identifier of the serving endpoint primarily used to set permissions and refer to this instance for other operations.
```

internal/acceptance/model_serving_test.go
```diff
@@ -122,6 +122,15 @@ func TestAccModelServing(t *testing.T) {
 					}
 				}
 			}
+
```
arpitjasa-db, line 125: I was informed this test is broken, so I'm surprised it works now?

alexott: I just modified code, code is still broken :-)

nfx (approved):

PR #2336, opened 2023-05-24
Add suppress diff for URL change that only change `/` (remove or add) in UC resources

### LINKED ISSUE #2335: [ISSUE] Issue with `databricks_catalog` and `databricks_external_location` resource and `storage_root`
Using both the 1.16.x and 1.17.x provider we are seeing that on every terraform plan that both catalogs and external locations are being recreated (catalogs) or updated (external locations) even when there are no changes to those resources. It looks like somewhere a trailing "/" is added to the storage_root and then terraform is wanting to change that to remove the trailing "/"

```text
  # databricks_catalog.business_units["enverus-databricks-unity-ba-ue1-dev"] must be replaced
-/+ resource "databricks_catalog" "business_units" {
      ~ id            = "sample_catalog_dev" -> (known after apply)
        name          = "sample_catalog_dev"
      ~ storage_root  = "s3://sample-catalog-dev/" -> "s3://sample-catalog-dev" # forces replacement
        # (4 unchanged attributes hidden)
    }
```

alexott:
## Changes

UC Engineering team changed a way how the storage URLs are normalized, and this caused an issue with existing UC resources (catalogs, schemas, ...) that used URLs like `s3://bucket` instead of `s3://bucket/`

This fixes #2335

## Tests

- [x] `make test` run locally
- [x] tested manually
- [x] relevant acceptance tests are passing - tested with `aws-prod-ucws`
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -3,10 +3,19 @@ package catalog
 import (
 	"context"
 	"fmt"
+	"log"
 
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
+
+func ucDirectoryPathSuppressDiff(k, old, new string, d *schema.ResourceData) bool {
+	if (new == (old + "/")) || (old == (new + "/")) {
+		log.Printf("[DEBUG] Ignoring configuration drift from %s to %s", old, new)
+		return true
+	}
+	return false
+}
 
 type CatalogsAPI struct {
 	client  *common.DatabricksClient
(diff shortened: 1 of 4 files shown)


catalog/resource_catalog.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func ucDirectoryPathSuppressDiff(k, old, new string, d *schema.ResourceData) bool {
+	return (new == (old + "/")) || (old == (new + "/"))
+}
+
```
tanmay-db, line 14: Thanks for the PR, can we add a comment mentioning why this was required? and add a log similar to the one below if it's true:

`log.Printf("[DEBUG] Ignoring configuration drift from %s to %s", old, new)`

tanmay-db (approved):

PR #2337, opened 2023-05-24
Add Pulumi user agent support

AaronFriel:
## Changes

Per a request via partnership, this enables the Pulumi Databricks provider to use a distinct user agent string, via setting `IsPulumi = true` via `ldflags` or a minimal patch.

If you'd prefer, we could modify the code to this instead:

```go
// IMPORTANT: this line cannot be changed, because it's used for
// internal purposes at Databricks.
var BaseUserAgent string = "databricks-tf-provider"

func init() {
	useragent.WithProduct(BaseUserAgent, common.Version())
}
```

But this is, I think, more fragile in that your code wouldn't contain the other Pulumi UA string and it might not be obvious to future engineers why this is a variable instead of a const, why it is factored out of `init()`, and so on.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -37,10 +37,16 @@ import (
 	"github.com/databricks/terraform-provider-databricks/workspace"
 )
 
+var IsPulumi string = "false"
+
 func init() {
 	// IMPORTANT: this line cannot be changed, because it's used for
 	// internal purposes at Databricks.
-	useragent.WithProduct("databricks-tf-provider", common.Version())
+	BaseUserAgent := "databricks-tf-provider"
+	if IsPulumi == "true" {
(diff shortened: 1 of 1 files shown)


AaronFriel:
I've been informed by that we could instead call `databricks.WithUserAgentExtra("pulumi", "x")`, if that's preferable, we'll proceed with that.

nfx:
@AaronFriel `databricks.WithUserAgentExtra` is more preferable

nfx (changes requested):
please use [`useragent.WithUserAgentExtra`](https://pkg.go.dev/github.com/databricks/databricks-sdk-go@v0.9.0/useragent#WithUserAgentExtra)

PR #2341, opened 2023-05-25
[WIP] Added support for File IO in Terraform

tanmay-db:
## Changes

Draft PR for discussion with Files IO team, it doesn't use Go SDK because its not present in Open API spec.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/storage/resource_files.go
+++ b/storage/resource_files.go
@@ -1,0 +1,90 @@
+package storage
+
+import (
+	"context"
+	"fmt"
+	"github.com/databricks/terraform-provider-databricks/common"
+)
+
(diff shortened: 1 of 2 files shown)


alexott:
Why not via Go SDK?

tanmay-db:
> Why not via Go SDK?

Hi @alexott This is currently not part of Open API Spec.

alexott:
Another thing that we need to consider is resource name - the `databricks_file` could be too generic, and people will mix up it with `databricks_dbfs_file` & `databricks_workspace_file`...

PR #2357, opened 2023-05-31
Migrate resource `databricks_catalog` to Go SDK

nkvuong:
## Changes

Migrate `databricks_catalog` resource to Go SDK.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/databricks/databricks-sdk-go/service/catalog"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -17,17 +18,8 @@ func ucDirectoryPathSuppressDiff(k, old, new string, d *schema.ResourceData) boo
 	return false
 }
 
-type CatalogsAPI struct {
-	client  *common.DatabricksClient
-	context context.Context
-}
-
-func NewCatalogsAPI(ctx context.Context, m any) CatalogsAPI {
-	return CatalogsAPI{m.(*common.DatabricksClient), context.WithValue(ctx, common.Api, common.API_2_1)}
-}
-
 type CatalogInfo struct {
-	Name         string            `json:"name" tf:"force_new"`
+	Name         string            `json:"name"`
 	Comment      string            `json:"comment,omitempty"`
 	StorageRoot  string            `json:"storage_root,omitempty" tf:"force_new"`
 	ProviderName string            `json:"provider_name,omitempty" tf:"force_new,conflicts:storage_root"`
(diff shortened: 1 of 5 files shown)


catalog/resource_catalog.go
```diff
@@ -80,39 +40,77 @@ func ResourceCatalog() *schema.Resource {
...
+			if err != nil {
+				return err
+			}
+			var updateCatalogRequest catalog.UpdateCatalog
+			common.DataToStructPointer(d, catalogSchema, &updateCatalogRequest)
+			ci, err := w.Catalogs.Update(ctx, updateCatalogRequest)
+			// We need to update the resource data because Name is updatable
+			// So if we don't update the field then the requests would be made to old Name which doesn't exists.
+			d.SetId(ci.Name)
+			if err != nil {
```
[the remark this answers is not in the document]
nkvuong: changed the code to only set the new Id after a successful update

catalog/resource_catalog.go
```diff
@@ -17,56 +18,16 @@ func ucDirectoryPathSuppressDiff(k, old, new string, d *schema.ResourceData) boo
...
-		return err
-	}
-	for _, s := range schemas.Schemas {
-		if s.Name != name+".information_schema" {
-			schemasAPI.forceDeleteSchema(s.FullName)
-		}
-	}
-
-	return a.client.Delete(a.context, "/unity-catalog/catalogs/"+name, nil)
+	Name          string            `json:"name"`
```
fjakobs, line 22: isn't the `tf:"force_new"` missing here?

catalog/resource_catalog.go
```diff
@@ -80,39 +41,79 @@ func ResourceCatalog() *schema.Resource {
...
 		},
-		Update: update,
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var updateCatalogRequest catalog.UpdateCatalog
+			common.DataToStructPointer(d, catalogSchema, &updateCatalogRequest)
+			ci, err := w.Catalogs.Update(ctx, updateCatalogRequest)
```
alexott, line 99: How it will work when the name is already updated in resource - as I see in the docs, we address catalog by name: https://docs.databricks.com/api/workspace/catalogs/update, so if the name in resource is updated, it won't find an old one to change it.

nkvuong: if the change happened out of bound, then there is not much we can do. what happens with the update operation here is that we call the Update API, update the resource Id to the new name, so that the Get API will use the new name and succeed

nkvuong: credit to @tanmay-db for getting this to worked with the Volume resource, so I am just backporting the fix

catalog/resource_catalog.go
nkvuong: catalog name is updatable through our API but we force recreating the resource because the provider was not able to handle the change in name. Same explanation as the below's comment from Alex

catalog/resource_catalog.go
alexott: If we allow name change, then SDK needs to support that pattern. Inside provider we can find if the name field has changed, and then use previous name (the `d.Id()`) in the URL

nkvuong: we need to do the same for all the UC resources then, as they are all addressed by names.

currently volume handles it like this https://github.com/databricks/terraform-provider-databricks/blob/master/catalog/resource_volume.go#L85

catalog/resource_catalog.go
```diff
@@ -17,56 +18,16 @@ func ucDirectoryPathSuppressDiff(k, old, new string, d *schema.ResourceData) boo
...
-type CatalogsAPI struct {
-	client  *common.DatabricksClient
-	context context.Context
-}
-
-func NewCatalogsAPI(ctx context.Context, m any) CatalogsAPI {
-	return CatalogsAPI{m.(*common.DatabricksClient), context.WithValue(ctx, common.Api, common.API_2_1)}
-}
-
 type CatalogInfo struct {
```
nfx, line 21: @nkvuong did you try using UpdateCatalog for the schema this way we won't have to worry about the new fields?...

catalog/resource_catalog.go
```diff
@@ -17,56 +18,16 @@ func ucDirectoryPathSuppressDiff(k, old, new string, d *schema.ResourceData) boo
...
-			schemasAPI.forceDeleteSchema(s.FullName)
-		}
-	}
-
-	return a.client.Delete(a.context, "/unity-catalog/catalogs/"+name, nil)
+	Name          string            `json:"name"`
+	Comment       string            `json:"comment,omitempty"`
+	StorageRoot   string            `json:"storage_root,omitempty" tf:"force_new"`
+	ProviderName  string            `json:"provider_name,omitempty" tf:"force_new,conflicts:storage_root"`
+	ShareName     string            `json:"share_name,omitempty" tf:"force_new,conflicts:storage_root"`
```
nfx, line 26: this can go into schema callback.

nfx (approved):

PR #2364, opened 2023-06-01
Added support for Unity Catalog `databricks_catalog_workspace_binding` resource

### LINKED ISSUE #2312: [FEATURE] Catalog-Workspace Bindings
### Use-cases

- Optionally binding created catalogs to specific workspaces
- Retrieving a a list of workspaces bound to a catalog
- Retrieve a list of all catalog-workspace bindings

### Attempted Solutions

N/A

### Proposal

Support for the new [Workspace Bindings API](https://docs.databricks.com/api-explorer/workspace/workspacebindings).

There are a few options for implementation but I see these as the most reasonable ones:

1. Add a new argument to the `databricks_catalog` resource
    - Advantages:
        - No new resources for users to keep track of
        - Ensuring that the api will always be invoked in a Unity Catalog context
    - Disadvantages:
        - Tight coupling with the catalog API which could potentially lead to issues down the road if, for example, the catalog api is enabled at account-level while the binding api isn't
        - Limited access to the binding api in that the catalog has to be created using Terraform to be able to also bind it using Terraform
        - There isn't a data source currently to retrieve details about catalog(s), apart from ids, which would mean the GET method has no consistent implementation for now.

2. Create a new resource - For example, `databricks_catalog_binding`
    - Advantages:
        - No coupling with other resources
        - Would be more consistent with a similar data source to retrieve catalog bindings. e.g. `data "databricks_catalog_binding"` and `data "databricks_catalog_bindings"`
    - Disadvantages:
        - Being a separate resource, it will require more arguments and checks compared to the first option such as specifying the metastore and catalog ids. 

3. Supporting both of the above options in a non-exclusive manner. The AWS provider implements a similar strategy for policy attachments in [`aws_iam_role_policy_attachment`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role_policy_attachment) and [`aws_iam_role`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role#managed_policy_arns)
    - Advantages:
        - Best of both worlds - Allowing flexibility for users to bind the catalogs as needed in at any point in the stack.
    - Disadvantages:
        - More work to implement
        - More user options = More developer issues

ehab-eb:
## Changes

Closes #2312 

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strconv"
 
 	"github.com/databricks/databricks-sdk-go/service/catalog"
 	"github.com/databricks/terraform-provider-databricks/common"
@@ -73,6 +74,26 @@ func ResourceCatalog() *schema.Resource {
 			_, err = w.Catalogs.Update(ctx, updateCatalogRequest)
 			if err != nil {
 				return err
+			}
+
+			// Bind the current workspace if the catalog is isolated, otherwise the read will fail
+			if d.Get("isolation_mode") == "ISOLATED" {
+				currentMetastoreAssignment, err := catalog.NewMetastores(c.DatabricksClient).Current(ctx)
+				if err != nil {
+					return err
+				}
+				currentWorkspaceId, err := strconv.ParseInt(currentMetastoreAssignment.WorkspaceId, 10, 64)
+				if err != nil {
+					return err
+				}
+				createBindingRequest := catalog.UpdateWorkspaceBindings{
+					Name:             ci.Name,
+					AssignWorkspaces: []int64{currentWorkspaceId},
+				}
+				_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, createBindingRequest)
+				if err != nil {
+					return err
+				}
 			}
 
 			return nil
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -80,7 +80,7 @@ func ResourceCatalog() *schema.Resource {
 				return nil
 			}
 			// Bind the current workspace if the catalog is isolated, otherwise the read will fail
-			currentMetastoreAssignment, err := catalog.NewMetastores(c.DatabricksClient).Current(ctx)
+			currentMetastoreAssignment, err := w.Metastores.Current(ctx)
 			if err != nil {
 				return err
 			}
@@ -88,11 +88,10 @@ func ResourceCatalog() *schema.Resource {
 			if err != nil {
 				return err
 			}
-			createBindingRequest := catalog.UpdateWorkspaceBindings{
+			_, err = w.WorkspaceBindings.Update(ctx, catalog.UpdateWorkspaceBindings{
 				Name:             ci.Name,
 				AssignWorkspaces: []int64{currentWorkspaceId},
-			}
-			_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, createBindingRequest)
+			})
 			if err != nil {
 				return err
 			}
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -76,24 +76,25 @@ func ResourceCatalog() *schema.Resource {
 				return err
 			}
 
+			if d.Get("isolation_mode") != "ISOLATED" {
+				return nil
+			}
 			// Bind the current workspace if the catalog is isolated, otherwise the read will fail
-			if d.Get("isolation_mode") == "ISOLATED" {
-				currentMetastoreAssignment, err := catalog.NewMetastores(c.DatabricksClient).Current(ctx)
-				if err != nil {
-					return err
-				}
-				currentWorkspaceId, err := strconv.ParseInt(currentMetastoreAssignment.WorkspaceId, 10, 64)
-				if err != nil {
-					return err
-				}
-				createBindingRequest := catalog.UpdateWorkspaceBindings{
-					Name:             ci.Name,
-					AssignWorkspaces: []int64{currentWorkspaceId},
-				}
-				_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, createBindingRequest)
-				if err != nil {
-					return err
-				}
+			currentMetastoreAssignment, err := catalog.NewMetastores(c.DatabricksClient).Current(ctx)
+			if err != nil {
+				return err
+			}
+			currentWorkspaceId, err := strconv.ParseInt(currentMetastoreAssignment.WorkspaceId, 10, 64)
+			if err != nil {
+				return err
+			}
+			createBindingRequest := catalog.UpdateWorkspaceBindings{
+				Name:             ci.Name,
+				AssignWorkspaces: []int64{currentWorkspaceId},
+			}
+			_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, createBindingRequest)
+			if err != nil {
+				return err
 			}
 
 			return nil
--- a/catalog/resource_catalog_workspace_binding.go
+++ b/catalog/resource_catalog_workspace_binding.go
@@ -20,11 +20,10 @@ func ResourceCatalogWorkspaceBinding() *schema.Resource {
 			if err != nil {
 				return err
 			}
-			createBindingRequest := catalog.UpdateWorkspaceBindings{
+			_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, catalog.UpdateWorkspaceBindings{
 				Name:             catalogName,
 				AssignWorkspaces: []int64{i64WorkspaceId},
-			}
-			_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, createBindingRequest)
+			})
 			return err
 		},
 		ReadContext: func(ctx context.Context, catalogName, workspaceId string, c *common.DatabricksClient) error {
@@ -39,18 +38,17 @@ func ResourceCatalogWorkspaceBinding() *schema.Resource {
 			if !contains(bindings.Workspaces, i64WorkspaceId) {
 				return apierr.NotFound("Catalog has no binding to this workspace")
 			}
-			return err
+			return nil
 		},
 		DeleteContext: func(ctx context.Context, catalogName, workspaceId string, c *common.DatabricksClient) error {
 			i64WorkspaceId, err := strconv.ParseInt(workspaceId, 10, 64)
 			if err != nil {
 				return err
 			}
-			removeBindingRequest := catalog.UpdateWorkspaceBindings{
+			_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, catalog.UpdateWorkspaceBindings{
 				Name:               catalogName,
 				UnassignWorkspaces: []int64{i64WorkspaceId},
-			}
-			_, err = catalog.NewWorkspaceBindings(c.DatabricksClient).Update(ctx, removeBindingRequest)
+			})
 			return err
 		},
 	})
--- a/catalog/resource_catalog_workspace_binding.go
+++ b/catalog/resource_catalog_workspace_binding.go
@@ -10,15 +10,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func contains_int64(s []int64, e int64) bool {
-	for _, a := range s {
-		if a == e {
-			return true
-		}
-	}
-	return false
-}
-
 func ResourceCatalogWorkspaceBinding() *schema.Resource {
 	return common.NewPairID("catalog_name", "workspace_id").Schema(func(
 		m map[string]*schema.Schema) map[string]*schema.Schema {
@@ -45,7 +36,7 @@ func ResourceCatalogWorkspaceBinding() *schema.Resource {
 			if err != nil {
 				return err
 			}
-			if !contains_int64(bindings.Workspaces, i64WorkspaceId) {
+			if !contains(bindings.Workspaces, i64WorkspaceId) {
 				return apierr.NotFound("Catalog has no binding to this workspace")
 			}
 			return err
--- a/catalog/resource_catalog_workspace_binding.go
+++ b/catalog/resource_catalog_workspace_binding.go
@@ -4,15 +4,11 @@ import (
 	"context"
 	"strconv"
 
+	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/service/catalog"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
-
-type catalogWorkspaceBindingInfo struct {
-	Name      string `json:"name" tf:"force_new"`
-	Workspace int64  `json:"workspace" tf:"force_new"`
-}
 
 func contains_int64(s []int64, e int64) bool {
 	for _, a := range s {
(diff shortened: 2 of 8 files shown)


ehab-eb:
@nkvuong It's far from done but I though I'd open a draft PR so you can see the direction I'm going in. 

The main challenge so far is that the terraform schema doesn't match the API schema so I'm trying to figure out the best way to work around that.

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			workspaces := []int64{d.Get("Workspace").(int64)}
+			var createBindingRequest catalog.UpdateWorkspaceBindings
+			createBindingRequest.UnassignWorkspaces = workspaces
+			d.Set("UnassignWorkspaces", d.Get("Workspace"))
```
nkvuong, line 57: don't need to do this

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+			common.DataToStructPointer(d, s, &createBindingRequest)
+			w.WorkspaceBindings.Update(ctx, createBindingRequest)
```
nkvuong, line 59: you need the below to capture error
```
_, err = w.WorkspaceBindings.Update(ctx, createBindingRequest)
```

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+		Schema: s,
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			workspaces := []int64{d.Get("Workspace").(int64)}
+			var createBindingRequest catalog.UpdateWorkspaceBindings
+			createBindingRequest.AssignWorkspaces = workspaces
+			d.Set("AssignWorkspaces", d.Get("Workspace"))
```
nkvuong, line 31: don't need this

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+			common.DataToStructPointer(d, s, &createBindingRequest)
```
nkvuong, line 32: don't need this either, since you've already constructed the request manually

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+			w.WorkspaceBindings.Update(ctx, createBindingRequest)
```
nkvuong, line 33: you need the below to capture error
```
_, err = w.WorkspaceBindings.Update(ctx, createBindingRequest)
```

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+			// FIX: Need to set an id that can identify a unique workspace-catalog combination
+			return nil
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			// FIX: This will return a full list of bound workspaces
+			w.WorkspaceBindings.GetByName(ctx, d.Id())
```
nkvuong, line 46: you will need to unpack the Id to get the catalog name to provide this method, and if the workspace id is not in the list, then we need to mark the resource as new, i.e. to be created next time.

catalog/resource_catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestCatalogWorkspaceBindingsCornerCases(t *testing.T) {
+	qa.ResourceCornerCases(t, ResourceCatalogWorkspaceBinding())
+}
+
+func TestCatalogWorkspaceBindingsAssign(t *testing.T) {
+	resource := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
```
nkvuong, line 16: arrange the fixture in the order they would be call

catalog/resource_catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestCatalogWorkspaceBindingsCornerCases(t *testing.T) {
+	qa.ResourceCornerCases(t, ResourceCatalogWorkspaceBinding())
+}
+
+func TestCatalogWorkspaceBindingsAssign(t *testing.T) {
```
nkvuong, line 14:
```suggestion
func TestCatalogWorkspaceBindings_Create(t *testing.T) {
```

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			workspaces := []int64{d.Get("Workspace").(int64)}
+			var createBindingRequest catalog.UpdateWorkspaceBindings
+			createBindingRequest.UnassignWorkspaces = workspaces
+			d.Set("UnassignWorkspaces", d.Get("Workspace"))
+			common.DataToStructPointer(d, s, &createBindingRequest)
```
nkvuong, line 58: don't need this either, as you're already constructing the request manually

nkvuong (commented):
Leave a couple of comments on how to proceed

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type CatalogWorkspaceBindingInfo struct {
```
nfx, line 11: un-export it: start name with lowercase

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+	Name      string `json:"name" tf:"force_new"`
+	Workspace int64  `json:"workspace" tf:"force_new"`
+}
+
+func ResourceCatalogWorkspaceBinding() *schema.Resource {
```
nfx, line 16: please use bind resource with the pair id. see example at https://github.com/databricks/terraform-provider-databricks/blob/master/scim/resource_group_role.go#L14-L30

nfx (changes requested):
1. use BindResource
2. add acceptance test
3. add documentation

ehab-eb:
Thank you both for the feedback! I've addressed most of it but I'm running into issues [here](https://github.com/databricks/terraform-provider-databricks/pull/2364/files#diff-8d3f13784519d57561fe1c9dcde1274dea6d2612c7299b0869f387de447265e3R54) with the pair id.

The workspace id is defined as `int64` and the `Unpack` function is expecting `string`. Since the unpacking infers the types from the schema I'm not entirely sure how to convert the type without affecting the terraform schema data types. Any advice?

nfx:
@ehab-eb use `BindResource` and convert between int64/string within the methods.

ehab-eb:
@nfx @nkvuong I've updated the code to use `BindResource` and the tests are now passing! Feels like the tests aren't covering all bases though, so I'm happy to discuss what other cases to implement.

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func contains_int64(s []int64, e int64) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
+}
```
ehab-eb, line 20: Is this the right place & name for this function?

ehab-eb: I can get rid of this function if we change the [`contains`](https://github.com/databricks/terraform-provider-databricks/blob/48e7be82bb1c77d99879b78281e61c335a2b6fae/catalog/update.go#L11-L18) function to use generics. For example,
```go
func contains[T comparable](s []T, e T) bool {
	for _, a := range s {
		if a == e {
			return true
		}
	}
	return false
}
```

nkvuong: it makes sense to change that function to use generics - it is only used in a single place within the catalog package

ehab-eb: Awesome! I can do that in a separate PR to keep the concerns separate?

catalog/resource_catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,43 @@
+package catalog
+
+import (
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestCatalogWorkspaceBindingsCornerCases(t *testing.T) {
```
ehab-eb, line 10: I'm not sure why but this test is extremely flaky - It passes locally maybe 9 out of 10 time but then fails randomly like it did in this action. https://github.com/databricks/terraform-provider-databricks/actions/runs/5176590327/jobs/9326664436?pr=2364

internal/acceptance/catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,33 @@
...
+			name = "test{var.RANDOM}"
+		}
+
+		resource "databricks_catalog" "prod" {
+			name = "prod{var.RANDOM}"
+		}
+
+		resource "databricks_catalog_workspace_binding" "dev" {
+			catalog_name = databricks_catalog.dev.name
+			workspace_id = {env.TEST_WORKSPACE_ID}
```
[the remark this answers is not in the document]
nkvuong: need this for acceptance tests to run correctly

internal/acceptance/catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,33 @@
...
+		}
+
+		resource "databricks_catalog_workspace_binding" "test" {
+			catalog_name = databricks_catalog.test.name
+			workspace_id = {env.TEST_WORKSPACE_ID}
```
nkvuong, line 29:
```suggestion
			workspace_id = {env.THIS_WORKSPACE_ID}
```

catalog/resource_catalog_workspace_binding.go
nkvuong: sure, feel free to raise a small PR for this one - should be quick to test & approve

catalog/resource_catalog_workspace_binding_test.go
nkvuong: I don't think it's flaky, rather than it always fails the first run, and then succeeds to following runs

nkvuong:
acceptance test is failing with `cannot read catalog workspace binding: Catalog has no binding to this workspace`. This is because the catalog isolation mode is not correctly set. That fix is in this PR #2357

ehab-eb:
> acceptance test is failing with `cannot read catalog workspace binding: Catalog has no binding to this workspace`. This is because the catalog isolation mode is not correctly set. That fix is in this PR #2357

Once your PR is merged should the isolation level be implicitly updated in this resource or should it rely on the catalog resource setting the correct isolation mode?

catalog/resource_catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,43 @@
...
+	qa.ResourceCornerCases(t, ResourceCatalogWorkspaceBinding(), qa.CornerCaseID("my_catalog|1234567890101112"))
```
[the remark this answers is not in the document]
ehab-eb: Is this because the corner cases don't set the isolation mode?

catalog/resource_catalog_workspace_binding.go
ehab-eb: Done #2373

nkvuong:
> should it rely on the catalog resource setting the correct isolation mode?
^

ehab-eb:
> > this is better, but then require the catalog to be created in Terraform,

Or just that the user has toggled any of the workspaces attached to this catalog using the UI.

> maybe worth checking that the catalog has the correct isolation mode during creation?

I was thinking the same thing but I'm wondering if it makes more sense to just let the API fail naturally?

Either way I'll add a note in the docs about the isolation mode.

nkvuong:
so we have a slight problem - creating catalog with `isolation_mode = "ISOLATED"` will result in GET API failing with the message `cannot read catalog: Catalog 'prod{var.random}' is not accessible in current workspace`

I'm going to check with our engineering team on how they expect it to work

nkvuong:
@ehab-eb sorry, copied the wrong error message, as I was trying to test locally as well

ehab-eb:
@nkvuong so you're saying that when a catalog is created in isolated mode and then we try to attach it to the current workspace the bindings API returns a 403 AND succeeds in making the binding?

ehab-eb:
Yep I agree. Option 2 makes the most sense. Will make the changes

ehab-eb:
@nkvuong I'm scratching my head a little bit trying to find a way to get the current workspace id. I can find the host url but not sure how to connect that to an actual id. Any tips on where to look?

I considered implicitly calling `mws_workspaces` data source but that's an account level auth and I imagine it needs a different set of permissions than creating a catalog.

nkvuong:
within go sdk, there is this method that should return the `workspace_id` - https://github.com/databricks/databricks-sdk-go/blob/main/service/catalog/api.go#LL783C25-L783C32

ehab-eb:
@nkvuong Thanks! That did it :)

internal/acceptance/catalog_workspace_binding_test.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		resource "databricks_catalog_workspace_binding" "dev" {
+			catalog_name = databricks_catalog.dev.name
+			workspace_id = {env.THIS_WORKSPACE_ID}
+		}
+
+		resource "databricks_catalog_workspace_binding" "test" {
+			catalog_name = databricks_catalog.test.name
+			workspace_id = {env.THIS_WORKSPACE_ID}
```
ehab-eb, line 31: Is there an env variable holding a different workspace id to test that it's able to attach to other workspaces too?

docs/resources/catalog_workspace_binding.md
```diff
@@ -0,0 +1,43 @@
...
+* `catalog_name` - Name of Catalog. Change forces creation of a new resource.
+* `workspace_id` - ID of the workspace. Change forces creation of a new resource.
+
+## Import
+
+This resource can be imported by name:
+
+```bash
+terraform import databricks_catalog_workspace_binding.this <name>
+```
```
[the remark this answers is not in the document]
nkvuong: we can just remove this

nkvuong:
@ehab-eb re-running acceptance tests, but there is an issue with `catalog.NewMetastores(c.DatabricksClient).Current(ctx)` where the struct has the wrong type, leading to `cannot unmarshal number into Go struct field MetastoreAssignment.workspace_id of type string`

I'll ask the engineering team to fix this

catalog/resource_catalog.go
```diff
@@ -75,6 +76,26 @@ func ResourceCatalog() *schema.Resource {
 				return err
 			}
 
+			// Bind the current workspace if the catalog is isolated, otherwise the read will fail
+			if d.Get("isolation_mode") == "ISOLATED" {
```
nfx, line 80: reduce the nesting by flipping the if statement - e.g. if it's not isolated catalog, return immediately

catalog/resource_catalog_workspace_binding.go
```diff
@@ -0,0 +1,57 @@
...
+	return common.NewPairID("catalog_name", "workspace_id").Schema(func(
+		m map[string]*schema.Schema) map[string]*schema.Schema {
+		return m
+	}).BindResource(common.BindResource{
+		CreateContext: func(ctx context.Context, catalogName, workspaceId string, c *common.DatabricksClient) error {
+			i64WorkspaceId, err := strconv.ParseInt(workspaceId, 10, 64)
+			if err != nil {
+				return err
+			}
+			createBindingRequest := catalog.UpdateWorkspaceBindings{
```
nfx, line 23: probably inlining this variable would be nicer for readability

nfx (changes requested):

ehab-eb:
> @ehab-eb re-running acceptance tests, but there is an issue with `catalog.NewMetastores(c.DatabricksClient).Current(ctx)` where the struct has the wrong type, leading to `cannot unmarshal number into Go struct field MetastoreAssignment.workspace_id of type string`
> 
> I'll ask the engineering team to fix this

@nkvuong I've made a change to how I identify the current metastore. I believe it's more or less the same call underneath but I'm not sure what the root cause was for the error you got so it might be worth running the acceptance tests again?

nkvuong:
@ehab-eb acceptance tests are still failing with
```
 Error: cannot create catalog: json: cannot unmarshal number into Go struct field MetastoreAssignment.workspace_id of type string
    
      with databricks_catalog.dev,
      on terraform_plugin_test.tf line 1, in resource "databricks_catalog" "dev":
       1: resource "databricks_catalog" "dev" {
    
    
    Error: cannot create catalog: json: cannot unmarshal number into Go struct field MetastoreAssignment.workspace_id of type string
    
      with databricks_catalog.test,
      on terraform_plugin_test.tf line 5, in resource "databricks_catalog" "test":
       5: resource "databricks_catalog" "test" {
```

nfx (approved):

thaiphv:
Thanks for implementing this resource. @nfx when do you plan to release this?

Hong-Kit:
@nfx Thanks this is awesome work! Would like to know about the release schedule too 🙏

PR #2371, opened 2023-06-04
Added `databricks_access_control_rule_set` resource for managing account-level access

gauthamsunjay:
## Changes

Added rule set resource. Rule set resource encapsulates grant_rules that customers can configure on accounts and service principals. It supports two APIs: Get rule set and update rule set. 

We support the API through a workspace level proxy to the accounts backend and an account level API. 

Workflow:
1. Create service principal in the workspace or account
2. Grant permissions on this service principal to other users

## Tests

- [x] `make test` run locally
- [x] using Go SDK

Manually tested e2e using the following terraform code

```
provider "databricks" {
    host = "https://dbc-31b63398-1bed.dev.databricks.com/"
    token = "<redacted_token>"
}

data "databricks_service_principal" "target_sp" {
    application_id = "0820032d-50fb-4d2d-a426-441db30343d4"
}

resource "databricks_rule_set" "sp_default" {
    name = "accounts/e77921da-37d2-41fa-ba78-49ac2fac7f61/servicePrincipals/${data.databricks_service_principal.target_sp.application_id}/ruleSets/default"
    grant_rules {
        principals = [
            "users/gautham.sunjay+wsa@databricks.com"
        ]
        role = "roles/servicePrincipal.manager"
    }

    grant_rules {
        principals = [
            "users/gautham.sunjay+ds@databricks.com",
            "users/saurabh.garg+ds@databricks.com"
        ]
        role = "roles/servicePrincipal.user"
    }
}
```


alexott:
Why it's not implemented via Go SDK?

gauthamsunjay:
> Why it's not implemented via Go SDK?

We have an account level API and a workspace level API. The SDK is not yet generated for the workspace level API since the OpenAPI PR is still not merged - https://github.com/databricks/universe/pull/320756.

But the idea is to support both account level and workspace level endpoints using terraform when the SDK is available.

rule_sets/resource_rule_set.go
```diff
@@ -0,0 +1,181 @@
+package rule_sets
```
nfx, line 1: this should go into `permissions` package in TF

nfx (changes requested):
we cannot accept this resource without Go SDK

docs/resources/rule_set.md
```diff
@@ -0,0 +1,138 @@
...
+
+# databricks_rule_set Resource
+
+This resource allows you to manage access rules on Databricks account level resources. For convenience we allow accessing this resource through the Databricks account and workspace.
+
+To configure rule sets through the Databricks account, the provider must be configured with `host = "https://accounts.cloud.databricks.com"` on AWS deployments or `host = "https://accounts.azuredatabricks.net"` and authenticate using [AAD tokens](https://registry.terraform.io/providers/databricks/databricks/latest/docs#special-configurations-for-azure) on Azure deployments
+
+## Example usage
+
+Rule set management through Databricks workspace:
```
alexott, line 13: Do we really need to do that through workspace as well? What is the benefit of it?

docs/resources/rule_set.md
```diff
@@ -0,0 +1,138 @@
...
+
+Arguments of the `grant_rules` block are:
+
+- `role` - (Required) Role to be granted. The following roles are supported:
+  * roles/servicePrincipal.manager - Manager of a service principal.
+  * roles/servicePrincipal.user - User of a service principal.
+- `principals` - (Required) a list of principals who are granted a role. The following format is supported:
+  * users/{username}
+  * groups/{groupname}
+  * servicePrincipals/{applicationId}
```
alexott, line 130: Same here - can we allow to specify `users`, `service_principals`, `groups` arguments & gneerate `principals` from it?

docs/resources/rule_set.md
gauthamsunjay: In this system, any user should be able to manage rule sets as long as they have sufficient permissions. It is not restricted to account admins only. Since non account admins still do not have good authentication mechanisms, we are exposing the APIs through the workspace where they can use personal access tokens.

docs/resources/rule_set.md
gauthamsunjay: I am new to Go and terraform.. Do you have any tips for how to do that?

nfx: @alexott Google provider does it as close to API as possible

https://registry.terraform.io/providers/hashicorp/google/4.22.0/docs/resources/api_gateway_api_iam

internal/acceptance/account_rule_set_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/service/iam"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/databricks/terraform-provider-databricks/permissions"
+)
+
+func TestAccountRuleSetsFullLifeCycle(t *testing.T) {
```
nfx, line 14:
```suggestion
func TestAccAccountRuleSetsFullLifeCycle(t *testing.T) {
```

internal/acceptance/account_rule_set_test.go
```diff
@@ -0,0 +1,46 @@
...
+			grant_rules {
+				principals = [
+					"groups/${databricks_group.this.display_name}"
+				]
+				role = "roles/servicePrincipal.manager"
+			}
+		}`,
+		Check: resourceCheck("databricks_rule_set.sp_rule_set",
+			func(ctx context.Context, client *common.DatabricksClient, id string) error {
+				getRuleSetReq := iam.GetRuleSetRequest{
```
nfx, line 34: use the `w, err := client.WorkspaceClient()` idioms

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type RuleSetApi struct {
+	client *common.DatabricksClient
+	ctx    context.Context
+}
+
+func NewRuleSetApi(ctx context.Context, m interface{}) RuleSetApi {
```
nfx, line 19: please don't define `RuleSetApi` in the TF, because we already have it in the SDK.

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+	return RuleSetApi{m.(*common.DatabricksClient), ctx}
```
nfx, line 20: and context always has to be used as the first argument.

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (a RuleSetApi) Read(getRuleSetReq iam.GetRuleSetRequest) (*iam.RuleSetResponse, error) {
```
nfx, line 23: try changing the signature to:
`func ReadFromWsOrAcc(ctx context.Context, client *common.DatabricksClient, req ) getRuleSetReq iam.GetRuleSetRequest) (*iam.RuleSetResponse, error)`

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+		return accountClient.AccessControl.GetRuleSet(a.ctx, getRuleSetReq)
+	}
+	workspaceClient, err := a.client.WorkspaceClient()
+	if err != nil {
+		return nil, err
+	}
+	return workspaceClient.AccountAccessControlProxy.GetRuleSet(a.ctx, getRuleSetReq)
+}
+
+func (a RuleSetApi) Update(updateRuleSetReq iam.UpdateRuleSetRequest) (*iam.RuleSetResponse, error) {
```
nfx, line 38: same comment as for `Read` ^

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+		ruleSetUpdateReq.RuleSet.Etag = ruleSetGetRes.Etag
+		ruleSetUpdateRes, err := rsApi.Update(ruleSetUpdateReq)
+		if err != nil {
+			return "", err
+		}
+		return ruleSetUpdateRes.Etag, nil
+	}
+	handleConflictAndUpdate := func(ruleSetUpdateReq iam.UpdateRuleSetRequest, rsApi RuleSetApi) (string, error) {
+		ruleSetUpdateRes, err := rsApi.Update(ruleSetUpdateReq)
+		if err != nil {
```
nfx, line 84: please typecast to an error and check on error code, because using exception message for logic is very fragile. See example on how to use the error code: https://github.com/databricks/cli/blob/35dd56c716eaa46c543d96a21d802614b7caf085/libs/filer/dbfs_client.go#L117-L131

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+				Name: name,
+				Etag: "",
+			}
+			if etag, ok := d.GetOk("etag"); ok {
+				getRuleSetReq = iam.GetRuleSetRequest{
+					Name: name,
+					Etag: fmt.Sprintf("%v", etag),
+				}
+			}
+			data, err := rsApi.Read(getRuleSetReq)
```
nfx, line 122:
```suggestion
			data, err := readFromWsOrAccoutn(ctx, c, iam.GetRuleSetRequest{
				Name: d.Id(),
				Etag: d.Get("etag").(string), // check if it still works, go "zeroval" should work
			})
```

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+				return err
+			}
+			return common.StructToData(data, s, d)
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			var ruleSetUpdateReq iam.UpdateRuleSetRequest
+			common.DataToStructPointer(d, s, &ruleSetUpdateReq.RuleSet)
+			ruleSetUpdateReq.Name = ruleSetUpdateReq.RuleSet.Name
+			rsApi := NewRuleSetApi(ctx, c)
+			if etag, ok := d.GetOk("etag"); ok {
```
nfx, line 133: you don't have to use `d.GetOk("etag")`, because it has to be available on `ruleSetUpdateReq` already, as that's what `DataToStructPointer` is doing

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+				updateRuleSetReq.RuleSet.Etag = fmt.Sprintf("%v", etag)
+				updatedEtag, err := handleConflictAndUpdate(updateRuleSetReq, rsApi)
+				if err != nil {
+					return err
+				}
+				d.Set("etag", updatedEtag)
+				return nil
+			}
+			etag, err := getAndUpdateRuleSet(updateRuleSetReq, rsApi)
```
nfx, line 169: here you've assumed that `etag` doesn't exist, but it's a false assumption, because there's always an `etag` in the state, `Create` added it

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+			if etag, ok := d.GetOk("etag"); ok {
+				ruleSetUpdateReq.RuleSet.Etag = fmt.Sprintf("%v", etag)
+				updatedEtag, err := handleConflictAndUpdate(ruleSetUpdateReq, rsApi)
+				if err != nil {
+					return err
+				}
+				d.Set("etag", updatedEtag)
+				return nil
+			}
+			etag, err := getAndUpdateRuleSet(ruleSetUpdateReq, rsApi)
```
nfx, line 142: here you've assumed that etag doesn't exist, but it's a false assumption, because there's always an etag in the state, Create added it. so `getAndUpdateRuleSet` probably makes no sense.

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+				return err
+			}
+			d.Set("etag", etag)
+			return nil
+		},
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			// we remove all grant rules. Account admins will still be able to update rule set
+			name := d.Id()
+			rsApi := NewRuleSetApi(ctx, c)
+			updateRuleSetReq := iam.UpdateRuleSetRequest{
```
nfx, line 153: inline the variable, you should use `DataToStructPointer` save some code :)

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+				Name: name,
+				RuleSet: iam.RuleSetUpdateRequest{
+					Name: name,
+					Etag: "",
+				},
+			}
+			if etag, ok := d.GetOk("etag"); ok {
```
nfx, line 160: most likely, the call to `DataToStructPointer` will also set the etag

permissions/resource_rule_set_test.go
```diff
@@ -0,0 +1,347 @@
...
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/iam"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+var (
+	TEST_ACCOUNT_ID                 = "cb376b18-60fa-4058-b2cd-dd85acf63165"
```
nfx, line 17: in golang, we don't use CONSTANT_CASE. we use `camelCase`. please don't trigger our linters.

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+				updateRuleSetReq.RuleSet.Etag = fmt.Sprintf("%v", etag)
```
nfx, line 161: type cast it to string:
```suggestion
				updateRuleSetReq.RuleSet.Etag = etag.(string)
```

permissions/resource_rule_set.go
nfx: and make this function confined into `ResourceRuleSet`.

nfx (changes requested):

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,177 @@
...
+
+func ResourceRuleSet() *schema.Resource {
+	s := common.StructToSchema(
+		iam.RuleSetUpdateRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["name"].Required = true
+			m["etag"] = &schema.Schema{
+				Computed: true,
+				Type:     schema.TypeString,
+			}
```
nfx, line 61: probably these can be removed as well

jiyunyao (approved):
@alexott @nfx We researched three ACL API design:

- Azure role assignment: "azurerm_role_assignment" ([terraform](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/role_assignment)), their design is simple but assume each role assignment and each resource, identity has a UUID. This is a platform unification where we don't have.
- GCP IAM v1: [terraform](https://registry.terraform.io/providers/hashicorp/google/4.22.0/docs/resources/notebooks_instance_iam). This means different IAM endpoints for every resource types. Not supporting deny and etc.
- GCP IAM v2: [terraform](https://registry.terraform.io/providers/hashicorp/google/4.22.0/docs/resources/iam_deny_policy), [API doc](https://cloud.google.com/iam/docs/reference/rest/v2/policies/update).

We are following GCP IAM v2 with slight tweaks. It's not the best user experience, but it's the newest design pattern in the industry. On the other hand, we will need each rule/ruleSet be addressable when these ACLs can be managed from SQL.

gauthamsunjay:
@nfx @alexott any update on this PR?

alexott:
@gauthamsunjay  please fix the failing tests

gauthamsunjay:
@alexott The tests failed due to a dependency on the SDK. The SDK has since been updated. How do I re-trigger the test run?

gauthamsunjay:
This PR requires https://github.com/databricks/terraform-provider-databricks/pull/2414 to get merged for dependent SDK changes.

gauthamsunjay:
@nfx @alexott all tests are now passing.

permissions/resource_rule_set.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+func ResourceRuleSet() *schema.Resource {
+	s := common.StructToSchema(
+		iam.RuleSetUpdateRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["etag"].Required = false
+			m["etag"].Computed = true
+			m["grant_rules"].Type = schema.TypeSet
+			common.MustSchemaPath(m, "grant_rules", "principals").Type = schema.TypeSet
```
alexott, line 22: I'm not sure that we should use `set` type instead of the `list` at least for the `grant_rules` blocks - we had problems in DLT resource when set was used for complex objects, then it caused nor necessary diffs, etc.

alexott (changes requested):
we need to at least update the docs

docs/resources/rule_set.md
```diff
@@ -0,0 +1,139 @@
...
+// account level group
+data "databricks_group" "ds" {
+  display_name = "Data Science"
+}
+
+resource "databricks_service_principal" "automation_sp" {
+  display_name = "SP_FOR_AUTOMATION"
+}
+
+resource "databricks_rule_set" "automation_sp_rule_set" {
```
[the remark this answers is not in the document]
gauthamsunjay: Name identifies a rule set resource, we also set ID to be equal to name. How does terraform pass the grant_rules to us?

permissions/resource_rule_set.go
gauthamsunjay: I did not notice any issues with my local testing. This is similar to what is done by UC - https://github.com/databricks/terraform-provider-databricks/blob/master/catalog/resource_grants.go#L39

alexott: See #1401 and PRs #1416 & #1508 about effect of using `TypeSet` with complex structures.

docs/resources/rule_set.md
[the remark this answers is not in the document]
alexott: My question was about following scenario - if I define two resources with the same name and different principals - what will happen?

```hcl
resource "databricks_rule_set" "automation_sp_rule_set1" {
  name = "accounts/${local.account_id}/servicePrincipals/${databricks_service_principal.automation_sp.application_id}/ruleSets/default"
  grant_rules {
    principals = ["groups/${data.databricks_group.ds1.display_name}"]
    role       = "roles/servicePrincipal.user"
  }
}

resource "databricks_rule_set" "automation_sp_rule_set2" {
  name = "accounts/${local.account_id}/servicePrincipals/${databricks_service_principal.automation_sp.application_id}/ruleSets/default"
  grant_rules {
    principals = ["groups/${data.databricks_group.ds2.display_name}"]
    role       = "roles/servicePrincipal.user"
  }
}
```

from the code I see that it should attempt to merge both definitions, but I want to be sure that I understand how it works, especially if you remove only one resource - what will happen with another?

Also, what will happen if I have intersecting principals, etc.?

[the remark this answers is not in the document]
tanmay-db: why is having resource with same name but different principal required?

permissions/resource_rule_set.go
tanmay-db: Thanks @alexott for the review. Going through the PRs, was it investigated why empty structs were generated? 

> But it should be investigated why we generated empty structs when using reflection

docs/resources/rule_set.md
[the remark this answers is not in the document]
gauthamsunjay: > if I define two resources with the same name and different principals - what will happen?

Ah yeah, this can be problematic. The behavior will be unpredictable because terraform possibly thinks the rule set resources can be created / updated in parallel. The warning suggestion sounds good to me. Let me add it. 

> Also, what will happen if I have intersecting principals, etc.?

As mentioned above, there will be unpredictable behavior.

permissions/resource_rule_set.go
gauthamsunjay: @alexott should I change the schema of grant_rules to list? The backend will handle duplicate entries of the same role so it should be fine from that perspective.

alexott: This is really a good question - and we need to check if backend may return the entries in the different order? Like, if I submit

```
["z1", "a1", "x1", "b1"]
```

may it return things in the other order?

```
["a1", "b1", "x1", "z1"]
```

if yes, then it may cause constant configuration drift.

gauthamsunjay: the backend will merge the entires..

```
{
  "role": "servicePrincipal/user",
  "principals": ["users/gautham@databricks.com"]
}

{
  "role": "servicePrincipal/user",
  "principals": ["groups/data_scientists"]
}
```

will become
```
{
  "role": "servicePrincipal/user",
  "principals": ["users/gautham@databricks.com", "groups/data_scientists"]
}
```

gauthamsunjay: Actually, let me verify this..

common/client.go
```diff
@@ -44,10 +44,25 @@ type DatabricksClient struct {
...
+		return c.cachedAccountClient, nil
+	}
+	a, err := databricks.NewAccountClient((*databricks.Config)(c.DatabricksClient.Config))
+	if err != nil {
+		return nil, err
+	}
+	c.cachedAccountClient = a
+	return a, nil
+}
+
```
tanmay-db, line 65: cc: @mgyucht since we were discussing adding account client and data.

permissions/resource_rule_set.go
tanmay-db: If it is definite (and not just this example instance done above) that backend returns in specified order, then what might be the issue with using this assumption?

tanmay-db: @gauthamsunjay I think this is the only unresolved issue right? Rest looks lgtm to me.

gauthamsunjay: Is it a blocker to have the grant_rules as a set? If the backend behavior changes there will be constant drift which we don't want as well. 

cc @alexott

alexott: let leave it as set - if you didn't find any problems in testing, then it could work. I simply had bad experience before. But we can change type if necessary

tanmay-db (approved):

tanmay-db:
@gauthamsunjay I have triggered an integration test run, please hold for few minutes before merging. Will update here.

tanmay-db:
@gauthamsunjay ran integration tests locally, please feel free to merge. It doesn't show up on our CI when running on same environment, now sure why but looking into it. That doesn't have to be a blocker.

tanmay-db:
Ah, I understand why this is happening, can you please add `Mws` in the test name: `TestAccAccountRuleSetsFullLifeCycle` -> `TestMwsAccAccountRuleSetsFullLifeCycle` @gauthamsunjay

tanmay-db:
Since there are concerns over this and blocks: https://github.com/databricks/terraform-provider-databricks/pull/2429, going forward with that PR since we would need AccountClient for Account Data, this PR would then need to rebase with master.

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,172 @@
...
+subcategory: "Security"
+---
+
+# databricks_access_control_rule_set Resource
+
+This resource allows you to manage access rules on Databricks account level resources. For convenience we allow accessing this resource through the Databricks account and workspace.
+
+-> **Note** Currently, we only support managing access rules on service principal resources through databricks_access_control_rule_set.
+
+-> **Warning** databricks_access_control_rule_set cannot be used to manage access rules for resources supported by [databricks_permission](permissions.md). Refer to it's documentation for more information.
```
gauthamsunjay, line 11: I have referenced permissions resource here

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,172 @@
...
+
+resource "databricks_rule_set" "automation_sp_rule_set" {
+  name = "accounts/${provider.databricks.account_id}/servicePrincipals/${databricks_service_principal.automation_sp.application_id}/ruleSets/default"
+
+  grant_rules {
+    principals = ["groups/${databricks_group.ds.display_name}"]
+    role       = "roles/servicePrincipal.user"
+  }
+}
+```
```
gauthamsunjay, line 130: Added example for GCP account

docs/resources/permissions.md
```diff
@@ -12,6 +12,8 @@ This resource allows you to generically manage [access control](https://docs.dat
 
 -> **Note** If multiple permission levels are specified for an identity (e.g. `CAN_RESTART` and `CAN_MANAGE` for a cluster), only the highest level permission is returned and will cause permanent drift.
 
+-> **Warning** To manage access control on service principals, use [databricks_access_control_rule_set](access_control_rule_set.md).
```
gauthamsunjay, line 15: Referenced the rule set resource in permissions

provider/provider.go
```diff
@@ -82,6 +82,7 @@ func DatabricksProvider() *schema.Provider {
 			"databricks_zones":                   clusters.DataSourceClusterZones(),
 		},
 		ResourcesMap: map[string]*schema.Resource{ // must be in alphabetical order
+			"databricks_access_control_rule_set":     permissions.ResourceAccessControlRuleSet(),
```
gauthamsunjay, line 85: changed name of the resource to `databricks_access_control_rule_set`

gauthamsunjay (commented):
@mgyucht @jiyunyao can you take another look? I have made changes as per feedback.

jiyunyao (approved):
Thanks for addressing the comments. LGTM

permissions/resource_access_control_rule_set.go
```diff
@@ -0,0 +1,139 @@
...
+		ruleSetUpdateRes, err := updateThroughWsOrAcc(ctx, c, ruleSetUpdateReq)
+		if err != nil {
+			var aerr *apierr.APIError
+			if !errors.As(err, &aerr) {
+				return "", err
+			}
+			if aerr.StatusCode == http.StatusConflict {
+				if aerr.ErrorCode == "RESOURCE_CONFLICT" {
+					// we need to get and update
+					etag, err := fetchLatestEtagAndUpdateRuleSet(ctx, c, ruleSetUpdateReq)
```
mgyucht, line 79: IIUC, this only retries once on resource conflict. Do you want to retry more times ever, or is one retry sufficient?

permissions/resource_access_control_rule_set.go
```diff
@@ -0,0 +1,139 @@
...
+			})
+			if err != nil {
+				return err
+			}
+			return common.StructToData(data, s, d)
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			var ruleSetUpdateReq iam.UpdateRuleSetRequest
+			common.DataToStructPointer(d, s, &ruleSetUpdateReq.RuleSet)
+			ruleSetUpdateReq.Name = ruleSetUpdateReq.RuleSet.Name
```
mgyucht, line 114: Not a blocker for this PR, but this is a bit unusual for an API. I think the API you're exposing in Terraform makes sense, as does the Read API, but why does Name need to be duplicated to the top-level structure for update and create? Why does there even need to be a wrapper structure at all? From my reading, your API should be able to accept a RuleSet directly in the update call since RuleSet contains its own name. In this case, the request/response body structure will be identical for create, read, and update.

permissions/resource_access_control_rule_set.go
```diff
@@ -0,0 +1,139 @@
...
+				Name: d.Id(),
+				RuleSet: iam.RuleSetUpdateRequest{
+					Name: d.Id(),
+					Etag: d.Get("etag").(string),
+				},
+			})
+			if err != nil {
+				return err
+			}
+			d.Set("etag", updatedEtag)
```
mgyucht, line 135: I'm pretty sure this is not necessary because resource state is removed after deleting a resource.

permissions/resource_access_control_rule_set_test.go
```diff
@@ -0,0 +1,356 @@
+package permissions
```
mgyucht, line 1: Great test coverage!

permissions/resource_access_control_rule_set.go
```diff
@@ -0,0 +1,139 @@
...
+			return "", err
+		}
+		return ruleSetUpdateRes.Etag, err
+	}
+	return common.Resource{
+		Schema: s,
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			var ruleSetUpdateReq iam.UpdateRuleSetRequest
+			common.DataToStructPointer(d, s, &ruleSetUpdateReq.RuleSet)
+			ruleSetUpdateReq.Name = ruleSetUpdateReq.RuleSet.Name
```
mgyucht, line 92: See comment in Update.

permissions/resource_access_control_rule_set.go
```diff
@@ -0,0 +1,139 @@
...
+		iam.RuleSetUpdateRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["etag"].Required = false
+			m["etag"].Computed = true
+			m["grant_rules"].Type = schema.TypeSet
+			common.MustSchemaPath(m, "grant_rules", "principals").Type = schema.TypeSet
+
+			return m
+		})
+	readFromWsOrAcc := func(ctx context.Context, c *common.DatabricksClient, getRuleSetReq iam.GetRuleSetRequest) (*iam.RuleSetResponse, error) {
```
mgyucht, line 26: @tanmay-db Another use case of two implementations based on whether the provider is account-level or workspace-level. It may be useful to expose some utility function like
```go
func ContextSpecificAPI[T any](func (context.Context, *databricks.WorkspaceClient) error, func (context.Context, *databricks.AccountClient) error) T, error
```
which handles checking for account ID, fetching the client and handling errors associated with that step, reducing the code to purely business logic for the resource implementer.

mgyucht (commented):
Thanks for these changes! I took another, more thorough pass and have a few more comments for you. I'm more or less happy with the implementation but have a few small questions. Some are suggestions you should be able to apply straight away in documentation. However, aside from this PR, I am very curious about the structure of your create/update APIs. See my comment in your implementation.

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,172 @@
...
+// account level group
+data "databricks_group" "ds" {
+  display_name = "Data Science"
+}
+
+resource "databricks_service_principal" "automation_sp" {
+  display_name = "SP_FOR_AUTOMATION"
+}
+
+resource "databricks_rule_set" "automation_sp_rule_set" {
```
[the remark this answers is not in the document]
gauthamsunjay: Good catch! Will do

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,172 @@
...
+  ]
+  role = "roles/servicePrincipal.user"
+}
+```
+
+Arguments of the `grant_rules` block are:
+
+- `role` - (Required) Role to be granted. The following roles are supported:
+  * roles/servicePrincipal.manager - Manager of a service principal.
+  * roles/servicePrincipal.user - User of a service principal.
```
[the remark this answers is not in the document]
gauthamsunjay: @mgyucht the manager roles are explained in the documentation - https://docs.databricks.com/security/auth-authz/access-control/service-principal-acl.html#service-principal-roles

permissions/resource_access_control_rule_set.go
gauthamsunjay: One retry is usually sufficient. Change to the rule set is expected to be an infrequent operation.

permissions/resource_access_control_rule_set.go
gauthamsunjay: This is good observation Miles. 

> but why does Name need to be duplicated to the top-level structure for update and create?

The top level name is used to reference the rule set that is being updated. And the wrapper structure is used to prevent customers from blinding passing the response from `Get` to `Update`. In our workspace permissions API, we have encountered situations where customers just passed the response from GET which included inherited permissions to PUT and caused a permission explosion. So we have intentionally tried to make sure the same mistake cannot happen in the rule set API. 

@jiyunyao can you also please take a look here?

permissions/resource_access_control_rule_set.go
gauthamsunjay: Got it, will remove!

docs/resources/access_control_rule_set.md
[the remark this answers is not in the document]
mgyucht: Great. Can we include that link in the docs?

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,170 @@
...
+  display_name = "Data Science"
+}
+
+resource "databricks_service_principal" "automation_sp" {
+  application_id = "00000000-0000-0000-0000-000000000000"
+  display_name   = "SP_FOR_AUTOMATION"
+}
+
+resource "databricks_access_control_rule_set" "automation_sp_rule_set" {
+  name = "accounts/${provider.databricks.account_id}/servicePrincipals/${databricks_service_principal.automation_sp.application_id}/ruleSets/default"
```
nkvuong, line 94: this `${provider.databricks.account_id}` interpolation is not allowed by Terraform, you'll need to factor out `account_id` as a variable/local

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,170 @@
...
+resource "databricks_group" "ds" {
+  display_name = "Data Science"
+}
+
+resource "databricks_service_principal" "automation_sp" {
+  display_name = "SP_FOR_AUTOMATION"
+}
+
+resource "databricks_access_control_rule_set" "automation_sp_rule_set" {
+  name = "accounts/${provider.databricks.account_id}/servicePrincipals/${databricks_service_principal.automation_sp.application_id}/ruleSets/default"
```
nkvuong, line 63: same here

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,170 @@
...
```
nkvuong, line 123: Same

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+```
+
+Rule set management through GCP Databricks account:
+
+```hcl
+// initialize provider at account-level
+provider "databricks" {
+  alias      = "mws"
```
nkvuong, line 108: remove this alias

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+```
+
+Rule set management through Azure Databricks account:
+
+```hcl
+// initialize provider at Azure account-level
+provider "databricks" {
+  alias      = "azure_account"
```
nkvuong, line 77: remove this alias

docs/resources/access_control_rule_set.md
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+```
+
+Rule set management through AWS Databricks account:
+
+```hcl
+// initialize provider at account-level
+provider "databricks" {
+  alias      = "mws"
```
nkvuong, line 46: remove this alias

nkvuong (changes requested):

docs/resources/access_control_rule_set.md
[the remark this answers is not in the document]
gauthamsunjay: Done, I have included the link in the docs!

permissions/resource_access_control_rule_set.go
mgyucht: I can see that happening since this API is PUT (i.e. replace), not PATCH (i.e. incremental change). I worry that changes like this will be harder to support in the future when we try to automate the generation of Terraform resource implementations, and SDK users also may be confused about what name parameter to use for https://github.com/databricks/databricks-sdk-go/blob/main/service/iam/model.go#L634 and for https://github.com/databricks/databricks-sdk-go/blob/main/service/iam/model.go#L610.

mgyucht (approved):
Overall I approve this PR. A couple of notes: The name parameter for this resource is not very user-friendly. However, as a first cut, it is simplest to maintain, and we can always introduce helper fields going forward in a backwards-compatible way to help users out in constructing the name. Especially when we do this, we should really rethink whether it is necessary for users to specify account ID when using this API. It is good that the API is generic enough to support the cross-account use case, but it is putting a burden on all other users of the API (for now, all users, since this use case is not yet supported) to include account ID in their templates directly, to expose vars to pass it through, and so on. Again, the API can be widened in the future and generally speaking in a backwards-compatible way if the account ID that customers authenticate to matches that in the name parameter.

As for the API, since that API is currently public, and its behavior is publicly documented this way today, we'll accept this PR. However, I do think you should reconsider whether diverging from REST API standards for updates is justified in this case from a usability perspective (and with respect to AIP standards).

gauthamsunjay:
> Overall I approve this PR. A couple of notes: The name parameter for this resource is not very user-friendly. However, as a first cut, it is simplest to maintain, and we can always introduce helper fields going forward in a backwards-compatible way to help users out in constructing the name. Especially when we do this, we should really rethink whether it is necessary for users to specify account ID when using this API. It is good that the API is generic enough to support the cross-account use case, but it is putting a burden on all other users of the API (for now, all users, since this use case is not yet supported) to include account ID in their templates directly, to expose vars to pass it through, and so on. Again, the API can be widened in the future and generally speaking in a backwards-compatible way if the account ID that customers authenticate to matches that in the name parameter.
> 
> As for the API, since that API is currently public, and its behavior is publicly documented this way today, we'll accept this PR. However, I do think you should reconsider whether diverging from REST API standards for updates is justified in this case from a usability perspective (and with respect to AIP standards).

Thanks Miles! @jiyunyao is working on a one pager to evaluate the option of not including accountId in the name. He will also weigh in on the update issue when he is back in office.

gauthamsunjay:
@tanmay-db @alexott @nfx can we merge this PR now?

permissions/resource_access_control_rule_set.go
jiyunyao: We are following AIP: https://google.aip.dev/134#etags where the resource body contains the name and update also has the name in URL: https://google.aip.dev/134#guidance.

This is similar design as https://cloud.google.com/iam/docs/reference/rest/v2/policies/update.

We don't intend to introduce PATCH for RuleSet update as it doesn't make sense (you can't partial update a list easily). We may support patch on individual rule update.

gauthamsunjay: @mgyucht I remember now. The name is part of the body because [App Infra](https://databricks.atlassian.net/browse/APP-1109) still doesn't support path parameters containing slash as recommended by the AIP @jiyunyao linked above.

tanmay-db (approved):

nkvuong (approved):

alexott (commented):
minor changes in the doc would be useful

nfx (approved):

alexott (approved):

PR #2378, opened 2023-06-05
Added an option disable_as_user_deletion to disable instead of delete `databricks_user` and `databricks_service_principal` by default

carletonmiya-db:
See: https://databricks.atlassian.net/browse/DECO-1046
One Decision Doc: https://docs.google.com/document/d/1qNn21xX0tl-MLDxJssUWMGwCLNntNjDbgTNLP6wrE0c/edit#

TODO: Wrote unit tests but need some help end-to-end testing.

## Changes
1. Add a flag so that SCIM DELETE will instead PATCH the user or service principal and set the active flag to false.
2. Added unit tests for disable_as_user_deletion.
3. Updated docs to document the new flag disable_as_user_deletion.
4. Docs on the flags force_delete_repos and force_delete_home_dir had some blank spaces after it. Just a minor fix to remove those.
5. testing.go was not setting the IsTesting flag in the Config for the fixtures. This causes Account SCIM testing to not work. Fixed testing.go by setting the flag using WithTesting().

## Tests
Unit tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/resource_service_principal.go
+++ b/scim/resource_service_principal.go
@@ -184,7 +184,7 @@ func ResourceServicePrincipal() *schema.Resource {
 			isDisable := d.Get("disable_as_user_deletion").(bool)
 			var err error
 			if isDisable {
-                        	r := PatchRequest("replace", "active", "false")
+				r := PatchRequest("replace", "active", "false")
 				err = spAPI.Patch(d.Id(), r)
 			} else {
 				err = spAPI.Delete(d.Id())
--- a/scim/resource_service_principal.go
+++ b/scim/resource_service_principal.go
@@ -210,7 +210,8 @@ func ResourceServicePrincipal() *schema.Resource {
 			// Force delete repos
 			if isForceDeleteRepos {
 				if isDisable {
-					return fmt.Errorf("force_delete_home_dir: internal error")
+					// Validations should prevent getting here. Double checking as this is destructive.
+					return fmt.Errorf("force_delete_repos: internal error")
 				}
 				err = workspace.NewNotebooksAPI(ctx, c).Delete(fmt.Sprintf("/Repos/%v", appId), true)
 				if err != nil {
@@ -220,6 +221,7 @@ func ResourceServicePrincipal() *schema.Resource {
 			// Force delete home dir
 			if isForceDeleteHomeDir {
 				if isDisable {
+					// Validations should prevent getting here. Double checking as this is destructive.
 					return fmt.Errorf("force_delete_home_dir: internal error")
 				}
 				err = workspace.NewNotebooksAPI(ctx, c).Delete(fmt.Sprintf("/Users/%v", appId), true)
--- a/scim/resource_service_principal.go
+++ b/scim/resource_service_principal.go
@@ -114,16 +114,15 @@ func ResourceServicePrincipal() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			}
-			m["disable_as_user_deletion"] = &schema.Schema{
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  true,
-			}
 			m["force_delete_repos"] = &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
 			}
 			m["force_delete_home_dir"] = &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			}
+			m["disable_as_user_deletion"] = &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
 			}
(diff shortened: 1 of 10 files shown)


scim/resource_user_test.go
```diff
@@ -440,7 +441,7 @@ func TestResourceUserDelete_Error(t *testing.T) {
 	}.ExpectError(t, "I'm a teapot")
 }
 
-func TestResourceUserDelete_NoErrorEmtpyParams(t *testing.T) {
+func TestResourceUserDelete_NoErrorNoDisable(t *testing.T) {
```
nfx, line 444: could you keep existing tests as is and create new ones with introduced invariants? `databricks_user` is one of the most-used resources in provider and this _removes code coverage_ for the standard critical paths.

scim/resource_user_test.go
```diff
@@ -585,11 +590,86 @@ func TestResourceUserDelete_NonExistingDir(t *testing.T) {
 		ID:       "abc",
 		HCL: `
 			user_name    = "abc"
+			disable_as_user_deletion = false
 			force_delete_home_dir = true
 		`,
 	}.Apply(t)
 	assert.EqualError(t, err, "force_delete_home_dir: Path (/Users/abc) doesn't exist.")
 }
+func TestResourceUserDeleteAsDisable(t *testing.T) {
```
nfx, line 599: run `make fmt`

scim/resource_user_test.go
```diff
@@ -585,11 +590,86 @@ func TestResourceUserDelete_NonExistingDir(t *testing.T) {
...
+		`,
+	}.ApplyNoError(t)
+}
+
+func TestResourceUserDeleteAsDisableWithForceDeleteRepos(t *testing.T) {
+	_, err := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "PATCH",
+				Resource: "/api/2.0/preview/scim/v2/Users/abc",
```
nfx, line 622: can you assert for body with `ExpectedRequest`? for this new functionality it's critical to assert on the body.

scim/users.go
```diff
@@ -86,6 +86,12 @@ func (a UsersAPI) Patch(userID string, r patchRequest) error {
 	return a.client.Scim(a.context, http.MethodPatch, fmt.Sprintf("/preview/scim/v2/Users/%v", userID), r, nil)
 }
 
+// Disable will disable the user given the user id
+func (a UsersAPI) Disable(userID string) error {
```
nfx, line 90: please make sure there's OpenAPI spec for this as well.

nfx (changes requested):
Please also make pull-request title changelog-ready

scim/resource_user_test.go
carletonmiya-db: Done

scim/resource_user_test.go
carletonmiya-db: Done

scim/resource_user_test.go
carletonmiya-db: Done

scim/users.go
carletonmiya-db: Removed this function and just did it inline in resource_user.go. I did not mean to pollute the API space here.

scim/resource_service_principal_test.go
```diff
@@ -376,9 +376,12 @@ func TestResourceServicePrincipalDelete(t *testing.T) {
 			},
 		},
 		Resource: ResourceServicePrincipal(),
-		HCL:      `display_name = "Squanchy"`,
 		Delete:   true,
 		ID:       "abc",
+		HCL: `
+			display_name = "Squanchy"
+			disable_as_user_deletion = false
```
nfx, line 383: can we make sure not to change the tests?..

nfx (changes requested):
can we make sure not to modify any existing tests, as they capture known behavior

scim/resource_service_principal_test.go
carletonmiya-db: Done

scim/resource_service_principal.go
```diff
@@ -122,6 +122,10 @@ func ResourceServicePrincipal() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			}
+			m["disable_as_user_deletion"] = &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
```
ramsuda-db, line 127: How come there is no `computed: ` value here like with users? How does the computed val get used?

ramsuda-db (commented):
Read the one decision doc, LGTM on logic, thanks a ton. Left one question about `computed`

ramsuda-db (approved):

scim/resource_service_principal.go
carletonmiya-db: Computed did not help as I needed to be able to tell the difference if a user actually set a value or they are taking the default (i.e. null). Only if taking the default, do I compute the value.

Issue #2224, opened 2023-04-17
Make sure that `common.StructToSchema` can accept aliases


nfx:
Make sure we can migrate the resources to Go SDK, that used to have `tf:"alias"`.

renaudhartert-db:
Hi @nfx, thanks for opening this issue. I'm doing a cleanup pass over our backlog and closing issues that haven't seen activity in a while.

This one hasn't had any comments or updates in over 2 years, so I'm closing it as part of the sweep. If the problem still reproduces on a recent version of the provider, please reopen the issue or create a new one.

Thanks!

PR #2389, opened 2023-06-08
Add `databricks_mlflow_experiment` data source

### LINKED ISSUE #2301: [FEATURE] Data Sources for MLflow experiment and model
### Use-cases

Current there is no data sources for MLflow experiment and model, which makes it difficult to refer to an existing experiment or model. Import resources in some cases are not very convenient when you have multiple environments.

### Attempted Solutions

One can declare an existing model or experiment as data sources:

```terraform

data "databricks_user" "john" {
  user_name = "john.doe@example.com"
}


data "databricks_mlflow_model" "test_model" {
  name = "test_model"
}

resource "databricks_permissions" "model_usage" {
  registered_model_id = data.databricks_mlflow_model.test_model.registered_model_id

  access_control {
    user_name        = data.databricks_user.john.user_name
  }
}

```

840:
## Changes
Add `databricks_mlflow_experiment` data source

## Tests
Created 3 tests for this data source

- [x] `make test` run locally
- [x] using Go SDK

Partially resolves #2301


840:
Currently you can only fetch the mlflow_experiment data source using the name. I believe it would be nice to provide the option to fetch using the ID, as well. I have to read a bit more into this.

nkvuong:
@840 thank you for raising the PR. a couple of thoughts for the pending tasks

you could look at https://github.com/databricks/terraform-provider-databricks/blob/master/clusters/data_cluster.go for example of searching both by name & id

the Go SDK has `GetExperiment` method which takes an `ExperimentId` as well.

For acceptance tests, you could create an mlflow experiment and try to read it in the same block, similar to https://github.com/databricks/terraform-provider-databricks/blob/master/internal/acceptance/data_job_test.go

mlflow/data_mlflow_experiment.go
```diff
@@ -0,0 +1,30 @@
...
+		Name             string `json:"name,omitempty" tf:"computed"`
+		ArtifactLocation string `json:"artifact_location,omitempty" tf:"computed"`
+		LifecycleStage   string `json:"lifecycle_stage,omitempty" tf:"computed"`
+	}, w *databricks.WorkspaceClient) error {
+		experimentResponse, err := w.Experiments.GetByName(ctx, ml.GetByNameRequest{ExperimentName: data.Name})
+		if err != nil {
+			return err
+		}
+		experiment := experimentResponse.Experiment
+		data.Name = experiment.Name
```
nkvuong, line 24: if `name` is provided, then you don't need to overwrite it

docs/data-sources/mlflow_experiment.md
```diff
@@ -0,0 +1,29 @@
...
+## Argument Reference
+
+* `name` - (Required) Path to experiment
+
+## Attribute Reference
+
+This data source exports the following attributes:
+
+* `artifact_location` - Location where artifacts for the experiment are stored.
+* `name` - Human readable name that identifies the experiment.
```
nkvuong, line 27: Need to be consistent with the `name` argument reference

docs/data-sources/mlflow_experiment.md
```diff
@@ -0,0 +1,29 @@
...
+* `lifecycle_stage` - Current life cycle stage of the experiment: "active" or "deleted". Deleted experiments are not returned by APIs.
```
nkvuong, line 28: Get by Id will work on deleted experiments

```suggestion
* `lifecycle_stage` - Current life cycle stage of the experiment: `active` or `deleted`. Deleted experiments are not returned by APIs.
```

840: I guess it has to be changed in the [Databricks documentation](https://docs.databricks.com/api/workspace/experiments/getexperiment#:~:text=Current%20life%20cycle%20stage%20of%20the%20experiment%3A%20%22active%22%20or%20%22deleted%22.%20Deleted%20experiments%20are%20not%20returned%20by%20APIs.), as well.

mlflow/data_mlflow_experiment.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"github.com/databricks/databricks-sdk-go/service/ml"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceExperiment() *schema.Resource {
+	return common.WorkspaceData(func(ctx context.Context, data *struct {
```
mgyucht, line 14: Can you just use *ml.Experiment from the Go SDK?

mgyucht (commented):
Quick first pass. LMK when no longer WIP.

mlflow/data_mlflow_experiment.go
840: At this moment it is not possible to annotate Terraform fields with the Go SDK model so this approach is not possible. Current approach is fine for now.

I did change `id` to `experiment_id` to make the transition to the Go SDK model backwards compatible.

Relevant issue for annotation support: #2224

840:
Will continue this week or early next week.

840:
@mgyucht WIP tag removed, should be ready!

alexott (approved):
Thank you!

PR #2392, opened 2023-06-10
Add warehouse_id parameter to `databricks_sql_table` resource

### LINKED ISSUE #2372: [FEATURE] Support SQL warehouses for CRUD operations on databricks_sql_table
### Use-cases

CRUD operations to a `databricks_sql_table` require a Compute cluster (the `cluster_id` argument). The majority of the time I feel SQL operations like this are done from SQL Warehouses, not all purpose clusters. As an added benefit, with the addition of SQL Serverless, this would greatly speed up execution for creating tables from a "cold start".

This feature request is to be able to use the `databricks_sql_table` resource with SQL Warehouses instead of Compute Clusters. 

### Attempted Solutions

I initially attempted to use a SQL warehouse ID in the `cluster_id` argument for the `databricks_sql_table` resource:

```terraform
resource "databricks_sql_table" "this" {
  catalog_name       = databricks_schema.this.catalog_name
  schema_name        = databricks_schema.this.name
  name               = "mytable"
  data_source_format = "PARQUET"
  storage_location   = var.storage_location
  cluster_id         = var.sql_warehouse_id # only works with Compute Cluster IDs, not SQL warehouses
  properties = {
    "tags.DeployedBy" = "Terraform"
  }
}
```

Provider version: 1.18.0
```
provider[registry.terraform.io/databricks/databricks] >= 1.18.0
```

### Proposal

Because there could be ambiguity differentiating Compute Cluster IDs from SQL Warehouse IDs, I'd like to propose a new optional string argument `warehouse_id` which should be mutually exclusive from `cluster_id`. And therefore, also should have backwards compatibility with the existing provider. 

For example, the usage above would become:

```terraform
resource "databricks_sql_table" "this" {
  catalog_name       = databricks_schema.this.catalog_name
  schema_name        = databricks_schema.this.name
  name               = "mytable"
  data_source_format = "PARQUET"
  storage_location   = var.storage_location
  # cluster_id = "" <-- cannot use this argument if using the `warehouse_id` argument
  warehouse_id         = var.sql_warehouse_id
  properties = {
    "tags.DeployedBy" = "Terraform"
  }
}
```

zcking:
## Tests

- [x] `make test` run locally
- [x] using Go SDK

I added some basic unit tests and followed the `commands` package closely for reference. 

I also installed the provider from my source, and used it in a real live terraform project to confirm it works, using `TF_LOG=debug` to monitor and troubleshoot the API requests as well. 

Closes #2372


sql/statement_execution.go
```diff
@@ -0,0 +1,117 @@
+package sql
```
mgyucht, line 1: Can you use the GoSDK for this? See https://github.com/databricks/databricks-sdk-go/blob/main/service/sql/api.go which already defines StatementExecutionAPI and WarehousesAPI.

catalog/resource_sql_table.go
```diff
@@ -88,27 +91,33 @@ func sqlTableIsManagedProperty(key string) bool {
...
 		if err != nil {
 			return
 		}
-		_, err = clustersAPI.StartAndGetInfo(ti.ClusterID)
-	}
-	if err != nil {
-		return
 	}
 	ti.exec = c.CommandExecutor(ctx)
+	ti.sqlExec = sql.NewStatementAPI(ctx, c)
```
mgyucht, line 120: Here, you should be able to do:
```go
w, err := c.WorkspaceClient()
if err != nil  {
  return err
}
ti.sqlExec = w.StatementExecution
```

mgyucht (changes requested):
Let's try to use the Go SDK for this, rather than reimplementing parts of it by hand.

If warehouse_id and cluster_id conflict, we need to add that to the schema definition in `func ResourceSqlTable()`:
```go
s["cluster_id"].ConflictsWith = []string{"warehouse_id"}
s["warehouse_id"].ConflictsWith = []string{"cluster_id"}
```

If exactly one of these fields is necessary, we can use ExactlyOneOf instead.

To confirm, is warehouse_id a computed field or not? It seems like it is optional rather than computed.

sql/statement_execution.go
zcking: Way better, thanks for the recommendation! I've made the suggested changes.

zcking: Since ExactlyOneOf has this behavior:
>  It will return a validation error if none are specified as well.

And the current behavior of this resource is that `cluster_id` is _optional_ (if not specified, a default cluster will be created named "terraform-sql-table") I believe the most correct constraint is your first suggestion:
```go
s["cluster_id"].ConflictsWith = []string{"warehouse_id"}
s["warehouse_id"].ConflictsWith = []string{"cluster_id"}
```

catalog/resource_sql_table.go
```diff
@@ -294,8 +308,24 @@ func (ti *SqlTableInfo) deleteTable() error {
 
 func (ti *SqlTableInfo) applySql(sqlQuery string) error {
 	log.Printf("[INFO] Executing Sql: %s", sqlQuery)
-	r := ti.exec.Execute(ti.ClusterID, "sql", sqlQuery)
+	if ti.WarehouseID != "" {
+		execCtx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+		defer cancel()
+		sqlRes, err := ti.sqlExec.ExecuteStatement(execCtx, sql.ExecuteStatementRequest{
+			Statement:   sqlQuery,
+			WaitTimeout: "10s",
```
mgyucht, line 316: How is this timeout determined? What if the user's query needs more time to run?

catalog/resource_sql_table.go
```diff
@@ -294,8 +308,24 @@ func (ti *SqlTableInfo) deleteTable() error {
 
 func (ti *SqlTableInfo) applySql(sqlQuery string) error {
 	log.Printf("[INFO] Executing Sql: %s", sqlQuery)
-	r := ti.exec.Execute(ti.ClusterID, "sql", sqlQuery)
+	if ti.WarehouseID != "" {
+		execCtx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
mgyucht, line 312: This is a very short timeout and may cause issues. For example, if there is a transient error and the request can be retried, this may be less forgiving. Even for requests which take a full 10 seconds,  We may want to use a longer timeout here, at least as long as `DatabricksClient.retryTimeout` (which defaults to 300 seconds). It's a bit screwy for the command execution API, which today captures the context in the `CommandsAPI` struct directly and doesn't set any timeouts.

sql/resource_sql_endpoint.go
```diff
@@ -106,7 +106,7 @@ func (a SQLEndpointsAPI) List() (lst EndpointList, err error) {
 
 // Start ..
 func (a SQLEndpointsAPI) Start(endpointID string, timeout time.Duration) error {
-	err := a.client.Post(a.context, fmt.Sprintf("/sql/warehouses/%s/start", endpointID), nil, nil)
+	err := a.client.Post(a.context, fmt.Sprintf("/sql/warehouses/%s/start", endpointID), `{}`, nil)
```
mgyucht, line 109: Is this needed? The start does not have any body AFAICT.

mgyucht (commented):
This change is looking much better! One last question on the timeout, then I think we're good to merge this.

alexott:
This should be covered by #2789

alexott:
Thank you for your contribution! This was superseded by #2789

PR #2401, opened 2023-06-14
Add experimental compute field to `databricks_job` resource

nordp:
## Changes
Add support for private preview functionality `compute` and `compute_key`.

## Tests
Add test case to `resource_job_test.go`

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/compute/compute_api.go
+++ b/compute/compute_api.go
@@ -1,5 +1,0 @@
-package compute
-
-type ComputeSpec struct {
-	Kind string `json:"kind"`
-}
--- a/compute/compute_api.go
+++ b/compute/compute_api.go
@@ -1,0 +1,6 @@
+package compute
+
+
+type ComputeSpec struct {
+    Kind string                                `json:"kind"`
+}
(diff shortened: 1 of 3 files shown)


compute/compute_api.go
```diff
@@ -0,0 +1,5 @@
+package compute
```
alexott, line 1: Will it be shared with clusters API as well?

alexott (approved):
looks good for me, only a comment on the `ComputeSpec` location

compute/compute_api.go
nordp: We're not sure yet, but I'd rather break it out sooner than later.

nfx: please don't add a package just with one `ComputeSpec`. 

use Go SDK
https://pkg.go.dev/github.com/databricks/databricks-sdk-go@v0.9.0/service/jobs

otherwise, move this type to `jobs/resource_job.go` for now

nfx (dismissed):

mgyucht (approved):
LGTM!

PR #2411, opened 2023-06-15
fix databricks_share Read

marekbrysa:
## Changes
If the API call returns 0 `object`s in `si` and there are `object`s in `d` (in existing tfstate), they are not deleted from `d` by `common.StructToData` so TF thinks there is no diff.

This causes an issue that there is no diff detected when the share returned by the API is empty.

This makes sure `object` is overwritten even when empty.

## Tests
Not sure how to fully reproduce it in resource_share tests.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_share.go
+++ b/catalog/resource_share.go
@@ -207,7 +207,7 @@ func ResourceShare() *schema.Resource {
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


alexott:
Fixed now with migration to the plugin framework (#3916)

PR #2412, opened 2023-06-16
Use separate suppress diff function for `databricks_sql_table` and `databricks_volume`

alexott:
## Changes

This is a follow up fix for #2408. It's needed because tables & volumes could be managed, and in this case API populate the `storage_location` with generated value and terraform was removing it because it's empty in the terraform code.

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -10,8 +10,16 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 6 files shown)


nkvuong (approved):
for completeness, do we want to add an extra test for `ucDirectoryPathSlashAndEmptySuppressDiff`

alexott:
I've added it to the existing test for suppress diffs

tanmay-db (approved):

PR #2422, opened 2023-06-23
Added `gcp_attributes.local_ssd_count` to `databricks_cluster` resource

alexott:
## Changes

New setting allows to specify number of local SSD disks to be attached to each node

Unfortunately we need to wait until Clusters GET is fixed because right now it doesn't return that attribute 🤦 

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK


mgyucht (commented):
Let's split into two PRs. I know it's a bit more overhead, but deprecations should be called ou explicitly in the changelog.

mgyucht (approved):

PR #2424, opened 2023-06-23
Feature/sql table replace if delta

danielsahal:
## Changes
Two changes:
- There's no need to keep "EXTERNAL" word in "CREATE EXTERNAL TABLE". If there's a LOCATION parameter provided, it will create EXTERNAL table by default,
- Databricks recommends to use REPLACE instead of dropping and recreating Delta Lake tables (https://docs.databricks.com/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html). The issue is with the external tables - if you for example remove column from a table, do the drop table and create a table in the same storage_location - it will throw you an error that the CT schema is different that the one on delta lake. Also REPLACE preserves the table history.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


mgyucht (commented):
Hi @danielsahal, thank you for raising this change. Several questions:
1. Is there an issue associated with these changes? It's not clear to me exactly what problems this is addressing. What is the exact Terraform workflow that this change addresses?
2. This change prevents Delta tables from being deleted entirely using this resource, which I believe is a breaking change. Consider non-external Delta tables: the user may actually want to delete the underlying data. What if the user's intention is to delete the table (which also matches the semantics of all other resources in this provider)?

nkvuong:
Reading through the PR, it updates the SQL sent by `databricks_sql_table` to match Databricks recommended syntax (i.e. removing unnecessary keyword `EXTERNAL` and using `CREATE OR REPLACE`)

Agreeing with @mgyucht on the 2nd point though, all tables should be dropped by this resource if removed by the users - otherwise it will cause confusion

nkvuong:
@danielsahal just want to check in with you on this PR

alexott (approved):

PR #2429, opened 2023-06-26
Added `AccountData`, `AccountClient` and define generic databricks data for workspace and account

tanmay-db:
## Changes

Define generic data for both account and workspace data, added account client for getting data for account level resources.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -45,6 +45,7 @@ type DatabricksClient struct {
 	// callback used to create API1.2 call wrapper, which simplifies unit tessting
 	commandFactory        func(context.Context, *DatabricksClient) CommandExecutor
 	cachedWorkspaceClient *databricks.WorkspaceClient
+	cachedAccountClient   *databricks.AccountClient
 	mu                    sync.Mutex
 }
 
@@ -64,6 +65,20 @@ func (c *DatabricksClient) WorkspaceClient() (*databricks.WorkspaceClient, error
 	})
 	c.cachedWorkspaceClient = w
 	return w, nil
+}
+
+func (c *DatabricksClient) AccountClient() (*databricks.AccountClient, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.cachedAccountClient != nil {
+		return c.cachedAccountClient, nil
+	}
+	acc, err := databricks.NewAccountClient((*databricks.Config)(c.DatabricksClient.Config))
+	if err != nil {
+		return nil, err
+	}
+	c.cachedAccountClient = acc
+	return acc, nil
 }
 
 // Get on path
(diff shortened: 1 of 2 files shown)


common/client.go
```diff
@@ -66,6 +67,20 @@ func (c *DatabricksClient) WorkspaceClient() (*databricks.WorkspaceClient, error
...
+func (c *DatabricksClient) AccountClient() (*databricks.AccountClient, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.cachedAccountClient != nil {
+		return c.cachedAccountClient, nil
+	}
+	acc, err := databricks.NewAccountClient((*databricks.Config)(c.DatabricksClient.Config))
+	if err != nil {
+		return nil, err
+	}
```
tanmay-db, line 79: AccountClient has AccountUsersService instead of currentUsersService that's why we cannot override implementation with cachedMe similar to account. 

This would require another type:

```
type cachedMeAccount struct {
	internalImpl iam.AccountUsersService
	cachedUser   *iam.User
	mu           sync.Mutex
}

func (a *cachedMeAccount) Me(ctx context.Context) (*iam.User, error) {
```

But we cannot do it because AccountUsersService requires other methods to be implemented such as Create

common/resource.go
```diff
@@ -300,6 +300,51 @@ func WorkspaceData[T any](read func(context.Context, *T, *databricks.WorkspaceCl
 	}
 }
 
+func AccountData[T any](read func(context.Context, *T, *databricks.AccountClient) error) *schema.Resource {
```
tanmay-db, line 303: We don't have any test for WorkspaceData. I assume it's because they are supposed to be tested when used in other resources and resources tests. Similar would be for AccountData.

It will be tested in: https://github.com/databricks/terraform-provider-databricks/pull/2017

mgyucht, line 303: Replace with genericData that parameterizes which client is used by read function.

common/resource.go
```diff
@@ -256,6 +256,29 @@ func DataResource(sc any, read func(context.Context, any, *DatabricksClient) err
...
+//		metastores, err := acc.Metastores.List(ctx)
+//		...
+//	})
+func AccountData[T any](read func(context.Context, *T, *databricks.AccountClient) error) *schema.Resource {
+	return GenericDatabricksData(func(c *DatabricksClient) (*databricks.AccountClient, error) {
+		return c.AccountClient()
+	}, read)
+}
+
+func GenericDatabricksData[T any, C any](getClient func(*DatabricksClient) (C, error), read func(context.Context, *T, C) error) *schema.Resource {
```
mgyucht, line 281: let's make this private: `genericDatabricksData`

mgyucht (approved):
LGTM modulo one comment.

PR #2444, opened 2023-06-29
Add `full_refresh` attribute to the `pipeline_task` in `databricks_job`

### LINKED ISSUE #2362: [FEATURE] allow to specify `full_refresh` on DLT task in `databricks_job`
**Use-cases**
There is an option in the UI to "trigger full refresh" in a pipeline task but not in Terraform Provider. As for the pipeline Terraform only has the option of providing pipeline ID in the public document.

### Proposal
There should be an option to "trigger full refresh" to update Tables and Views in Delta Live Table

alexott:
## Changes

This allows to force full refresh of the pipeline from the job

this fixes #2362

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK


mgyucht (changes requested):
Impl looks good to me, but would you mind adding a unit test for this case?

alexott:
What would it test? it then should be integration test, with checking DLT logs, etc.

mgyucht (approved):
Just double checked the OpenAPI spec and this LGTM! We need to prioritize moving Jobs to use Go SDK though.

PR #2447, opened 2023-06-29
Add ability to subscribe to dashboards & alert SQL tasks in `databricks_job`

### LINKED ISSUE #2436: [FEATURE] Add subscriptions to SQL Alert job tasks
### Use-cases

Currently there is no way of adding subscriptions (destinations and users) to an alert itself. This must be done via the GUI. However API also does not appear to have this functionality for alerts. There are SQL alert tasks in a job where you can also define subscriptions. These jobs also allow you to schedule when to check the alert (feature also missing on just an alert object in the api and thus terraform). However still lacks the ability to define those subscriptions in the terraform resource. It is possible to define them if creating the job via the API.

So ultimately the use case is being able to add subscriptions to alerts beyond just the creator of them.

### Attempted Solutions

No solution possible given existing features. Must use the API directly

### Proposal

Enable the functionality of the [Job API](https://docs.databricks.com/api/workspace/jobs/create) that adds subscriptions to the existing [SQL Alert Task](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/job#sql_task-configuration-block) resource

alexott:
## Changes

It's now possible to subscribe to updates of individual dashboards & alerts in the SQL tasks of Databricks Workflows

This fixes #2436

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK


mgyucht (changes requested):
Impl LGTM, but can we add a unit test verifying that the generated request has the expected fields?

mgyucht (approved):

PR #2451, opened 2023-06-30
Fix workspace binding acceptance tests

mgyucht:
## Changes
This PR fixes the workspace binding acceptance tests. There are two changes:
1. it simplifies the test cases considerably: two tests, one where a binding is created for the current workspace, and one where a binding is created for a separate workspace.
2. it fixes the delete implementation for catalog to add the current workspace to the workspace bindings for the catalog if the catalog is isolated. If the workspace is not in the bindings for an isolated catalog, even metastore admins cannot delete it via the workspace-level API.

Note that isolated catalogs automatically use the workspace binding mechanism to add the current workspace to the catalog bindings on creation, so the second change is only in case of the workspace being listed in a workspace_binding resource. In this case, the workspace binding will be cleaned up first, removing this workspace from bindings, so it must be added back first before deleting the catalog.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -137,6 +137,21 @@ func ResourceCatalog() *schema.Resource {
 				return err
 			}
 			force := d.Get("force_destroy").(bool)
+			// If the workspace has isolation mode ISOLATED, we need to add the current workspace to its
+			// bindings before deleting.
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm

PR #2452, opened 2023-06-30
Exporter: make resource names more unique to avoid duplicate resources errors

alexott:
## Changes

This includes following changes:

* Add user ID to the `databricks_user` resource name to avoid clashes on names like, `user+1` and `user_1`
* Add user/sp/group ID to the name of the `databricks_group_member` resource
* Remove too aggressive name normalization pattern that also leads to the generation of the duplicate resource names for different resources

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -92,10 +92,10 @@ type mount struct {
 var nameFixes = []regexFix{
 	{regexp.MustCompile(`[0-9a-f]{8}[_-][0-9a-f]{4}[_-][0-9a-f]{4}` +
 		`[_-][0-9a-f]{4}[_-][0-9a-f]{12}[_-]`), ""},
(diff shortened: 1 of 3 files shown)


mgyucht (commented):
I think this is OK, but can you give some examples of the impact of the changes (as I'm still gaining context on this part of the provider)? Also, if you could include your manual test in the PR description, that would be super helpful.

alexott:
Exporter uses the user's name to generate a resource name, but resource name should follow the Terraform syntax that doesn't allow characters like `+`, `@`, etc. So Exporter is replacing special characters with `_`, etc. But if you have name like, `user+1@domain.com` and `user_1@domain.com`, then they are both normalized to `user_1@domain.com` and Terraform complains about duplicate resources.

nfx (approved):

PR #2456, opened 2023-07-03
Added `databricks_mlflow_model` data source

840:
## Changes
Add `databricks_mlflow_model` data source

## Tests

- [x] `make test` run locally
- [x] using Go SDK

Partially resolves #2301


nkvuong:
acceptance tests passed on all 3 clouds (unrelated test failed on Azure)

docs/data-sources/mlflow_model.md
```diff
@@ -0,0 +1,35 @@
+---
+subcategory: "MLflow"
+---
+# databricks_mlflow_model Data Source
+
+-> **Note** If you have a fully automated setup with workspaces created by [databricks_mws_workspaces](../resources/mws_workspaces.md) or [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace), please make sure to add [depends_on attribute](../index.md#data-resources-and-authentication-is-not-configured-errors) in order to prevent _default auth: cannot configure default credentials_ errors.
+
+Retrieves the settings of [databricks_mlflow_model](../resources/mlflow_model.md) by name.
+
+## Example Usage
```
nkvuong, line 10: could you add examples of how this data source can potentially be used with other mlflow resources?

docs/data-sources/mlflow_model.md
```diff
@@ -0,0 +1,35 @@
...
+data "databricks_mlflow_model" "this" {
+  name = "My MLflow Model"
+  version = "1"
+}
+```
+
+## Argument Reference
+
+* `name` - (Required) Name of the registered model.
+* `version` - (Optional) Model version number.
```
mgyucht, line 27: What is the default behavior if version is unspecified?

mlflow/data_mlflow_model.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/ml"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceModel() *schema.Resource {
+	return common.WorkspaceData(func(ctx context.Context, data *struct {
+		Name         string            `json:"name,omitempty" tf:"computed"`
+		Version      string            `json:"version,omitempty" tf:"computed"`
```
mgyucht, line 15: These are not computed, right? Or am I missing something?

mgyucht: Also, I think only Version is optional. Can you try removing `omitempty` from Name? Then you won't need the check on line 20, if I'm reading it properly.

mlflow/data_mlflow_model.go
```diff
@@ -0,0 +1,45 @@
...
+		ModelVersion []ml.ModelVersion `json:"model_versions,omitempty" tf:"computed"`
```
mgyucht, line 16: This only returns a single model. Why not use *ml.ModelVersion?

mgyucht (commented):
Couple questions, otherwise LGTM!

docs/data-sources/mlflow_model.md
```diff
@@ -0,0 +1,35 @@
+---
+subcategory: "MLflow"
+---
+# databricks_mlflow_model Data Source
+
+-> **Note** If you have a fully automated setup with workspaces created by [databricks_mws_workspaces](../resources/mws_workspaces.md) or [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace), please make sure to add [depends_on attribute](../index.md#data-resources-and-authentication-is-not-configured-errors) in order to prevent _default auth: cannot configure default credentials_ errors.
+
+Retrieves the settings of [databricks_mlflow_model](../resources/mlflow_model.md) by name.
```
mgyucht, line 8: In other data sources, getting by ID is also typically supported. Should we add that in this one as well?

mlflow/data_mlflow_model.go
840: A model can have an array of model versions. (Ref: [document](https://docs.databricks.com/api/workspace/modelregistry/getmodel))

If no version is specified, all latest version models are returned. If a model version is specified, it will only return that version.

I think it is more user friendly if it stays as an array even if only one model is returned to avoid confusion. But if you think there is a better approach to this, please let me know.

docs/data-sources/mlflow_model.md
840: The default behaviour is that it will return all latest version models which is the same behaviour described in the [documentation](https://docs.databricks.com/api/workspace/modelregistry/getlatestversions).

docs/data-sources/mlflow_model.md
840: I reviewed the documentation, it looks like [fetching a single model](https://docs.databricks.com/api/workspace/modelregistry/getmodel) does return the ID. However, [listing models](https://docs.databricks.com/api/workspace/modelregistry/listmodels) in order to search by ID does not seem to be present.

internal/acceptance/data_mlflow_model_test.go
```diff
@@ -0,0 +1,34 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+func TestAccDataSourceMlflowModel(t *testing.T) {
+	workspaceLevel(t, step{
```
mgyucht, line 8: Just to make sure this works, can we have two steps:
1. create the model
2. create the data source.
Then we can verify that we got back the model we expected.

docs/data-sources/mlflow_model.md
```diff
@@ -0,0 +1,60 @@
+---
+subcategory: "MLflow"
+---
+# databricks_mlflow_model Data Source
```
mgyucht, line 4: IIUC, this is actually a datasource for model versions, not for models. Should we call this `databricks_mlflow_model_versions`? We may want to reserve `databricks_mlflow_model` for a datasource that returns models directly rather than model versions.

mgyucht (commented):
We should probably distinguish between the singular version (model version for a model at a particular version) and the plural version (all model versions of a model). In the future, we may want to add other options (like filtering model versions created since some date) to the plural version that don't really make sense in the singular one. I think this is a practice we have elsewhere in the provider (e.g. databricks_sql_warehouse and databricks_sql_warehouses) and is common in other providers as well. 

So in summary, let's split this resource into databricks_mlflow_model_version and databricks_mlflow_model_versions. The first takes the model version as a required parameter and returns a single model version. The second takes no parameters and returns a list of model versions. WDYT?

internal/acceptance/data_mlflow_model_test.go
```diff
@@ -0,0 +1,37 @@
...
+				key   = "key2"
+				value = "value2"
+			  }
+			}`,
+		},
+		step{
+			Template: `
+			data "databricks_mlflow_model" "this" {
+			  depends_on = [databricks_mlflow_model.this]
+			  name       = "My MLflow Model"
```
tanmay-db, line 30: This would need to be updated accordingly as well

docs/data-sources/mlflow_model.md
tanmay-db: Hi @840, was there any other discussion regarding this?

docs/data-sources/mlflow_model.md
tanmay-db: @840 +1 to this as well.

docs/data-sources/mlflow_model.md
840: For now this PR will be for `databricks_mlflow_model`. I changed the logic so it won't use the `models_versions` API. I will add the other data sources in a separate PR.

`data_mlflow_model_version` - Will only reference a specific version of a model ([/api/2.0/mlflow/model-versions/get](https://docs.databricks.com/api/workspace/modelregistry/getmodelversion))
`data_mlflow_model_versions` - Will return the latest versions from ([/api/2.0/mlflow/registered-models/get-latest-versions](https://docs.databricks.com/api/workspace/modelregistry/getlatestversions))

docs/data-sources/mlflow_model.md
840: Added

internal/acceptance/data_mlflow_model_test.go
```diff
@@ -0,0 +1,37 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+func TestAccDataSourceMlflowModel(t *testing.T) {
```
tanmay-db, line 7: Hi @840 , the test is failing with:
````
2023/10/12 14:19:35 [INFO] ❌ TestAccDataSourceMlflowModel (1.970s)
2023/10/12 14:19:35 [INFO][TestAccDataSourceMlflowModel] === RUN   TestAccDataSourceMlflowModel
    init_test.go:43: ***
=== PAUSE TestAccDataSourceMlflowModel
=== CONT  TestAccDataSourceMlflowModel
    init_test.go:238: Step 2/2 error: Error running pre-apply refresh: exit status 1
        
        Error: Reference to undeclared resource
        
          on terraform_plugin_test.tf line 2, in data "databricks_mlflow_model" "this":
           2:   depends_on = [databricks_mlflow_model.this]
        
        A managed resource "databricks_mlflow_model" "this" has not been declared in
        the root module.
        
        Did you mean the data resource data.databricks_mlflow_model.this?
    testing_new.go:[85](https://github.com/databricks/eng-dev-ecosystem/actions/runs/6496639828/job/17644027311#step:8:86): Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Reference to undeclared resource
        
          on terraform_plugin_test.tf line 2, in data "databricks_mlflow_model" "this":
           2:   depends_on = [databricks_mlflow_model.this]
        
        A managed resource "databricks_mlflow_model" "this" has not been declared in
        the root module.
        
        Did you mean the data resource data.databricks_mlflow_model.this?
--- FAIL: TestAccDataSourceMlflowModel (1.[97](https://github.com/databricks/eng-dev-ecosystem/actions/runs/6496639828/job/17644027311#step:8:98)s)
````
can you please take a look? Thank you

File: internal/acceptance/data_mlflow_model_test.go [hunk 10]
```diff
@@ -0,0 +1,52 @@
...
+			}			
+
+			data "databricks_mlflow_model" "this" {
+			  depends_on = [databricks_mlflow_model.this]
+			  name       = databricks_mlflow_model.this.name
+			}
+			
+			output "model" {
+			  value = data.databricks_mlflow_model.this
+			}`,
```
tanmay-db, line 50: Rather than just outputting, it would be good if we can assert check on some values. For reference, we can do something similar to `TestAccDataSourceMwsWorkspaces`

internal/acceptance/data_mlflow_model_test.go
```diff
@@ -0,0 +1,52 @@
+package acceptance
+
+import (
+	"testing"
+)
+
+func TestAccMlflowModel(t *testing.T) {
```
tanmay-db, line 7: There is already a test with the name `TestAccMlflowModel`, since this is for data source -- let's rename it to `TestAccDataMlflowModel`

```suggestion
func TestAccDataMlflowModel(t *testing.T) {
```

tanmay-db (approved):
Small comment on testing, rest LGTM.

PR #2458, opened 2023-07-03
Allow search SQL Warehouses by name in `databricks_sql_warehouse` data source

### LINKED ISSUE #2443: [FEATURE] `databricks_sql_warehouse` should support returning a SQL warehouse by name
### Use-cases

Requiring that we refer SQL warehouse only by ID limits the portability of the terraform code. It's possible to get it by name by using a combination of `databricks_sql_warehouse` with `databricks_sql_warehouses`, but it's making code more verbose

alexott:
## Changes

Right now it's possible to search only by the warehouse ID, but it's not always convenient although it's possible by using `databricks_sql_warehouses` data source + explicit filtering.  This PR adds a capability to search by either SQL warehouse name or ID.

This fixes #2443

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK


sql/data_sql_warehouse.go
```diff
@@ -31,11 +32,36 @@ func DataSourceWarehouse() *schema.Resource {
...
-		err := c.Get(ctx, fmt.Sprintf("/sql/warehouses/%s", data.ID), nil, data)
+		var id string
+		endpointsAPI := NewSQLEndpointsAPI(ctx, c)
+		if data.Name != "" {
+			endpoints, err := endpointsAPI.List()
+			if err != nil {
+				return err
+			}
+			selected := []SQLEndpoint{}
+			lname := strings.ToLower(data.Name)
```
mgyucht, line 43: Why not exact match? Is this a pattern we have for data sources when looking up by name today? It would fail if two warehouses have the same name modulo case, e.g. "test" and "Test".

mgyucht (commented):
Few comments, but thanks for adding support for this!

docs/data-sources/sql_warehouse.md
```diff
@@ -19,18 +19,25 @@ data "databricks_sql_warehouse" "all" {
...
 
-* `id` - (Required) The ID of the SQL warehouse
+* `id` - (Required, if `name` isn't specified) The ID of the SQL warehouse.
+* `name` - (Required, if `ud` isn't specified) Name of the SQL warehouse to search (case-insensitive).
 
 ## Attribute reference
 
 This data source exports the following attributes:
 
-* `name` - Name of the SQL warehouse. Must be unique.
```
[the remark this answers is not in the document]
alexott: yes, makes sense - I'll update tomorrow.

sql/data_sql_warehouse.go
alexott: It's just for convenience. Our UI doesn't allow to create warehouses with the same name, but different cases. But yes, it looks like other resources aren't supporting that - jobs & clusters use exact matching.

alexott: will remove that tomorrow.

sql/data_sql_warehouse.go
```diff
@@ -31,11 +31,35 @@ func DataSourceWarehouse() *schema.Resource {
 
 	return common.DataResource(SQLWarehouseInfo{}, func(ctx context.Context, e interface{}, c *common.DatabricksClient) error {
 		data := e.(*SQLWarehouseInfo)
-		err := c.Get(ctx, fmt.Sprintf("/sql/warehouses/%s", data.ID), nil, data)
+		var id string
```
mgyucht, line 34: Hm, this is interesting. Maybe I'm misunderstanding the API, but I think that for either warehouse ID or name, you just need to make a single API call to https://docs.databricks.com/api/workspace/datasources/list and can filter the result list based on the "name" or "warehouse_id" fields.

mgyucht (commented):
Does it work if we just use the list data sources API and filter based on name or warehouse ID?

sql/data_sql_warehouse.go
alexott: We need to have `Get` call anyway because the data sources APIs don't have all necessary information about warehouse that data source returns

mgyucht: I see. Then maybe one small suggestion would be to do these requests in the opposite order. That way you could get the datasource ID and warehouse ID from the data sources API, filtering by warehouse name/ID, then look up the warehouse by ID.

sql/data_sql_warehouse.go
```diff
@@ -31,15 +32,37 @@ func DataSourceWarehouse() *schema.Resource {
...
+			log.Printf("[DEBUG] source=%v", source)
+			if data.Name != "" && source.Name == data.Name {
+				selected = append(selected, source)
+			} else if data.ID != "" && source.EndpointID == data.ID {
+				selected = append(selected, source)
+				break
+			}
+		}
+		if len(selected) == 0 {
+			return fmt.Errorf("can't find SQL warehouse with the name '%s'", data.Name)
```
mgyucht, line 55: suggestion: let's tune the error message based on whether ID or Name was specified.

sql/data_sql_warehouse.go
```diff
@@ -31,15 +32,37 @@ func DataSourceWarehouse() *schema.Resource {
...
+		endpointsAPI := NewSQLEndpointsAPI(ctx, c)
+		selected := []DataSource{}
+		dataSources, err := endpointsAPI.listDataSources()
 		if err != nil {
 			return err
 		}
-		endpointsAPI := NewSQLEndpointsAPI(ctx, c)
-		data.DataSourceID, err = endpointsAPI.ResolveDataSourceID(data.ID)
+		for _, source := range dataSources {
+			log.Printf("[DEBUG] source=%v", source)
```
mgyucht, line 46: Remove this?

mgyucht (approved):
Mostly LGTM, just two comments that we should address before merging.

PR #2460, opened 2023-07-03
Fixed model serving integration test

shreyas-goenka:
## Changes
getStatus was removed from the newer versions of the mlflow library. This PR fixes this test

## Tests
Integration test passes now

PR #2461, opened 2023-07-03
Use default cluster for model serving integration test

shreyas-goenka:
## Changes
Uses the default cluster we have in our test infra to run the test. This can cut down on the test time needed to spin up a new cluster
## Tests
Self testing. Integration test passes now

Builds on top of https://github.com/databricks/terraform-provider-databricks/pull/2460


alexott (approved):
lgtm

PR #2472, opened 2023-07-09
Fix handling of comments in `databricks_sql_table` resource

### LINKED ISSUE #2396: [ISSUE] Issue with `databricks_sql_table` resource for column comments
### Configuration
```hcl
resource "databricks_sql_table" "abc" {
  name               = "abc"
  catalog_name       = local.catalog
  schema_name        = databricks_schema.dummy_schema.name
  table_type         = "MANAGED"
  data_source_format = "DELTA"

  column {
    name    = "a"
    type    = "STRING"
    comment = "A comment"
  }
  column {
    name    = "b"
    type    = "STRING"
    comment = "Another comment that doesn't work."
  }
  comment = "Here the comment works"
}
```

### Expected Behavior
It needs single quoting for column comments as below:
```sql
-- Error: cannot create sql table: cannot execute
CREATE TABLE dummy_catalog.dummy_schema.abc (
a STRING COMMENT 'A comment.',
b STRING COMMENT 'Another comment that doesn't work.'
) USING DELTA
COMMENT 'Here the comment works';
```
Here is the line which creates the bug: https://github.com/databricks/terraform-provider-databricks/blob/v1.18.0/catalog/resource_sql_table.go#L149 It should be like this:
```go
comment = fmt.Sprintf(" COMMENT '%s'", col.Comment)
```

How should we proceed? Shall I open an fix PR?

### Actual Behavior
Currently the resource returns the query below which doesn't have quoting in comment for columns.
```sql
-- Error: cannot create sql table: cannot execute
CREATE TABLE dummy_catalog.dummy_schema.abc (
a STRING COMMENT A comment.,
b STRING COMMENT Another comment that doesn't work.
) USING DELTA
COMMENT 'Here the comment works';
```

### Steps to Reproduce

1. `terraform apply`

Terraform v0.15.1 on darwin_amd64
+ provider registry.terraform.io/databricks/databricks v1.18.0

karolusz:
## Changes
Closes #2396 

The SQL column comment is now single quotation literal (the same as table comment).
In addition, small function `parseComment` is added to avoid delimiter collision in the column and table comment strings. 

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK


catalog/resource_sql_table.go
```diff
@@ -57,6 +57,10 @@ func (ti *SqlTableInfo) FullName() string {
 	return fmt.Sprintf("%s.%s.%s", ti.CatalogName, ti.SchemaName, ti.Name)
 }
 
+func parseComment(s string) string {
+	return strings.ReplaceAll(s, "'", `\'`)
```
alexott, line 61: what if we already have it escaped by `\`?

alexott:
Please also add unit tests...

catalog/resource_sql_table.go
karolusz: To not break comments that have single quotes escaped by `\`,, the following change in commit 6bbe0b12d8489b3c409bac270a553b9255c7bf4b was made:
```go
func parseComment(s string) string {
	return strings.ReplaceAll(strings.ReplaceAll(s, `\'`, `'`), `'`, `\'`)
}
```

Right now the user can make sure that the single quotation mark will be escaped in the SQL query in the following ways in Terraform:
```terraform
  comment = <<-EOT
    single quote in the comment \'
  EOT
```
or 
```terraform
comment = "single quote in the comment \\'"
```
However, the comment will be flagged as an update, next time Terraform script is executed, as the string read from the table will not include the escape backslash:
```terrafrom
  ~ resource "databricks_sql_table" "abc" {
      ~ comment            = <<-EOT
          - single quote in the comment '
          + single quote in the comment \'
        EOT
```

The proposed change will not change the behavior for users who already escape with `\\'` (in Terrafrom double-quoted string) or with `\'` in Terrafrom heredoc string. For new table or column comments, users will no longer be required to escape the single quotation and comments will not be falsely flagged as updates for that reason.

karolusz: To not break comments that have single quotes escaped by `\`,, the following change in commit 6bbe0b12d8489b3c409bac270a553b9255c7bf4b was made:
```go
func parseComment(s string) string {
	return strings.ReplaceAll(strings.ReplaceAll(s, `\'`, `'`), `'`, `\'`)
}
```

Right now the user can make sure that the single quotation mark will be escaped in the SQL query in the following ways in Terraform:
```terraform
  comment = <<-EOT
    single quote in the comment \'
  EOT
```
or 
```terraform
comment = "single quote in the comment \\'"
```
However, the comment will be flagged as an update, next time Terraform script is executed, as the string read from the table will not include the escape backslash:
```terrafrom
  ~ resource "databricks_sql_table" "abc" {
      ~ comment            = <<-EOT
          - single quote in the comment '
          + single quote in the comment \'
        EOT
```

The proposed change will not change the behavior for users who already escape with `\\'` (in Terrafrom double-quoted string) or with `\'` in Terrafrom heredoc string. For new table or column comments, users will no longer be required to escape the single quotation and comments will not be falsely flagged as updates for that reason.

karolusz:
> Please also add unit tests...

unit tests added in commit 980c4b466f05b808cb05120f1550ff1c77923d8c

nfx (approved):

karolusz:
Corrected formatting in `resource_sql_table_test.go`

mgyucht (approved):
Seems fine for now

PR #2487, opened 2023-07-18
[SqlEndpoint] Force recreation when EnableServerlessCompute is modified

larrywax:
## Changes

Add flag force_new on parameter EnableServerlessCompute.
Modifying this parameter has no effect unless the Sql endpoint is recreated.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/sql/resource_sql_endpoint.go
+++ b/sql/resource_sql_endpoint.go
@@ -29,7 +29,7 @@ type SQLEndpoint struct {
 	MaxNumClusters          int             `json:"max_num_clusters,omitempty" tf:"default:1"`
 	NumClusters             int             `json:"num_clusters,omitempty" tf:"suppress_diff"`
 	EnablePhoton            bool            `json:"enable_photon" tf:"optional,default:true"`
-	EnableServerlessCompute bool            `json:"enable_serverless_compute,omitempty" tf:"suppress_diff"`
+	EnableServerlessCompute bool            `json:"enable_serverless_compute,omitempty" tf:"suppress_diff,force_new"`
 	InstanceProfileARN      string          `json:"instance_profile_arn,omitempty"`
 	State                   string          `json:"state,omitempty" tf:"computed"`
 	JdbcURL                 string          `json:"jdbc_url,omitempty" tf:"computed"`


alexott:
Why it should be recreated? API allows update of this flag.

larrywax:
> Why it should be recreated? API allows update of this flag.

Because it has no effect if you switch it from enabled to disabled, sadly

alexott:
It's a known issue, but forcing recreation is too dangerous as it may require cascading updates

mgyucht:
It sounds like this is a bug in the API rather than in the TF provider. I'll create a follow-up ticket for them to take a look.

alexott:
@mgyucht It's a kind of bug in the provider related to the handling of `omitempty`. The problem is that we can't send some default value on each request if we mark that field as `tf:optional` because it will break the existing API behavior. If the value is not set and the workspace is enabled, the warehouse is created as serverless, and vice versa.

mgyucht:
@larrywax thanks for reporting this. We believe the root cause of this issue to be tied to the treatment of optional parameters in the Go SDK. Today, all fields are modeled as non-pointer types in the Go SDK, and optional fields are tagged with `json:"...,omitempty"`. This means that it is impossible to explicitly send a parameter with the zero value. We will look into changing these optional values to be pointer types to be able to distinguish whether an optional field is set or not. All optional parameters are subject to this behavior across Terraform for resources using the Go SDK. We'll work on a fix for this in the coming month or two.

mgyucht:
Just to close the loop here: as I alluded to above, we're currently making a change in our Go SDK to make it possible to serialize fields with the zero value: see https://github.com/databricks/terraform-provider-databricks/pull/2753. I think this change (and others with the same root cause) can be addressed there. I'll close this PR, and we can follow-up on #2753.

alexott:
> cc @hectorcast-db here's an API that we can try to set ForceSendFields with your change to the Go SDK.

@mgyucht be careful - this specific use case isn't implemented yet because behavior needs to match actual behavior in the workspace.

hectorcast-db:
@alexott We are working on sending default fields if they are set in the TF resource by the customer. We will still not send all default values.

alexott:
I would be interested in seeing how it will work for a specific use case - `GetOk` doesn't help here

PR #2492, opened 2023-07-19
Added support for Unity Catalog `databricks_metastore` data source

tanmay-db:
## Changes

Enable fetching account level metastore information through id for a single metastore. 

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/docs/data-sources/metastore.md
+++ b/docs/data-sources/metastore.md
@@ -12,7 +12,9 @@ Retrieves information about metastore for a given id of [databricks_metastore](.
 MetastoreInfo response for a given metastore id
 
 ```hcl
-data "databricks_metastore" "this" {}
+data "databricks_metastore" "this" {
+  id = 
+}
 
 output "some_metastore" {
   value = data.databricks_metastore.this.metastore_info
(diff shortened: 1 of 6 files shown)


docs/data-sources/metastore.md
```diff
@@ -0,0 +1,53 @@
...
+  force_destroy = true
+}
+
+data "databricks_metastore" "this" {
+  id = databricks_metastore.this.id
+}
+
+output "some_metastore" {
+  value = data.databricks_metastore.this.metastore_info[0]
+}
```
tanmay-db, line 30: We have this in other places as well: https://github.com/databricks/terraform-provider-databricks/blob/master/docs/data-sources/job.md?plain=1#L20. Have brought this up internally, not sure if there already exists a ticket for this.

nkvuong: to be more precise, `metastore_info` is rendered as a list by the Terraform provider, so we need to use [0] to retrieve

docs/data-sources/metastores.md
tanmay-db, line 1: @mgyucht There is one concern here. We are mentioning: 
`Data resource will error in case of metastores with duplicate names.`

Me and @nkvuong discussed this and it's unlikely that users will have same metastore names, mostly they suffix / prefix them with region. We can update this to be a map of name -> list of ids but that would be bad UX for most of the people and make general usage complicated. 

What do you think?

tanmay-db: We have this for `databricks_jobs` as well i.e. `Name` for `JobSettings`

mgyucht: For some resources, it is possible to have multiple distinct resources with the same name, like jobs. While this is not a good practice and we're trying to move away from it as a company, it is something that happens. Since we already have this pattern in Jobs, I think it's OK to use it for metastores as well. In parallel, we can follow up with the UC team to understand if this case is even allowed.

mgyucht (approved):
LGTM!

PR #2493, opened 2023-07-19
Added exporter for `databricks_workspace_file` resource

alexott:
## Changes

It's now possible to export `databricks_workspace_file` resources either directly or when they are used inside clusters/jobs/pipelines definitions (as init scripts).

Besides that change, this PR includes following:

- Fixed emitting of the permissions for notebooks & directories as they were dependent on the specific format of the resource name
- Avoid code duplication for names generation and listing of notebooks, workspace files and directories
- make `NotebooksAPI.List` to return directories as well when called in the recursive mode (same as non-recursive behavior)
- Because of that, remove the separate `ListDirectories`
- Extend `workspace.ObjectStatus` with additional fields (will be required for incremental notebooks export)
- Cache listing of all workspace objects, and then use it for all operations - list notebooks, list directories, list workspace files 

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK


exporter/importables.go
```diff
@@ -752,7 +763,7 @@ var resourcesMap map[string]importable = map[string]importable{
 				displayName := d.Get("display_name").(string)
 				return applicationID == displayName
 			}
-			return pathString == "home" || pathString == "repos" || (pathString == "application_id" && !ic.Client.IsAzure())
```
mgyucht, line 755: This is just a bug I guess? These fields don't seem to belong to databricks_service_principal?

exporter/importables.go
```diff
@@ -1208,34 +1188,79 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			{Path: "source", File: true},
+			{Path: "path", Resource: "databricks_user", Match: "home", MatchType: MatchPrefix},
+			{Path: "path", Resource: "databricks_service_principal", Match: "home", MatchType: MatchPrefix},
+		},
+	},
+	"databricks_workspace_file": {
+		Service: "notebooks",
+		Name:    workspaceObjectResouceName,
+		List:    createListWorkspaceObjectsFunc(workspace.File, "databricks_workspace_file", "workspace_file"),
+		Import: func(ic *importContext, r *resource) error {
```
mgyucht, line 1228: This is also almost exactly the same as for databricks_notebook. Is there a good reason for their implementations to diverge over time, or would it be better to have one function generator for importing both of these resources (like you do for list)?

exporter/util.go
```diff
@@ -108,21 +120,15 @@ func (ic *importContext) emitUserOrServicePrincipalForPath(path, prefix string)
...
-				return false
-			}
-		} else {
-			return false
-		}
-	} else {
+	if !strings.HasPrefix(path, prefix) {
 		return false
 	}
+	parts := strings.SplitN(path, "/", 4)
```
mgyucht, line 126: thank you! old code hurts my eyes to read a bit

exporter/util.go
```diff
@@ -142,12 +148,24 @@ func (ic *importContext) emitNotebookOrRepo(path string) {
...
+			if v.ObjectType == workspace.Directory {
+				ic.allDirectories = append(ic.allDirectories, v)
+			}
+		}
 	}
 	return ic.allDirectories
 }
 
+func (ic *importContext) getAllWorkspaceObjects() []workspace.ObjectStatus {
+	if len(ic.allWorkspaceObjects) == 0 {
```
mgyucht, line 162: I suggest we use a mutex here to protect access to this field

alexott: Right now exporter is single threaded, so no need for the mutex. I'm thinking on how to make it running in parallel, but it requires more thinking.

exporter/importables.go
alexott: I was thinking about unifying, but there are a bit different implementations for writing the data & emitting permissions. I also have a request to export notebooks not only as source, but as `.ipynb` as well, so they will diverge more. But I need to think, maybe there could be a common part + function for actual data writing, etc.

exporter/util.go
mgyucht: Gotcha. At least this is in one method. We'll address multithreading later then.

mgyucht (approved):
LGTM, though at some point we need to really revisit the exporter.

PR #2494, opened 2023-07-19
Fix provider after updating SDK to 0.13

fjakobs:
## Changes
Fix provider after updating SDK to 0.13


sql/resource_sql_alerts.go
```diff
@@ -12,7 +12,7 @@ import (
 type AlertOptions struct {
 	Column        string `json:"column"`
 	Op            string `json:"op"`
-	Value         string `json:"value"`
+	Value         any    `json:"value"`
```
tanmay-db, line 15: I suspect we probably can't directly use `any` here without other changes down the line.

alexott: We can't change type of the existing resource. See this PR on how would you handle it: https://github.com/databricks/terraform-provider-databricks/commit/b2208bb8c941d2519359e3ac3dc196206c5b7c99

fjakobs: This didn't work indeed. I think I have a solution for now. It's a pity that we don't have union types in OpenAPI (e.g. string or number).

sql/resource_sql_alerts.go
```diff
@@ -70,15 +72,25 @@ func (a *AlertEntity) fromAPIObject(apiAlert *sql.Alert, s map[string]*schema.Sc
...
 		CustomBody:    apiAlert.Options.CustomBody,
 		CustomSubject: apiAlert.Options.CustomSubject,
 	}
 
+	// value can be a string or a float64 - unfortunately this can't be encoded in OpenAPI yet
+	switch value := apiAlert.Options.Value.(type) {
+	case string:
+		a.Options.Value = value
+	case float64:
+		a.Options.Value = strconv.FormatFloat(value, 'f', 0, 64)
```
alexott, line 89: should we convert back to float when in ` toCreateAlertApiObject` ?

mgyucht: How do we know if it is supposed to be a float?

It seems that TF mandates that everything has a type, so we aren't allowed to declare that any type can be provided in the HCL and let the provider will deal with it. If we want to handle this, we will need to expose additional fields on AlertOptions like `NumericValue`/`BooleanValue` to let users choose the type of the value they are setting. For now, let's just handle the string case, since that is all that is supported in TF.

fjakobs: I think we need to. Let me look into it

fjakobs: The cleanest approach would be to map `value: string | number` to `StringValue: string` and `NumberValue: number` in HCL and make both optional. 

The problem is that this would be a backward incompatible change since we currently have `value: string` in HCL. IF we assume that nobody ever used this because it was always broken then we are good but otherwise, we'd have to keep value, map it to `StringValue` and mark it as deprecated.

Makefile
```diff
@@ -12,11 +12,11 @@ fmt-docs:
 
 lint: vendor
 	@echo "✓ Linting source code with https://staticcheck.io/ ..."
-	@staticcheck ./...
+	@go run honnef.co/go/tools/cmd/staticcheck@v0.4.0 ./...
```
mgyucht, line 15: We're using 0.3.2, but latest release is 0.4.3. While I'm happy to bump the version of this linter, should we do this in a separate PR?

If we do this, we should remove https://github.com/databricks/terraform-provider-databricks/blob/remove-legacy-dlvLoadConfig/.github/workflows/push.yml#L40-L41.

sql/resource_sql_alerts.go
mgyucht: Do we potentially need to handle boolean values?

Makefile
fjakobs: done

sql/resource_sql_alerts.go
fjakobs: No, just strings and numbers

sql/resource_sql_alerts_test.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+		Resource: ResourceSqlAlert(),
+		Read:     true,
+		ID:       "xyz",
+	}.Apply(t)
+
+	assert.NoError(t, err)
+	assert.Equal(t, "xyz", d.Id(), "Resource ID should not be empty")
+
+	d, err = qa.ResourceFixture{
```
mgyucht, line 41: Can we split this into a separate test function?

sql/resource_sql_alerts_test.go
```diff
@@ -0,0 +1,70 @@
...
+				},
+			},
+		},
+		Resource: ResourceSqlAlert(),
+		Read:     true,
+		ID:       "xyz",
+	}.Apply(t)
+
+	assert.NoError(t, err)
+	assert.Equal(t, "xyz", d.Id(), "Resource ID should not be empty")
```
mgyucht, line 39: We can test that the right value is stored in the alert by doing `assert.Equal(t, "10", d.Get("options.value").(string), "Value should be correctly propagated")`. Let's do a similar check in the second case as well

mgyucht (commented):
LGTM with three small changes:
1. Remove the unnecessary docker-it file.
2. Make two separate test cases.
3. Add some assertion that the value is correctly parsed (see example in my comment)

sql/resource_sql_alerts_test.go
fjakobs: done

mgyucht (approved):
LGTM with last test fix

sql/resource_sql_alerts_test.go
fjakobs: I couldn't make this work as the framework it removing default fields.

PR #2495, opened 2023-07-19
Added `databricks_file` resource

tanmay-db:
## Changes

Adding support for Files API in terraform provider. Relevant Changes in Go SDK: https://github.com/databricks/databricks-sdk-go/pull/423

## Tests

- [x] `make test` run locally
- [x] using Go SDK


docs/resources/files.md
```diff
@@ -0,0 +1,54 @@
...
+This is a resource that lets you upload and download files upto 2GB in octet-stream.
+
+## Example Usage
+
+In order to manage a file on Databricks File System with Terraform, you must specify the `source` attribute containing the full path to the file on the local filesystem.
+
+```hcl
+resource "databricks_files" "this" {
+  source = "${path.module}/main.tf"
+  path   = "/tmp/main.tf"
```
tanmay-db, line 15: to check: only volumes?

internal/acceptance/files_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"testing"
+)
+
+func TestAccFiles(t *testing.T) {
+	workspaceLevel(t, step{
+		Template: `resource "databricks_files" "this" {
+			source = "{var.CWD}/../../storage/testdata/tf-test-python.py"
+			path = "/Shared/provider-test/xx_{var.RANDOM}"
```
tanmay-db, line 11: to check: only volumes? -- if yes -- create volumes first then files

docs/resources/files.md
```diff
@@ -0,0 +1,54 @@
...
+
+```bash
+$ terraform import databricks_files.this <path>
+```
+
+## Related Resources
+
+The following resources are often used in the same context:
+
+* [databricks_workspace_file](./workspace_file.md)
```
tanmay-db, line 47: confusing, 

databricks workspace file
dbfs file
databricks files

confusing -- what should customers use? and what is the difference? This needs to be documented clearly

docs/resources/files.md
```diff
@@ -0,0 +1,54 @@
...
+  source = "${path.module}/main.tf"
+  path   = "/tmp/main.tf"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `source` - The full absolute path to the file. Conflicts with `content_base64`.
```
alexott, line 23: we don't have `content_base64` in this resource, but it would be really useful to have it similar to `databricks_dbfs_file`, especially for small things, like, init scripts

docs/resources/files.md
```diff
@@ -0,0 +1,54 @@
+---
+subcategory: "Storage"
+---
+# databricks_files Resource
+
+This is a resource that lets you upload and download files upto 2GB in octet-stream.
+
+## Example Usage
+
+In order to manage a file on Databricks File System with Terraform, you must specify the `source` attribute containing the full path to the file on the local filesystem.
```
alexott, line 10: `Databricks File System` ? is it copy-paste? We need to specify to where these files could be uploaded

storage/resource_databricks_files.go
```diff
@@ -0,0 +1,103 @@
...
+
+			path := data.Id()
+			fileInfo, err := w.Files.GetStatus(ctx, files.GetStatusRequest{Path: path})
+			if err != nil {
+				return err
+			}
+			data.Set("modification_time", fileInfo.ModificationTime)
+			data.Set("file_size", fileInfo.FileSize)
+
+			downloadResponse, err := w.Files.Download(ctx, files.DownloadRequest{FilePath: path})
```
alexott, line 64: For big files, reading files back could be a big bottleneck...

alexott (changes requested):

docs/resources/files.md
tanmay-db: Added

storage/resource_databricks_files.go
tanmay-db: true but there isn't any workaround at the moment.

docs/resources/files.md
```diff
@@ -0,0 +1,108 @@
+---
+subcategory: "Storage"
+---
+# databricks_files Resource
+
+This resource allows uploading and downloading files upto 2GB in octet-stream.
+
+## Example Usage
+
+In order to manage a file on Unity Catalog Volumes with Terraform, you must specify the `source` attribute containing the full path to the file on the local filesystem.
```
alexott, line 10: if it's only for UC Volumes, should we incorporate this into the resource name?

storage/resource_databricks_files.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		reader = io.NopCloser(bytes.NewReader(decodedString))
+		if err != nil {
+			return nil, err
+		}
+	}
+	return reader, err
+}
+
+func ResourceFiles() *schema.Resource {
```
nkvuong, line 44: it should be singular to be consistent with other file resource

```suggestion
func ResourceFile() *schema.Resource {
```

docs/resources/files.md
nkvuong: now that I've thought about it, would make sense to call it `databricks_volume_file`

nkvuong (commented):
one minor change to doc, and a comment on the name of the resource

docs/resources/files.md
tanmay-db: @alexott @nkvuong Files for now only supports UC Volumes but is designed to be extended for other use cases as well so if we incorporate that in name then it will be a breaking change in future.

nkvuong: if that's the case, then let's call it `databricks_file` (singular)

tanmay-db: Updated

nkvuong (approved):
lgtm

docs/resources/files.md
```diff
@@ -0,0 +1,108 @@
...
+* [End to end workspace management](../guides/workspace-management.md) guide.
+* [databricks_dbfs_file](../data-sources/dbfs_file.md) data to get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_dbfs_file_paths](../data-sources/dbfs_file_paths.md) data to get list of file names from get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_global_init_script](global_init_script.md) to manage [global init scripts](https://docs.databricks.com/clusters/init-scripts.html#global-init-scripts), which are run on all [databricks_cluster](cluster.md#init_scripts) and [databricks_job](job.md#new_cluster).
+* [databricks_library](library.md) to install a [library](https://docs.databricks.com/libraries/index.html) on [databricks_cluster](cluster.md).
+* [databricks_mount](mount.md) to [mount your cloud storage](https://docs.databricks.com/data/databricks-file-system.html#mount-object-storage-to-dbfs) on `dbfs:/mnt/name`.
+* [databricks_workspace_conf](workspace_conf.md) to manage workspace configuration for expert usage.
```
alexott, line 108: The `workspace_conf` isn't relevant here.

docs/resources/files.md
```diff
@@ -0,0 +1,108 @@
...
+
+The following resources are often used in the same context:
+
+* [databricks_workspace_file](./workspace_file.md)
+* [End to end workspace management](../guides/workspace-management.md) guide.
+* [databricks_dbfs_file](../data-sources/dbfs_file.md) data to get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_dbfs_file_paths](../data-sources/dbfs_file_paths.md) data to get list of file names from get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_global_init_script](global_init_script.md) to manage [global init scripts](https://docs.databricks.com/clusters/init-scripts.html#global-init-scripts), which are run on all [databricks_cluster](cluster.md#init_scripts) and [databricks_job](job.md#new_cluster).
+* [databricks_library](library.md) to install a [library](https://docs.databricks.com/libraries/index.html) on [databricks_cluster](cluster.md).
+* [databricks_mount](mount.md) to [mount your cloud storage](https://docs.databricks.com/data/databricks-file-system.html#mount-object-storage-to-dbfs) on `dbfs:/mnt/name`.
```
alexott, line 107: I suggest to remove mention of mounts at all - it's a deprecated pattern

docs/resources/files.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Related Resources
+
+The following resources are often used in the same context:
+
+* [databricks_workspace_file](./workspace_file.md)
+* [End to end workspace management](../guides/workspace-management.md) guide.
+* [databricks_dbfs_file](../data-sources/dbfs_file.md) data to get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_dbfs_file_paths](../data-sources/dbfs_file_paths.md) data to get list of file names from get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_global_init_script](global_init_script.md) to manage [global init scripts](https://docs.databricks.com/clusters/init-scripts.html#global-init-scripts), which are run on all [databricks_cluster](cluster.md#init_scripts) and [databricks_job](job.md#new_cluster).
```
alexott, line 105: This is also not really relevant

storage/resource_databricks_files.go
alexott, line 1: Let rename source file to match the resource name

storage/resource_databricks_files.go
```diff
@@ -0,0 +1,127 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/databricks/databricks-sdk-go/service/files"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/databricks/terraform-provider-databricks/workspace"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func getContentReader(data *schema.ResourceData) (io.ReadCloser, error) {
```
alexott, line 20: Maybe it makes sense to unify it with https://github.com/databricks/terraform-provider-databricks/blob/master/workspace/file_resource.go#L36? That code is also calculating the MD5 of the source file...
Although we can do it at the later point of time

workspace/file_resource.go
```diff
@@ -146,6 +146,15 @@ func FileContentSchema(extra map[string]*schema.Schema) map[string]*schema.Schem
...
 				}
+				if strings.HasPrefix(v, "Volume") {
+					return diag.Diagnostics{
+						{
+							Summary:       "Path should start with /Volumes",
+							Severity:      diag.Error,
+							AttributePath: p,
+						},
+					}
+				}
```
alexott, line 157: Hmmm, do we really have that invalid usage?

alexott (approved):
overall looks good, just some comments

workspace/file_resource.go
tanmay-db: Yes, if path start with Volume instead of /Volumes the it throws the error.

storage/resource_databricks_files.go
tanmay-db: Yeah, it would need some refactoring, I will try to do it when migrating other resources which will also require some refactoring so will do them simultaneously.

PR #2506, opened 2023-07-24
SQL alert values can also be booleans

fjakobs:
## Changes
SQL alert values can also be booleans

follow up to https://github.com/databricks/terraform-provider-databricks/pull/2494


mgyucht (approved):
LGTM, let's add one more test case for this path

PR #2509, opened 2023-07-24
Add terraform support for Job Parameters (Private Preview)

geekswaroop:
## Changes

Job Parameters is in Private Preview. Documentation Link can be found [here](https://openapi.dev.databricks.com/pr-350680/api/workspace/jobs/create)

- [x] `make test` run locally
- [x] using Go SDK


jobs/resource_job.go
```diff
@@ -268,16 +268,17 @@ type JobSettings struct {
...
+	Continuous           *ContinuousConf         `json:"continuous,omitempty"`
+	Trigger              *Trigger                `json:"trigger,omitempty"`
+	MaxConcurrentRuns    int32                   `json:"max_concurrent_runs,omitempty"`
+	EmailNotifications   *EmailNotifications     `json:"email_notifications,omitempty" tf:"suppress_diff"`
+	WebhookNotifications *WebhookNotifications   `json:"webhook_notifications,omitempty" tf:"suppress_diff"`
+	NotificationSettings *NotificationSettings   `json:"notification_settings,omitempty"`
+	Tags                 map[string]string       `json:"tags,omitempty"`
+	Queue                *Queue                  `json:"queue,omitempty"`
+	RunAs                *JobRunAs               `json:"run_as,omitempty"`
+	Parameter            *JobParameterDefinition `json:"parameters,omitempty"`
```
gaborratky-db, line 281: The field name needs to stay plural and its type needs to be an array:
```suggestion
	Parameters           []JobParameterDefinition `json:"parameters,omitempty"`
```

jobs/resource_job.go
```diff
@@ -342,7 +356,8 @@ type JobRun struct {
 	Trigger     string   `json:"trigger,omitempty"`
 	RuntType    string   `json:"run_type,omitempty"`
 
-	OverridingParameters RunParameters `json:"overriding_parameters,omitempty"`
+	OverridingParameters RunParameters  `json:"overriding_parameters,omitempty"`
+	JobParameter         []JobParameter `json:"job_parameters,omitempty"`
```
gaborratky-db, line 360: The field name needs to stay plural.
```suggestion
	JobParameters        []JobParameter `json:"job_parameters,omitempty"`
```

gaborratky-db (changes requested):

jobs/resource_job.go
```diff
@@ -268,16 +268,17 @@ type JobSettings struct {
...
+	Continuous           *ContinuousConf          `json:"continuous,omitempty"`
+	Trigger              *Trigger                 `json:"trigger,omitempty"`
+	MaxConcurrentRuns    int32                    `json:"max_concurrent_runs,omitempty"`
+	EmailNotifications   *EmailNotifications      `json:"email_notifications,omitempty" tf:"suppress_diff"`
+	WebhookNotifications *WebhookNotifications    `json:"webhook_notifications,omitempty" tf:"suppress_diff"`
+	NotificationSettings *NotificationSettings    `json:"notification_settings,omitempty"`
+	Tags                 map[string]string        `json:"tags,omitempty"`
+	Queue                *Queue                   `json:"queue,omitempty"`
+	RunAs                *JobRunAs                `json:"run_as,omitempty"`
+	Parameters           []JobParameterDefinition `json:"parameters,omitempty"`
```
gaborratky-db, line 281: This field, similar to `Tasks` needs an additional annotation: `f:"alias:job_parameter"`.

We want users to be able to define one or more job parameters with the `job_parameter` block. The documentation should also reflect that.

docs/resources/job.md
```diff
@@ -207,6 +207,13 @@ This block controls notification settings for both email & webhook notifications
 * `no_alert_for_skipped_runs` - (Optional) (Bool) don't send alert for skipped runs.
 * `no_alert_for_canceled_runs` - (Optional) (Bool) don't send alert for cancelled runs.
 
+###  job_parameters Configuration Block
```
gaborratky-db, line 210: We use singular names for the blocks that represent one or more entities (e.g., the `task` block).

```suggestion
###  job_parameter Configuration Block
```

docs/resources/job.md
```diff
@@ -207,6 +207,13 @@ This block controls notification settings for both email & webhook notifications
...
+
+This block controls job-level parameters for a given job:
```
gaborratky-db, line 212:
```suggestion
This block defines a job-level parameter for the job. You can define several job-level parameters for the job. Supported options are:
```

gaborratky-db (changes requested):
I just noticed that the block is not consistently named ("job_parameters" vs. "job_parameter"). We should also call out in the documentation that several `job_parameter` can be used to defined parameters of the job.

jobs/resource_job.go
```diff
@@ -350,7 +364,8 @@ type JobRun struct {
 	Trigger     string   `json:"trigger,omitempty"`
 	RuntType    string   `json:"run_type,omitempty"`
 
-	OverridingParameters RunParameters `json:"overriding_parameters,omitempty"`
+	OverridingParameters RunParameters  `json:"overriding_parameters,omitempty"`
+	JobParameters        []JobParameter `json:"job_parameters,omitempty"`
```
gaborratky-db, line 368: FTR, we defined `JobParameters`, seeing that `OverridingParameters` was also defined but see no usage anywhere in the provider. There is no Terraform data source to retrieve job runs and the only place where runs are triggered/polled is when the job resource controls the job's state through `always_running` or `control_run_state`.

jobs/resource_job.go
```diff
@@ -286,6 +286,7 @@ type JobSettings struct {
 	Queue                *Queue                        `json:"queue,omitempty"`
 	RunAs                *JobRunAs                     `json:"run_as,omitempty"`
 	Health               *JobHealth                    `json:"health,omitempty"`
+	Parameters           []JobParameterDefinition      `json:"parameters,omitempty" tf:"alias:parameter"`
```
alexott, line 289: we may need to stop using aliases as they may prevent from porting to Go SDK

@mgyucht WDYT?

mgyucht: We should still be able to use the Go SDK as any tags/convenience methods here can be implemented in the closure passed to `common.StructToSchema`. E.g. `alias(old, new string)` would do `schema[new] = schema[old]; delete(old, schema)`.

mgyucht: For the in-between period, we can define helper methods. Long-term, these attributes will need to be included in the OpenAPI spec itself.

jobs/resource_job.go
mgyucht: IIUC this keeps this struct in sync with OpenAPI.

mgyucht (approved):

PR #2510, opened 2023-07-25
Use prefix of the error message to check existence of `databricks_user` with `force = true`

### LINKED ISSUE #2500: [ISSUE] Issue with `databricks_user`  resource on account level
Creating a `databricks_user` resource at the account level with `force = true` but still getting the user already exists in the account error that this setting is meant to avoid according to docs: https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/user#argument-reference

Error: cannot create user: User with email alice@bob.com already exists in this account. Note: emails are case insensitive (i.e abc@domain.com and ABC@DOMAIN.COM are considered equivalent)
This had initially worked as expected back in March (using version 1.10.0), but is failing now in version 1.20.0 - is there any new change in provider / api that could break this? cc: 
[@Mihai-Bogdan Zanfir](https://databricks.slack.com/team/U02CUQE06F6)
 
[@Shuyu Cao](https://databricks.slack.com/team/U05FZLFG6AU)

### Plan
```
  # databricks_user.shuyu_cao will be created
  + resource "databricks_user" "alice" {
      + active                     = true
      + allow_cluster_create       = false
      + allow_instance_pool_create = false
      + databricks_sql_access      = false
      + disable_as_user_deletion   = (known after apply)
      + display_name               = (known after apply)
      + force                      = true
      + home                       = (known after apply)
      + id                         = (known after apply)
      + repos                      = (known after apply)
      + user_name                  = "alice@databricks.com"
      + workspace_access           = false
    }
```

### Expected Behavior
No error if user already exists

### Actual Behavior
This error:

```
Error: cannot create user: User with email alice@bob.com already exists in this account. Note: emails are case insensitive (i.e abc@domain.com and ABC@DOMAIN.COM are considered equivalent)
```

### Important Factoids
This code fails when creating a user on the **account** level.

alexott:
## Changes

This fixes #2500

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/resource_user.go
+++ b/scim/resource_user.go
@@ -167,7 +167,8 @@ func createForceOverridesManuallyAddedUser(err error, d *schema.ResourceData, us
 	}
 	// corner-case for overriding manually provisioned users
 	userName := strings.ReplaceAll(u.UserName, "'", "")
-	if (err.Error() != userExistsErrorMessage(userName, false)) && (err.Error() != userExistsErrorMessage(userName, true)) {
+	if (!strings.HasPrefix(err.Error(), userExistsErrorMessage(userName, false))) &&
+		(!strings.HasPrefix(err.Error(), userExistsErrorMessage(userName, true))) {
 		return err
 	}
 	userList, err := usersAPI.Filter(fmt.Sprintf("userName eq '%s'", userName), true)


nkvuong (approved):
lgtm

mgyucht (approved):

tsutariya-98:
I used `force = true` in v1.20.0, and instead of giving me an error during `terraform validate` saying it's not a valid or appropriate attribute, it only threw an error during `terraform apply` saying "user already exists". I think it should instead tell me that this attribute is not available in this version and that I should upgrade to a higher version, or something like that.

PR #2528, opened 2023-07-28
Add `databricks_connection` resource to support Lakehouse Federation

### LINKED ISSUE #2575: [FEATURE] Add Lakehouse Federation Resources
### Use-cases
Create the resources described [here](https://docs.databricks.com/en/query-federation/index.html#) using the terraform provider.

### Attempted Solutions

These resources do not seem to exist yet, or it is not documented.

### Proposal

Add the resources for Lakehouse Query Federation, or implement them into existing resources and document the solution.

### References
N/A

nkvuong:
## Changes
Add support for Lakehouse Federation
- Add a new resource `databricks_connection` that represent Unity Catalog connections
- Extend `databricks_catalog` to support foreign catalogs

Closes #2575

## Tests

- [x] `make test` run locally
- [x] using Go SDK


catalog/resource_connection.go
```diff
@@ -0,0 +1,100 @@
...
+	PropertiesKvpairs map[string]string `json:"properties_kvpairs,omitempty" tf:"alias:properties,force_new"`
+	// If the connection is read only.
+	ReadOnly bool `json:"read_only,omitempty" tf:"force_new"`
+}
+
+func ResourceConnection() *schema.Resource {
+	s := common.StructToSchema(ConnectionInfo{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			return m
+		})
```
alexott, line 38: I think that we have multiple places like this, so it makes sense to move this to the `common` package

docs/resources/connection.md
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `name` - Name of the Connection
+- `connection_type` - Connection type. `MYSQL` `POSTGRESQL` `SNOWFLAKE` `REDSHIFT` `SQLDW` `SQLSERVER` or `DATABRICKS`.
```
alexott, line 37: It makes sense to point to docs or API for the latest list of the supported connection types

alexott (approved):
looks good, minor comments that we can address later

catalog/resource_connection_test.go
```diff
@@ -0,0 +1,315 @@
...
+				Method:   http.MethodPost,
+				Resource: "/api/2.1/unity-catalog/connections",
+				ExpectedRequest: catalog.CreateConnection{
+					Name:           "testConnectionName",
+					ConnectionType: catalog.ConnectionType("testConnectionType"),
+					Comment:        "This is a test comment.",
+					OptionsKvpairs: map[string]string{
+						"host": "test.com",
+					},
+					PropertiesKvpairs: map[string]string{
```
andrewli81, line 30: This needs to be a list of options:
```
"options":[
  {"key":"host","value":"mysqk.fakedb.com"},
  {"key":"port","value":"1234"},
  {"key":"user","value":"user123"},
  {"key":"password","value":"password123"}
]
```

andrewli81: Right now it is not a list:
"options_kvpairs": {
   "host": "test.com"
}

andrewli81 (approved):
LGTM, one comment on why the integration test is failing. 

This is an example good json input:
```
curl -vvv -H "Authorization: Bearer xxx" -H "Content-Type: application/json" -X POST https://dbc-01eafff1-4ea3.dev.databricks.com/api/2.0/unity-catalog/connections -d '{"name":"test_mysql","connection_type":"MYSQL","options":[{"key":"host","value":"mysqk.fakedb.com"},{"key":"port","value":"1234"},{"key":"user","value":"user123"},{"key":"password","value":"password123"}]}'
```

catalog/resource_connection_test.go
alexott: Is it really expected that it will be repeated keys? Other APIs are accepting objects as well, in addition to key/value pairs...

andrewli81: I believe the api docs are wrong, this is something we need to correct. There's no difference between the connection api code and other apis.

andrewli81: cc @Doug-Luce - I am not sure if the open api spec actually works - the spec mentions the options in a KV pair way:
```
"options_kvpairs": {
   "host": "test.com",
   "port": "1234",
   "user": "user123",
   "password": "password123"
}
```
but I always know this works:
```
"options":[
  {"key":"host","value":"mysqk.fakedb.com"},
  {"key":"port","value":"1234"},
  {"key":"user","value":"user123"},
  {"key":"password","value":"password123"}
]
```
It could be possible that we have an implementation deficiency of the connection API that caused us to not be able to take the kv style of input, but I think most likely it's a bug in our API doc.

catalog/resource_connection.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// This structure contains the fields of catalog.UpdateConnection and catalog.CreateConnection
```
mgyucht, line 11: We need better utilities for converting Go SDK structs to TF schemas so we don't need to inline stuff like this.

catalog/resource_connection.go
```diff
@@ -0,0 +1,100 @@
...
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var createConnectionRequest catalog.CreateConnection
+			var alias ConnectionInfo
+			common.DataToStructPointer(d, s, &createConnectionRequest)
+			common.DataToStructPointer(d, s, &alias)
+			//workaround as cannot set tf:"alias" for the Go SDK struct
```
mgyucht, line 50: micronit: Can you move this comment up one line?

catalog/resource_connection.go
```diff
@@ -0,0 +1,100 @@
...
+			createConnectionRequest.OptionsKvpairs = alias.OptionsKvpairs
+			createConnectionRequest.PropertiesKvpairs = alias.PropertiesKvpairs
+			conn, err := w.Connections.Create(ctx, createConnectionRequest)
+			if err != nil {
+				return err
+			}
+			d.SetId(conn.Name)
```
mgyucht, line 57: Are connection names globally unique?

catalog/resource_connection_test.go
mgyucht: Please fix the OpenAPI spec, as we (and customers!) depend heavily on its accuracy to be able to use your team's API.

BTW: do we really want this to be a repeated list of KV pairs? We should be critical about these interfaces and think from the customer's perspective, aiming to do what is natural and intuitive where possible, rather than force all customers to adapt to an unusual pattern because there is some small technical issue. In the worst case, you should be able to define your proto using the [Any type](https://protobuf.dev/programming-guides/proto3/#any) and deserialize yourself (scalapb docs here: https://scalapb.github.io/api/com/google/protobuf/any/Any.html).

go.sum
```diff
@@ -51,6 +51,8 @@ github.com/databricks/databricks-sdk-go v0.12.0 h1:VgMJpvEiyRRrJ0mQx22Rkc73zjxUe
 github.com/databricks/databricks-sdk-go v0.12.0/go.mod h1:h/oWnnfWcJQAotAhZS/GMnlcaE/8WhuZ5Vj7el/6Gn8=
 github.com/databricks/databricks-sdk-go v0.13.0 h1:Npi4laUUmcOPDPdJf2ZMGFUtybpf4LK6n5NQY56Ya2Q=
 github.com/databricks/databricks-sdk-go v0.13.0/go.mod h1:0iuEtPIoD6oqw7OuFbPskhlEryt2FPH+Ies1UYiiDy8=
+github.com/databricks/databricks-sdk-go v0.14.1 h1:s9x18c2i6XbJxem6zKdTrrwEUXQX/Nzn0iVM+qGlRus=
+github.com/databricks/databricks-sdk-go v0.14.1/go.mod h1:0iuEtPIoD6oqw7OuFbPskhlEryt2FPH+Ies1UYiiDy8=
```
mgyucht, line 55: Remove this

mgyucht (commented):
First batch of comments. Thanks for contributing this!

catalog/resource_connection.go
nkvuong: connection names must be unique within a metastore

catalog/resource_connection_test.go
mgyucht: I queried the REST API and saw the following response:
```
<     {
<       "connection_id": "<REDACTED>",
<       "connection_type": "POSTGRESQL",
<       "created_at": 1681945142627,
<       "created_by": "<REDACTED>",
<       "credential_type": "USERNAME_PASSWORD",
<       "full_name": "<REDACTED>",
<       "metastore_id": "<REDACTED>",
<       "name": "<REDACTED>",
<       "options": {
<         "host": "<REDACTED>,
<         "port": "5432"
<       },
<       "owner": "<REDACTED>",
<       "provisioning_info": {
<         "state": "ACTIVE"
<       },
<       "read_only": true,
<       "securable_kind": "CONNECTION_POSTGRESQL",
<       "securable_type": "CONNECTION",
<       "updated_at": 1681945142627,
<       "updated_by": "<REDACTED>",
<       "url": "jdbc://<REDACTED>:5432/"
<     }
```
Options is named `options` and is a dictionary of strings. This response had no properties, but that field is modeled identically internally.

I'll file a ticket to correct the OpenAPI specification. In the meantime, it seems like we should model this as a `map[string]string`.

Doug-Luce: I'm updating the OpenAPI spec to fix the naming of options as well as to reflect that properties_kvpairs is not a part of the response. I'll also add a couple of missing fields.

For creating the connection, options are specified as `map[string][string]`.
This is valid:
```
"options_kvpairs": {
   "host": "test.com",
   "port": "1234",
   "user": "user123",
   "password": "password123"
}
```

mgyucht: I think it should still be named options, as that is the name of the nested field inside after inlining?

Doug-Luce: Yes, it would be options and not options_kvpairs. Copy-pasted from Andrew's post, but this is the body I used to create the connection.
```
{
    "name": "openapi_connection_spec_test_2",
    "connection_type": "MYSQL",
    "options": {
        "host": "test.com",
        "port": "5432",
        "user": "dougluce",
        "password": "redacted"
    },
    "properties": {
        "some_prop": "test1",
        "some_prop2": "test2"
    }
}
```

mgyucht (commented):
Couple questions, otherwise seems mostly alright. Main question is what is the form of the ID of a connection in TF state. If it is not globally unique, we may want to combine it with the metastore's ID to get a globally unique ID. Thoughts?

catalog/resource_connection_test.go
```diff
@@ -0,0 +1,315 @@
...
+						"host": "test.com",
+					},
+					Properties: map[string]string{
+						"purpose": "testing",
+					},
+				},
+			},
+			{
+				Method:   http.MethodGet,
+				Resource: "/api/2.1/unity-catalog/connections/testConnectionName?",
```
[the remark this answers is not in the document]
nkvuong: our SDK generates GET method with a trailing `?` for parameters

catalog/resource_connection.go
andrewli81: Yeah if you want global uniqueness we should add metastore id into the scope.

catalog/resource_connection.go
```diff
@@ -0,0 +1,112 @@
...
+	// connection.
+	Properties map[string]string `json:"properties,omitempty" tf:"force_new"`
+	// If the connection is read only.
+	ReadOnly bool `json:"read_only,omitempty" tf:"force_new,computed"`
+}
+
+// suppress diff for sensitive options, which are not returned by the server
+func suppressSensitiveOptions(k, old, new string, d *schema.ResourceData) bool {
+	//this list will expand as other auth may have different sensitive options
+	sensitiveOptions := []string{"user", "password"}
```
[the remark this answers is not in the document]
mgyucht: Yes, but the question is how to determine whether user/password options are different locally than what is configured (imagine someone changed user/password manually in Databricks or via a separate process). How can we compute whether or not the resource needs to be updated? Typically, TF reads resources to compare them to its state & the configuration supplied by the user to compute a diff. If `user` and `password` are not returned, it may not be possible to know whether the connection needs to be updated.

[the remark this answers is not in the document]
andrewli81: Is it possible to always update the username and password? There are basically 2 worlds:
1. Silently fail to update when user changes username/password
2. Always update username / password.

nkvuong:
> What happens if a user defines two connections with the same name in two different metastores? Should the name be combined with the metastore ID to make a universally unique identifier?

@mgyucht this is actually a good point, and we may need to retroactively do this for all UC resources 😨

catalog/resource_catalog.go
```diff
@@ -27,15 +27,16 @@ func ucDirectoryPathSlashAndEmptySuppressDiff(k, old, new string, d *schema.Reso
...
-	Owner         string            `json:"owner,omitempty" tf:"computed"`
-	IsolationMode string            `json:"isolation_mode,omitempty" tf:"computed"`
-	MetastoreID   string            `json:"metastore_id,omitempty" tf:"computed"`
+	Name           string            `json:"name"`
+	Comment        string            `json:"comment,omitempty"`
+	StorageRoot    string            `json:"storage_root,omitempty" tf:"force_new"`
+	ProviderName   string            `json:"provider_name,omitempty" tf:"force_new,conflicts:storage_root"`
+	ShareName      string            `json:"share_name,omitempty" tf:"force_new,conflicts:storage_root"`
+	ConnectionName string            `json:"connection_name,omitempty" tf:"force_new,conflicts:storage_root"`
+	Properties     map[string]string `json:"properties,omitempty"`
```
andrewli81, line 36: CatalogInfo also has options - our open api spec needs updating.

catalog/resource_connection.go
andrewli81, line 38: The source of truth for the list of secrets can be retrieved by the type manifest API:
https://temp.corp.databricks.com/?114480149a090e3e#QR0tqKyrycfLQ0OH3ZEqsyZOuWEwEtNMb+o7S4bdpXs=

Can we also add `personalAccessToken`, `access_token`, `client_secret`, and `OAuthPvtKey`

catalog/resource_catalog.go
andrewli81: cc @Doug-Luce

andrewli81 (commented):
Three comments before merging:
1. List of secret options will need to expand a bit - you can always get the list of sensitive options from the type-manifest API, I posted an example: https://temp.corp.databricks.com/?114480149a090e3e#QR0tqKyrycfLQ0OH3ZEqsyZOuWEwEtNMb+o7S4bdpXs=

I will also point future folks who add new connectors to update this list in TF.

2. Can we make the "secret" options always update if specified by user? Silently failing to update might be surprising from user perspective.

3. CreateCatalog will need options too other than connection name for foreign catalogs to work, our API spec is a bit outdated, we will update that as well.

catalog/resource_connection.go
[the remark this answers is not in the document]
mgyucht: I am inclined not to accept either option. The former would lead to an exceptionally poor user experience, where they would need to do manual changes to rotate a credential. The latter is also not great, because it will indicate that there is always a diff even when there is no change needed to the resource, and in general will block users from using this resource if they have checks to ensure that there are no diffs between configured and deployed state.

My preference would be to override the CustomizeDiff function to: 1) diff all other fields, and then 2) check whether options that are considered secrets are different between config and state via `(r *schema.ResourceDiff) GetChange(k string)` method.

We do need to figure out a consistent story for dealing with secrets. In workspaces, we have secret management. Have we thought about how this would work for UC at the account-level?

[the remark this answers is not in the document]
andrewli81: @mgyucht 

>My preference would be to override the CustomizeDiff function to: 1) diff all other fields, and then 2) check whether options that are considered secrets are different between config and state via (r *schema.ResourceDiff) GetChange(k string) method.

Sorry I am a bit ignorant - how does `(r *schema.ResourceDiff) GetChange(k string)` method decide there's a diff? Via some hash object?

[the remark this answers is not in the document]
mgyucht: Terraform maintains a state file with all of the attributes from each resource it creates. The Terraform lifecycle is divided into two parts roughly speaking:
1. Figure out what resources need to be created, updated and/or deleted based on the differences between the configuration, the TF state file, and the state of the remote resource;
2. Apply those changes.
`GetChange()` by my reading returns, for a given field in the TF schema for the resource, the "old" value from the state and the "new" value from the config + remote state. I have to think through how this works a bit carefully, but we might be able to do something like: check if the config specifically has differences from the state for secret option keys, and for other keys, just check if there is a difference between the config and the remote state.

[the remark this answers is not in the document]
andrewli81: I see, so terraform remembers the state of last time that customer creates, makes sense.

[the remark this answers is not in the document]
andrewli81: Yeah then I would say maybe let's just go with the simple approach and just always refresh.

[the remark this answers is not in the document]
mgyucht: Given that we have the choice between forcing users to make manual changes if secret fields are updated vs updating them every time, I agree that we should lean towards always refreshing. In light of moving forward with this resource and unblocking customers, let's do this. @nkvuong what do you think, would this be a viable option from the Terraform perspective? IIUC to do this, we treat the secret options in the same way as non-secret options, and they will be included in the diff every time. 

@andrewli81 I really recommend that your team revisits the secret management story for UC connections. We don't need each team to reinvent the wheel when it comes to secret management, and when this happens, it increases complexity for consumers and results in worse user experiences. For inspiration, you can look at how [secrets are referenced in Spark configurations](https://docs.databricks.com/en/security/secrets/secrets.html#use-a-secret-in-a-spark-configuration-property-or-environment-variable). I understand that there is a workspace-level/account-level divide that will need to be addressed. There may be other use-cases for account-level secrets in the future as well (or even today: consider account-level service principal passwords).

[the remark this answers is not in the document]
nkvuong: > we treat the secret options in the same way as non-secret options, and they will be included in the diff every time.

this will cause permanent diff in the Terraform apply - we will need to add to doc that users should add `ignore_changes` to their resource definition to avoid that.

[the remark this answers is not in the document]
alexott: we have similar problem with MLflow webhooks, and on read we simply restore data from the state: https://github.com/databricks/terraform-provider-databricks/blob/master/mlflow/resource_mlflow_webhook.go#L79 - we can try to do the same here

nkvuong: done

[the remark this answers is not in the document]
andrewli81: >I understand that there is a workspace-level/account-level divide that will need to be addressed. There may be other use-cases for account-level secrets in the future as well (or even today: consider account-level service principal passwords).

We have to be unblocked on shipping features that need to store secrets in UC account level objects in lieu of a team that is willing to fund an account level secret manager, I will probably write a vision doc for this because I think we will need it for other use cases as well.

catalog/resource_catalog.go
andrewli81: @nkvuong Can you address this one as well? Thanks!

nkvuong: @andrewli81 how would the options be split between connections & catalogs? and would there be sensitive options to be handled here as well? 
once this is fixed in OpenAPI, we just need to regenerate the SDK, so should not take long

nkvuong:
@mgyucht all comments should be now addressed, just need to wait for the OpenAPI spec for Catalog to be updated, i.e. adding `options` parameter to `databricks_catalog`

mgyucht (commented):
One small question, but the implementation looks good to me.

catalog/resource_catalog.go
andrewli81: @nkvuong 

> @andrewli81 how would the options be split between connections & catalogs? and would there be sensitive options to be handled here as well?

The options of catalogs are mainly used to identify databases. This is because database systems like postgresql uses 3 layer namespaces (same as UC), so customer will need to specify which database to import into a UC catalog. 

https://docs.databricks.com/en/query-federation/postgresql.html#language-SQL:~:text=Run%20the%20following%20SQL%20command%20in%20a%20notebook%20or%20Databricks%20SQL%20editor.%20Items%20in%20brackets%20are%20optional.%20Replace%20the%20placeholder%20values%3A

The open API spec has been fixed: https://docs.databricks.com/api/workspace/connections/create

Can we also fix it here?

PR #2547, opened 2023-08-02
Update metastore.md

charlottevdscheun:
## Changes
The wrong argument was put in the example of data source metastore

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


charlottevdscheun:
> thank you! Maybe we really need to support `id` as well

Should make sense, most of the data sources have the same attributes. `id` would be a nice generic one, same as `name`

PR #2549, opened 2023-08-02
Fixed `databricks_job` update from instance pool to regular node type

marekbrysa:
## Changes
When a job cluster was updated from instance pool to regular, the computed `driver_instance_pool_id` was still being sent, which then caused failure of the cluster to start with error in the UI "Worker must use a pool if driver uses a pool".

This resets the driver_instance_pool_id if instance_pool_id is not set.

Also I noticed there was a function FixInstancePoolChangeIfAny that was only used on `databricks_cluster`. Looking at https://github.com/databricks/terraform-provider-databricks/pull/850, I believe it's also relevant for jobs clusters, so applying it too.

## Tests
Tested updating a jobs cluster from regular to pool and back.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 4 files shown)


mgyucht (approved):
Nice! Any chance you can do a manual test of this to verify?

marekbrysa:
@mgyucht yeah, I already did, it seems to work: "Tested updating a jobs cluster from regular to pool and back."

PR #2555, opened 2023-08-03
Added `acl_principal_id` to data sources: `databricks_user`, `databricks_service_principal`, `databricks_group`, `databricks_current_user`

alexott:
## Changes

This is to have parity between resources & data sources

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/data_current_user.go
+++ b/scim/data_current_user.go
@@ -41,6 +41,10 @@ func DataSourceCurrentUser() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"acl_principal_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 		ReadContext: func(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
 			c := m.(*common.DatabricksClient)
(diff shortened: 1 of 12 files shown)


scim/data_current_user.go
```diff
@@ -55,6 +59,11 @@ func DataSourceCurrentUser() *schema.Resource {
 			d.Set("user_name", me.UserName)
 			d.Set("home", fmt.Sprintf("/Users/%s", me.UserName))
 			d.Set("repos", fmt.Sprintf("/Repos/%s", me.UserName))
+			if strings.Contains(me.UserName, "@") {
```
mgyucht, line 62: @gauthamsunjay FYI: this is how we tell if the current principal is a user or service principal. It would be nice to expose this in the SCIM /Me response if it isn't already there.

mgyucht (approved):
LGTM!

scim/data_current_user.go
gauthamsunjay: @baishen-db can you take a look here? Is SCIM/Me exposed for users and service principals?

alexott: It's exposed to both. The main question is how to distinguish between them...

gauthamsunjay: I just spoke to @baishen-db and @michaelduan-db offline. `SCIM/Me` doesn't currently distinguish between user and a service principal.

PR #2563, opened 2023-08-07
Exporter: Incremental export of notebooks, SQL objects and some other resources

alexott:
## Changes

This draft PR is providing implementation of incremental export of notebooks, SQL objects (dashboards, alerts, queries) and some other resources.

Example of usage:

```sh
./terraform-provider-databricks exporter -debug -directory output -skip-interactive \ 
  -listing sql-queries,sql-dashboards,sql-alerts 
  -incremental -updated-since '2023-07-24T00:00:00Z'
```

### Current limitations

- when modified resource has dependencies they are exported as well. No check (yet) if it's modified or not (we don't store last modified date for dependent objects yet). Also, all resources that don't have last modified date will be exported.
- there is no way for detecting deletion of objects - workaround would be to do a periodic full export.

### Implementation

Two new options were added - `-incremental` to trigger incremental behavior, and `-updated-since` to specify a timestamp (see TODOs).

When incremental mode is set, then list operations of objects having last modified timestamp will do the check and emit only objects that are modified at or after a given timestamp. 

When all resources are exported, the new set of resources is merged with following objects:

- existing TF resources are loaded from the corresponding source files, and that resources that aren't a part of new export are added to corresponding `hclwrite.File` object (see `context.updateExportedWithIncrementals` function)
- existing variables are also merged with new variables (see `context.generateVariables` function)
- existing `import.sh` is analyzed, and its content is merged with new resources.

### TODOs

- [x] Add tests
- [x] Make sure that all resources that have "last modified date" are exported (global init scripts, IP Access Lists, etc.)
- [x] store last run timestamp in some configuration file inside the output diretory and load from it if `-updated-since` isn't provided
- (to discuss) should we store last modified date in some configuration file that will be loaded?  Like, to avoid re-downloading of notebooks if DLT pipeline is modified and notebook isn't...

## Tests

No tests yet, will be added as we proceed with code review

- [x] `make test` run locally
- [ ] ~using Go SDK~

### DIFF
--- a/exporter/exporter_test.go
+++ b/exporter/exporter_test.go
@@ -2128,6 +2128,7 @@ func TestIncrementalDLTAndMLflowWebhooks(t *testing.T) {
 				Response: ml.ListRegistryWebhooks{
 					Webhooks: webhooks,
 				},
+				ReuseRequest: true,
 			},
 			{
 				Method:   "GET",
@@ -2135,6 +2136,7 @@ func TestIncrementalDLTAndMLflowWebhooks(t *testing.T) {
 				Response: ml.ListRegistryWebhooks{
 					Webhooks: webhooks,
 				},
+				ReuseRequest: true,
 			},
 			{
 				Method:   "GET",


workspace/resource_notebook.go
```diff
@@ -41,15 +41,28 @@ var extMap = map[string]notebookLanguageFormat{
...
+	ObjectType            string                 `json:"object_type,omitempty" tf:"computed"`
+	Path                  string                 `json:"path"`
+	Language              string                 `json:"language,omitempty"`
+	CreatedAt             int64                  `json:"created_at,omitempty"`
+	ModifiedAt            int64                  `json:"modified_at,omitempty"`
+	ModifiedAtInteractive *ModifiedAtInteractive `json:"modified_at_interactive,omitempty"`
+	Size                  int64                  `json:"size,omitempty"`
+}
+
+func (obs *ObjectStatus) GetModifiedAt() int64 {
```
nfx, line 61: don't add it to a model, as it'll be replaced by SDK. move to exporter

exporter/context.go
```diff
@@ -143,13 +152,52 @@ func newImportContext(c *common.DatabricksClient) *importContext {
 		allDirectories:      []workspace.ObjectStatus{},
 		allWorkspaceObjects: []workspace.ObjectStatus{},
 		workspaceConfKeys:   workspaceConfKeys,
+		shImports:           make(map[string]bool),
 	}
 }
 
+func getLastRunString(fileName string) string {
```
nfx, line 159: can you also parse durations, e.g. `1w`, `1M`, `1d`, `1h` - the equivalents of `time.Duration.String`.

nfx (dismissed):

workspace/resource_notebook.go
alexott: Ok. will do.

exporter/context.go
alexott: Maybe, let see if it will be required - I want to get feedback from customer on UX. But really, the duration here doesn't matter much, maybe except the case when it's scheduled from cron, but then it should be provided via `updated-since` parameter, not from file

exporter/context.go
```diff
@@ -205,7 +253,24 @@ func (ic *importContext) Run() error {
...
+		if err == nil {
+			fileScanner := bufio.NewScanner(shFile)
+			fileScanner.Split(bufio.ScanLines)
+			for fileScanner.Scan() {
+				line := fileScanner.Text()
+				if strings.HasPrefix(line, "terraform import ") {
+					ic.shImports[strings.TrimRight(line, "\n")] = true
+				}
+			}
+			shFile.Close()
```
pietern, line 268: Not closed in the err path (may be problematic on Windows).

```go
defer shFile.close()
```

pietern (commented):
I recommend making incremental the default mode and doing a full export with `updatedSinceMs == 0`. That would save quite a bit of branching and debugging easier.

exporter/context.go
alexott: but `shFile` will be non-nil only if `err == nil`

alexott:
@mgyucht, can you please start a new merge? For some reason, one of the tests began to fail. It's now fixed

PR #2567, opened 2023-08-08
Refresh `databricks_grants` with latest permissible grants

nkvuong:
## Changes
Refresh `databricks_grants` with latest permissible grants

## Tests

- [x] `make test` run locally
- [ ] ~covered with integration tests in `internal/acceptance`~
- [ ] ~using Go SDK~

### DIFF
--- a/catalog/resource_grants.go
+++ b/catalog/resource_grants.go
@@ -102,6 +102,9 @@ func getPermissionEndpoint(securable, name string) string {
 	if securable == "share" {
 		return fmt.Sprintf("/unity-catalog/shares/%s/permissions", name)
 	}
+	if securable == "the_connection" {
+		return fmt.Sprintf("/unity-catalog/permissions/foreign_connection/%s", name)
+	}
 	return fmt.Sprintf("/unity-catalog/permissions/%s/%s", securable, name)
 }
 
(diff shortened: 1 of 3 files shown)


catalog/resource_grants.go
```diff
@@ -102,6 +102,9 @@ func getPermissionEndpoint(securable, name string) string {
 	if securable == "share" {
 		return fmt.Sprintf("/unity-catalog/shares/%s/permissions", name)
 	}
+	if securable == "the_connection" {
```
alexott, line 105: is `the_connection` correct name for securable?

nkvuong: typo 😶‍🌫️

nkvuong: I've added a test to confirm

mgyucht (commented):
BROWSE, CREATE_MODEL, CREATE_FOREIGN_SCHEMA, and CREATE_FOREIGN_TABLE are not checked into OpenAPI. Where did these come from?

nkvuong:
@mgyucht some of those permissions are in private preview, so they are in the UC service but not in OpenAPI - we need to discuss how to handle this better going forward, e.g. leave it to UC service to do the validation but that means tf plan won't show errors.

mgyucht:
Understood. Sounds like we'll need a way to integrate it into the OpenAPI spec somehow.

mgyucht (approved):
LGTM!

PR #2586, opened 2023-08-15
Add support for Databricks Account Console IP Access Lists

### LINKED ISSUE #2530: [FEATURE] Add support for Databricks Account Console IP Access Lists
### Use-cases

Customer is looking to add Databricks Account Console IP Access Lists via Terraform during account level configuration, but cannot as the existing provider only support IP Access Lists on Workspaces.

### Attempted Solutions

Currently the only solution is to configure this manually on the UI or via API calls.

### Proposal

Add this support to the DB TF provider.

### References

https://docs.databricks.com/security/network/ip-access-list-account.html#ip-access-lists-for-the-account-console
https://docs.databricks.com/api/account/accountipaccesslists

nkvuong:
## Changes
Extend `databricks_ip_access_list` to support Account Console.

Closes #2530

## Tests

- [x] `make test` run locally
- [x] using Go SDK


nkvuong:
there's an issue with API sometimes returning 409 Conflict, which needs to be retried

access/resource_ip_access_list.go
```diff
@@ -31,47 +32,77 @@ func ResourceIPAccessList() *schema.Resource {
...
-			if err != nil {
-				return err
-			}
-			common.StructToData(status.IpAccessList, s, d)
-			return nil
+			return c.AccountOrWorkspaceRequest(func(acc *databricks.AccountClient) error {
+				_, err := acc.IpAccessLists.GetByIpAccessListId(ctx, d.Id())
+				if err != nil {
+					return err
+				}
```
alexott, line 68: If it returns blank response, do we need this piece of code?

alexott (approved):
lgtm

access/resource_ip_access_list.go
nkvuong: this is to check if the access list id exists, before listing all the access lists and find the right one

PR #2591, opened 2023-08-17
Fixed databricks_access_control_rule_set integration test in Azure

mgyucht:
## Changes
This PR makes two changes:
1. Use string prefix to match on the cloud environment rather than exact match. This fixes the underlying issue.
2. Load environment before calling `accountTest`. This allows users to debug this test from VS Code without needing to run it in CI.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott (approved):
lgtm

PR #2606, opened 2023-08-22
add `databricks_system_schema` resource

### LINKED ISSUE #2663: [FEATURE] Enable system tables through terraform for UC
### Use-cases
We manage many databricks accounts across our organisation.  We are starting to use system tables for a few of our internal processes such as billing auditing etc. 

At the moment we need to run different curl commands per account to enable/disable system tables per UC, which is not very scalable. 
### Attempted Solutions
We're in the process of automating this working around the fact that there's no way (or at least I couldn't find it in the docs) to do this through terraform as we do with the rest of our databricks config.
### Proposal

Having a new resources that looks like similar to 
```
resource "databricks_metastore_systemschema" {
  name = "billing"
  status = "enable"
}
```
would be really useful

nkvuong:
## Changes
- add `databricks_system_schema` resource to enable [system tables](https://docs.databricks.com/en/administration-guide/system-tables/index.html)
close #2663

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_system_schema.go
+++ b/catalog/resource_system_schema.go
@@ -4,29 +4,34 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/databricks/databricks-sdk-go/service/catalog"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"golang.org/x/exp/slices"
 )
 
 func ResourceSystemSchema() *schema.Resource {
-	create := func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-		o, n := d.GetChange("system_schema")
-		old, okOld := o.([]any)
-		new, okNew := n.([]any)
+	systemSchema := common.StructToSchema(catalog.SystemSchemaInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		m["metastore_id"] = &schema.Schema{
+			Type:     schema.TypeString,
+			Computed: true,
+		}
+		m["state"].Computed = true
+		return m
+	})
+	pi := common.NewPairID("metastore_id", "schema").Schema(
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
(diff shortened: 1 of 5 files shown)


catalog/resource_system_schema.go
```diff
@@ -0,0 +1,116 @@
...
+				}
+			}
+		}
+		d.SetId(metastoreSummary.MetastoreId)
+		return nil
+	}
+	return common.Resource{
+		Schema: map[string]*schema.Schema{
+			"system_schema": {
+				Type:     schema.TypeList,
```
alexott, line 65: Wouldn't it be better to use `TypeSet` to avoid changes when order changes, handling duplicate elements, etc.?

catalog/resource_system_schema.go
```diff
@@ -0,0 +1,116 @@
...
+	"log"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/exp/slices"
+)
+
+func ResourceSystemSchema() *schema.Resource {
+	create := func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
```
alexott, line 15: call it createOrUpdate?

alexott (approved):
LGTM, some minor comments

mgyucht (commented):
IIUC, this PR models system schemas as a single resource with a `system_schema` field that is a list of strings. However, looking at this from the REST perspective, the resource should probably just correspond to a single system schema. Users can use for_each if they want to define multiple schemas in a metastore rather than us hand-coding the implementation here. The reason I suggest this direction is because it aligns Terraform with the REST API, which improves the chances that we can autogenerate this resource in the future without custom implementations. WDYT?

nkvuong:
@mgyucht it certainly simplify our codebase

I will add a separate PR for a data source that list all available system schemas, to be used for the `for_each`

tanmay-db (approved):
LGTM as the comments have been addressed.

PR #2626, opened 2023-08-29
Detect run_as drift in job resource

dvkuznietsov:
## Changes
run_as field in jobs is asymmetric: it's JobSettings.RunAs in create/update requests and RunAsUserName in read requests. As a result, when the job drifts, terraform can't detect the drift. To fix that, adjust the Read function in terraform to fill in the RunAs struct based on the RunAsUsername field.

## Tests
Tested locally E2E

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -574,6 +574,21 @@ func (a JobsAPI) Read(id string) (job Job, err error) {
 		job.Settings.sortTasksByKey()
 		job.Settings.sortWebhooksByID()
 	}
+
+	if job.RunAsUserName != "" && job.Settings != nil {
+		userNameIsEmail := strings.Contains(job.RunAsUserName, "@")
+
+		if userNameIsEmail {
+			job.Settings.RunAs = &JobRunAs{
+				UserName: job.RunAsUserName,
+			}
+		} else {
+			job.Settings.RunAs = &JobRunAs{
+				UserName: job.RunAsUserName,
+			}
+		}
+	}
+
 	return
 }
 
--- a/jobs/resource_job_test.go
+++ b/jobs/resource_job_test.go
@@ -87,6 +87,9 @@ func TestResourceJobCreate(t *testing.T) {
 							PauseStatus:          "PAUSED",
 						},
 						Queue: &Queue{},
+						RunAs: &JobRunAs{
+							UserName: "user@mail.com",
+						},
 					},
 				},
 			},


jobs/resource_job.go
```diff
@@ -574,6 +574,21 @@ func (a JobsAPI) Read(id string) (job Job, err error) {
...
+		userNameIsEmail := strings.Contains(job.RunAsUserName, "@")
+
+		if userNameIsEmail {
+			job.Settings.RunAs = &JobRunAs{
+				UserName: job.RunAsUserName,
+			}
+		} else {
+			job.Settings.RunAs = &JobRunAs{
+				UserName: job.RunAsUserName,
+			}
```
tanmay-db, line 588: Both the statements in if else is same?

dvkuznietsov: sorry, didn't push the latest commit

dvkuznietsov:
@tanmay-db 

> Hi, can you mention how you tested this E2E locally?

- I've built the provider from sources, configured terraform to use the binary and ran `terraform apply`. 
- then I logged into the workspace and verified the job was created with correct run as
- then I changed the run as of the job
- lstly, I ran `terraform` plan. Terraform detected the drift

tanmay-db:
I have triggered an integration test run. Please wait until it's finished before this can be merged.

jobs/resource_job.go
```diff
@@ -574,6 +574,21 @@ func (a JobsAPI) Read(id string) (job Job, err error) {
...
+	if job.RunAsUserName != "" && job.Settings != nil {
+		userNameIsEmail := strings.Contains(job.RunAsUserName, "@")
+
+		if userNameIsEmail {
+			job.Settings.RunAs = &JobRunAs{
+				UserName: job.RunAsUserName,
+			}
+		} else {
+			job.Settings.RunAs = &JobRunAs{
+				ServicePrincipalName: job.RunAsUserName,
```
tanmay-db, line 587: @dvkuznietsov  Two questions: 
* Is it certain that if a job is not userEmail then it is service principal? 
* Is checking for @ for email enough? 

Hi @nkvuong, do you know if there is some existing utility for checking valid email in terraform?

dvkuznietsov: > Is it certain that if a job is not userEmail then it is service principal?
> Is checking for @ for email enough?

as of now, run_as_user_name is either an email or a UUID. String parsing is not ideal and that's why we moved on to run_as field in write path, but that's what we work with for now

dvkuznietsov: and, to answer your question, I beleive that's enough

mgyucht: @dvkuznietsov is it too much to ask the Jobs API to use run_as for reads and writes rather than force clients to do this mapping themselves? Thinking RESTfully, when a user sets a field on a resource, they would expect to get that field as they set it when getting that resource back. However, here, Jobs may be breaking that pattern. WDYT?

dvkuznietsov: @mgyucht the fields are not symmetric: run_as is an optional field that we have recently added and run_as_user_name is an old field that is almost always filled in.

The intention for this change is if users don't provide run_as, we don't need to catch drifts in it, but if they do provide it, we do need to detect drifts

mgyucht: Can you include `run_as` in the get response though? I guess what I don't understand is why it is a write-only field.

dvkuznietsov: we can't. It would be a breaking change for customers that pipe GET responses to create/reset calls

mgyucht: Forgive me for asking, but why would that be a breaking change? Does the Jobs API not tolerate both of these parameters being set at the same time in a create/reset call?

jobs/resource_job_test.go
```diff
@@ -87,6 +87,9 @@ func TestResourceJobCreate(t *testing.T) {
 							PauseStatus:          "PAUSED",
 						},
 						Queue: &Queue{},
+						RunAs: &JobRunAs{
+							UserName: "user@mail.com",
+						},
```
mgyucht, line 92: I thought RunAs is not included in get requests?

dvkuznietsov:
@tanmay-db could you re-run integration tests please?

jobs/resource_job_test.go
tanmay-db: Just to confirm @dvkuznietsov what was the conclusion on these comments? I am guessing these were discussed offline?

tanmay-db (approved):
LGTM, unrelated integration tests are failing and shouldn't be a blocker for this.

Issue #1864, opened 2022-12-20
[ISSUE] Issue with `databricks_mount` resource when changing admin clusters


PeterDowdy:
### Configuration
```hcl
resource "databricks_cluster" "admin" {
  cluster_name            = "staging - Admin"
  spark_version           = data.databricks_spark_version.latest_lts.id
  node_type_id            = data.databricks_node_type.smallest.id
  autotermination_minutes = 30
  num_workers             = 0
  spark_conf = {
    # Single-node
    "spark.databricks.cluster.profile" : "singleNode"
    "spark.master" : "local[*]"
  }
  custom_tags = {
    "ResourceClass" = "SingleNode"
  }
  spark_env_vars = {
    GITLAB_REGISTRY_TOKEN = "{{secrets/gitlab-package-registry/gitlab-registry-token}}"
  }
  aws_attributes {
    instance_profile_arn = module.databricks_instance_admin_profile.id
  }

  depends_on = [
    module.databricks_instance_admin_profile
  ]
}
### S3 buckets

module "output_bucket" {
  source                  = "./modules/bio_team_bucket"
  bucket_name             = "a_bucket_name"
  deployment_stage  = var.deployment_stage
  data_encryption_key_arn = "arn:aws:kms:a_key_goes_here"
}

# ## DBFS mounts
resource "databricks_mount" "output_bucket_mount" {
  name            = module.output_bucket.bucket_id
  encryption_type = "sse-kms"

  cluster_id = databricks_cluster.admin.cluster_id
  s3 {
    bucket_name = module.output_bucket.bucket_id
  }
}
```

### Expected Behavior
The mount should re-create with the new cluster

### Actual Behavior
The mount fails to create with this error message: `instance profile is required to re-create mounting cluster`. When I inspect the terraform state for the mount, it still has the ID of the old admin cluster. Since that cluster is gone and that ID is irretrievable, the mount keeps failing when it 400s while requesting the cluster by ID. It then ignores the cluster_id specified in the terraform and raises an error message.

### Steps to Reproduce
1. Create a cluster in terraform
2. Create an S3 mount using that cluster
3. Delete that cluster
4. Create a new admin cluster and set the s3 mount to use that cluster

### Terraform and provider versions
Terraform v1.3.6
on darwin_arm64
+ provider registry.terraform.io/databricks/databricks v1.7.0
+ provider registry.terraform.io/hashicorp/aws v4.15.1
[...]

PR #2634, opened 2023-08-30
Fixed mount read failing if cluster no longer exists

shreyas-goenka:
## Changes
This PR fixes issue: https://github.com/databricks/terraform-provider-databricks/issues/1864

Databricks does not have an API for mounts. To work around this issue and offer a terraform resource for it we would instead rely on an existing cluster for CRUD (using dbutils). 

An issue with this is if the cluster used to last create / update the mount no longer exists, it's not possible to read the current remote state of mount (or delete it). This is because during READ / DELETE terraform operations, the databricks provider application does not have access to any HCL configruation. 

Thus the solution here is to assume that a mount has been deleted if the cluster no longer exists. This is fine because mount create function in the provider is idempotent.

This PR also makes some tangentially related changes to the testing helper function, to make the acceptance tests work here, namely:
1. Adds a PreConfig field. This is required to delete the cluster during the pinned cluster test run.
2. Passes all state fields instead of just the ids during the default read checks. This is required (and nice) because of two reasons:
       - Maps to real terraform behaviour where the entire state is availble during read operations.
       - The ID is not sufficient to read a `databricks_mount` resource. For example: a S3 databricks mount resource where you also need a instance_profile or cluster_id.

## Tests
Using unit tests, integration tests and manually. 

For manual testing:
Steps:
1. Deploy a mount
4. Delete the cluster used to create the mount
5. run terraform plan/apply which will cause a read request to be sent to the mount.

Before:
This would fail with the error
```
│ Error: instance profile is required to re-create mounting cluster
│ 
│   with databricks_mount.my_mount,
│   on main.tf line 34, in resource "databricks_mount" "my_mount":
│   34: resource "databricks_mount" "my_mount" {
│ 
╵
```

After:
The mount is "recreated" on apply.

- [x] `make test` run locally
- [x] using Go SDK


storage/resource_mount_test.go
```diff
@@ -108,6 +108,78 @@ func TestResourceAwsS3MountGenericCreate(t *testing.T) {
 	assert.Equal(t, testS3BucketPath, d.Get("source"))
 }
 
+func TestResourceAwsS3MountGenericCreateWithInvalidClusterId(t *testing.T) {
```
mgyucht, line 111: This case is not affected by your PR, right?

storage/resource_mount.go
```diff
@@ -37,6 +39,34 @@ func (cb mountCallback) preProcess(r *schema.Resource) func(
...
+// because without a running cluster we do not have a way to do CRUD on mounts.
+// The argument being that possibly orphaning mounts is a better experience than
+// requiring manual edits to tfstate.
+func validateCluster(cb func(context.Context, *schema.ResourceData, any) diag.Diagnostics) func(context.Context, *schema.ResourceData, any) diag.Diagnostics {
+	return func(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+		clusterId := d.Get("cluster_id").(string)
+		if clusterId != "" {
+			clustersAPI := clusters.NewClustersAPI(ctx, m)
+			_, err := clustersAPI.Get(clusterId)
+			if apierr.IsMissing(err) {
```
mgyucht, line 60: Thank you for using apierr :)

storage/resource_mount.go
```diff
@@ -37,6 +39,34 @@ func (cb mountCallback) preProcess(r *schema.Resource) func(
...
+// that the mount also does not exist.
+//
+// In cases of update requests this would lead to an attempted recreation of the
+// mount
+// In cases of delete, this could orphan the mount
+//
+// This is needed to resolve issue: https://github.com/databricks/terraform-provider-databricks/issues/1864
+// because without a running cluster we do not have a way to do CRUD on mounts.
+// The argument being that possibly orphaning mounts is a better experience than
+// requiring manual edits to tfstate.
```
pietern, line 53: What would happen if we only set the `cluster_id` to the empty string if it doesn't exist?

Is it possible that might trigger creation of a temporary cluster to perform the CRUD after all?

pietern (commented):
Code LGTM, one question on the approach.

I think this is the only way we can make this work in all cases, because downstream code relies on a cluster ID being set. If it isn't, it will fail on read, and the only way to fix that is with manual state manipulation.

Still, it would be good to include a short discussion of this case in a comment.

storage/resource_mount.go
```diff
@@ -62,8 +92,8 @@ func ResourceMount() *schema.Resource {
 	tpl := GenericMount{}
 	r := commonMountResource(tpl, ResourceDatabricksMountSchema())
 	r.CreateContext = mountCallback(mountCreate).preProcess(r)
-	r.ReadContext = mountCallback(mountRead).preProcess(r)
-	r.DeleteContext = mountCallback(mountDelete).preProcess(r)
+	r.ReadContext = validateCluster(mountCallback(mountRead).preProcess(r))
+	r.DeleteContext = validateCluster(mountCallback(mountDelete).preProcess(r))
```
mgyucht, line 96: Is this needed on delete as well?

storage/resource_mount.go
mgyucht: Do you need an instance profile to read mounts? If so, I think this is a good option. We only need the temporary cluster for the read pathway.

shreyas-goenka: Re: are instance profiles required? 
Not sure. One disadvantage that this approach has over the current approach is though users will need cluster_creation permissions to apply. This could be a regression for some customers.

Re: Setting cluster_id to the empty string. 
I believe that is what happens in the current code path. If the cluster_id is empty then the provider tries to use an instance profile if you have one defined. 
see: 
https://github.com/databricks/terraform-provider-databricks/blob/8dc13d4d703fee2ad85b6c60fedba72f0b0b87b1/storage/s3.go#L70
AND
https://github.com/databricks/terraform-provider-databricks/blob/8dc13d4d703fee2ad85b6c60fedba72f0b0b87b1/storage/s3.go#L79

storage/resource_mount_test.go
shreyas-goenka: Yup

storage/resource_mount.go
shreyas-goenka: You are welcome. Copy pasted this piece, but I'll take the credit :)

storage/resource_mount.go
shreyas-goenka: Yes. Since the resources does not have a update method, what happens instead is a "delete" and then "create".

During delete, we need to validate if the cluster is stale.

tanmay-db (approved):
went through the PR, LGTM. Adding changes to doc folder would be good for potential user facing changes 

Triggered an integration test run, please wait until that is completed.

storage/resource_mount.go
```diff
@@ -37,6 +39,41 @@ func (cb mountCallback) preProcess(r *schema.Resource) func(
 	}
 }
 
+// If a cluster_id is specified for the mount, this function validates that the
+// cluster actually exists. If the cluster does not exist, we assume (incorrectly)
+// that the mount also does not exist.
+//
+// In cases of update requests this would lead to an attempted recreation of the
+// mount
```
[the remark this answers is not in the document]
shreyas-goenka: done :)

shreyas-goenka:
NOTE: added a note about this in the docs.

alexott (approved):

storage/resource_mount.go
```diff
@@ -37,6 +39,41 @@ func (cb mountCallback) preProcess(r *schema.Resource) func(
...
+// because without a running cluster we do not have a way to do CRUD on mounts.
+// The argument being that possibly orphaning mounts is a better experience than
+// requiring manual edits to tfstate.
+//
+// Why do we need a valid cluster_id to be set here?
+// Answer: Downstream read and delete code relies on a valid cluster_id being set
+// in the terraform state. If the cluster_id is invalid then there is no way to update
+// it using native terraform apply workflows. The only workaround is to manually edit the
+// tfstate file replacing the existing invalid cluster_id with a new valid one.
+func validateCluster(cb func(context.Context, *schema.ResourceData, any) diag.Diagnostics) func(context.Context, *schema.ResourceData, any) diag.Diagnostics {
```
mgyucht, line 61: I was very confused by this method but actually it makes sense. I just think we need to rename it to something like `removeFromStateIfClusterDoesNotExist`. The description should be updated to reflect this as well, something like: "removeFromStateIfClusterDoesNotExist marks the resource as deleted if the cluster with the given cluster ID specified for the mount point does not exist." It is kind of in your description above, but it would be good to be more explicit about what this does. (What does "we assume that the mount does not exist" mean?)

mgyucht (approved):
Several nits, but the implementation looks good to me. Sorry about the delay here, as it took me a bit of time to totally grok the issue, and this has been a bit lower down on my priority list.

docs/resources/mount.md
```diff
@@ -7,6 +7,8 @@ This resource will [mount your cloud storage](https://docs.databricks.com/data/d
 
 **Note** When `cluster_id` is not specified, it will create the smallest possible cluster in the default availability zone with name equal to or starting with `terraform-mount` for the shortest possible amount of time. To avoid mount failure due to potentially quota or capacity issues with the default cluster, we recommend specifying a cluster to use for mounting.
 
+**Note** CRUD operations on a databricks mount require a running cluster. Due to limitations with terraform and the databricks mounts APIs, if the cluster the mount was most recently created / updated using no longer exists AND the mount is destroyed as a part of a terraform apply, we mark it as deleted without cleaning it up from the workspace.
```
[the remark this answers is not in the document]
shreyas-goenka: done

internal/acceptance/mounts_test.go
```diff
@@ -0,0 +1,145 @@
...
+			  
+				# Test cluster to create the mount using.
+				resource "databricks_cluster" "this" {
+					cluster_name = "acc-test-mounts-{var.RANDOM}"
+					spark_version = data.databricks_spark_version.latest.id
+					instance_pool_id = "{env.TEST_INSTANCE_POOL_ID}"
+					num_workers = 1
+
+					aws_attributes {
+					instance_profile_arn = "{env.TEST_INSTANCE_PROFILE_ARN}"
```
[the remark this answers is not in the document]
shreyas-goenka: done

internal/acceptance/mounts_test.go
```diff
@@ -0,0 +1,145 @@
...
+				cluster := resources["databricks_cluster.this"]
+				if cluster == nil {
+					return fmt.Errorf("expected to find databricks_cluster.this in resources keys: %v", maps.Keys(resources))
+				}
+				clusterId2 = cluster.Primary.ID
+
+				// Assert cluster was indeed recreated
+				assert.NotEmpty(t, clusterId1)
+				assert.NotEmpty(t, clusterId2)
+				assert.NotEqual(t, clusterId1, clusterId2)
```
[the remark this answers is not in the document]
shreyas-goenka: This asserts an invariant for this test, ie the old cluster was "deleted" IMO nice to assert this case works exactly as expected.

storage/resource_mount.go
shreyas-goenka: done

PR #2641, opened 2023-08-31
updated share and recipient docs; terrafmt

owlleg6:
## Changes

Updated `databricks_share` documentation, an owner attribute has to be outside of `object` configuration block.

Added more detailed description to `databricks_recipient` attribute reference `token` attribute.

Updated all docs using 'terrafmt' .

## Tests

Validated using `make fmt-docs`. Only documentation files are changed.


mgyucht (approved):
Thank you for doing this!

PR #2649, opened 2023-09-04
Remove callback field from acceptance test framework

shreyas-goenka:
## Changes
Internal terraform helpers provide a `Callback` field that is not a part of the terraform SDK. We already have an alternate available used everywhere else but for `group_member_test.go`

This PR switches to using the `Check` field instead, and removes the callback field.

It's nice to remove this field because it encourages a bad pattern where the assertion is made on the first resource defined in the test template, which is captured using regex capture groups.

## Tests
The tests still pass. Also manually tested using debugger that the assertions now in `Check` are still called.


tanmay-db (approved):
LGTM, is there already an integration test run for it?

shreyas-goenka:
Ran the integration tests locally, they passed.

PR #2667, opened 2023-09-09
Add deprecation warning for init scripts from DBFS in `databricks_cluster`, `databricks_job` & `databricks_pipeline` resources

alexott:
## Tests

- [x] `make test` run locally
- [ ] using Go SDK


clusters/resource_cluster.go
```diff
@@ -15,6 +15,8 @@ import (
 // DefaultProvisionTimeout ...
 const DefaultProvisionTimeout = 30 * time.Minute
 
+const DbfsDeprecationWarning = "For init scripts use 'volumes', 'workspace' or cloud storage location instead of 'dbfs'."
```
tanmay-db, line 18: Thanks for doing this, a question - can single quotes within quotes cause an issue?

tanmay-db (approved):

clusters/resource_cluster.go
alexott: No. We have such things in outer places as well.

PR #2680, opened 2023-09-12
Exporter: improve exporting of `databricks_cluster_policy` resource

### LINKED ISSUE #2664: Custom cluster policy not getting exported using tf databricks exporter
### Configuration
terraform {
required_providers {
databricks = {
version = "1.21.0"
}
}
}

### Expected Behavior
Custom cluster policy should get exported using tf databricks exporter

### Actual Behavior
Custom cluster policy not getting exported using tf databricks exporter

### Steps to Reproduce

1. Create a new custom cluster policy using primary workspace >> Compute >> Policies >> Create Policy
2. Use terraform databricks exporter to export this policy 
3. We will see that only built in policies gets exported and we don't see the custom polices being exported

$ terraform version
Terraform v1.1.7
on windows_amd64

provider registry.terraform.io/databricks/databricks v1.21.0

alexott:
## Changes

This includes:

- Listing of all cluster policies, not only that are referenced in the jobs/clusters.
- Emit secret scopes when they are referenced in Spark Conf or environment variables

Also, fixes generation of non-existent secret scopes that happens when we emit non-existent secret scope (for example, it's a dangling reference in cluster policy or old job configuration).

This fixes #2664

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -13,6 +13,7 @@ import (
 
 	"golang.org/x/exp/slices"
 
+	"github.com/databricks/databricks-sdk-go/service/compute"
 	"github.com/databricks/databricks-sdk-go/service/ml"
 	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/terraform-provider-databricks/clusters"
@@ -552,9 +553,37 @@ var resourcesMap map[string]importable = map[string]importable{
 		},
 	},
 	"databricks_cluster_policy": {
-		Service: "compute",
+		Service: "policies",
 		Name: func(ic *importContext, d *schema.ResourceData) string {
 			return d.Get("name").(string)
+		},
+		List: func(ic *importContext) error {
+			w, err := ic.Client.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			policies, err := w.ClusterPolicies.ListAll(ic.Context, compute.ListClusterPoliciesRequest{})
+			if err != nil {
+				return err
+			}
+			for offset, policy := range policies {
+				log.Printf("[INFO] Scanning %d:  %v", offset+1, policy)
+				if slices.Contains(predefinedClusterPolicies, policy.Name) {
+					continue
+				}
+				if !ic.MatchesName(policy.Name) {
+					log.Printf("[DEBUG] Policy %s doesn't match %s filter", policy.Name, ic.match)
+					continue
+				}
+				ic.Emit(&resource{
+					Resource: "databricks_cluster_policy",
+					ID:       policy.PolicyId,
+				})
+				if offset%10 == 0 {
+					log.Printf("[INFO] Scanned %d of %d cluster policies", offset+1, len(policies))
+				}
+			}
+			return nil
 		},
 		Import: func(ic *importContext, r *resource) error {
 			if ic.meAdmin {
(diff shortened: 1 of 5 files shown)


exporter/importables.go
```diff
@@ -534,10 +535,38 @@ var resourcesMap map[string]importable = map[string]importable{
...
+		Service: "policies",
 		Name: func(ic *importContext, d *schema.ResourceData) string {
 			return d.Get("name").(string)
 		},
+		List: func(ic *importContext) error {
+			w, err := ic.Client.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			policies, err := w.ClusterPolicies.ListAll(ic.Context, compute.ListClusterPoliciesRequest{})
```
tanmay-db, line 547: Why was listing all policies not required before?

exporter/importables.go
```diff
@@ -534,10 +535,38 @@ var resourcesMap map[string]importable = map[string]importable{
...
+				}
+				if !ic.MatchesName(policy.Name) {
+					log.Printf("[INFO] Policy %s doesn't match %s filter", policy.Name, ic.match)
+					continue
+				}
+				ic.Emit(&resource{
+					Resource: "databricks_cluster_policy",
+					ID:       policy.PolicyId,
+				})
+				if offset%10 == 0 {
```
tanmay-db, line 564: Is this for better readability on console?

exporter/importables.go
```diff
@@ -556,7 +585,8 @@ var resourcesMap map[string]importable = map[string]importable{
 				defaultValue, dok := policy["defaultValue"]
 				typ := policy["type"]
 				if !vok && !dok {
-					log.Printf("[INFO] Skipping policy element as it doesn't have both value and defaultValue")
+					log.Printf("[DEBUG] Skipping policy element as it doesn't have both value and defaultValue. k='%v', policy='%v'",
```
tanmay-db, line 588: If the above line is at INFO

> log.Printf("[INFO] Policy %s doesn't match %s filter", policy.Name, ic.match)

Then I think this should also be INFO?

exporter/importables.go
```diff
@@ -582,6 +612,15 @@ var resourcesMap map[string]importable = map[string]importable{
 						ID:       eitherString(value, defaultValue),
 					})
 				}
+				if typ == "fixed" && (strings.HasPrefix(k, "spark_conf.") || strings.HasPrefix(k, "spark_env_vars.")) {
```
tanmay-db, line 615: For my understanding can you explain this a bit more, why `fixed` is required?

exporter/importables.go
alexott: yeah, it should be DEBUG

exporter/importables.go
alexott: Because there are other types, like, `regexp` - in this case we won't be able to handle it

exporter/importables.go
alexott: It was an initial design decision from the beginning of development, but now customers are starting to ask for it. The new implementation can behave the old way - just don't provide `policies` in the `-listing` switch

exporter/importables.go
alexott: It will make things too noisy...

exporter/importables.go
alexott: fixed

tanmay-db (approved):

PR #2691, opened 2023-09-18
Exporter: parallel listing of workspace objects

alexott:
## Changes

Right now listing of workspace objects (files/directories) is a bottleneck, especially in the incremental export mode. This PR tries to solve this problem by using multiple goroutines listing multiple directories in parallel

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/workspace/resource_notebook.go
+++ b/workspace/resource_notebook.go
@@ -3,9 +3,13 @@ package workspace
 import (
 	"context"
 	"encoding/base64"
+	"log"
+	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/databricks/terraform-provider-databricks/common"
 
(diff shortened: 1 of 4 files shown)


workspace/resource_notebook.go
```diff
@@ -139,6 +143,86 @@ func (a NotebooksAPI) Mkdirs(path string) error {
...
+	numWorkers := 5
+	if os.Getenv("EXPORTER_WS_LIST_PARALLELISM") != "" {
+		t, _ := strconv.ParseInt(os.Getenv("EXPORTER_WS_LIST_PARALLELISM"), 10, 32)
+		numWorkers = int(t)
+	}
+	channelSize := 100000
+	if os.Getenv("EXPORTER_CHANNEL_SIZE") != "" {
+		t, _ := strconv.ParseInt(os.Getenv("EXPORTER_CHANNEL_SIZE"), 10, 32)
+		channelSize = int(t)
+	}
```
tanmay-db, line 200: I think abstracting this logic outside might be good since we are duplicating getting the os.Getenv()
````
func getEnvAsInt(envName string, defaultValue int) int {
	if val, exists := os.LookupEnv(envName); exists {
		parsedVal, err := strconv.Atoi(val)
		if err == nil {
			return parsedVal
		}
	}
	return defaultValue
}
````

File: workspace/resource_notebook.go [hunk 2]
```diff
@@ -139,6 +143,86 @@ func (a NotebooksAPI) Mkdirs(path string) error {
...
+			time.Sleep(15 * time.Millisecond)
+		}
+	}
+}
+
+func (a NotebooksAPI) ListParallel(path string, recursive bool) ([]ObjectStatus, error) {
+	var answer syncAnswer
+	wg := &sync.WaitGroup{}
+
+	numWorkers := 5
```
tanmay-db, line 191: Also, let's define constants outside like:
````
const (
	Directory                      = "Directory"
	DefaultWorkers                 = 5
	DefaultChannelSize             = 100000
	directyListingMaxAttempts      = 3
	EnvParallelism                 = "EXPORTER_WS_LIST_PARALLELISM"
	EnvChannelSize                 = "EXPORTER_CHANNEL_SIZE"
)

````

tanmay-db (approved):
Couple comments, rest LGTM.

alexott:
Thank you - it makes sense! I'll update the code tomorrow.

Issue #2693, opened 2023-09-18
[DOC] AWS Unity Catalog Service Principal Reference Typo


metrocavich:
### Affected Resource(s)
https://registry.terraform.io/providers/databricks/databricks/latest/docs/guides/unity-catalog

### Expected Details
Second link, `[Service Principal](resources/service_principal.md)` should be `[Service Principal](../resources/service_principal.md)`

### List of things to potentially add/remove:
Fix link

### Important Factoids
Missing a ../ causes it to direct to https://registry.terraform.io/providers/databricks/databricks/latest/docs/guides/docs/resources/service_principal
Which is invalid as it's under guides.

### References
#2694

metrocavich:
Fixed by #2694

PR #2694, opened 2023-09-18
Fix Service Principal reference

metrocavich:
Typo in path fixed

## Changes
Simple typo fix in reference to #2693


alexott (approved):
thank you

PR #2699, opened 2023-09-19
Exporter: added support for exporting `databricks_access_control_rule_set` on account level

alexott:
## Changes

Added exporting of a given resource. 

Also:
* added some documentation inside the code
* possibility to specify if a resource is a workspace or account level
* fixed bug in the regex matching

## Tests

There were some problems with emulating account-level APIs, so the unit test will be added a bit later.

- [x] `make test` run locally
- [x] tested manually
- [x] using Go SDK

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -287,15 +287,15 @@ func (ic *importContext) Run() error {
 	} else if !info.IsDir() {
 		return fmt.Errorf("the path %s is not a directory", ic.Directory)
 	}
-	w, err := ic.Client.WorkspaceClient()
-	if err != nil {
-		return err
-	}
 
 	ic.accountLevel = ic.Client.Config.IsAccountClient()
 	if ic.accountLevel {
 		ic.meAdmin = true
 	} else {
+		w, err := ic.Client.WorkspaceClient()
+		if err != nil {
+			return err
+		}
 		me, err := w.CurrentUser.Me(ic.Context)
 		if err != nil {
 			return err
@@ -326,7 +326,11 @@ func (ic *importContext) Run() error {
 			continue
 		}
 		if ic.accountLevel && !ir.AccountLevel {
-			log.Printf("[DEBUG] %s (%s service) is not account level", resourceName, ir.Service)
+			log.Printf("[DEBUG] %s (%s service) is not a account level resource", resourceName, ir.Service)
+			continue
+		}
+		if !ic.accountLevel && !ir.WorkspaceLevel {
+			log.Printf("[DEBUG] %s (%s service) is not a workspace level resource", resourceName, ir.Service)
 			continue
 		}
 		ic.waitGroup.Add(1)
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -14,6 +14,7 @@ import (
 	"golang.org/x/exp/slices"
 
 	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/databricks-sdk-go/service/iam"
 	"github.com/databricks/databricks-sdk-go/service/ml"
 	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/terraform-provider-databricks/clusters"
@@ -154,7 +155,8 @@ func generateMountBody(ic *importContext, body *hclwrite.Body, r *resource) erro
 
 var resourcesMap map[string]importable = map[string]importable{
 	"databricks_dbfs_file": {
-		Service: "storage",
+		WorkspaceLevel: true,
+		Service:        "storage",
 		Name: func(ic *importContext, d *schema.ResourceData) string {
 			fileNameMd5 := fmt.Sprintf("%x", md5.Sum([]byte(d.Id())))
 			s := strings.Split(d.Id(), "/")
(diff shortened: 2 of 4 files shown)


exporter/importables.go
```diff
@@ -1951,4 +2002,75 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			return nil
+		},
+		Depends: []reference{
+			{Path: "grant_rules.principals", Resource: "databricks_user", Match: "acl_principal_id"},
+			{Path: "grant_rules.principals", Resource: "databricks_group", Match: "acl_principal_id"},
+			{Path: "grant_rules.principals", Resource: "databricks_service_principal", Match: "acl_principal_id"},
+			{Path: "name", Resource: "databricks_service_principal", Match: "application_id", MatchType: MatchRegexp,
+				Regexp: regexp.MustCompile("^accounts/[^/]+/servicePrincipals/([^/]+)/ruleSets/default$")},
+			{Path: "name", Resource: "databricks_group", MatchType: MatchRegexp,
+				Regexp: regexp.MustCompile("^accounts/[^/]+/groups/([^/]+)/ruleSets/default$")},
```
gauthamsunjay, line 2066: we also support the account rule set btw - `accounts/{accountId}/ruleSets/default`. Should that be included here?

alexott: These are replacement rules that convert something like `accounts/111111/servicePrincipals/123456/ruleSets/default` into Terraform code with reference to another resource `accounts/111111/servicePrincipals/${databricks_service_principal.123456.application_id}/ruleSets/default`.

Default ruleset is handled in the actual export logic [here](https://github.com/databricks/terraform-provider-databricks/pull/2699/files#diff-ab7238b14b515ceaca6b7f28691827b1044cb9a521f517753a17aa87b20c795bR2014).

gauthamsunjay (approved):

alexott:
@tanmay-db @mgyucht can we get this reviewed - besides new functionality, this PR contains a fix for a bug that leads to a crash when generating references - I hit it in #2742 last week as well.

exporter/context.go
```diff
@@ -326,7 +326,11 @@ func (ic *importContext) Run() error {
 			continue
 		}
 		if ic.accountLevel && !ir.AccountLevel {
-			log.Printf("[DEBUG] %s (%s service) is not account level", resourceName, ir.Service)
+			log.Printf("[DEBUG] %s (%s service) is not a account level resource", resourceName, ir.Service)
```
tanmay-db, line 329: nit: `an account`

exporter/importables.go
```diff
@@ -1971,4 +2022,75 @@ var resourcesMap map[string]importable = map[string]importable{
 			// {Path: "http_url_spec.authorization", Variable: true},
 		},
 	},
+	"databricks_access_control_rule_set": {
+		AccountLevel: true,
+		Service:      "access",
+		// Name: func(ic *importContext, d *schema.ResourceData) string {
```
tanmay-db, line 2028: we don't require `Name` for access_control?

exporter/importables.go
```diff
@@ -1971,4 +2022,75 @@ var resourcesMap map[string]importable = map[string]importable{
...
+		// 	return "webhook_" + d.Id()
+		// },
+		List: func(ic *importContext) error {
+			accountId := ic.Client.Config.AccountID
+			// emit default ruleset
+			ic.Emit(&resource{
+				Resource: "databricks_access_control_rule_set",
+				ID:       fmt.Sprintf("accounts/%s/ruleSets/default", accountId),
```
tanmay-db, line 2036: Trying to understand this a bit better, why do we emit default ruleset in List?

exporter/importables.go
```diff
@@ -1971,4 +2022,75 @@ var resourcesMap map[string]importable = map[string]importable{
...
+					if len(parts) != 2 {
+						log.Printf("[WARN] Incorrect principal: '%s'", principal)
+						continue
+					}
+					switch parts[0] {
+					case "users":
+						ic.Emit(&resource{
+							Resource:  "databricks_user",
+							Attribute: "user_name",
+							Value:     parts[1],
```
tanmay-db, line 2057: Are there any checks we can put on Value? and throw error if not met?

tanmay-db (approved):
Some comments but otherwise LGTM.

I had a question on testing. 

> There were some problems with emulating account-level APIs, so the unit test will be added a bit later.

What were the issues? Do we already have a ticket to track unit-tests? I think we would want to have them as soon as possible otherwise this could lead to regression.

exporter/importables.go
alexott: No, it's not strictly required - if the `Name` function isn't defined, then a default name is generated

exporter/importables.go
alexott: Multiple checks are done inside the Emit function (if the service is enabled, etc.), and ImportResource - if a resource is already imported, if a resource exists, etc. It's better to make these checks in a single place (resource itself) instead of checking in each place that is doing E.

exporter/importables.go
alexott: We need an entry point to bootstrap the import process

mgyucht (approved):

PR #2706, opened 2023-09-21
Added logging package and fixed issue with API calls not being shown in DEBUG or lower log levels.

tanmay-db:
## Changes

Terraform never had logging configured correctly (used to log all levels irrespective of log level specified). The issue (api calls not being shown in `DEBUG` log level) surfaced because in the TF version bump to 1.20.9 which had Go SDK version bump to 0.10.0, the PR: https://github.com/databricks/databricks-sdk-go/pull/426 introduced INFO as basic log level leading to debug level never being enabled even if it is passed through the environment variable. 

## Tests

Unit tests added, Integration tests are running...

Also tested manually
```
 terraform {
  required_providers {
    databricks = {
      version = "1.26.0"
    }
  }
}

provider "databricks" {
    profile = <redacted>
}

data "databricks_catalogs" "all" {}

output "all_catalogs" {
  value = data.databricks_catalogs.all
}
```

shows the api calls

> 2023-09-21T03:03:41.340+0200 [DEBUG] provider.terraform-provider-databricks: GET /api/2.1/unity-catalog/catalogs
< HTTP/2.0 

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -1,0 +1,69 @@
+package logging
+
+import (
+	"os"
+
+	"github.com/databricks/databricks-sdk-go/logger"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
+)
+
+// Interface to mock os.Getenv() operations
+type envOperations interface {
+	Getenv(key string) string
+}
+
+type realEnvOperations struct{}
+
+func (r *realEnvOperations) Getenv(key string) string {
+	return os.Getenv(key)
+}
+
+// Unfortunately LogLevel() in terraform plugin sdk v2 is not test friendly so we have to mock it since it calls os.GetEnv() underneath
+type logLevelGetter interface {
+	GetLogLevel() string
+}
+
+type realLogLevelGetter struct{}
+
+func (r *realLogLevelGetter) GetLogLevel() string {
+	return logging.LogLevel()
+}
+
+func SetLogger() {
+	logger.DefaultLogger = &logger.SimpleLogger{
+		Level: terraformToSDKLogLevel(&realEnvOperations{}, &realLogLevelGetter{}),
+	}
+}
+
+// Golang doesn't have mutable maps, having global mutable variable is less safe as compared to a function returning map
+func terraformToSDKLogLevelMapping() map[string]logger.Level {
+	return map[string]logger.Level{
+		"TRACE": logger.LevelTrace,
+		"DEBUG": logger.LevelDebug,
+		"INFO":  logger.LevelInfo,
+		"WARN":  logger.LevelWarn,
+		"ERROR": logger.LevelError,
+	}
+}
+
+func terraformToSDKLogLevel(envOps envOperations, logGetter logLevelGetter) logger.Level {
+	// Exception for unknown log levels are handled in underlying LogLevel() method
+	// Defaulting to TRACE in such cases
+	terraformLogLevel := logGetter.GetLogLevel()
+
+	// If TF_LOG environment variable is not set
+	if terraformLogLevel == "" {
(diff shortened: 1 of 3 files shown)


logging/logging.go
```diff
@@ -0,0 +1,69 @@
...
+	terraformLogLevel := logGetter.GetLogLevel()
+
+	// If TF_LOG environment variable is not set
+	if terraformLogLevel == "" {
+		// If TF_ACC_LOG_PATH is set, we use TRACE log level to be consistent with LogOutput() method
+		if envOps.Getenv(logging.EnvAccLogFile) != "" {
+			return logger.Level(logger.LevelTrace)
+		} else {
+			// In all other cases we use INFO log level to be consistent with Go SDK
+			return logger.Level(logger.LevelInfo)
```
tanmay-db, line 61: There seems to be another issue with the provider / how terraform provider is configured. If TF_LOG="" or is not set then INFO level isn't used. 

This is true even without these changes.

tanmay-db: It works if TF_LOG=INFO is set

tanmay-db: I can try to investigate why this is happening but given it was happening before and can easily be mitigated with TF_LOG=info being set, we could go ahead with changes and file a bug fix for this to do later.

provider/provider.go
```diff
@@ -41,6 +42,7 @@ func init() {
 	// IMPORTANT: this line cannot be changed, because it's used for
 	// internal purposes at Databricks.
 	useragent.WithProduct("databricks-tf-provider", common.Version())
+	logging.SetLogger()
```
mgyucht, line 45: I don't know that this is great to do in `init()`, since this assumes that the logging configuration is done fully statically by TF (and I'm a bit nervous to put things in `init()` that don't have to be there). Maybe we move this to ConfigureContextFunc.

logging/logging.go
```diff
@@ -0,0 +1,69 @@
...
+		"TRACE": logger.LevelTrace,
+		"DEBUG": logger.LevelDebug,
+		"INFO":  logger.LevelInfo,
+		"WARN":  logger.LevelWarn,
+		"ERROR": logger.LevelError,
+	}
+}
+
+func terraformToSDKLogLevel(envOps envOperations, logGetter logLevelGetter) logger.Level {
+	// Exception for unknown log levels are handled in underlying LogLevel() method
```
mgyucht, line 50: Interesting. This probably works fine given the current SDK logging architecture and gets us back to where we were, but I think it's a bit of a hack to try to go around TF logging infra and read from the environment like this. For instance, users can't enable debug logging for just our provider (otherwise, we would also need to check TF_LOG_PROVIDER_DATABRICKS as well).

Another approach would be to define a struct and implement the logging.Logger interface, then configure the SDK to use this logger. This is what is done in the AWS TF provider implementation. They [implement a custom logger](https://github.com/hashicorp/aws-sdk-go-base/blob/main/logging/tf_logger.go) that implements their [Logger interface](https://github.com/hashicorp/aws-sdk-go-base/blob/main/logging/logger.go) and then [configure the client to use that logger at configuration time](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/conns/config.go#L64). The advantage of this approach is that you let TF determine what gets printed to the log (so, for instance, TF could change the log level while the provider was running, or provide different ways for users to specify log level rather than through environment variables).

mgyucht (commented):
While this would solve the immediate need, I don't think it's the right way to solve this problem. Instead, we should implement the Logger interface from the SDK using the `tflog` package and set the SDK logger to use that.

tanmay-db:
Thanks @mgyucht for review, closing this in favour of: https://github.com/databricks/terraform-provider-databricks/pull/2747. This PR is still WIP, need to clean it up and add tests but it is working fine tested manually.

PR #2710, opened 2023-09-21
Add Terraform Provider for DefaultNamespaceSettings

hectorcast-db:
## Changes
Add Terraform Provider for DefaultNamespaceSettings

## Tests
- [X] `make test` run locally
  - `common`, `mws` and `provider` tests are failing, but also fail in master. Needs investigation.
- [x] using Go SDK


catalog/data_metastores.go
```diff
@@ -14,12 +14,12 @@ func DataSourceMetastores() *schema.Resource {
 		Ids map[string]string `json:"ids,omitempty" tf:"computed"`
 	}
 	return common.AccountData(func(ctx context.Context, data *metastoresData, acc *databricks.AccountClient) error {
-		metastores, err := acc.Metastores.List(ctx)
+		metastores, err := acc.Metastores.ListAll(ctx)
```
hectorcast-db, line 17: Just quickly changes this to have the code compile with the new SDK. Not the correct change (test fails). I will fix it.

common/client.go
```diff
@@ -83,27 +83,32 @@ func (c *DatabricksClient) AccountClient() (*databricks.AccountClient, error) {
 
 // Get on path
 func (c *DatabricksClient) Get(ctx context.Context, path string, request any, response any) error {
-	return c.Do(ctx, http.MethodGet, path, request, response, c.addApiPrefix)
+	return c.Do(ctx, http.MethodGet, path, nil, request, response, c.addApiPrefix)
```
hectorcast-db, line 86: SDK adds a new header parameter.

catalog/resource_connection_test.go
```diff
@@ -30,7 +30,6 @@ func TestConnectionsCreate(t *testing.T) {
 					Properties: map[string]string{
 						"purpose": "testing",
 					},
-					Owner: "InitialOwner",
```
hectorcast-db, line 33: Doesn't exist anymore in the SDK. Needs investigation.

provider/provider.go
```diff
@@ -127,6 +128,7 @@ func DatabricksProvider() *schema.Provider {
 			"databricks_mws_storage_configurations":  mws.ResourceMwsStorageConfigurations(),
 			"databricks_mws_vpc_endpoint":            mws.ResourceMwsVpcEndpoint(),
 			"databricks_mws_workspaces":              mws.ResourceMwsWorkspaces(),
+			"databricks_namespace_settings":          settings.ResourceNamespaceSettings(),
```
deryasusm-db, line 131: `default_namespace_setting` instead of `namespace_settings`

settings/resource_settings_test.go
```diff
@@ -0,0 +1,215 @@
...
+					},
+					SettingName: "default",
+				},
+			},
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/settings/types/default_namespace_ws/names/default?etag=etag2",
+				Status:   200,
+				Response: &settings.DefaultNamespaceSetting{
+					Etag: "etag3",
```
deryasusm-db, line 58: This GET would still return etag2, since the PATCH above returns etag2. (GET does not create a new version)

settings/resource_settings_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+func TestQueryRead(t *testing.T) {
+	d, err := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/settings/types/default_namespace_ws/names/default?etag=etag1",
+				Status:   200,
+				Response: &settings.DefaultNamespaceSetting{
+					Etag: "etag2",
```
deryasusm-db, line 91: This GET would still return etag1, since the setting is not patched in this test

settings/resource_settings_test.go
```diff
@@ -0,0 +1,215 @@
...
+					},
+					SettingName: "default",
+				},
+			},
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/settings/types/default_namespace_ws/names/default?etag=etag3",
+				Status:   200,
+				Response: &settings.DefaultNamespaceSetting{
+					Etag: "etag4",
```
deryasusm-db, line 161: This GET would still return etag3 (GET does not create a new version, PATCH and DELETE do)

settings/resource_settings_test.go
```diff
@@ -0,0 +1,215 @@
...
+				Status:   200,
+				Response: &settings.DefaultNamespaceSetting{
+					Etag: "etag2",
+					Namespace: settings.StringMessage{
+						Value: "namespace_value",
+					},
+					SettingName: "default",
+				},
+			},
+			{
```
deryasusm-db, line 203: Delete also returns etag (https://src.dev.databricks.com/databricks/universe/-/blob/settings-policy/api/openapi/settings-policy.openapi.jsonnet?L69-79). Could we update the test?

settings/resource_settings.go
```diff
@@ -0,0 +1,153 @@
...
+	res, err := a.Read(etag)
+	if err != nil {
+		// If settings is not set-up, we will get an error with the etag in it.
+		// This will happen, for instance, during the first ever call to GET.
+		return a.getEtagFromError(err)
+	}
+	return res.Etag, nil
+}
+
+func (a SettingsAPI) getEtagFromError(err error) (string, error) {
```
deryasusm-db, line 37: Just so you know, not every exception will contain the metadata (which contains the etag), only certain 404 cases and the conflict case (409). Let's make sure we're not always anticipating metadata to be present on the error.

common/client.go
```diff
@@ -117,6 +117,11 @@ func (c *DatabricksClient) Patch(ctx context.Context, path string, request any)
 	return c.Do(ctx, http.MethodPatch, path, nil, request, nil, c.addApiPrefix)
 }
 
+// Patch on path
+func (c *DatabricksClient) PatchWithResponse(ctx context.Context, path string, request any, response any) error {
```
deryasusm-db, line 121: To confirm, FieldMask will be sent as part of `request`, right?

settings/resource_settings.go
```diff
@@ -0,0 +1,153 @@
...
+		metadata := errorInfos[0].Metadata
+		if etag, ok := metadata["etag"]; ok {
+			return etag, nil
+		}
+	}
+	return "", fmt.Errorf("error fetching the default workspace namespace settings: %w", err)
+}
+
+func (a SettingsAPI) Create(setting settings.UpdateDefaultWorkspaceNamespaceRequest) (string, error) {
+	return a.Update("", setting)
```
deryasusm-db, line 49: Call this `request`, not `setting`

settings/resource_settings.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+	return a.client.Delete(a.context, "/settings/types/default_namespace_ws/names/default", map[string]string{
+		"etag": etag,
+	})
+}
+
+func (a SettingsAPI) Update(etag string, setting settings.UpdateDefaultWorkspaceNamespaceRequest) (string, error) {
+	etag, err := a.getCurrentEtag(etag)
```
deryasusm-db, line 71: Call it `request` instead of `setting`

settings/resource_settings_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	assert.NoError(t, err)
+
+	assert.Equal(t, "etag2", d.Id())
+	assert.Equal(t, "default", d.Get("setting_name"))
+	res := d.Get("namespace").([]interface{})[0].(map[string]interface{})
+	assert.Equal(t, "namespace_value", res["value"])
+}
+
+func TestQueryUpdate(t *testing.T) {
```
deryasusm-db, line 118: Just for my understanding, TF tests actually seem to be unit tests with mocks, right? I had the impression that these would be e2e tests working with actual endpoints/services.

settings/resource_settings_test.go
hectorcast-db: Updated

settings/resource_settings_test.go
hectorcast-db: In this particular test I am only checking the GET command. The setting could have been changed separately by something else. I changed the rest of the tests which do a "patch + get" just after each other. But I will leave this one as it is.

settings/resource_settings_test.go
hectorcast-db: Udated

settings/resource_settings_test.go
hectorcast-db: Updated

settings/resource_settings.go
hectorcast-db: Right now I am returning an error to the user if that happens. What should I do? As far as I understand, we need the Etag to do the operation, so returning the error seems the only option for me.

common/client.go
hectorcast-db: Yes, you can see it on the tests.

docs/resources/settings.md
alexott, line 1: We need to rename `settings` to `namespace_settings` to match resource name. Same for the source code.

docs/resources/settings.md
```diff
@@ -0,0 +1,30 @@
+---
+subcategory: "Settings"
+---
+
+# databricks_namespace_settings Resource
+
+The `databricks_namespace_settings` resource allows you to set the settings for the default namespace.
```
alexott, line 7: Should we explain what `default namespace` is? At least we need to have a link to the docs.

settings/resource_settings.go
deryasusm-db: 409 (RESOURCE_CONFLICT) will get raised from the API if there are concurrent writes (patch/delete) to the setting. In such a case, the request would need to be retried by using the fresh etag that'll be present in the 409 response. I think we can do it like access mgmt API did (which also uses etags): https://github.com/databricks/terraform-provider-databricks/pull/2371/files#diff-7eef2f7ec147f91e0b6597efd71eeece37d9dbf900e83f03d8f5f00d6011b7e7R116. See `handleConflictAndUpdate`. I suggest we do that in PATCH and DELETE, it's like an additional retry on conflict. @hectorcast-db

settings/resource_default_namespace_settings.go
```diff
@@ -0,0 +1,162 @@
...
+		metadata := errorInfos[0].Metadata
+		if etag, ok := metadata["etag"]; ok {
+			return etag, nil
+		}
+	}
+	return "", fmt.Errorf("error fetching the default workspace namespace settings: %w", err)
+}
+
+func (a DefaultNamespaceSettingsAPI) Create(request settings.UpdateDefaultWorkspaceNamespaceRequest) (string, error) {
+	return a.Update("", request)
```
deryasusm-db, line 47: We don't expose a separate CREATE. Our UPDATE (patch) creates the resource if it does not exist. So the user will not get an error back if they call this CREATE method for a setting that already exists - it will be treated as yet another UPDATE. What is the expectation from TF perspective? By the way, access mgmt's CREATE is also that way - CREATE is treated as an update (https://github.com/databricks/terraform-provider-databricks/pull/2371/files#diff-7eef2f7ec147f91e0b6597efd71eeece37d9dbf900e83f03d8f5f00d6011b7e7R89-R99) @hectorcast-db

settings/resource_default_namespace_settings.go
```diff
@@ -0,0 +1,162 @@
...
+	return DefaultNamespaceSettingsAPI{client, ctx}
+}
+
+// DefaultNamespaceSettingsAPI exposes the Default Namespace Settings API
+type DefaultNamespaceSettingsAPI struct {
+	client  *common.DatabricksClient
+	context context.Context
+}
+
+func (a DefaultNamespaceSettingsAPI) getCurrentEtag(etag string) (string, error) {
```
deryasusm-db, line 25: getLatestEtag to be consistent with Access Mgmt's naming

settings/resource_default_namespace_settings_test.go
```diff
@@ -0,0 +1,215 @@
...
+import (
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/settings"
+	"github.com/databricks/terraform-provider-databricks/qa"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestQueryCreateDefaultNameSetting(t *testing.T) {
```
deryasusm-db, line 12: Can we add an internal/acceptance test that uses the TF template? We can run them against a test shard until we enable the UC settings public API in prod. @hectorcast-db

docs/resources/default_namespace_settings.md
```diff
@@ -0,0 +1,26 @@
+---
+subcategory: "Settings"
+---
+
+# databricks_namespace_settings Resource
+
```
deryasusm-db, line 6: I chatted with @mgyucht on the doc part. Let's put a note here saying that this API is private preview mode currently. We will remove that "private preview mode" note and update this doc to include proper documentation after we enable the UC setting public API in production. @hectorcast-db

alexott: Previously we just released PrPr functionality without public documentation, but included TF docs into PrPr user guide. I think that we need to follow that pattern here as well.

Cc: @mgyucht

settings/resource_default_namespace_settings.go
hectorcast-db: As you can read in this code, the Create call just calls the Update one. I created a separate method because it feels more readable to me. But I can remove it.

settings/resource_settings.go
hectorcast-db: I always get the latest tag before updating. So there should be no conflict.

alexott: Terraform executes things in parallel, so there is always a chance that some other resource may do an update.

hectorcast-db: Updated

settings/resource_default_namespace_settings.go
deryasusm-db: @mgyucht Is this behavior above acceptable for CREATE?

hectorcast-db: I have done some changes, so this comment may not be applicable anymore. 
But what do you mean exactly?

docs/resources/default_namespace_settings.md
mgyucht: I see, thanks for the context @alexott. Happy to follow the existing process, I didn't know what the status quo was here.

settings/resource_default_namespace_settings.go
deryasusm-db: CREATE is based on UPDATE: https://github.com/databricks/terraform-provider-databricks/pull/2710/files#diff-d0b20ee3d8f9ca5a7dd83ed86b87a701fb91c302b9119d54945ecc13a40b696aR123. We do not have a POST RPC in our API, we only have PATCH and it does an UpSert (insert or update). What this means is that, the CREATE method on TF will not really be different from the UPDATE method and it will not fail if you call CREATE on an already existing setting (it will just update the existing entity). Is that the expected behavior? One can argue that CREATE should fail if the resource already exists.

I am looking at http://go/terraform-friendly and I don't see a note on CREATE to return ALREADY_EXISTS. In fact, the statement below hints that CREATE can "recreate" the resource instead of failing, if the resource already exists. So I believe we're fine here.

> An update could be omitted, but the resource is recreated with every change.

settings/resource_default_namespace_settings.go
```diff
@@ -0,0 +1,178 @@
...
+	client  *common.DatabricksClient
+	context context.Context
+}
+
+func (a DefaultNamespaceSettingsAPI) isEtagVersionError(err error) bool {
+	var aerr *apierr.APIError
+	if !errors.As(err, &aerr) {
+		return false
+	}
+	return aerr.StatusCode == http.StatusNotFound || (aerr.StatusCode == http.StatusConflict || aerr.ErrorCode == "RESOURCE_CONFLICT")
```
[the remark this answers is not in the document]
hectorcast-db: We usually checking both for the 409 (conflict) case (see https://github.com/databricks/terraform-provider-databricks/pull/2371/files#diff-7eef2f7ec147f91e0b6597efd71eeece37d9dbf900e83f03d8f5f00d6011b7e7R76).

It may not be necessary, but it should not make a difference.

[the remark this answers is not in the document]
deryasusm-db: > We usually checking both for the 409 (conflict) case (see https://github.com/databricks/terraform-provider-databricks/pull/2371/files#diff-7eef2f7ec147f91e0b6597efd71eeece37d9dbf900e83f03d8f5f00d6011b7e7R76).

That's an AND though - so it should be `(aerr.StatusCode == http.StatusConflict **&&** aerr.ErrorCode == "RESOURCE_CONFLICT")` instead of `(aerr.StatusCode == http.StatusConflict **||** aerr.ErrorCode == "RESOURCE_CONFLICT")`

internal/acceptance/default_namespace_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/service/settings"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAccDefaultNamespaceSetting(t *testing.T) {
```
deryasusm-db, line 12: If Acc is for account, it's not correct - this is a workspace level setting

internal/acceptance/default_namespace_test.go
```diff
@@ -0,0 +1,34 @@
+package acceptance
```
deryasusm-db, line 1: Could we add a test case that deletes an existing setting? The verification is to call GET and get back NOT_FOUND after deletion

docs/resources/settings.md
deryasusm-db: There's no public documentation in Prod yet for UC as default.

internal/acceptance/default_namespace_test.go
hectorcast-db: Acc is for Acceptance. 2 lines below you can see we are targetting a workspace.

internal/acceptance/default_namespace_test.go
hectorcast-db: Added

settings/resource_default_namespace_settings.go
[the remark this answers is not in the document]
hectorcast-db: Updated

settings/resource_default_namespace_settings.go
```diff
@@ -0,0 +1,178 @@
...
+	var setting settings.DefaultNamespaceSetting
+	err := a.client.Get(a.context, "/settings/types/default_namespace_ws/names/default", map[string]string{
+		"etag": etag,
+	}, &setting)
+	return setting, err
+}
+
+var resourceSchema = common.StructToSchema(settings.DefaultNamespaceSetting{},
+	func(s map[string]*schema.Schema) map[string]*schema.Schema {
+		delete(s, "etag")
```
deryasusm-db, line 104: @hectorcast-db Question about these fields: I stumbled upon https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed, it looks like `computed` is used to indicate resource fields that are not user configurable. For all settings, etag can not be created by user - user can only pass in an etag returned from the API. For this setting, the setting name is also not user configurable since it's a singleton setting.

What's the reason of using delete here? Shouldn't we use the computed keyword?

docs/resources/default_namespace_settings.md
```diff
@@ -0,0 +1,26 @@
...
+The `databricks_default_namespace_settings` resource allows you to set the settings for the default namespace.
```
deryasusm-db, line 7: Let's use the below as the explanation @hectorcast-db

          _The `databricks_default_namespace_settings` resource allows you to operate the setting configuration for the default namespace in the Databricks workspace.
          Setting the default catalog for the workspace determines the catalog that is used when queries do not reference
          a fully qualified 3 level name. For example, if the default catalog is set to 'retail_prod' then a query
          'SELECT * FROM myTable' would reference the object 'retail_prod.default.myTable'
          (the schema 'default' is always assumed).
          This setting requires a restart of clusters and SQL warehouses to take effect. Additionally, the default namespace only applies when using Unity Catalog-enabled compute._

hectorcast-db: Updated the docs

common/client.go
```diff
@@ -112,11 +112,21 @@ func (c *DatabricksClient) Delete(ctx context.Context, path string, request any)
 	return c.Do(ctx, http.MethodDelete, path, nil, request, nil, c.addApiPrefix)
 }
 
+// Delete on path
+func (c *DatabricksClient) DeleteWithResponse(ctx context.Context, path string, request any, response any) error {
```
tanmay-db, line 116: This might lead to confusion down the line as to which Delete method we should for the client. It's not immediately obvious if we should use Delete or DeleteWithResponse. It would be good to add a little more context in comments for this and PatchWithResponse

internal/acceptance/default_namespace_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/settings"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAccDefaultNamespaceSetting(t *testing.T) {
+	t.Skip("Automatic test disabled until feature is enabled")
```
tanmay-db, line 15: We should probably create a ticket to track this and put a reminder for when we need to enable it otherwise it would be easy to miss.

internal/acceptance/default_namespace_test.go
```diff
@@ -0,0 +1,71 @@
...
+		`,
+		Check: resourceCheck("databricks_default_namespace_settings.this", func(ctx context.Context, client *common.DatabricksClient, id string) error {
+			ctx = context.WithValue(ctx, common.Api, common.API_2_1)
+			w, err := client.WorkspaceClient()
+			assert.NoError(t, err)
+			res, err := w.Settings.ReadDefaultWorkspaceNamespace(ctx, settings.ReadDefaultWorkspaceNamespaceRequest{
+				Etag: id,
+			})
+			assert.NoError(t, err)
+			assert.Equal(t, res.Namespace.Value, "namespace_value")
```
tanmay-db, line 32: nit: what does this assert check? Is this necessary?

settings/resource_default_namespace_settings.go
```diff
@@ -0,0 +1,178 @@
...
+	if len(errorInfos) > 0 {
+		metadata := errorInfos[0].Metadata
+		if etag, ok := metadata["etag"]; ok {
+			return etag, nil
+		}
+	}
+	return "", fmt.Errorf("error fetching the default workspace namespace settings: %w", err)
+}
+
+func (a DefaultNamespaceSettingsAPI) DeleteWithRetry(etag string) (string, error) {
```
tanmay-db, line 49: should we use: https://github.com/databricks/databricks-sdk-go/blob/main/retries/retries.go?

common/client.go
hectorcast-db: Added comment

internal/acceptance/default_namespace_test.go
hectorcast-db: https://databricks.atlassian.net/browse/DECO-6049

internal/acceptance/default_namespace_test.go
hectorcast-db: Line 31 checks that the resource exists and can be loaded. Line 32 checks the value of the resource (i.e., that it has been created correctly)

docs/resources/default_namespace_settings.md
hectorcast-db: Added the documentation shared by slack.

tanmay-db (approved):
LGTM

docs/resources/default_namespace_settings.md
```diff
@@ -0,0 +1,30 @@
+---
+subcategory: "Settings"
+---
+
+# databricks_namespace_settings Resource
```
alexott, line 5: Can we add a note that it's a workspace-level resource.

docs/resources/default_namespace_settings.md
```diff
@@ -0,0 +1,30 @@
...
+
+The `databricks_default_namespace_settings` resource allows you to operate the setting configuration for the default namespace in the Databricks workspace.
+Setting the default catalog for the workspace determines the catalog that is used when queries do not reference
+a fully qualified 3 level name. For example, if the default catalog is set to 'retail_prod' then a query
+'SELECT * FROM myTable' would reference the object 'retail_prod.default.myTable'
```
alexott, line 10: maybe it's better to use backticks for the SQL query and table names?

Issue #2678, opened 2023-09-12
[ISSUE] Issue with `databricks_permissions` resource when using with a SQL Warehouse endpoint


thaiphv:
### Configuration
```hcl
resource "databricks_sql_endpoint" "this" {
  auto_stop_mins       = var.auto_stop_mins
  cluster_size         = var.cluster_size
  max_num_clusters     = var.max_num_clusters
  name                 = local.endpoint_name
  spot_instance_policy = var.spot_instance_policy
  warehouse_type       = "PRO"
  channel {
    name = "CHANNEL_NAME_CURRENT"
  }
}

resource "databricks_group" "this" {
  display_name = "${local.endpoint_name} Usage"
  provider = databricks.mws
}

resource "databricks_permissions" "this" {
  sql_endpoint_id = databricks_sql_endpoint.this.id

  access_control {
    group_name       = databricks_group.this.display_name
    permission_level = "CAN_USE"
  }
}
```

### Expected Behavior
The `databricks_permissions` resource is created successfully.

### Actual Behavior
Terraform failed to apply with the error:
```
Error: cannot create permissions: PUT requests for warehouse a2325222d375d1cb with no existing owner must provide a new owner.
  with module.shared_small.databricks_permissions.this,
```

If I followed the error message and granted "IS_OWNER" to the service user, Terraform failed with a different error:
```
Error: cannot create permissions: it is not possible to decrease administrative permissions for the current user: <redacted>
  with module.shared_small.databricks_permissions.this,
```

### Steps to Reproduce
1. `terraform apply`-->

### Terraform and provider versions
Terraform: 1.4.6
Databricks provider: 1.25.0

### Debug Output
```
[...]

Issue #2268, opened 2023-05-08
[ISSUE] Issue with `databricks_permissions` resource


daniprado:
### Configuration
```hcl
resource "databricks_sql_global_config" "this" {
  security_policy           = "DATA_ACCESS_CONTROL"
  data_access_config        = {}
  sql_config_params         = {}
  enable_serverless_compute = true
}

resource "databricks_sql_endpoint" "this" {
  for_each = local.databricks_sql_warehouses

  name = each.value.name

  cluster_size              = try(each.value.cluster_size, "Small")
  min_num_clusters          = try(each.value.min_num_clusters, 1)
  max_num_clusters          = try(each.value.max_num_clusters, 1)
  auto_stop_mins            = try(each.value.auto_stop_mins, 10)
  spot_instance_policy      = try(each.value.spot_instance_policy, "COST_OPTIMIZED")
  enable_serverless_compute = try(each.value.enable_serverless_compute, true)
  warehouse_type            = "PRO"

  dynamic "channel" {
    for_each = try(each.value.channel_name, null) != null ? [ each.value.channel_name ] : []

    content {
      name = channel.value
    }
  }

  depends_on = [
    databricks_sql_global_config.this
  ]
}

resource "databricks_permissions" "sql_warehouse" {
  for_each = databricks_sql_endpoint.this

  sql_endpoint_id = each.value.id

  access_control {
    service_principal_name = local.databricks_sql_warehouses[each.key].manager
    permission_level = "CAN_MANAGE"
  }

  dynamic "access_control" {
    for_each = try(local.databricks_sql_warehouses[each.key].users, [])
    content {
      group_name       = access_control.value
      permission_level = "CAN_USE"
    }
  }
}
```

### Expected Behavior
As `IS_OWNER` permission on `access_control` block cannot be provided, it should be completely ignored by the Terraform execution.

### Actual Behavior
Whenever a `terraform apply` is executed, the `IS_OWNER` block is marked for deletion. This does not happen so that the block will appear again next time, causing the resource to be "modified" on every execution.

**UPDATE**: This apparently happens only when the current owner of the SQL Warehouse is not the user executing the Terraform plan/apply.

### Steps to Reproduce
1. Create a `databricks_sql_endpoint`
2. Create a `databricks_permissions` resource attached to it.
3. Execute plan/apply for the same code a second time.

### Terraform and provider versions
```
Terraform v1.4.0
on linux_amd64
[...]

PR #2719, opened 2023-09-25
Fixed `databricks_permissions` resource for correct permissions update.

tanmay-db:
## Changes

Don't auto set `IS_OWNER` permissions but allows user to set it explicitly. Resolves: https://github.com/databricks/terraform-provider-databricks/issues/2678

## Tests

Integration tests passed.

Successful terraform apply on:
```
terraform {
  required_providers {
    databricks = {
      version = "1.26.0"
    }
  }
}

provider "databricks" {
    profile = "aws-prod"
}

  auto_stop_mins       = 10
  cluster_size         = "Medium"
  max_num_clusters     = 3
  name                 = "tanmay-test"
  spot_instance_policy = "ON_DEMAND_ONLY"
  warehouse_type       = "PRO"
  channel {
    name = "CHANNEL_NAME_CURRENT"
  }
}


resource "databricks_group" "this" {
  display_name = "TanmayTest2"
}

  sql_endpoint_id = databricks_sql_endpoint.this.id

  access_control {
    group_name       = databricks_group.this.display_name
    permission_level = "CAN_USE"
  }
}
```

Tested for: https://github.com/databricks/terraform-provider-databricks/issues/2268 as well, TF apply works on:
````
resource "databricks_permissions" "test1" {
  sql_endpoint_id = databricks_sql_endpoint.this.id

  access_control {
    group_name = databricks_group.this.display_name
  }
}

resource "databricks_permissions" "test2" {
  sql_endpoint_id = databricks_sql_endpoint.this.id

  access_control {
    user_name = "<redacted>"
    permission_level = "IS_OWNER"
  }
}
````
- [ ] `make test` run locally
- [ ] using Go SDK

### CODE REVIEW

**Review by alexott — APPROVED** [review]

File: permissions/resource_permissions_test.go [hunk 1]
```diff
@@ -825,8 +826,67 @@ func TestResourcePermissionsCreate_SQLA_Endpoint(t *testing.T) {
...
+
+func TestResourcePermissionsCreate_SQLA_Endpoint_WithOwner(t *testing.T) {
+	d, err := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			me,
+			{
+				Method:   "PUT",
+				Resource: "/api/2.0/permissions/sql/warehouses/abc",
+				ExpectedRequest: AccessControlChangeList{
+					AccessControlList: []AccessControlChange{
```
tanmay-db, line 881: Can ordering of list items be a problem? cc: @nkvuong

alexott: potentially yes, because it's a list

permissions/resource_permissions.go
```diff
@@ -245,15 +245,6 @@ func (a PermissionsAPI) Delete(objectID string) error {
 			UserName:        job.CreatorUserName,
 			PermissionLevel: "IS_OWNER",
 		})
-	} else if strings.HasPrefix(objectID, "/sql/warehouses") {
```
tanmay-db, line 248: Can this be a regression if we are not setting owner automatically?

permissions/resource_permissions_test.go
tanmay-db: Just to check this should always have been a problem right? and not something we are introducing as part of this PR
@alexott

alexott: I think yes

permissions/resource_permissions.go
tanmay-db: We are prioritising https://github.com/databricks/terraform-provider-databricks/issues/2678 since many people are blocked on this. There might be an issue with DBSQL API that we need to followup with. I have verified that setting IS_OWNER works for a username so this would only affect people that want to have owner set automatically.

Even in that case, people won't be unblocked -- they can set the permission in TF explicitly. 

So going ahead with this change would be good.

nkvuong (approved):
thank you for pushing this fix

permissions/resource_permissions_test.go
tanmay-db: Given this is already present, to minimise the changes (additional behaviour change) we are leaving this as is for now.

pietern (approved):
This comment suggests CAN_MANAGE is implicitly added:
https://github.com/databricks/terraform-provider-databricks/issues/2678#issuecomment-1731460835

That doesn't block this fix, but might cause future issues with state drift.

alexott:
Tested myself as well, works as it should be

tanmay-db:
Thank you @alexott

PR #2724, opened 2023-09-26
Add `empty_result_state` attribute to the `databricks_sql_alert` resource

### LINKED ISSUE #2662: [FEATURE] Support default DBSQL alert state when query returns empty results
### Use-cases
Configure default DBSQL alert state when query returns empty rows.

### Proposal
For `databricks_sql_alert` resource, support `empty_result_state` under `option` for default DBSQL alert state when query returns empty rows. This is already supported in Databricks API: https://github.com/databricks/universe/pull/380668.

alexott:
## Changes

This fixes #2662

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/sql/resource_sql_alerts.go
+++ b/sql/resource_sql_alerts.go
@@ -13,12 +13,13 @@ import (
 )
 
 type AlertOptions struct {
-	Column        string `json:"column"`
-	Op            string `json:"op"`
-	Value         string `json:"value"`
-	Muted         bool   `json:"muted,omitempty"`
-	CustomBody    string `json:"custom_body,omitempty"`
-	CustomSubject string `json:"custom_subject,omitempty"`
+	Column           string `json:"column"`
+	Op               string `json:"op"`
+	Value            string `json:"value"`
+	Muted            bool   `json:"muted,omitempty"`
+	CustomBody       string `json:"custom_body,omitempty"`
+	CustomSubject    string `json:"custom_subject,omitempty"`
+	EmptyResultState string `json:"empty_result_state,omitempty"`
 }
 
 type AlertEntity struct {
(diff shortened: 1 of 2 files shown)


sql/resource_sql_alerts.go
```diff
@@ -46,14 +47,18 @@ func (a *AlertEntity) toCreateAlertApiObject(s map[string]*schema.Schema, data *
 		Op:            a.Options.Op,
 		Value:         a.Options.Value,
 	}
-
-	return ca, nil
+	// This is a workaround for Go SDK problem, will be fixed there.
+	var err error
+	if a.Options.EmptyResultState != "" {
+		err = ca.Options.EmptyResultState.Set(a.Options.EmptyResultState)
```
tanmay-db, line 53: We should throw an error if the value is not supported

> Currently supported values are `unknown`, `triggered`, `ok`

sql/resource_sql_alerts.go
```diff
@@ -46,14 +47,18 @@ func (a *AlertEntity) toCreateAlertApiObject(s map[string]*schema.Schema, data *
 		Op:            a.Options.Op,
 		Value:         a.Options.Value,
 	}
-
-	return ca, nil
+	// This is a workaround for Go SDK problem, will be fixed there.
```
tanmay-db, line 50: Which problem is this?

tanmay-db (approved):

sql/resource_sql_alerts.go
alexott: The problem with SDK that it doesn't accept empty string for optional enums

sql/resource_sql_alerts.go
alexott: The check should be done on the backend. The main problem right now is that I can't simply do `ca.Options.EmptyResultState.Set(a.Options.EmptyResultState)` if the value is omitted - SDK returns an error.

sql/resource_sql_alerts.go
alexott: Tracked here: https://github.com/databricks/databricks-sdk-go/issues/630

junwei-db:
@alexott I noticed the documentation for this change is incorrect -- the new `empty_result_state` attribute should be specified under `databricks_sql_alert.options`, not directly under `databricks_sql_alert`. Could you help fix this? Thanks

alexott:
@junwei-db it's just a doc problem, implementation is in the correct place

junwei-db:
Right, that's what I was referring to. Could you help fix the doc?

PR #2725, opened 2023-09-26
Add `dashboard_filters_enabled` attribute to `databricks_sql_dashboard` resource

### LINKED ISSUE #2172: [FEATURE] SQL dashboard filters
Hopefully, I am not just overlooking how to do this - I didn't see anything in the provider docs. I also recognize that the contributors are probably aware of this feature not existing but I didn't see any issues formally requesting this and was also wondering how far off this addition might be.

### Use-cases
- When we edit a dashboard through the Databricks UI, it is possible to create a new dashboard widget that filters one or more queries
- We want to manage our dashboards through terraform in order to version them
- If we want to fully manage our dashboards through terraform, then this is a missing feature which means that a terraform managed dashboard has less functionality than a manually created one

### Attempted Solutions

Not necessarily a solution but here is what I have done up to the point where I am blocked:
- I previously created a dashboard by writing queries and creating the dashboard, widgets, and visualizations all through the Databricks UI
- On my dashboard, I added some filters
image

Then we wanted to keep our dashboard under version control:
- I used the terraform exporter tool to produce the terraform of my dashboard for me
- Everything except the dashboard filters is exported correctly and is useable and the dashboard filters are `databricks_sql_widget` resources without any connections/functionality
```
  resource "databricks_sql_widget" "abc" {
    title = "site"
    position {
      size_y = 3
      size_x = 3
      pos_y  = 64
    }
    dashboard_id = databricks_sql_dashboard.xyz.id
  }
```
image

### Proposal
I'm really not familiar enough with Terraform to propose anything. Sorry about that.

### References
N/A

alexott:
## Changes

When `dashboard_filters_enabled` is set to `true` then filters are shown on the SQL Dashboard level. In this case, the `databricks_sql_widget` should have following block to use it:

```
  parameter {
    type   = "dashboard-level"
    name   = "Limit"
    map_to = "Limit"
  }
```

This fixes #2172

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/sql/api/dashboard.go
+++ b/sql/api/dashboard.go
@@ -4,11 +4,12 @@ import "encoding/json"
 
 // Dashboard ...
 type Dashboard struct {
-	ID        string            `json:"id"`
-	Name      string            `json:"name"`
-	Tags      []string          `json:"tags,omitempty"`
-	Widgets   []json.RawMessage `json:"widgets,omitempty"`
-	Parent    string            `json:"parent,omitempty"`
-	CreatedAt string            `json:"created_at,omitempty"`
-	UpdatedAt string            `json:"updated_at,omitempty"`
+	ID                      string            `json:"id"`
+	Name                    string            `json:"name"`
+	Tags                    []string          `json:"tags,omitempty"`
+	Widgets                 []json.RawMessage `json:"widgets,omitempty"`
+	Parent                  string            `json:"parent,omitempty"`
+	CreatedAt               string            `json:"created_at,omitempty"`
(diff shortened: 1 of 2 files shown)


mgyucht (approved):
Code change LGTM, but we should really migrate this resource to the Go SDK.

mgyucht:
Any chance we can write an integration test for this?

alexott:
> Code change LGTM, but we should really migrate this resource to the Go SDK.

I would really wait until LakeView APIs are available...

PR #2738, opened 2023-09-28
Bump databricks-sdk-go dependency to 0.21.0

hectorcast-db:
## Changes
Bump databricks-sdk-go dependency to 0.21.0

## Tests
- [X] `make test` run locally
- [X] using Go SDK

### DIFF
--- a/catalog/resource_connection.go
+++ b/catalog/resource_connection.go
@@ -39,6 +39,7 @@ var sensitiveOptions = []string{"user", "password", "personalAccessToken", "acce
 func ResourceConnection() *schema.Resource {
 	s := common.StructToSchema(ConnectionInfo{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["owner"].Deprecated = "owner field is deprecated due the Catalog API changes."
 			return m
 		})
 	pi := common.NewPairID("metastore_id", "name").Schema(
--- a/catalog/resource_connection.go
+++ b/catalog/resource_connection.go
@@ -25,8 +25,6 @@ type ConnectionInfo struct {
 	NameArg string `json:"-" url:"-"`
 	// A map of key-value properties attached to the securable.
 	Options map[string]string `json:"options" tf:"sensitive"`
-	// Username of current owner of the connection.
-	Owner string `json:"owner,omitempty" tf:"force_new,suppress_diff"`
 	// An object containing map of key-value properties attached to the
 	// connection.
 	Properties map[string]string `json:"properties,omitempty" tf:"force_new"`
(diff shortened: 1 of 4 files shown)


catalog/resource_connection.go
```diff
@@ -39,6 +37,12 @@ var sensitiveOptions = []string{"user", "password", "personalAccessToken", "acce
 func ResourceConnection() *schema.Resource {
 	s := common.StructToSchema(ConnectionInfo{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["owner"] = &schema.Schema{
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: `owner field is deprecated due the Catalog API changes. Owner cannot be specified when creating a connection.`,
+				Default:    "",
```
alexott, line 44: it's not required, it will be empty string anyway

alexott (approved):
small nit

catalog/resource_connection.go
```diff
@@ -39,6 +37,11 @@ var sensitiveOptions = []string{"user", "password", "personalAccessToken", "acce
 func ResourceConnection() *schema.Resource {
 	s := common.StructToSchema(ConnectionInfo{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["owner"] = &schema.Schema{
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: `owner field is deprecated due the Catalog API changes. Owner cannot be specified when creating a connection.`,
+			}
```
alexott, line 44: @nkvuong pointed that we shouldn't remove Owner - we just need to set it via a separate API call

alexott (changes requested):

catalog/resource_connection.go
hectorcast-db: Added the resource back. As discussed externally, there is currently no API to update the owner.

catalog/resource_connection_test.go
```diff
@@ -30,7 +30,6 @@ func TestConnectionsCreate(t *testing.T) {
 					Properties: map[string]string{
 						"purpose": "testing",
 					},
-					Owner: "InitialOwner",
```
hectorcast-db, line 33: Since the field has been removed from the OpenAPI spec, it won't be send during the Create call, so we update the test. The field will be added to the Update call in the OpenAPI spec, so we are keeping the TF Resource field.

go.sum
```diff
@@ -23,6 +23,8 @@ github.com/cloudflare/circl v1.3.3/go.mod h1:5XYMA4rFBvNIrhs50XuiBJ15vF2pZn4nnUK
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/databricks/databricks-sdk-go v0.20.0 h1:KyVQkvyFYgIlTBjQbazAtW/Y5tZaHy2pF5DRpC35s0s=
```
nkvuong, line 24: could you run `go mod tidy`

nkvuong (approved):

alexott (approved):

PR #2740, opened 2023-09-28
GitHub Actions workflow to compute provider schema diff

pietern:
## Changes

A workflow to compute provider schema diff between two versions of the provider. If triggered manually (through the actions UI), it takes inputs for the base ref and head ref to compare. If triggered from a pull request, it compares the schema on the main branch with the schema in the pull request.

The current behavior doesn't fail the workflow if there is a diff. This means folks need to actively look at the output. We can switch this by removing "continue-on-error: true" on the diff command itself.

## Tests

Example run on my own fork (for the diff between v1.25 and v1.27): https://github.com/pietern/terraform-provider-databricks/actions/runs/6337272638/job/17212054575

### DIFF
--- a/.github/workflows/schema.yml
+++ b/.github/workflows/schema.yml
@@ -1,0 +1,96 @@
+name: schema
+
+on:
+  workflow_dispatch:
+
+jobs:
+  compute_current:
+    name: "Generate current"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


hectorcast-db (approved):
I think it's good to not fail on diff, since that would also block all non-breaking changes.

mgyucht (approved):
Awesome!

mgyucht:
I think you added continue-on-error after this change, so actually this should be good to go.

PR #2741, opened 2023-09-28
Add enabled field for queueing

geekswaroop:
## Changes
With the GA of queueing feature, we now require the enabled field to be set mandatorily if you want queueing as part of the job settings. The openAPI changes have been merged

## Tests

- Updated tests

- [x] `make test` run locally
- [x] using Go SDK


alexott (commented):
The code looks good, but my biggest worry is that this will be a breaking change - because `enabled` is the required attribute, then existing TF code will require fixing.  We can set default value for `enabled` to `true` to mitigate this.

@gaborratky-db WDYT?

gaborratky-db:
Do you know roughly how many customers would be affected by the breaking change? I share your concern, but I also don't want to allow `queue { }` in TF indefinitely and prefer API/TF/DABs/etc. to be consistent. In the API, we are exempting private preview customers for 30 days but then start enforcing it.

For TF, I'd just start requiring `enabled` field. It's a simple fix for the few customers who use `queue { }` from TF today (all private preview customers).

alexott:
I don't have stats, but I think that you can estimate from your telemetry...

tanmay-db (approved):

tanmay-db (approved):
Given this is GA and hence broken for most customers in TF (except private preview which have exemption period of 30 days) going ahead with having enabled as mandatory.

alexott (approved):

Issue #2574, opened 2023-08-09
[FEATURE] Exporter: flexible filtering (inclusing/exclusion) of exported resources


alexott:
Right now we have the `-match` command line parameter that allows to list & export only resources with the matching name. But sometimes we may need an ability to exclude resources, like, notebooks in `/Users/user@domain/.ide/` folder that is used by VSCode extension, AutoML experiments, etc. It would be useful to add an ability to:

* Specify different patterns for different resources
* Specify multiple patterns per resource
* Support for more complex matches, like, regular expressions, etc.

This may require addition of functionality to load such things from the config file, not only specify options in the command-line.

PR #2742, opened 2023-09-28
Export: parallel export of resources

alexott:
## Changes

This change significantly speeds up the exporting of resources. I.e. for a workspace with approximately 5k TF resources export now takes ~4 minutes vs 50+ minutes.

The following changes are made:

- Isolate mutable parts & protect different mutable objects by dedicated mutexes. 
- Actual export is done in parallel by running listing & exporting operations in separate goroutines. Exporting goroutines are allocated by resource type, with multiple routines per type (configurable). Each resource type also has dedicated channels.
- Code generation is the single-threaded operation performed after export is done

Other changes include:

- Fixed panic when reading SQL Alert without Query attribute (happens when the underlying query is deleted)
- added `-noformat` option to skip running `terraform fmt` on the exported data
- skipping auxiliary directories to reduce the number of objects to export: `.ide`, `.bundle`, `__pycache__`. (It will be further expanded by implementing#2574)
- added caching of users’ and service principals’ information to decrease load on the SCIM backend (additional work is required there)
- export all jobs, not only that were run "recently" (It will be further expanded by implementing #2574)

P.S. The code coverage drop could be related to the use of goroutines - the relevant tests didn't change.

TODOs:

- [x] make sure tests are running
- [x] investigate panics when exporting SQL alerts
- [x] validate exported results
- [x] update docs

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/guides/experimental-exporter.md
+++ b/docs/guides/experimental-exporter.md
@@ -5,9 +5,9 @@ page_title: "Experimental resource exporter"
 
 -> **Note** This tooling is experimental and provided as is. It has an evolving interface, which may change or be removed in future versions of the provider.
 
--> **Note** Use the same user who did the exporting to import the exported templates.  Otherwise it could cause changes in the jobs ownership.
+-> **Note** Use the same user who did the exporting to import the exported templates.  Otherwise, it could cause changes in the ownership of the objects.
 
-Generates `*.tf` files for Databricks resources as well as `import.sh` to run import state. Available as part of provider binary. The only possible way to authenticate is through [environment variables](../index.md#Environment-variables). It's best used when you need to quickly export Terraform configuration for an existing Databricks workspace. After generating configuration, we strongly recommend manually review all created files.
+Generates `*.tf` files for Databricks resources as well as `import.sh` that is used to import objects into the Terraform state. Available as part of provider binary. The only possible way to authenticate is through [environment variables](../index.md#Environment-variables). It's best used when you need to quickly export Terraform configuration for an existing Databricks workspace. After generating the configuration, we strongly recommend manually reviewing all created files.
 
 ## Example Usage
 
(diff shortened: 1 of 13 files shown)


docs/guides/experimental-exporter.md
```diff
@@ -79,6 +80,15 @@ Services are just logical groups of resources used for filtering and organizatio
 
 For security reasons, [databricks_secret](../resources/secret.md) cannot contain actual plaintext secrets. Importer will create a variable in `vars.tf`, that would have the same name as secret. You are supposed to [fill in the value of the secret](https://blog.gruntwork.io/a-comprehensive-guide-to-managing-secrets-in-your-terraform-code-1d586955ace1#0e7d) after that.
 
+## Parallel execution
+
+To speedup export, Terraform Exporter performs many operations, such as, listing & actual data exporting, in parallel using Goroutines.  There are built-in defaults controlling the parallelism, but it's also possible to tune some parameters using environment variables specific to exporter:
```
mgyucht, line 85:
```suggestion
To speed up export, Terraform Exporter performs many operations, such as listing & actual data exporting, in parallel using Goroutines.  There are built-in defaults controlling the parallelism, but it's also possible to tune some parameters using environment variables specific to the exporter:
```

docs/guides/experimental-exporter.md
```diff
@@ -79,6 +80,15 @@ Services are just logical groups of resources used for filtering and organizatio
...
+
+* `EXPORTER_WS_LIST_PARALLELISM` (default: `5`) controls how many Goroutines are used to perform parallel listing of Databricks Workspace objects (notebooks, directories, workspace files, ...).
+* `EXPORTER_DIRECTORIES_CHANNEL_SIZE` (default: `100000`) controls capacity of the channel that is used when listing workspace objects. Please make sure that this value is big enough (default value should be ok), otherwise there is a chance of deadlock.
```
mgyucht, line 88: How big is big enough? Does it need to be >= the number of directories in your workspace?

docs/guides/experimental-exporter.md
```diff
@@ -79,6 +80,15 @@ Services are just logical groups of resources used for filtering and organizatio
...
+
+* `EXPORTER_WS_LIST_PARALLELISM` (default: `5`) controls how many Goroutines are used to perform parallel listing of Databricks Workspace objects (notebooks, directories, workspace files, ...).
+* `EXPORTER_DIRECTORIES_CHANNEL_SIZE` (default: `100000`) controls capacity of the channel that is used when listing workspace objects. Please make sure that this value is big enough (default value should be ok), otherwise there is a chance of deadlock. 
+* `EXPORTER_PARALLELISM_NNN` - number of Goroutines used to process resources of specific type (replace `NNN` with resource name, for example, `EXPORTER_PARALLELISM_databricks_notebook=10` sets number of Goroutines for `databricks_notebook` resource to `10`).  Defaults for some resources are defined by the `goroutinesNumber` map in `exporter/context.go`, or equal to `2` if there is no value there.  *Don't increase default values too much to avoid REST API throttling!*
```
mgyucht, line 89: Is `NNN` case-sensitive?

sql/resource_sql_alerts.go
```diff
@@ -72,29 +73,38 @@ func (a *AlertEntity) toEditAlertApiObject(s map[string]*schema.Schema, data *sc
 func (a *AlertEntity) fromAPIObject(apiAlert *sql.Alert, s map[string]*schema.Schema, data *schema.ResourceData) error {
 	a.Name = apiAlert.Name
 	a.Parent = apiAlert.Parent
-	a.QueryId = apiAlert.Query.Id
+	if apiAlert.Query != nil {
+		a.QueryId = apiAlert.Query.Id
+	} else {
+		log.Printf("[WARN] Query object is nil in alert '%s' (id: %s) ", apiAlert.Name, apiAlert.Id)
```
mgyucht, line 79: I think we should use [tflog.Warn](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-log/tflog#example-Warn) for logging in the provider. If we're not calling this from `terraform` directly, we may need to call `SetProviderRootLogger` from github.com/hashicorp/terraform-plugin-log/internal/logging.

sql/resource_sql_alerts.go
```diff
@@ -72,29 +73,38 @@ func (a *AlertEntity) toEditAlertApiObject(s map[string]*schema.Schema, data *sc
...
+			a.Options.Value = value
+		case float64:
+			a.Options.Value = strconv.FormatFloat(value, 'f', 0, 64)
+		case bool:
+			a.Options.Value = strconv.FormatBool(value)
+		default:
+			return fmt.Errorf("unexpected type for value: %T", value)
+		}
+	} else {
+		log.Printf("[WARN] Options object is nil in alert '%s' (id: %s) ", apiAlert.Name, apiAlert.Id)
```
mgyucht, line 106: ditto

sql/resource_sql_alerts.go
```diff
@@ -132,6 +142,7 @@ func ResourceSqlAlert() *schema.Resource {
 			}
 			apiAlert, err := w.Alerts.GetByAlertId(ctx, data.Id())
 			if err != nil {
+				log.Printf("[WARN] error getting alert by ID: %v", err)
```
mgyucht, line 145: ditto

workspace/resource_notebook.go
```diff
@@ -205,10 +216,14 @@ func getEnvAsInt(envName string, defaultValue int) int {
 	return defaultValue
 }
 
-func (a NotebooksAPI) ListParallel(path string, recursive bool) ([]ObjectStatus, error) {
```
mgyucht, line 208: Hm, where was this parameter used before?

exporter/util.go
```diff
@@ -664,9 +708,15 @@ func (ic *importContext) getUpdatedSinceStr() string {
...
-	}
 	return ic.updatedSinceMs
 }
+
+func getEnvAsInt(envName string, defaultValue int) int {
+	if val, exists := os.LookupEnv(envName); exists {
+		parsedVal, err := strconv.Atoi(val)
+		if err == nil {
+			return parsedVal
+		}
```
mgyucht, line 719: Potentially we should log an error if we couldn't parse the env variable. The user intended to set this but we couldn't parse it.

exporter/util.go
```diff
@@ -149,21 +150,42 @@ func (ic *importContext) emitNotebookOrRepo(path string) {
...
-				ic.allDirectories = append(ic.allDirectories, v)
+		ic.wsObjectsMutex.Lock()
+		if len(ic.allDirectories) == 0 {
+			for _, v := range objects {
+				if v.ObjectType == workspace.Directory {
+					ic.allDirectories = append(ic.allDirectories, v)
+				}
 			}
 		}
+		ic.wsObjectsMutex.Unlock()
```
mgyucht, line 161: In general, I think it's preferable to do `defer ic.wsObjectsMutex.Unlock()` earlier up, in case of an early return or panic.

exporter/model.go
```diff
@@ -30,7 +36,42 @@ type resourceApproximation struct {
 }
 
 type stateApproximation struct {
-	Resources []resourceApproximation `json:"resources"`
+	mutex sync.RWMutex
+	// TODO: use map by type -> should speedup Has function?
+	resources []resourceApproximation
```
mgyucht, line 41: How big is the slice in general? Iterating through a list can be faster than a map. May want to benchmark/check out generated assembly for insights here.

exporter/exporter_test.go
```diff
@@ -204,9 +204,10 @@ func TestImportingMounts(t *testing.T) {
 			err := ic.Importables["databricks_mount"].List(ic)
 			assert.NoError(t, err)
 
-			for i := 0; i < len(ic.Scope); i++ {
+			resources := ic.Scope.Sorted()
+			for i := 0; i < len(resources); i++ {
```
mgyucht, line 208: nit:
```suggestion
			for i := range resources {
```

mgyucht (approved):
Overall seems fine. I'm not looking too carefully at the exporter code/parallelism itself, more at the code in the TF provider. Main point is that we should probably use TF native logging primitives in the provider so that our provider only logs when users request it.

docs/guides/experimental-exporter.md
alexott: yes, it's exact name of the resource.

sql/resource_sql_alerts.go
alexott: It works fine, taking into account `TF_LOG`, etc. The biggest pain point right now is not logging REST API payloads

workspace/resource_notebook.go
alexott: Really it wasn't used except for "compatibility" with the non-parallel implementation. But it makes no sense to use parallel implementation for non-recursive listing.

exporter/model.go
alexott: it's proportional to the number of exported objects. In big workspaces it could be easily in range of hundreds of thousands

sql/resource_sql_alerts.go
mgyucht: That issue should be fixed by https://github.com/databricks/terraform-provider-databricks/pull/2747.

Does this print logs even when the user hasn't configured them?

alexott: yes, right now it's the default behavior. I'll file a separate GH Issue to rework logging in exporter.

alexott: We have them everywhere in the provider. I.e., https://github.com/databricks/terraform-provider-databricks/blob/master/clusters/clusters_api.go#L683

mgyucht: Oh I see... then let's leave it as is, and we can add a lint rule at the same time that we actually address this issue.

PR #2747, opened 2023-09-29
Added logger package and fixed issue with API calls not being shown in DEBUG or lower log levels

tanmay-db:
## Changes

Terraform never had logging configured correctly (used to log all levels irrespective of log level specified). The issue (api calls not being shown in `DEBUG` log level) surfaced because in the TF version bump to 1.20.9 which had Go SDK version bump to 0.10.0, the PR: https://github.com/databricks/databricks-sdk-go/pull/426 introduced INFO as basic log level leading to debug level never being enabled even if it is passed through the environment variable. 

Note: `terraform apply` does not show INFO level logs by default. For that `TF_LOG=INFO` needs to be passed which I think is the expected (correct) behaviour.

## Tests

- Tested Manually with using TF_LOG=DEBUG (TF_LOG=debug also works) env which works, not using the env doesn't show logs (correctly), using TF_LOG=TRACE shows right level (TRACE) logs.
- Manually running `TF_LOG=DEBUG terraform apply` shows API calls, 
`````
2023-09-29T14:51:24.653+0200 [DEBUG] provider.terraform-provider-databricks: GET /api/2.1/unity-catalog/catalogs
< HTTP/2.0 200 OK
< {
<   "catalogs": [
<     {
<       "browse_only": false,
<       "catalog_type": "MANAGED_CATALOG",
<       "comment": "",
`````
- Added unit test
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -14,7 +14,7 @@ type TfLogger struct {
 }
 
 // This function is always enabled because TfLogger implements the Logger interface from Go SDK and there we check
-// the logging is enabled based on level (which default to Info). This however isn't required in terraform
+// if the logging is enabled based on level (which default to Info). This however isn't required here
 // since we use the tflog package which automatically reads from TF_LOG environment variable.
 func (tfLogger *TfLogger) Enabled(_ context.Context, _ logger.Level) bool {
 	return true
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -14,8 +14,8 @@ type TfLogger struct {
 }
 
 // This function is always enabled because we extend the Logger from Go SDK and there we check
-// the logging based on level (defaulting to Info). This however isn't required in terraform since we use the tflog package
-// which doesn't require this.
+// the logging is enabled based on level (which default to Info). This however isn't required in terraform
+// since we use the tflog package which automatically reads from TF_LOG environment variable.
 func (tfLogger *TfLogger) Enabled(_ context.Context, _ logger.Level) bool {
 	return true
 }
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -15,7 +15,7 @@ type TfLogger struct {
 // This function is always enabled because TfLogger implements the Logger interface from Go SDK and there we check
 // if the logging is enabled based on level (which default to Info).
 // This however isn't possible here since tflog isn't enabled / disabled based on log level.
-// Ommiting is done internally through the `ShouldOmit` method that filters based on logger configurations.
+// Omitting is done internally through the `ShouldOmit` method that filters based on logger configurations.
 func (tfLogger *TfLogger) Enabled(_ context.Context, _ logger.Level) bool {
 	return true
 }
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -2,6 +2,7 @@ package logger
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 
 	"github.com/databricks/databricks-sdk-go/logger"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: logger/logger.go [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/databricks/databricks-sdk-go/logger"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+type TfLogger struct {
+	Name  string
+	Level logger.Level
+}
+
+func (tfLogger *TfLogger) Enabled(_ context.Context, level logger.Level) bool {
```
mgyucht, line 16: Can you leave a comment here about why this is always enabled?

logger/logger.go
```diff
@@ -0,0 +1,71 @@
...
+	return true
+}
+
+func (tfLogger *TfLogger) Tracef(ctx context.Context, format string, v ...any) {
+	if tfLogger == nil {
+		tflog.Trace(ctx, format, convertToMap(v))
```
mgyucht, line 22: I don't think this is the right way to do this. IIUC, the extra parameters are meant to be interpolated into the provided format string. This will leave the format string unprocessed and log the parameters as extra k-v pairs. Instead, you should use fmt.Sprintf to format the string.

mgyucht (commented):
I think this is the right approach, but we need to fix format strings.

logger/logger.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (tfLogger *TfLogger) Errorf(ctx context.Context, format string, v ...any) {
+	if tfLogger == nil {
+		tflog.Error(ctx, fmt.Sprintf(format, v...), convertToMap(v))
+	} else {
+		tflog.SubsystemError(ctx, tfLogger.Name, fmt.Sprintf(format, v...), convertToMap(v))
+	}
+}
+
+func convertToMap(args ...interface{}) map[string]interface{} {
```
tanmay-db, line 63: Not sure if there is a better way to do this.

tanmay-db: I think it should be fine since we are already formatting the format string and then passing it to tflog. This function mostly is to make method signature valid (accepts `map[string]interface{}` instead of `any` that we use in Go SDK)

logger/logger_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+	"testing"
+
+	goLogger "github.com/databricks/databricks-sdk-go/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTfLogger_Enabled(t *testing.T) {
```
tanmay-db, line 11: @mgyucht Regarding unit test — mocking the tflog will make code a bit complicated (mocking interface) and I think in our earlier discussion we didn’t want to go that way. In Go SDK logging_test we only check if a certain level is enabled (not asserting the actual logged message), 

Please let me know if we should keep things simple as is or go ahead with full unit tests, right now am only checking if logger is enabled by default and default logger is set.

logger/logger.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+type TfLogger struct {
+	Name string
+}
+
+// This function is always enabled because TfLogger implements the Logger interface from Go SDK and there we check
+// if the logging is enabled based on level (which default to Info). This however isn't required here
+// since we use the tflog package which automatically reads from TF_LOG environment variable.
```
mgyucht, line 18: It isn't that this is not required. `tflog` simply does not expose what log level is supported for a logger. If it did, we would delegate to `tflog`.

logger/logger.go
mgyucht: I don't think this is needed at all. Any extra parameters during logging are expected to be inserted into the format string. IMO, you should pass `nil` as the extra argument everywhere. As is, this is not very useful, it's just a list of the arguments to the format string.

logger/logger_test.go
mgyucht: I don't think this is possible to unit test, unless tflog exposes some internal details. I don't think unit tests should be necessary here to be honest. Only other test you may want to have is that the logger is set when the provider is initialized.

mgyucht (commented):
Some nits, but we're getting closer.

mgyucht (approved):

alexott:
@tanmay-db btw, it doesn't work for exporter :-(

PR #2753, opened 2023-10-02
Handle ForceSendFields

hectorcast-db:
## Changes
Populate ForceSendFields for SDK based resources

## Tests

- [X] `make test` run locally
- [ ] N/A covered with integration tests in `internal/acceptance`
- [X] using Go SDK


hectorcast-db:
> I'm not really sure that this change is required - we already can send fields - just don't use `omitempty` for `json`, and add `tf:"optional"` to have field optional in the Terraform. But it doesn't help in the following cases:
> 
> * for disabling serverless SQL - if we send `false` by default, then it will change behavior of the SQL warehouse creation when serverless is enabled on the workspace level - instead of creating serverless by default it will be disabled.
> * single node DLT clusters - if we use defaults, then it will create single-node clusters by default, but DLT by default creates autoscale cluster of 1 to 5 workers.

Discussed offline. In short, ForceSendFields is dynamically filled based on the TF content. We only send the field if it was explicitly added to the TF resource by the user. 
The default is still not being send to avoid the problems described.
This way we overcome the limitation of Go and the `omitempty` flag.

common/reflect_resource.go
```diff
@@ -575,6 +583,34 @@ func readReflectValueFromData(path []string, d attributeGetter,
...
+		return nil
+	}
+
+	if !field.CanSet() || field.Kind() != reflect.Slice {
+		return errors.New("cannot set field")
+	}
+
+	if len(presentFields) == 0 {
+		return nil
+	}
```
mgyucht, line 609: Let's check this first.

mgyucht (approved):
Basically looks good to me. Would love one test case covering the zero value scenario, but otherwise this seems great!

catalog/resource_metastore_test.go
```diff
@@ -123,6 +123,7 @@ func TestCreateMetastore_DeltaSharing(t *testing.T) {
 					DeltaSharingScope: "INTERNAL_AND_EXTERNAL",
 					DeltaSharingRecipientTokenLifetimeInSeconds: 0,
 					DeltaSharingOrganizationName:                "acme",
+					ForceSendFields:                             []string{"DeltaSharingRecipientTokenLifetimeInSeconds"},
```
hectorcast-db, line 126: This test compares the marshalled send data. Since now we are sending DeltaSharingRecipientTokenLifetimeInSeconds, we need to add it to FSF to the expected request.

catalog/resource_metastore_data_access_test.go
```diff
@@ -209,7 +209,7 @@ func TestCreateAccountDacWithAws(t *testing.T) {
 			},
 			{
 				Method:   "GET",
-				Resource: "/api/2.0/accounts/100/metastores/abc/storage-credentials/?",
+				Resource: "/api/2.0/accounts/100/metastores/abc/storage-credentials/bcd?",
```
hectorcast-db, line 212: This is caused by SDK update, not the ForceSendFields

alexott (commented):
it's best to rebase to the `master` that has new Go SDK already

common/reflect_resource.go
```diff
@@ -524,9 +538,17 @@ func DataToReflectValue(d *schema.ResourceData, r *schema.Resource, rv reflect.V
...
-	return iterFields(rv, path, s, func(fieldSchema *schema.Schema,
+	names := getJsonToFieldNameMap(rv.Type())
+	forceSendFields := []string{}
+	err := iterFields(rv, path, s, func(fieldSchema *schema.Schema,
 		path []string, valueField *reflect.Value) error {
 		fieldPath := strings.Join(path, ".")
+		alias := path[len(path)-1]
+		// GetOk returns false for default fields.
+		// Using GetOkExists determine which fields should be added to ForceSendFields.
+		if _, exists := d.GetOkExists(fieldPath); exists && isBasicType(fieldSchema.Type) {
```
hectorcast-db, line 549: GetOkExists is technically deprecated, but there is no replacement and won't be removed until a replacement has been provided

https://github.com/hashicorp/terraform-plugin-sdk/pull/350#issuecomment-597888969

catalog/resource_metastore_test.go
mgyucht: That's awesome. So before, the test wasn't actually checking if this attribute was in the request (as we can't distinguish between absent and set to zero value in Go), but now we can assert that it was sent and set to the zero value.

common/reflect_cache.go
```diff
@@ -0,0 +1,31 @@
...
+var mutex sync.Mutex
+
+var mapCache = map[reflect.Type]map[string]string{}
+
+func getJsonToFieldNameMap(structType reflect.Type) map[string]string {
+	if structType.Kind() != reflect.Ptr && structType.Kind() != reflect.Struct {
+		return nil
+	}
+	mutex.Lock()
+	defer mutex.Unlock()
```
mgyucht, line 17: This is probably fine, but if we want to slightly decrease lock contention here, we can actually release the lock while we're doing reflection and acquire it again only when adding the value to the cache.

common/reflect_resource.go
```diff
@@ -575,6 +597,35 @@ func readReflectValueFromData(path []string, d attributeGetter,
...
+	}
+
+	if rv.Kind() != reflect.Ptr && rv.Kind() != reflect.Struct {
+		return nil
+	}
+
+	field := rv.FieldByName("ForceSendFields")
+
+	if !field.IsValid() {
+		return nil
```
mgyucht, line 618: This is for compatibility with non-SDK structs?

mgyucht (approved):
Overall looks good, just a couple small nits/suggestions.

catalog/resource_metastore_test.go
hectorcast-db: Correct. Before it was just checking that it was "not non-zero".

common/reflect_resource.go
hectorcast-db: That, and also for any struct which does not have the ForceSendFields field. (SDK or not). In such case, `FieldByName` returns the zero(default) Value object, which is not "valid".

PR #2760, opened 2023-10-03
(chore) Make error messages more explicit for `databricks_metastore_data_access`

nfx:


common/client.go
```diff
@@ -87,13 +87,21 @@ func (c *DatabricksClient) AccountOrWorkspaceRequest(accCallback func(*databrick
...
+		return nil
 	} else {
 		ws, err := c.WorkspaceClient()
 		if err != nil {
 			return err
 		}
-		return wsCallback(ws)
+		err = wsCallback(ws)
+		if err != nil {
+			return fmt.Errorf("workspace: %w", err)
```
tanmay-db, line 102: we would need to fix the tests accordingly for these as well.

alexott:
@nkvuong should we improve this?

PR #2765, opened 2023-10-05
Fix: don't rely on having `@` to check if it's user or SP

alexott:
## Changes

There are some installations where the user name isn't equal to email - for them, our detection of username vs SP ID is broken, and leads to some problems, like, the generation of erroneous `run_as` blocks in job definitions, etc.

This PR fixes this by validating a given string as UUID, and if it matches, then uses it as SP ID, and the rest as user name.  Changes affect the following:

* `databricks_job` API - besides validation, it also doesn't fill the `RunAs` structure on read when creator user name == run as user name.
* `databricks_current_user` data source - should fix generation of `acl_principal_id`

There will be changes in the exporter as well, but they will be done later, as part of the SCIM refactoring that needs to be done anyway...

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/data_job_test.go
+++ b/jobs/data_job_test.go
@@ -84,6 +84,97 @@ func TestDataSourceQueryableJobMatchesId(t *testing.T) {
 	})
 }
 
+func TestDataSourceQueryableJobRunAsSP(t *testing.T) {
+	spID := "3f670caf-9a4b-4479-8143-1a0878da8f57"
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/jobs/get?job_id=234",
+				Response: Job{
+					JobID: 234,
+					Settings: &JobSettings{
+						Name: "Second",
+					},
+					CreatorUserName: "user@domain.com",
+					RunAsUserName:   spID,
+				},
+			},
+		},
+		Resource:    DataSourceJob(),
+		Read:        true,
+		New:         true,
+		NonWritable: true,
+		HCL:         `job_id = "234"`,
+		ID:          "234",
+	}.ApplyAndExpectData(t, map[string]any{
+		"job_id":                         "234",
+		"id":                             "234",
+		"job_settings.0.settings.0.name": "Second",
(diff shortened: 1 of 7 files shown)


dvkuznietsov (approved):
lgtm job

jobs/data_job_test.go
```diff
@@ -84,6 +84,97 @@ func TestDataSourceQueryableJobMatchesId(t *testing.T) {
...
+		ID:          "234",
+	}.ApplyAndExpectData(t, map[string]any{
+		"job_id":                             "234",
+		"id":                                 "234",
+		"job_settings.0.settings.0.name":     "Second",
+		"job_settings.0.settings.0.run_as.0": map[string]any{},
+	})
+}
+
+func TestDataSourceQueryableJobRunAsAnoterUser(t *testing.T) {
```
mgyucht, line 148:
```suggestion
func TestDataSourceQueryableJobRunAsAnotherUser(t *testing.T) {
```

jobs/resource_job.go
```diff
@@ -572,16 +572,14 @@ func (a JobsAPI) Read(id string) (job Job, err error) {
 		job.Settings.sortWebhooksByID()
 	}
 
-	if job.RunAsUserName != "" && job.Settings != nil {
-		userNameIsEmail := strings.Contains(job.RunAsUserName, "@")
-
-		if userNameIsEmail {
+	if job.Settings != nil && job.RunAsUserName != "" && job.RunAsUserName != job.CreatorUserName {
```
mgyucht, line 575: Why the addition of `job.RunAsUserName != job.CreatorUserName`? I see that it is happening but don't quite understand why.

mgyucht (commented):
Seems OK to me, just want to understand the extra change in the Job resource.

jobs/resource_job.go
alexott: Right now `job.RunAsUserName` is populated every time, even without `run_as` specified. As a result, we generate not required blocks in the exported code.

gaborratky-db (commented):
defer to @dvkuznietsov

mgyucht (approved):
LGTM, now that we have an integration test.

PR #2768, opened 2023-10-05
Add MLflow models to the exporter

michael-berk:
## Changes
* Add MLflow model exporter functionality
* Add an exporter test

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


mberk06:
Still to do
* Resolve empty resource export
* Update docs

exporter/importables.go
```diff
@@ -1971,4 +1971,46 @@ var resourcesMap map[string]importable = map[string]importable{
 			// {Path: "http_url_spec.authorization", Variable: true},
 		},
 	},
+	"databricks_mlflow_model": {
+		Service: "mlflow-models", // why model instead of mlflow
+		Name: func(ic *importContext, d *schema.ResourceData) string {
+			return d.Id()
```
alexott, line 1977: Let's change this to `return d.Id() + "_" + d.Get("registered_model_id").(string)` because there could be different models with similar names, but having `_` instead of `-`, etc.

alexott (changes requested):
let's add docs, dependencies, etc. This should be rebased on master after #2853 is merged
Also add dependency between model serving & this resource.

mberk06:
Blocked by #2853

alexott:
@michael-berk #2853 is merged, please test

tanmay-db:
@michael-berk any update on this? We have a release planned for today. cc: @mgyucht

michael-berk:
> @michael-berk any update on this? We have a release planned for today. cc: @mgyucht

I need to do some more integration testing. Will mark ready for review once complete. Thanks

michael-berk:
@alexott regarding the dependency with model serving, we need `model_version` to host a model for inference. It actually holds the URI to the model artifact.

We don't have [model_version](https://mlflow.org/docs/latest/rest-api.html#create-modelversion) resource or even a `model_version` field in the the model resource itself, so I think the dependency mapping is not possible. Should we push this change?

PR #2771, opened 2023-10-06
Added `databricks_registered_model` resource

arpitjasa-db:
## Changes

This PR adds [Models in UC](https://docs.databricks.com/en/mlflow/models-in-uc.html) as a Terraform resource to the [Databricks Terraform Provider](https://registry.terraform.io/providers/databricks/databricks/latest), named in a future-proof way and to disambiguate from the soon-to-be deprecated [workspace MLflow models](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/mlflow_model).

## Tests

- [X] `make test` run locally
- [X] using Go SDK

Also tested manually using the `make install`-version of the Terraform provider that includes these changes and the Dogfood staging workspace.
Screenshot 2023-10-06 at 10 54 46 AM
Screenshot 2023-10-06 at 11 55 21 AM
Screenshot 2023-10-06 at 11 52 26 AM

### DIFF
--- a/catalog/resource_registered_model.go
+++ b/catalog/resource_registered_model.go
@@ -52,11 +52,9 @@ func ResourceRegisteredModel() *schema.Resource {
 			if err != nil {
 				return err
 			}
-			var m catalog.CreateRegisteredModelRequest
 			var u catalog.UpdateRegisteredModelRequest
-			common.DataToStructPointer(d, s, &m)
+			common.DataToStructPointer(d, s, &u)
 			u.FullName = d.Id()
-			u.Comment = m.Comment
 			_, err = w.RegisteredModels.Update(ctx, u)
 			return err
 		},
--- a/catalog/resource_registered_model.go
+++ b/catalog/resource_registered_model.go
@@ -45,11 +45,7 @@ func ResourceRegisteredModel() *schema.Resource {
 			if err != nil {
 				return err
 			}
-			err = common.StructToData(*model, s, d)
-			if err != nil {
-				return err
-			}
-			return nil
+			return common.StructToData(*model, s, d)
 		},
 		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			w, err := c.WorkspaceClient()
--- a/catalog/resource_registered_model.go
+++ b/catalog/resource_registered_model.go
@@ -1,0 +1,78 @@
+package catalog
+
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func ResourceRegisteredModel() *schema.Resource {
+	s := common.StructToSchema(
+		catalog.CreateRegisteredModelRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["name"].ForceNew = true
+			m["catalog_name"].ForceNew = true
+			m["schema_name"].ForceNew = true
+			m["storage_location"].ForceNew = true
+			m["storage_location"].Computed = true
+
+			return m
+		})
+
+	return common.Resource{
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var m catalog.CreateRegisteredModelRequest
+			common.DataToStructPointer(d, s, &m)
+			model, err := w.RegisteredModels.Create(ctx, m)
+			if err != nil {
+				return err
+			}
+			d.SetId(model.FullName)
+			return nil
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			model, err := w.RegisteredModels.GetByFullName(ctx, d.Id())
+			if err != nil {
+				return err
+			}
+			err = common.StructToData(*model, s, d)
+			if err != nil {
+				return err
+			}
+			return nil
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var m catalog.CreateRegisteredModelRequest
+			var u catalog.UpdateRegisteredModelRequest
+			common.DataToStructPointer(d, s, &m)
+			u.FullName = d.Id()
+			u.Comment = m.Comment
+			_, err = w.RegisteredModels.Update(ctx, u)
+			return err
+		},
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			return w.RegisteredModels.DeleteByFullName(ctx, d.Id())
+		},
+		StateUpgraders: []schema.StateUpgrader{},
+		Schema:         s,
+		SchemaVersion:  0,
+	}.ToResource()
+}
--- a/catalog/resource_registered_model_test.go
+++ b/catalog/resource_registered_model_test.go
@@ -1,0 +1,228 @@
+package catalog
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestRegisteredModelCornerCases(t *testing.T) {
+	qa.ResourceCornerCases(t, ResourceRegisteredModel())
+}
+
+func TestRegisteredModelCreate(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   http.MethodPost,
+				Resource: "/api/2.1/unity-catalog/models",
+				ExpectedRequest: catalog.CreateRegisteredModelRequest{
+					Name:        "model",
+					CatalogName: "catalog",
+					SchemaName:  "schema",
+					Comment:     "comment",
+				},
+				Response: catalog.RegisteredModelInfo{
+					Name:        "model",
+					CatalogName: "catalog",
+					SchemaName:  "schema",
+					FullName:    "catalog.schema.model",
+					Comment:     "comment",
+				},
+			},
+			{
+				Method:   http.MethodGet,
+				Resource: "/api/2.1/unity-catalog/models/catalog.schema.model?",
+				Response: catalog.RegisteredModelInfo{
+					Name:        "model",
+					CatalogName: "catalog",
+					SchemaName:  "schema",
+					FullName:    "catalog.schema.model",
+					Comment:     "comment",
+				},
+			},
+		},
+		Resource: ResourceRegisteredModel(),
+		HCL: `
+			name = "model"
+			catalog_name = "catalog"
+			schema_name = "schema"
+			comment = "comment"
+			`,
+		Create: true,
+	}.ApplyNoError(t)
+}
+
+func TestRegisteredModelCreate_Error(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   http.MethodPost,
+				Resource: "/api/2.1/unity-catalog/models",
+				Response: apierr.APIErrorBody{
+					ErrorCode: "INVALID_REQUEST",
+					Message:   "Internal error happened",
+				},
+				Status: 400,
+			},
+		},
+		Resource: ResourceRegisteredModel(),
+		Create:   true,
+	}.ExpectError(t, "Internal error happened")
+}
+
+func TestRegisteredModelRead(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   http.MethodGet,
+				Resource: "/api/2.1/unity-catalog/models/catalog.schema.model?",
+				Response: catalog.RegisteredModelInfo{
+					Name:        "model",
+					CatalogName: "catalog",
+					SchemaName:  "schema",
+					FullName:    "catalog.schema.model",
+					Comment:     "comment",
+				},
+			},
+		},
+		Resource: ResourceRegisteredModel(),
+		Read:     true,
+		ID:       "catalog.schema.model",
+	}.ApplyNoError(t)
+}
+
+func TestRegisteredModelRead_Error(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   http.MethodGet,
+				Resource: "/api/2.1/unity-catalog/models/catalog.schema.model?",
+				Response: apierr.APIErrorBody{
+					ErrorCode: "INVALID_REQUEST",
+					Message:   "Internal error happened",
+				},
+				Status: 400,
+			},
+		},
+		Resource: ResourceRegisteredModel(),
+		Read:     true,
+		ID:       "catalog.schema.model",
+	}.ExpectError(t, "Internal error happened")
+}
+
+func TestRegisteredModelUpdate(t *testing.T) {
+	qa.ResourceFixture{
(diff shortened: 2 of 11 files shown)


docs/resources/grants.md
```diff
@@ -194,6 +194,24 @@ resource "databricks_grants" "volume" {
...
+  grant {
+    principal  = "Data Engineers"
+    privileges = ["APPLY_TAG", "EXECUTE"]
+  }
+  grant {
+    principal  = "Data Analysts"
+    privileges = ["EXECUTE"]
+  }
+}
+```
```
alexott, line 213: Why it's called `function`? Won't be people confused by it?

docs/index.md
```diff
@@ -52,10 +52,12 @@ Databricks SQL
...
 * Provide [global configuration for all SQL warehouses](docs/resources/sql_global_config.md)
 
-MLFlow
+Machine Learning
 
-* Create [MLFlow models](resources/mlflow_model.md).
-* Create [MLFlow experiments](resources/mlflow_experiment.md).
+* Create [models in Unity Catalog](resources/registered_model.md).
+* Create [MLflow experiments](resources/mlflow_experiment.md).
+* Create [workspace models](resources/mlflow_model.md).
```
smurching, line 59:
```suggestion
* Create [models in the workspace model registry](resources/mlflow_model.md).
```

To match docs

catalog/resource_registered_model.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+func ResourceRegisteredModel() *schema.Resource {
+	s := common.StructToSchema(
+		catalog.CreateRegisteredModelRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["name"].ForceNew = true
+			m["catalog_name"].ForceNew = true
+			m["schema_name"].ForceNew = true
+			m["storage_location"].ForceNew = true
```
alexott, line 18: If it's computed, we may not need `ForceNew`.

docs/resources/grants.md
```diff
@@ -194,6 +194,24 @@ resource "databricks_grants" "volume" {
 }
 ```
 
+## Function grants
```
smurching, line 197:
```suggestion
## Registered model grants
```

docs/resources/grants.md
```diff
@@ -194,6 +194,24 @@ resource "databricks_grants" "volume" {
...
+
+You can grant `ALL_PRIVILEGES`, `APPLY_TAG`, and `EXECUTE` privileges to [_`catalog.schema.function`_](registered_model.md) specified in the `function` attribute.
```
smurching, line 199:
```suggestion
You can grant `ALL_PRIVILEGES`, `APPLY_TAG`, and `EXECUTE` privileges to [_`catalog.schema.model`_](registered_model.md) specified in the `function` attribute.
```

See views/tables for a similar example (a view is a kind of table)

docs/resources/grants.md
smurching: I think this is because the underlying securable_type (which is the field used to construct the [UC permissions REST API](https://docs.databricks.com/api/workspace/grants/update) path) is `function`, but I don't see any downside to calling the resource `model` here. Similarly in SQL we'll support things like `GRANT EXECUTE ON MODEL mycatalog.myschema.mymodel`.

catalog/resource_registered_model.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func ResourceRegisteredModel() *schema.Resource {
+	s := common.StructToSchema(
+		catalog.CreateRegisteredModelRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
```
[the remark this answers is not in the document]
smurching: Oh I see we document that `comment` is supported - does that work if it's not part of the Schema defined here?

docs/resources/grants.md
arpitjasa-db: Unfortunately the underlying [docs](https://docs.databricks.com/api/workspace/grants/get)/Go SDK don't have `model` in the enum because as you said it's not a securable type, so we get the failure seen below.
Screenshot 2023-10-06 at 10 55 05 AM

We could create a custom mapping from `model` to `function`, but that might be weird/complicate the code. I don't think this is that odd considering our [docs page](https://docs.databricks.com/api/workspace/registeredmodels) also has a note to use `function` for the securable type.

```
Note: The securable type for models is "FUNCTION". When using REST APIs (e.g. tagging, grants) that specify a securable type, use "FUNCTION" as the securable type.
```

catalog/resource_registered_model.go
arpitjasa-db: Yeah, but if someone manually sets this field after it's been computed, it still tries to modify the field in-place instead of replacing the whole resource. That's why I added both attributes

catalog/resource_registered_model.go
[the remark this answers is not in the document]
arpitjasa-db: Yup it works since it's part of `catalog.CreateRegisteredModelRequest`! The schema attributes you see here simply add to the schema defined there. You can see examples of `comment` working in the tests

catalog/resource_registered_model.go
```diff
@@ -0,0 +1,78 @@
...
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var m catalog.CreateRegisteredModelRequest
+			var u catalog.UpdateRegisteredModelRequest
+			common.DataToStructPointer(d, s, &m)
+			u.FullName = d.Id()
+			u.Comment = m.Comment
+			_, err = w.RegisteredModels.Update(ctx, u)
```
nkvuong, line 64: does this mean only comment is updateable?

catalog/resource_registered_model_test.go
```diff
@@ -0,0 +1,228 @@
...
+					SchemaName:  "schema",
+					FullName:    "catalog.schema.model",
+					Comment:     "comment",
+				},
+			},
+		},
+		Resource: ResourceRegisteredModel(),
+		Read:     true,
+		ID:       "catalog.schema.model",
+	}.ApplyNoError(t)
```
nkvuong, line 95: should use `ApplyAndExpectData` to verify that the data returned is correct

internal/acceptance/registered_model_test.go
```diff
@@ -0,0 +1,43 @@
...
+		step{
+			Template: fmt.Sprintf(`
+			resource "databricks_registered_model" "model" {
+				name = "%[1]s"
+				catalog_name = "main"
+				schema_name = "default"
+			}
+			
+			resource "databricks_grants" "model_grants" {
+				function = "main.default.%[1]s"
```
nkvuong, line 23:
```suggestion
				function = databricks_registered_model.model.id
```

docs/resources/grants.md
nkvuong: we already had to build custom mapping for a few securables, so it shouldn't be too much work - https://github.com/databricks/terraform-provider-databricks/blob/master/catalog/resource_grants.go#L101

catalog/resource_registered_model.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+			model, err := w.RegisteredModels.GetByFullName(ctx, d.Id())
+			if err != nil {
+				return err
+			}
+			err = common.StructToData(*model, s, d)
+			if err != nil {
+				return err
+			}
+			return nil
```
nkvuong, line 52:
```suggestion
			return common.StructToData(*model, s, d)
```

nkvuong (changes requested):
couple of suggested changes

catalog/resource_registered_model.go
arpitjasa-db: Yes exactly

docs/resources/grants.md
arpitjasa-db: @nkvuong that's perfect! Added `model` there and it seems to work now

catalog/resource_registered_model.go
nkvuong: so would it make sense to do the below, and don't use the `m` variable
```suggestion
			common.DataToStructPointer(d, s, &u)
			u.FullName = d.Id()
			_, err = w.RegisteredModels.Update(ctx, u)
```

arpitjasa-db: Yup it does, updated the code

smurching (approved):
LGTM, thanks @arpitjasa-db !

vladimirk-db (approved):

catalog/resource_registered_model.go
```diff
@@ -0,0 +1,72 @@
...
+				return err
+			}
+			return common.StructToData(*model, s, d)
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var u catalog.UpdateRegisteredModelRequest
```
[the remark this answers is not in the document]
arpitjasa-db: I don't think we should allow modifying `catalog` OWNER from the `model` resource. I think you meant owner of the `model` which I don't think we need to support at this time. We can loop back into adding this when we add tag support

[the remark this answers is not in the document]
shreyas-goenka: Ah yes, the owner of the model. Sounds good, support for owner can be a followup.

internal/acceptance/registered_model_test.go
```diff
@@ -0,0 +1,43 @@
...
+			
+		`, name),
+		},
+		step{
+			Template: fmt.Sprintf(`
+			resource "databricks_registered_model" "model" {
+				name = "%[1]s"
+				catalog_name = "main"
+				schema_name = "default"
+				comment = "new comment"
```
tanmay-db, line 38: Can you add another test here which updates the `name`. This would be good because name is part of the full_name that is used in the request https://docs.databricks.com/api/workspace/registeredmodels/update.

internal/acceptance/registered_model_test.go
```diff
@@ -0,0 +1,43 @@
+package acceptance
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+)
+
+func TestAccRegisteredModel(t *testing.T) {
```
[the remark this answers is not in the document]
arpitjasa-db: Updated!

internal/acceptance/registered_model_test.go
```diff
@@ -0,0 +1,43 @@
...
+	name := fmt.Sprintf("terraform-test-registered-model-%[1]s",
+		acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))
+	workspaceLevel(t,
+		step{
+			Template: fmt.Sprintf(`
+			resource "databricks_registered_model" "model" {
+				name = "%[1]s"
```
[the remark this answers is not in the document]
arpitjasa-db: Updated!

internal/acceptance/registered_model_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+)
+
+func TestAccRegisteredModel(t *testing.T) {
+	name := fmt.Sprintf("terraform-test-registered-model-%[1]s",
+		acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))
+	workspaceLevel(t,
```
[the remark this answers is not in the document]
arpitjasa-db: Updated!

docs/resources/model_serving.md
```diff
@@ -103,9 +103,10 @@ $ terraform import databricks_model_serving.this <model-serving-endpoint-name>
 
 The following resources are often used in the same context:
 
+* [databricks_registered_model](registered_model.md) to create [Models in Unity Catalog](https://docs.databricks.com/en/mlflow/models-in-uc.html) in Databricks.
 * [End to end workspace management](../guides/workspace-management.md) guide.
 * [databricks_directory](directory.md) to manage directories in [Databricks Workspace](https://docs.databricks.com/workspace/workspace-objects.html).
-* [databricks_mlflow_model](mlflow_model.md) to create [MLflow models](https://docs.databricks.com/applications/mlflow/models.html) in Databricks.
+* [databricks_mlflow_model](mlflow_model.md) to create models in the [workspace model registry](https://docs.databricks.com/en/mlflow/model-registry.html) in Databricks.
```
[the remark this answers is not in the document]
arpitjasa-db: Added the note!

internal/acceptance/registered_model_test.go
arpitjasa-db: Added the test

nkvuong (approved):

internal/acceptance/registered_model_test.go
```diff
@@ -0,0 +1,48 @@
...
+			}
+			
+			resource "databricks_grants" "model_grants" {
+				model = databricks_registered_model.model.id
+			  
+				grant {
+				  principal = "account users"
+				  privileges = ["EXECUTE"]
+				}
+
```
nkvuong, line 24:
```suggestion
			}
```

nkvuong:
integration tests passed (failed tests are for `databricks_sql_query`)

tanmay-db (approved):
LGTM, comments have also been addressed.

docs/resources/mlflow_model.md
```diff
@@ -5,6 +5,8 @@ subcategory: "MLflow"
 
 This resource allows you to create [MLflow models](https://docs.databricks.com/applications/mlflow/models.html) in Databricks.
 
+**Note** This documentation covers the Workspace Model Registry. Databricks recommends using [Models in Unity Catalog](registered_model.md). Models in Unity Catalog provides centralized model governance, cross-workspace access, lineage, and deployment. Workspace Model Registry will be deprecated in the future.
```
pietern, line 8: The deprecation notice can be left out if we don't know when this will happen (or ever).

pietern (commented):
Non-blocking comment.

Note that the `FullName` field is a read-only field only present on the create response and as parameter to the other API calls, so it doesn't show up in the resource schema (only as ID).

I looked at the schema diff and this is the result:
```json
{
  "block": {
    "attributes": {
      "catalog_name": {
        "description_kind": "plain",
        "required": true,
        "type": "string"
      },
      "comment": {
        "description_kind": "plain",
        "optional": true,
        "type": "string"
      },
      "id": {
        "computed": true,
        "description_kind": "plain",
        "optional": true,
        "type": "string"
      },
      "name": {
        "description_kind": "plain",
        "required": true,
        "type": "string"
      },
      "schema_name": {
        "description_kind": "plain",
        "required": true,
        "type": "string"
      },
      "storage_location": {
        "computed": true,
        "description_kind": "plain",
        "optional": true,
        "type": "string"
      }
    },
    "description_kind": "plain"
  },
  "version": 0
}
```

catalog/resource_registered_model.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func ResourceRegisteredModel() *schema.Resource {
+	s := common.StructToSchema(
+		catalog.CreateRegisteredModelRequest{},
```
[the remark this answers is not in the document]
mgyucht: Another note: `owner` is not part of the create structure but is part of the update, so users won't be able to set an owner for the model. We can revisit this later if need be.

docs/resources/registered_model.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the registered model.
+* `catalog_name` - (Required) The name of the catalog where the schema and the registered model reside.
+* `schema_name` - (Required) The name of the schema where the registered model resides.
+* `comment` - The comment attached to the registered model.
```
mgyucht, line 25: Do you not want to document storage_location? Can users set this explicitly?

mgyucht (approved):
To unblock the release of this resource, I'll approve. I think we can expand the set of fields accepted by the input and those provided for users to read after creation as a follow-up.

docs/resources/registered_model.md
mgyucht: For now, I will merge as is, and we can add more documentation in a follow-up.

PR #2773, opened 2023-10-06
Exporter: Follow recommendations of identity team on decreasing load on SCIM API

alexott:
## Changes

Reimplement search for users & service principals to decrease a load on the SCIM backend

## Tests

- [x] `make test` run locally
- [x] using Go SDK


mgyucht (commented):
Seems fine, but I see that databricks_service_principal_role is not covered in an integration test. Can you add a simple one to internal/acceptance/service_principal_test.go?

scim/resource_service_principal.go
```diff
@@ -63,8 +68,8 @@ func (a ServicePrincipalsAPI) Patch(servicePrincipalID string, r patchRequest) e
 }
 
 // Update replaces resource-friendly-entity
-func (a ServicePrincipalsAPI) Update(servicePrincipalID string, updateRequest User) error {
-	servicePrincipal, err := a.Read(servicePrincipalID)
+func (a ServicePrincipalsAPI) Update(servicePrincipalID string, attributes string, updateRequest User) error {
+	servicePrincipal, err := a.Read(servicePrincipalID, attributes)
```
mgyucht, line 72: This method depends on the `groups` response, but that isn't part of `userAttributes`. Can you try the following (or whatever attribute should be present to get group membership info back)?
```suggestion
func (a ServicePrincipalsAPI) Update(servicePrincipalID string, updateRequest User) error {
	servicePrincipal, err := a.Read(servicePrincipalID, "groups")
```

mgyucht (commented):
I think there is a small bug in the SP Update method, PTAL.

alexott:
@mgyucht I've added an integration test, but I can't find any environment that exposes `TEST_EC2_INSTANCE_PROFILE` (we have a lot of other tests skipped because of this)

alexott:
The new `TestAccServicePrincipalResourceOnAwsWithRole` test is passing...

scim/resource_service_principal.go
```diff
@@ -33,8 +34,12 @@ func (a ServicePrincipalsAPI) Create(rsp User) (sp User, err error) {
...
 }
 
-func (a ServicePrincipalsAPI) Read(servicePrincipalID string) (sp User, err error) {
-	servicePrincipalPath := fmt.Sprintf("/preview/scim/v2/ServicePrincipals/%v", servicePrincipalID)
+func (a ServicePrincipalsAPI) Read(servicePrincipalID string, attributes string) (sp User, err error) {
+	attrs := ""
+	if attributes != "" {
+		attrs = "?attributes=" + attributes
+	}
+	servicePrincipalPath := fmt.Sprintf("/preview/scim/v2/ServicePrincipals/%v%s", servicePrincipalID, attrs)
```
tanmay-db, line 42: @alexott how does this reduce the load? Is there backend logic based on attributes?

exporter/util.go
```diff
@@ -284,6 +309,33 @@ const (
...
+			users, err = ic.accountClient.Users.ListAll(ic.Context, iam.ListAccountUsersRequest{
+				Attributes: "userName,id",
+			})
+		} else {
+			users, err = ic.workspaceClient.Users.ListAll(ic.Context, iam.ListUsersRequest{
+				Attributes: "userName,id",
+			})
+		}
+		if err != nil {
+			log.Printf("[WARN] can't fetch list of users")
```
tanmay-db, line 329: This should be [ERROR] since below we throw:

> err = fmt.Errorf("there is no user '%s'", name)

exporter/util.go
```diff
@@ -292,30 +344,61 @@ func (ic *importContext) findUserByName(name string) (u scim.User, err error) {
...
+			sps, err = ic.accountClient.ServicePrincipals.ListAll(ic.Context, iam.ListAccountServicePrincipalsRequest{
+				Attributes: "id,userName",
+			})
+		} else {
+			sps, err = ic.workspaceClient.ServicePrincipals.ListAll(ic.Context, iam.ListServicePrincipalsRequest{
+				Attributes: "id,userName",
+			})
+		}
+		if err != nil {
+			log.Printf("[WARN] can't fetch list of service principals")
```
tanmay-db, line 393: Same

scim/resource_service_principal.go
alexott: Yes, some attributes, like, `roles` are requiring multiple additional queries against multiple tables, and it's proportional to the number of roles attached. 

But the biggest issue was in the listing of service principals - it was always trying to pull all attributes for all service principals, and this call was timed out. Now we're doing the listing by fetching only application ID & SP ID and then fetching individual instances one by one.

tanmay-db (approved):
LGTM, previous comments have also been addressed

scim/resource_service_principal.go
tanmay-db: I see, thanks for the clarification.

PR #2786, opened 2023-10-10
Suppress diff for `user_name` in `databricks_user` when the changes only in character case

alexott:
The identity management team rolled out the changes for AWS workspaces to force user names be lower-cased by default (to match behavior on GCP & Azure).  But these changes are causing the permanent configuration drift because `user_name` is marked as `force_new`.  To mitigate this problem, added a suppress diff function that ignores changes in case.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/resource_user.go
+++ b/scim/resource_user.go
@@ -29,6 +29,7 @@ func ResourceUser() *schema.Resource {
 	userSchema := common.StructToSchema(entity{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			addEntitlementsToSchema(&m)
+			m["user_name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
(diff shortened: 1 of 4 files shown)


scim/resource_user.go
```diff
@@ -29,6 +29,7 @@ func ResourceUser() *schema.Resource {
 	userSchema := common.StructToSchema(entity{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			addEntitlementsToSchema(&m)
+			m["user_name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
tanmay-db, line 32: Should we update our docs mentioning that the `user_name` is case insensitive?

tanmay-db (approved):

PR #2810, opened 2023-10-17
Use token-based pagination for Jobs List operations

### LINKED ISSUE #2807: [ISSUE] Exporter: Issue with `databricks_jobs` resource import
Tried to import our workspace jobs using the exporter module using the below CLI. The exporter module seems to be using the deprecated `offset` parameter and whenever it exceeds the value 1000, it encounters an `INVALID_PARAMETER_VALUE` error and the final report says `[ERROR] no resources to import`.
```bash
export DATABRICKS_HOST=https://<workspace_name>.cloud.databricks.com/
export DATABRICKS_TOKEN=<token>
../admin/.terraform/providers/registry.terraform.io/databricks/databricks/1.28.0/darwin_amd64/terraform-provider-databricks_v1.28.0 exporter \
  -skip-interactive \
  -services=jobs \
  -listing=databricks_jobs \
  -last-active-days=30 \
  -debug
```

### Expected Behavior

All the jobs should have been imported(irrespective of the total job count. We have > 1000 jobs btw).

### Actual Behavior

No jobs get imported even though 1000 jobs were successfully listed.

Databricks TF provider versions: 1.19.0, 1.25.1, 1.28.0
Terraform version: 1.3.5

### Debug Output
```
[DEBUG] GET /api/2.1/jobs/list?expand_tasks=false&limit=25&offset=1025
< HTTP/2.0 400 Bad Request
< {
<   "message": "The supplied offset 1025 has exceeded the value of 1000 allowed for this parameter.Please use th... (88 more bytes)"
< }
```
### Important Factoids

1. Fails in workspaces where there are more than 1k jobs.
2. `offset` parameter used in ‎`JobsAPI.ListByName` in `jobs/resource_job.go` needs to be replaced with the `page_token` param as instructed in the API doc [here](https://docs.databricks.com/api/workspace/jobs/list).

alexott:
## Changes

As we couldn't easily switch `databricks_job` resource to use Go SDK, we were still using offset-based pagination that has a significant load onto the control plane.  This PR changes list operation to use recommended token-based approach until we switch to Go SDK.

This fixes #2807

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -310,8 +310,10 @@ func (js *JobSettings) sortWebhooksByID() {
 
 // JobListResponse returns a list of all jobs
 type JobListResponse struct {
-	Jobs    []Job `json:"jobs"`
-	HasMore bool  `json:"has_more,omitempty"`
+	Jobs          []Job  `json:"jobs"`
+	HasMore       bool   `json:"has_more,omitempty"`
+	NextPageToken string `json:"next_page_token,omitempty"`
+	PrevPageToken string `json:"prev_page_token,omitempty"`
 }
 
 // Job contains the information when using a GET request from the Databricks Jobs api
@@ -416,12 +418,14 @@ func (a JobsAPI) ListByName(name string, expandTasks bool) ([]Job, error) {
 	if name != "" {
 		params["name"] = name
 	}
-	offset := 0
 
+	nextPageToken := ""
 	ctx := context.WithValue(a.context, common.Api, common.API_2_1)
 	for {
 		var resp JobListResponse
-		params["offset"] = offset
+		if nextPageToken != "" {
+			params["page_token"] = nextPageToken
+		}
 		err := a.client.Get(ctx, "/jobs/list", params, &resp)
 		if err != nil {
 			return nil, err
@@ -430,7 +434,7 @@ func (a JobsAPI) ListByName(name string, expandTasks bool) ([]Job, error) {
 		if !resp.HasMore {
 			break
 		}
-		offset += len(resp.Jobs)
+		nextPageToken = resp.NextPageToken
 	}
 	return jobs, nil
 }
--- a/jobs/resource_job_test.go
+++ b/jobs/resource_job_test.go
@@ -2184,7 +2184,7 @@ func TestJobsAPIList(t *testing.T) {
 	qa.HTTPFixturesApply(t, []qa.HTTPFixture{
 		{
 			Method:   "GET",
-			Resource: "/api/2.1/jobs/list?expand_tasks=false&limit=25&offset=0",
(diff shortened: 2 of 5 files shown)


jobs/resource_job_test.go
```diff
@@ -2205,26 +2205,26 @@ func TestJobsAPIListMultiplePages(t *testing.T) {
...
 			Method:   "GET",
-			Resource: "/api/2.1/jobs/list?expand_tasks=false&limit=25&offset=1",
+			Resource: "/api/2.1/jobs/list?expand_tasks=false&limit=25&page_token=aaaa",
 			Response: JobListResponse{
 				Jobs: []Job{
 					{
 						JobID: 2,
 					},
 				},
-				HasMore: false,
```
oleksiidav, line 2227: Nit: The API would still return `HasMore: false`, we can keep it

jobs/resource_job.go
```diff
@@ -430,7 +434,7 @@ func (a JobsAPI) ListByName(name string, expandTasks bool) ([]Job, error) {
 		if !resp.HasMore {
 			break
 		}
-		offset += len(resp.Jobs)
+		nextPageToken = resp.NextPageToken
```
oleksiidav, line 437: This can't be nil, right? Then the check at line 426 is safe

oleksiidav (approved):
LGTM with minor nits

jobs/resource_job_test.go
alexott: it will be implicitly initialized to `false`

jobs/resource_job.go
oleksiidav: Yes

jobs/resource_job.go
```diff
@@ -310,8 +310,10 @@ func (js *JobSettings) sortWebhooksByID() {
 
 // JobListResponse returns a list of all jobs
 type JobListResponse struct {
-	Jobs    []Job `json:"jobs"`
-	HasMore bool  `json:"has_more,omitempty"`
+	Jobs          []Job  `json:"jobs"`
+	HasMore       bool   `json:"has_more,omitempty"`
+	NextPageToken string `json:"next_page_token,omitempty"`
+	PrevPageToken string `json:"prev_page_token,omitempty"`
```
mgyucht, line 316: Just to confirm, you're adding PrevPageToken for completeness, even though we don't use it in this PR.

mgyucht (commented):
Implementation seems good to me. Can we extend TestJobsData to make a second request with the page token?

jobs/resource_job.go
alexott: Yes, only for completeness...

mgyucht (approved):
LGTM! Missed the addition of the next_page_token

PR #2816, opened 2023-10-18
Specify that Exporter is experimental

tanmay-db:
## Changes

Exporter is experimental and although we mention that in https://registry.terraform.io/providers/databricks/databricks/latest/docs/guides/experimental-exporter but we don't have mention of that in our repository. Adding message. 

## Tests

Ran: `./terraform-provider-databricks exporter` and Warning is printed.
image

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/command.go
+++ b/exporter/command.go
@@ -76,6 +76,7 @@ func (ic *importContext) interactivePrompts() {
 // Run import according to flags
 func Run(args ...string) error {
 	log.SetOutput(&logLevel)
+	log.Printf("[WARN] This tooling is experimental and provided as is. It has an evolving interface, which may change or be removed in future versions of the provider.")
 	client, err := client.New(&config.Config{})
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


mgyucht (approved):
LGTM. Let's wait for @alexott to review as well before merging.

exporter/command.go
```diff
@@ -76,6 +76,7 @@ func (ic *importContext) interactivePrompts() {
 // Run import according to flags
 func Run(args ...string) error {
 	log.SetOutput(&logLevel)
+	log.Printf("[WARN] This tooling is experimental and provided as is. It has an evolving interface, which may change or be removed in future versions of the provider.")
```
alexott, line 79: I'm not sure that adding this to the code is required... Exporter functionality isn't something that people just find in the docs, and it's already covered in the docs. Otherwise we'll need to add such warnings to most of our tooling - CLI, SDKs, etc.

tanmay-db: We want to highlight that it is experimental so it would also be nice to have this in the code.

PR #2821, opened 2023-10-19
Add deployment field to resource_job

shreyas-goenka:
## Changes
This PR adds the deployment field to the Job resource. This allows jobs created to be aware that they are a part of a bundle, as well as communicate bundle metadata with the job.

## Tests
Unit test


alexott (commented):
code-wise is ok, but should we document it?

shreyas-goenka:
I'd recommend against documenting it because:
1.  It's in private preview
2. The only usecase today and forseeable future is for Jobs to know they are part of a bundle. We don't want users to set it manually.

alexott (approved):

PR #2826, opened 2023-10-20
Revert "Refactor `databricks_recipient` to Go SDK"

nkvuong:
Reverts databricks/terraform-provider-databricks#2757

Integration test failed with `Error: cannot create recipient: panic: reflect: call of reflect.Value.SetString on interface Value`. This is due to an issue with OpenAPI spec.

We either needs to revert the PR or generate a new version of Go SDK before the next release


tanmay-db (approved):
Thanks for the PR. I think going ahead with the revert is safer approach than fix forward.

PR #2837, opened 2023-10-25
Point master to main everywhere in TF

mgyucht:
## Changes
Currently, `main` and `master` point to the same commit. This PR rewrites all references to branch `master` to `main` in the `main` branch. Once this is merged, I will change the default branch in the TF repo from master to main.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, main ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, main ]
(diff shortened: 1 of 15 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, main ]
```
tanmay-db, line 5: curious -- why do we want to keep master here? Longer term I think master would be deleted and only merging to main would be allowed so maybe we should start doing it from now?

tanmay-db (approved):

PR #2841, opened 2023-10-26
Change Catalog Volume Storage Location to Computed

### LINKED ISSUE #2840: [ISSUE] Issue with `databricks_volume` `storage_location` attribute
The `storage_location` attribute of the `databricks_volume` resource should be optional and computed. When unspecified, this value is populated from Databricks from the auto-generated storage location of the metastore. This attribute is used to copy data into from other locations and needs to be exposed as calucated for plans to execute correctly. Otherwise, a plan will attempt to utilize a null valued string instead of waiting for the deployed resource's information.

https://github.com/databricks/terraform-provider-databricks/blob/b60be7bbaf0b262d4a5ed4cc5106a99b1c10e902/catalog/resource_volume.go#L27

image

nerddtvg:
## Changes
Enabling the `StorageLocation` property of `databricks_volume` to be computed as well as optional. This will allow the computed value, when no value is provided, to be used in plans.

## Tests
Added `StorageLocation` tests to `resource_volume_test.go` however those are only scaffolded responses.

I'm not sure how best to test that a value is returned in `internal/acceptance`. Current tests are AWS only with hardcoded storage locations.

- [x] `make test` run locally
- [x] relevant change in `docs/` folder (Not Applicable)
- [x] using Go SDK

Fixes #2840


nkvuong:
@nerddtvg you could add a test that creates a managed volume, which will not require a storage location and return one in the response

PR #2843, opened 2023-10-26
Revert "Handle ForceSendFields (#2753)"

mgyucht:
This reverts commit cff5d8037f21e6542bca85ec9d589d766a4c20fe.

## Changes
This change introduced a regression for resources where the zero value of primitive fields ought not to be sent under a specific circumstance (when the field is present in the state but not in the config, for example, if the field were previously set to some value and then removed from the config during the next run). Instead, we'll manually set ForceSendFields where needed to correct the behavior throughout the provider. A proper fix will need to wait until we adopt the plugin framework, which does expose the resource config to provider authors.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott (approved):

mgyucht:
nightlies pass, merging now.

PR #2852, opened 2023-10-27
Add NumWorkers to ForceSendFields in `databricks_pipeline` if `spark_conf` contains config for single-node cluster

### LINKED ISSUE #1875: [ISSUE] Cannot set `num_workers` to 0 `databricks_pipeline` in resource
If you set `num_workers` to 0 in a `databricks_pipeline` resource - the field is considered "unset" and this is omitted. This is a mistake as 0 is a supported value for a DLT pipeline (0 implies a single-node cluster as per https://docs.databricks.com/workflows/delta-live-tables/delta-live-tables-configuration.html#cluster-config). When you set 0 - you end up with a legacy auto-scaling cluster of 1 -5 workers.

hectorcast-db:
## Changes
Add NumWorkers to ForceSendFields in Pipelines if AutoScaling is null

This fixes #1875

## Tests

- [X] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/pipelines/resource_pipeline.go
+++ b/pipelines/resource_pipeline.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/marshal"
 	"github.com/databricks/terraform-provider-databricks/clusters"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/databricks/terraform-provider-databricks/libraries"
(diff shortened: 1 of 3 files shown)


hectorcast-db:
Created this PR using the standard solution for this kind of issues. But this resrouce is special because if nothing is specified, then DLT creates a cluster with enhanced autoscaling from 1 to 5 nodes. This PR would change that.

mgyucht (approved):
Looks great!

mgyucht:
Could we add this to an E2E test as well?

pipelines/resource_pipeline.go
```diff
@@ -199,12 +212,22 @@ func (a PipelinesAPI) Create(s PipelineSpec, timeout time.Duration) (string, err
 	return id, nil
 }
 
+func adjustForceSendFields(s *PipelineSpec) {
+	for i := range s.Clusters {
+		cluster := &s.Clusters[i]
+		if cluster.SparkConf["spark.databricks.cluster.profile"] == "singleNode" {
+			cluster.ForceSendFields = append(cluster.ForceSendFields, "NumWorkers")
```
andrewnester, line 219: Can't we always send this instead? My understanding is that if it's not explicitly set it won't be sent

hectorcast-db: TF SDK doesn't have a method to know if a field is set or not. In this case, not set means "use the default autostaling, with min 2- max 5 workers". While NumWorkers 0 means use 0 workers.

We are instead using this spark config which tells Spark that there are no workers. This config only makes sense if NumWorkers is zero, so we use it to determine that we need to send the field.

I will add a comment.

andrewnester: What would happens if we mark NumWorkers as ForceSendFields always without the condition?
In particular if it's not set in TF but marked as ForceSendField, which value will be sent?

hectorcast-db: Then we would send 0, which is the default GoLang value for Int.

PR #2853, opened 2023-10-27
Fix Read method of the `databricks_mlflow_model` resource

alexott:
## Changes

`StructToData` did use wrong structure to deserialize data, so struct was empty after read. Also, changed to use `CreateModelRequest` as a base struct because we don't use all fields of the `Model` anyway.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/mlflow/resource_mlflow_model.go
+++ b/mlflow/resource_mlflow_model.go
@@ -10,14 +10,9 @@ import (
 
 func ResourceMlflowModel() *schema.Resource {
 	s := common.StructToSchema(
-		ml.Model{},
+		ml.CreateModelRequest{},
 		func(s map[string]*schema.Schema) map[string]*schema.Schema {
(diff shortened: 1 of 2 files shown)


tanmay-db (commented):
Thanks for the PR, a question:

> Also, changed to use CreateModelRequest as a base struct because we don't use all fields of the Model anyway.

Can this cause a regression? To me, I think CreateModelRequest as a base model sounds bad because it should only be for Creation.

Why are the rest of the fields not used? I think if they are there, they are probably used in some API call?

alexott:
The rest of the fields are computed, if we use `DatabricksModel` as a base, we will need to delete most of the fields, like it was done in previous implementation

alexott:
Also, in the update API only two fields could be updated.

tanmay-db (approved):

PR #2854, opened 2023-10-27
Exporter: add case-insensitive match type for dependencies

### LINKED ISSUE #2848: Exporter: Add ability to perform case-insensitive comparison when resolving dependencies
Right now, the exporter resolves the dependencies using the exact matching, but this doesn't always work correctly in cases when comparison should be case-insensitive, like, in the following case when the first user's email is capitalized in the job's settings, but SCIM API returns lower-cased name.

```
email_notifications {
    on_success = ["User1@domain.com", databricks_user.user2.user_name]
}
```

alexott:
## Changes

This was necessary because we need to match user emails in a case-insensitive manner for cases when email is populated manually, like, jobs notifications, etc.

This fixes #2848

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


exporter/exporter_test.go
```diff
@@ -880,6 +880,9 @@ func TestImportingJobs_JobList(t *testing.T) {
 							UserName:             "user@domain.com",
 							ServicePrincipalName: "0000-1111-2222-3333-4444-5555",
 						},
+						EmailNotifications: &jobs.EmailNotifications{
+							OnFailure: []string{"user@domain.com"},
+						},
```
tanmay-db, line 885: what does this match against? for case insensitivity.

tanmay-db (approved):

exporter/exporter_test.go
alexott: it will match against the `userName`, but will do it in a case-insensitive manner.

PR #2857, opened 2023-10-27
Exporter: disable generation of empty or default blocks for `databricks_job` and `databricks_cluster`

alexott:
## Changes

* Disable generation of empty blocks for `email_notifications`, `webhook_notifications`, and `notification_settings` in `databricks_job`
* Don't produce `spark_conf` & `spark_env_vars` in clusters if they are default values produced by the REST API backend

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"reflect"
 	"regexp"
 	"sort"
 	"strconv"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/databricks/databricks-sdk-go/service/compute"
 	"github.com/databricks/databricks-sdk-go/service/iam"
+	sdk_jobs "github.com/databricks/databricks-sdk-go/service/jobs"
 	"github.com/databricks/databricks-sdk-go/service/ml"
 	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/terraform-provider-databricks/clusters"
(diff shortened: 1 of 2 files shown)


exporter/importables.go
```diff
@@ -561,6 +563,43 @@ var resourcesMap map[string]importable = map[string]importable{
...
+				if js.NotificationSettings != nil {
+					return reflect.DeepEqual(*js.NotificationSettings, sdk_jobs.JobNotificationSettings{})
+				}
+			}
+			if strings.HasPrefix(pathString, "task.") {
+				parts := strings.Split(pathString, ".")
+				if len(parts) > 2 {
+					taskIndex, err := strconv.Atoi(parts[1])
+					if err == nil && taskIndex >= 0 && taskIndex < len(js.Tasks) {
+						switch parts[len(parts)-1] {
```
tanmay-db, line 587: what is `parts[len(parts)-1]`? Might be good to define it in variable so it's more code readable.

exporter/importables.go
```diff
@@ -561,6 +563,43 @@ var resourcesMap map[string]importable = map[string]importable{
...
+						case "notification_settings":
+							if js.Tasks[taskIndex].NotificationSettings != nil {
+								return reflect.DeepEqual(*js.Tasks[taskIndex].NotificationSettings, sdk_jobs.TaskNotificationSettings{})
+							}
+						case "email_notifications":
```
tanmay-db, line 592: we don't need a case for `webhook_notifications` here?

tanmay-db (approved):
Few comments, otherwise LGTM, just to check were these changes tested manually with exporter?

exporter/importables.go
alexott: webhook is defined only on a job level

exporter/importables.go
alexott: `parts[len(parts)-1]` - block name

alexott:
@tanmay-db yes, tested manually.

exporter/importables.go
tanmay-db: I think doing something like would be more readable:
`blockName := parts[len(parts)-1]`

exporter/importables.go
tanmay-db: I see

alexott: let me add it

tanmay-db: but wasn't this only defined at job level? We do we need to add this?

alexott: We have different things:

- webhooks could be defined only on a job level
- `notification_settings` and `email_notifications` could be defined on both task & job level

PR #2885, opened 2023-11-07
Authenticate client during configuration

mgyucht:
## Changes
Currently, the client in the provider authenticates when making its first request. This request can have a context with a timeout or cancel func. However, because the OAuth2 caches this context, future token refreshes are subject to this cancellation/timeout. Instead, we call `Config.Authenticate()` during provider configuration. This serves 2 benefits:
1. The provider will fail to be configured if it cannot authenticate, rather than spitting out 1000s of resources with default auth problems.
2. The OAuth2 token source will be constructed with `context.Background()`, so token refreshes will never timeout.

Ideally, we would use the context associated with the request for token refreshes, but the OAuth2 library doesn't currently support this.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/provider/provider_test.go
+++ b/provider/provider_test.go
@@ -3,10 +3,13 @@ package provider
 import (
 	"context"
 	"fmt"
+	"log"
 	"net/http"
+	"net/http/httptest"
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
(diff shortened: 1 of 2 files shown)


provider/provider_test.go
```diff
@@ -388,6 +392,59 @@ func TestConfig_CorruptConfig(t *testing.T) {
...
+	} else if r.RequestURI == "/api/2.0/clusters/get?cluster_id=123" {
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprint(w, `{"cluster_id": "123"}`)
+		return
+	}
+	w.WriteHeader(http.StatusNotFound)
+}
+
+// Start a local webserver to serve tokens. Configure the CLI to fetch a token from this endpoint.
+// Then, using a context with a short timeout, call an API. Wait for 2x the timeout, then call
```
hectorcast-db, line 413: nit. IIUC, it's not the context which is short lived, but the token itself.

hectorcast-db (approved):

alexott (approved):
looks good. I've tested it for 3 hours

alexott:
I've tested it with following code and reduced rate limit & TF parallelism:

```hcl
data "databricks_current_user" "me" {
}

locals {
  cnt = 3000
}

resource "databricks_notebook" "this" {
  count = local.cnt
  content_base64 = base64encode(<<-EOT
    # created from ${abspath(path.module)}
    display(spark.range(10))
    EOT
  )
  path     = "${data.databricks_current_user.me.home}/OAuthTest/nb-${count.index}"
  language = "PYTHON"
}

resource "databricks_permissions" "notebook_usage" {
  count = local.cnt
  notebook_path = databricks_notebook.this[count.index].path

  access_control {
    group_name       = "users"
    permission_level = "CAN_READ"
  }
}
```

It run for ~3 hours without any problems

PR #2887, opened 2023-11-07
Added `databricks_instance_profiles` data source

### LINKED ISSUE #2330: [FEATURE] databricks_instance_profile datasource
### Use-cases

When using features such as  https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/service_principal_role, it would be really useful to be able to query Databricks for the registered instance profiles, for example by name.

Is not clear whether using the https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_instance_profile would be an alternative solution, since the provider seems only to take an ID, which sounds like an internal ID in databricks.

ehab-eb:
## Changes
Added a new data source `databricks_instance_profiles` to list all instance profiles and export their properties as a set of objects.

Fixes #2330

Planning to add some filtering options in a separate PR but would like to have a discussion about the design of the filters :) 

## Tests

- [x] `make test` run locally
- [x] using Go SDK


aws/data_instance_profiles.go
```diff
@@ -0,0 +1,38 @@
...
+	}, w *databricks.WorkspaceClient) error {
+		instanceProfiles, err := w.InstanceProfiles.ListAll(ctx)
+		if err != nil {
+			return err
+		}
+		for _, v := range instanceProfiles {
+			arnSlices := strings.Split(v.InstanceProfileArn, "/")
+			name := arnSlices[len(arnSlices)-1]
+			data.InstanceProfiles = append(data.InstanceProfiles, instanceProfileData{
+				Name:    name,
```
ehab-eb, line 30: `name` isn't returned by the API but it's added to the response to match what users see in the UI.

PR #2892, opened 2023-11-08
Allow authentication during configure to fail

mgyucht:
## Changes
As a follow-up to https://github.com/databricks/terraform-provider-databricks/pull/2885, we should tolerate auth failures for dynamically configured providers during the planning phase. This allows users to create workspaces and then configure a provider to manage resources in that workspace as part of a single template. 

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -2,12 +2,15 @@ package provider
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"log"
 	"net/http"
 	"reflect"
 	"sort"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
@@ -226,7 +229,7 @@ func configureDatabricksClient(ctx context.Context, d *schema.ResourceData) (any
 		}
 	}
 	sort.Strings(attrsUsed)
-	log.Printf("[INFO] Explicit and implicit attributes: %s", strings.Join(attrsUsed, ", "))
+	tflog.Info(ctx, fmt.Sprintf("Explicit and implicit attributes: %s", strings.Join(attrsUsed, ", ")))
 	if cfg.AuthType != "" {
 		// mapping from previous Google authentication types
 		// and current authentication types from Databricks Go SDK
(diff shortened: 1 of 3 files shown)


provider/provider.go
```diff
@@ -247,8 +250,16 @@ func configureDatabricksClient(ctx context.Context, d *schema.ResourceData) (any
...
-	err = cfg.Authenticate(r.WithContext(context.Background()))
-	if err != nil {
+	err = cfg.Authenticate(r)
+	// When a user creates a workspace and then configures the Databricks provider in the same
+	// module for that workspace, the workspace does not exist at planning time, so computed
+	// attributes like the host are not included in the ConfigureProvider call. In this case,
+	// authentication will fail, but it is safe to continue: the client will not be used, as
+	// no resources yet exist in the as-yet-nonexisting workspace, so Terraform will not attempt
+	// to fetch any resources.
+	if errors.Is(err, config.ErrCannotConfigureAuth) {
```
hectorcast-db, line 260: Is this strict enough? I see a lot of tests failing.

provider/provider.go
```diff
@@ -226,7 +229,7 @@ func configureDatabricksClient(ctx context.Context, d *schema.ResourceData) (any
 		}
 	}
 	sort.Strings(attrsUsed)
-	log.Printf("[INFO] Explicit and implicit attributes: %s", strings.Join(attrsUsed, ", "))
+	tflog.Info(ctx, fmt.Sprintf("Explicit and implicit attributes: %s", strings.Join(attrsUsed, ", ")))
```
mgyucht, line 232: This should be a no-op, but I needed to make this change so that I could enable logging for just our provider using `TF_LOG_PROVIDER_DATABRICKS=INFO` (which according to the docs, only enables `tflog.*` logs and not `log.Printf` logs). In general, however, we should be using `tflog.*` everywhere for logging to fit into TF's logging framework.

provider/provider_test.go
```diff
@@ -460,6 +460,13 @@ func configureProviderAndReturnClient(t *testing.T, tt providerFixture) (*common
...
 	client := p.Meta().(*common.DatabricksClient)
-
+	r, err := http.NewRequest("GET", "", nil)
+	if err != nil {
+		return nil, err
+	}
+	err = client.Config.Authenticate(r)
+	if err != nil {
+		return nil, err
+	}
```
mgyucht, line 470: This was removed in https://github.com/databricks/terraform-provider-databricks/commit/8842527bba1942d19cad5bb3bca57edb2bf9e416#diff-978ff4fdc176f5a02cba8681ff7f0c943bf0740a2ca7a68ff4e9f08d1efd3744L406-L410, but we add it back here to ensure existing test cases work. Those test cases make sure the provider auth works as expected when auth ought to work, but now we tolerate auth failing during provider construction, so we need to retry again to get the underlying auth error.

hectorcast-db (approved):

PR #2902, opened 2023-11-09
Improve performance of `databricks_notebook` and `databricks_workspace_file` resources

### LINKED ISSUE #2898: [ISSUE] Issue with provider performance
Increasing `-parallelism` doesn't have an effect on the provider's performance. When I'm creating a significant number of resources, I see that planning and applying changes are done in parallel, but API execution is sequential, often with significant delay. 

### Expected Behavior

Resources are created in parallel.

### Steps to Reproduce

Generate a terraform file with hundreds of notebooks, and run the `terraform apply`. The performance will be the same independent of set `-parallelism`.

`master` branch. Tested with Terraform 1.5.7 & 1.3.10

### Debug Output

Will share internally

alexott:
## Changes

There were problems with the performance of `databricks_notebook` resource:

* There is an explicit Mutex there to coordinate the creation & deletion of directories - use of it led to the situation when notebooks were created sequentially even with high parallelism.
* Both `databricks_notebook` and `databricks_workspace_file` are always called `mkdirs` API to make sure that the parent directory exists.

This PR includes the following changes:

* In both `databricks_notebook` and `databricks_workspace_file`, the `mkdirs` API is called only when the error about "Parent folder doesn't exist". This should decrease the number of API calls.
* Decrease the use of locks in `databricks_notebook`

This fixes #2898

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/workspace/resource_notebook.go
+++ b/workspace/resource_notebook.go
@@ -300,7 +300,7 @@ func (a NotebooksAPI) list(path string) ([]ObjectStatus, error) {
 // Delete will delete folders given a path and recursive flag
 func (a NotebooksAPI) Delete(path string, recursive bool) error {
 	if recursive {
-		log.Printf("[DEBUG] Doing recursive delete...")
+		log.Printf("[DEBUG] Doing recursive delete of path '%s'", path)
 		mtx.Lock()
 		defer mtx.Unlock()
 	}
@@ -389,12 +389,10 @@ func ResourceNotebook() *schema.Resource {
 			if err != nil {
 				if isParentDoesntExistError(err) {
 					parent := filepath.ToSlash(filepath.Dir(path))
-					if parent != "/" {
-						log.Printf("[DEBUG] Parent folder '%s' doesn't exist, creating...", parent)
-						err = notebooksAPI.Mkdirs(parent)
-						if err != nil {
-							return err
-						}
+					log.Printf("[DEBUG] Parent folder '%s' doesn't exist, creating...", parent)
+					err = notebooksAPI.Mkdirs(parent)
+					if err != nil {
+						return err
 					}
 					err = notebooksAPI.Create(createNotebook)
 				}
--- a/workspace/resource_notebook.go
+++ b/workspace/resource_notebook.go
@@ -101,8 +101,10 @@ var mtx = &sync.Mutex{}
 
 // Create creates a notebook given the content and path
 func (a NotebooksAPI) Create(r ImportPath) error {
-	mtx.Lock()
-	defer mtx.Unlock()
+	if r.Format == "DBC" {
+		mtx.Lock()
+		defer mtx.Unlock()
+	}
 	return a.client.Post(a.context, "/workspace/import", r, nil)
 }
 
@@ -297,8 +299,11 @@ func (a NotebooksAPI) list(path string) ([]ObjectStatus, error) {
 
 // Delete will delete folders given a path and recursive flag
 func (a NotebooksAPI) Delete(path string, recursive bool) error {
-	mtx.Lock()
-	defer mtx.Unlock()
+	if recursive {
+		log.Printf("[DEBUG] Doing recursive delete...")
+		mtx.Lock()
+		defer mtx.Unlock()
+	}
 	return a.client.Post(a.context, "/workspace/delete", DeletePath{
 		Path:      path,
 		Recursive: recursive,
@@ -363,13 +368,6 @@ func ResourceNotebook() *schema.Resource {
 			}
 			notebooksAPI := NewNotebooksAPI(ctx, c)
 			path := d.Get("path").(string)
-			parent := filepath.ToSlash(filepath.Dir(path))
-			if parent != "/" {
-				err = notebooksAPI.Mkdirs(parent)
-				if err != nil {
-					return err
-				}
-			}
 			createNotebook := ImportPath{
 				Content:   base64.StdEncoding.EncodeToString(content),
 				Language:  d.Get("language").(string),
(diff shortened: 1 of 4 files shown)


workspace/resource_notebook.go
```diff
@@ -101,8 +101,10 @@ var mtx = &sync.Mutex{}
 
 // Create creates a notebook given the content and path
 func (a NotebooksAPI) Create(r ImportPath) error {
-	mtx.Lock()
-	defer mtx.Unlock()
+	if r.Format == "DBC" {
+		mtx.Lock()
```
nfx, line 105: can you switch to tree-based locking? this was added to prevent errors when creating notebooks along the tree

nfx (dismissed):

workspace/resource_notebook.go
alexott: DBC is kind of a specific case... I'm not sure how this will behave in the case of big trees, and/or in the case of multiple workspaces

tanmay-db: Hi Alex, thanks for the PR, a question -- what make case for .dbc files specific wrt to parallelism?

alexott: DBC files may contain directories, so we need to make sure that the files that are potentially with directories, had a lock

alexott: But I haven't seen DBC files deployed via TF in the real-life...

workspace/resource_notebook.go
```diff
@@ -101,8 +101,10 @@ var mtx = &sync.Mutex{}
 
 // Create creates a notebook given the content and path
 func (a NotebooksAPI) Create(r ImportPath) error {
-	mtx.Lock()
-	defer mtx.Unlock()
+	if r.Format == "DBC" {
```
mgyucht, line 104: @alexott do we need to check if Format is `AUTO`? This could also be a DBC notebook and run into this race condition.

workspace/resource_notebook.go
mgyucht: From my reading of the comment next to the mutex, this applies to race conditions when creating/deleting directories, not notebooks. There should generally not be two notebook resources with the same path, otherwise there will always be a race condition. While it would be more precise to lock on a per-path basis, you would never get a diff-free plan after refreshing if two notebooks have the same path. So IMO this change is sufficiently safe.

workspace/resource_notebook.go
```diff
@@ -297,8 +299,11 @@ func (a NotebooksAPI) list(path string) ([]ObjectStatus, error) {
 
 // Delete will delete folders given a path and recursive flag
 func (a NotebooksAPI) Delete(path string, recursive bool) error {
-	mtx.Lock()
-	defer mtx.Unlock()
+	if recursive {
+		log.Printf("[DEBUG] Doing recursive delete...")
```
mgyucht, line 303: Suggestion: include the path name here?

workspace/resource_notebook.go
```diff
@@ -389,7 +387,20 @@ func ResourceNotebook() *schema.Resource {
 			}
 			err = notebooksAPI.Create(createNotebook)
 			if err != nil {
-				return err
+				if isParentDoesntExistError(err) {
+					parent := filepath.ToSlash(filepath.Dir(path))
+					if parent != "/" {
```
mgyucht, line 392: Will we ever have the else variant of this case? `/` is guaranteed to exist, so we will not enter the if block starting on 390.

workspace/resource_notebook.go
```diff
@@ -431,7 +442,13 @@ func ResourceNotebook() *schema.Resource {
...
-			return NewNotebooksAPI(ctx, c).Delete(d.Id(), true)
+			objType := d.Get("object_type")
+			return NewNotebooksAPI(ctx, c).Delete(d.Id(), !(objType == Notebook || objType == File))
 		},
 	}.ToResource()
 }
+
+func isParentDoesntExistError(err error) bool {
+	errStr := err.Error()
+	return strings.HasPrefix(errStr, "The parent folder ") && strings.HasSuffix(errStr, " does not exist.")
```
mgyucht, line 453: Do we not have an error code for this? Note: once the next Go SDK is released, we can do `errors.is(err, databricks.ErrResourceDoesNotExist)` to check for this case (merged today)!

workspace/resource_notebook.go
```diff
@@ -431,7 +442,13 @@ func ResourceNotebook() *schema.Resource {
 			})
 		},
 		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			return NewNotebooksAPI(ctx, c).Delete(d.Id(), true)
+			objType := d.Get("object_type")
+			return NewNotebooksAPI(ctx, c).Delete(d.Id(), !(objType == Notebook || objType == File))
```
mgyucht, line 446: So, if objType is Directory, Library, or Repo, we do want to recursively delete?

workspace/resource_workspace_file.go
```diff
@@ -37,21 +38,28 @@ func ResourceWorkspaceFile() *schema.Resource {
...
 				Path:      path,
 				Overwrite: true,
-			})
+			}
+			err = client.Workspace.Import(ctx, importReq)
 			if err != nil {
-				return err
+				if isParentDoesntExistError(err) {
+					parent := filepath.ToSlash(filepath.Dir(path))
+					if parent != "/" {
```
mgyucht, line 51: Ditto, we probably can inline the if body here since it will always be true

mgyucht (commented):
Couple of nits, questions and suggestions. Glad to do another round today.

workspace/resource_notebook.go
alexott: We don't allow `AUTO` as format: https://github.com/databricks/terraform-provider-databricks/blob/master/workspace/resource_notebook.go#L329

workspace/resource_notebook.go
alexott: No, we don't have error code at this level.

workspace/resource_workspace_file.go
alexott: No - this is to handle of cases if I notebooks/files are created at root level

workspace/resource_notebook.go
alexott: done

workspace/resource_notebook.go
mgyucht: Right, but my point is the REST API will never return an error that / doesn't exist.

mgyucht: I guess there isn't a problem with this per se, but it seems unnecessarily defensive, unless I'm missing something.

alexott: ah, yes - misunderstood the case. let me change it...

alexott: Fixed

workspace/resource_notebook.go
alexott: it's only Directory that could be here.

mgyucht (approved):

PR #2914, opened 2023-11-14
Do not persist state from failed updates in  databricks_workspace_conf 

shreyas-goenka:
## Changes
The Terraform SDK by default persists the state, even if apply fails. This PR sets the config back to it's previous value to prevent values not applied being persisted in state. 

This also solves the issue where entering typos for key values would brick the terraform state. 

## Tests
Manually and new integration test


alexott (commented):
In general looks good, can we add a unit test for this change?

shreyas-goenka:
Added unit test

alexott (approved):
lgtm

PR #2929, opened 2023-11-17
Exporter: fix generation of `branch` & `tag` attributes in `databricks_repo`

### LINKED ISSUE #2927: Exporter: incomplete handling of Databricks Repos
The customer reported a few issues with handling repos in exporter:

- information about branches and tags wasn't generated (most probably because they are marked as `computed`)
- no permissions were generated  - just checked, works just fine

alexott:
## Changes

These attributes were omitted from the generated code because they are marked as `computed` in the resource.  This PR fixes this.

This fixes #2927

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -1265,8 +1265,13 @@ var resourcesMap map[string]importable = map[string]importable{
 			return nil
 		},
 		ShouldOmitField: func(ic *importContext, pathString string, as *schema.Schema, d *schema.ResourceData) bool {
-			if pathString == "path" {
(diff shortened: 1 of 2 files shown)


alexott:
@hectorcast-db It's not the resource computed, but the attribute. Like, `branch` is returned by API even if it's not set explicitly.

hectorcast-db (approved):
I was missing a bit of context on the imported resources. I have checked the code and docs and it makes sense now.

PR #2931, opened 2023-11-17
Force sending the `content` field when creating empty `databricks_workspace_file` resources

alexott:
## Changes

The issue was reported by the customer who used the exporter to migrate the Azure workspace that they exported many empty `__init__.py` files, but couldn't import them because Terraform gave them an error.  The reason for this error is that in API spec the `content` field is marked as `optional` and correspondingly the `omitempty` was generated in Go SDK, but in fact, this field is required by the API.

To work around this problem until the API spec is fixed, the `Content` field was explicitly added to the `ForceSendFields` array to send the empty content.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/acceptance/workspace_file_test.go
+++ b/internal/acceptance/workspace_file_test.go
@@ -18,6 +18,15 @@ func TestAccWorkspaceFile(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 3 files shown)


mgyucht (approved):

alexott:
I'm waiting for acceptance tests passing

Issue #2939, opened 2023-11-21
[ISSUE] run_as_role Issue with `databricks_sql_query` resource


SlevinBE:
### Configuration

```hcl
resource "databricks_sql_query" "sample_query" {
  provider = databricks.created_workspace
  run_as_role    = "viewer"
  query          = <<EOT
    select * from test_table;
  EOT
  parent         = "folders/${databricks_directory.operational.object_id}"
  name           = "sample_query"
  data_source_id = databricks_sql_endpoint.this.data_source_id
}
```

### Expected Behavior
The query should've been created with a run as "viewer" role.

### Actual Behavior
The query is created with the default run as role, which is `owner`. This means we have a constant Terraform drift when running TF apply.
```
# databricks_sql_query.sample_query will be updated in-place
  ~ resource "databricks_sql_query" "sample_query" {
        id             = "65bd4581-5f07-4ef6-8f27-262ccc89b46d"
        name           = "sample_query"
      ~ run_as_role    = "owner" -> "viewer"
        # (5 unchanged attributes hidden)
    }
```

### Steps to Reproduce
1. `terraform apply`using the provided `databricks_sql_query` resource definition
2. once applied, run `terraform plan`, which results in the TF config drift.

### Terraform and provider versions
terraform_version: 1.5.5
databricks provider version: latest

### Is it a regression?
no

### Important Factoids
In the Terraform provider code, we found that it tries to create the query resource using an incorrect API call. Specifically in [this piece of code](https://github.com/databricks/terraform-provider-databricks/blob/master/sql/resource_sql_query.go#L296), it's trying to set the `run_as_role` property on the `options` object. However, looking at the [Databricks API definition for creating a query resource](https://docs.databricks.com/api/workspace/queries/create), the options block is "Exclusively used for storing a list parameter definitions", so it doesn't include the same number of properties as the response or GET object.  

For example, when creating a SQL Query via the API with the following body, similar to what the TF provider currently does:
```
{
    "data_source_id": "8ac1d7df-01d9-4218-935f-6ec6a1539f34",
    "query": "SELECT * from sample_query",
    "name": "test query",
    "parent": "folders/123456",
    "description": "Tests the creation of a query",
    "options": {
        "parameters": [
        ],
        "run_as_role": "viewer"
    }
}
```
We get the following response with an incorrect run_as_role configuration:
```
{
    "id": "2853db7e-3edc-48be-840a-36feff665e82",
    "version": 1,
[...]

PR #2940, opened 2023-11-21
fix: resolved issue with run_as_role not being applied on databricks_sql_query resource

SlevinBE:
Fix for: https://github.com/databricks/terraform-provider-databricks/issues/2939

## Changes
The `run_as_role` property wasn't being applied correctly on a databricks_sql_query resource, as documented in https://github.com/databricks/terraform-provider-databricks/issues/2939. This PR tries to solve this issue by aligning the TF code with the [Databricks Query API](https://docs.databricks.com/api/workspace/queries/create), so that run_as_role is only specified and queried from the top-level object, and not from the `options` block anymore.

## Tests
Specifically, the `sql/resource_sql_query_test.go` was updated and run to verify this fix. A proper integration test run is still required to validate the fix.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/sql/api/query.go
+++ b/sql/api/query.go
@@ -14,6 +14,7 @@ type Query struct {
 	Query        string `json:"query"`
 	// Deprecated: Use databricks_job resource to schedule a Query
 	Schedule       *QuerySchedule    `json:"schedule"`
+	RunAsRole      string            `json:"run_as_role,omitempty"`
 	Options        *QueryOptions     `json:"options,omitempty"`
 	Tags           []string          `json:"tags,omitempty"`
 	Visualizations []json.RawMessage `json:"visualizations,omitempty"`
@@ -46,8 +47,6 @@ type QuerySchedule struct {
 type QueryOptions struct {
 	Parameters    []any             `json:"-"`
 	RawParameters []json.RawMessage `json:"parameters,omitempty"`
-
-	RunAsRole string `json:"run_as_role,omitempty"`
 }
 
 // MarshalJSON ...
--- a/sql/resource_sql_query.go
+++ b/sql/resource_sql_query.go
@@ -290,10 +290,7 @@ func (q *QueryEntity) toAPIObject(schema map[string]*schema.Schema, data *schema
 	}
 
 	if q.RunAsRole != "" {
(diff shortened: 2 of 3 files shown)


sql/api/query.go
```diff
@@ -14,6 +14,7 @@ type Query struct {
 	Query        string `json:"query"`
 	// Deprecated: Use databricks_job resource to schedule a Query
 	Schedule       *QuerySchedule    `json:"schedule"`
+	RunAsRole      string            `json:"run_as_role,omitempty"`
```
alexott, line 17: If the API fills `run_as_role` with default `owner`, then it makes sense to add `suppress_diff`, like this: https://github.com/databricks/terraform-provider-databricks/blob/master/sql/resource_sql_dashboard.go#L19

alexott (changes requested):

sql/api/query.go
SlevinBE: Hi Alex, can you explain what the `suppress_diff` exactly does? I've read the documentation in the [Contributor docs](https://github.com/databricks/terraform-provider-databricks/blob/master/CONTRIBUTING.md#adding-a-new-resource), but it's not quite clear to me yet.

alexott: It's suppressing the diff from a default value to an empty string.  It's needed because if you declare the resource without `run_as_role`, then API will fill it with default `owner`, but when you run apply next time, because `run_as_role` is empty in the query, then it will be detected as drift, and will try to reset it back.

SlevinBE: Ah, makes sense. However, `suppress_diff` is already defined on the [QueryEntity](https://github.com/KOR-Financial/terraform-provider-databricks/blob/fix/2939_sql_query_run_as_role/sql/resource_sql_query.go#L27)'s RunAsState property. Isn't the Query api object you're referencing here used to represent the Databricks API objects, so unrelated to the Terraform state, and therefore requiring no `suppress_diff` config here?

alexott: ah, sorry - misread it

alexott (approved):
thank you for contribution!

hectorcast-db (approved):

PR #2941, opened 2023-11-21
Allow to override built-in `databricks_cluster_policy` resources

### LINKED ISSUE #2935: [FEATURE] Allow to override built-in cluster policies
Databricks UI allows to override built-in cluster policies (based on the cluster families). If we try to do the same with Terraform then we'll get an error like this:

```
Cluster Policy with name `Personal Compute` already exists
```.

We need to support this as the following:

1. If we see an override for built-in cluster policy we need to allow this (kind of a `force` flag for users/service principals)
2. on destroy, we need to only remove the policy override, not remove the policy completely.

### PR DESCRIPTION
**Pull request by alexott**
## Changes

Databricks is shipped with a number of built-in cluster policies that are inherited from the cluster policy families.  For built-in cluster policies we need to provide an override, not the `definition`.  We also need to handle deletes for them differently - just remove the overrides because we can't remove built-in policies.  We also need to distinguish cases when we override built-in policy from cases when we create a new policy based on the cluster policy family.

This PR makes following changes:

* Discovers a list of cluster policies families dynamically instead of relying on hardcoded list of policies
* When creating a new `databricks_cluster_policy`, check if it's built-in policy or not, and if it's built-in, then update it with policy override instead of trying to create.
* When deleting a created `databricks_cluster_policy`, check if it's built-in policy or not, and if it's built-in, then reset policy overrides to empty string to return back to the original state.

This fixes #2935

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/policies/resource_cluster_policy.go
+++ b/policies/resource_cluster_policy.go
@@ -3,11 +3,26 @@ package policies
 import (
 	"context"
 
+	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/service/compute"
 	"github.com/databricks/terraform-provider-databricks/common"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
+
+func isBuiltinPolicyFamily(ctx context.Context, w *databricks.WorkspaceClient, familyId, familyName string) (bool, error) {
+	// Fetch supported policy families, and check against it
+	families, err2 := w.PolicyFamilies.ListAll(ctx, compute.ListPolicyFamiliesRequest{})
+	if err2 != nil {
+		return false, err2
+	}
+	for _, family := range families {
+		if familyId == family.PolicyFamilyId && familyName == family.Name {
+			return true, nil
+		}
+	}
+	return false, nil
+}
 
 // ResourceClusterPolicy ...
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by alexott** [comment]
@mgyucht this is required for already merged exporter functionality

File: policies/resource_cluster_policy.go [hunk 1]
```diff
@@ -66,6 +101,9 @@ func ResourceClusterPolicy() *schema.Resource {
 			var request compute.EditPolicy
 			common.DataToStructPointer(d, s, &request)
 			request.PolicyId = d.Id()
+			if request.PolicyFamilyId != "" {
+				request.Definition = ""
```
hectorcast-db, line 105: Why this? Is to fix previously updated definitions?

hectorcast-db (approved):

policies/resource_cluster_policy.go
alexott: The problem is the following:

1. When we override a cluster policy, then we need to use the `policy_override` field, instead of `definition`, otherwise we'll get an error.
2. When we read policy back after creation, this operation also brings back the `definition`
3. When we're trying to update an instance, then if we send the `definition` that we have in state, then we get an error saying that we need to use `policy_override` with cluster policy families.

PR #2388, opened 2023-06-08
Add run_as and run_as_user_name field to job definition 

gaborratky-db:
## Changes

Added read-only runAsUserName field to job definition and write-only runAs field to job settings definition.

## Tests

- [x] `make test` run locally
- [x] relevant change in `docs/` folder
- [ ] covered with integration tests in `internal/acceptance`
- [ ] relevant acceptance tests are passing
- [ ] using Go SDK

PR #2943, opened 2023-11-21
Remove diff suppression for a job's `run_as` block

pietern:
## Changes

Diff suppression meant that once a nested key was set (either `user_name` or `service_principal_name`), it was unable to unset it.  If both are specified, they are ignored. This meant that once a job was configured with a `run_as` user, it was not possible to configure it with a service principal, and vice versa.

Related PRs:
* #2388
* #2626
* #2765

## Tests

The new integration test passes.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/job_test.go
+++ b/internal/acceptance/job_test.go
@@ -279,3 +279,23 @@ func TestAccJobRunAsServicePrincipal(t *testing.T) {
 		Template: runAsTemplate(`service_principal_name = "` + spId + `"`),
 	})
 }
+
+func TestAccJobRunAsMutations(t *testing.T) {
+	loadDebugEnvIfRunsFromIDE(t, "ucws")
+	spId := GetEnvOrSkipTest(t, "ACCOUNT_LEVEL_SERVICE_PRINCIPAL_ID")
+	unityWorkspaceLevel(
+		t,
+		// Provision job with service principal `run_as`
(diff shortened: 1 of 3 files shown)


pietern:
@dvkuznietsov Do you recall why you needed to use `suppress_diff` in #2626?

dvkuznietsov:
@pietern run_as is a write-only field with a matching read-only field, run_as_username. We want to diff run_as with run_as_username only if run_as is configured by the customer and suppress_diff seemed appropriate to achieve that, similarly to run_if

alexott (approved):

alexott (approved):

tanmay-db (approved):

pietern:
@dvkuznietsov Thanks for the context. After chatting with @mgyucht about this, we figured making it a computed field is more appropriate, given that there always is a value, and only if the user-specified value is different, do we need to trigger an update. This also means that if you specify a `run_as` block with the identity of the authenticated user, it won't need to update.

mgyucht (approved):

shreyas-goenka:
@pietern Is this fine?
> This also means that if you specify a run_as block with the identity of the authenticated user, it won't trigger an update
I guess the implicit assumption we are making here is in most automation cases the user who created the job is the authentication user.

PR #2949, opened 2023-11-22
Migrate jobs read to use go-sdk

edwardfeng-db:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -663,46 +663,46 @@ func gitSourceSchema(r *schema.Resource, prefix string) {
 
 var jobGoSdkSchema = common.StructToSchema(jobs.JobSettings{},
 	func(s map[string]*schema.Schema) map[string]*schema.Schema {
-		jobSettingsSchema(&s, "")
-		jobSettingsSchema(&s["task"].Elem.(*schema.Resource).Schema, "task.0.")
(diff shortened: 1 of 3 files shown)


alexott:
How it's planned to handle aliases?

jobs/resource_job.go
```diff
@@ -895,7 +895,15 @@ func ResourceJob() *schema.Resource {
...
+				return err
+			}
+			jobID, err := parseJobId(d.Id())
+			if err != nil {
+				return err
+			}
+			job, err := w.Jobs.GetByJobId(ctx, jobID)
 			if err != nil {
 				return err
 			}
```
hectorcast-db, line 909: In line 911 we do 
```
return common.StructToData(*job.Settings, jobSchema, d)
```
job.Settings is now a different struct, with different tf tags. It may be lack of tests, but I am a bit surprised that this change works this easily.

alexott: I would recommend to not to start with jobs until we figure out how to solve a problem with aliases

hectorcast-db: I don't have context on the aliases and the problems. Can you share a reference?

alexott: The job resource heavily uses `tf:"alias"` to rename fields, for example, in `git` block, we rename `git_url` into `url`, etc.

PR #2952, opened 2023-11-23
Fix databricks_workspace_conf nightly race

mgyucht:
## Changes
There is a race condition causing nightly tests to fail for databricks_workspace_conf. `TestAccWorkspaceConfFullLifecycle` sets enableIpAccessList to true, and `TestAccInvalidWorkspaceConfAreIgnored` sets this to false. However, both tests are operating on the same workspace conf in parallel.

To fix this, I combined both test cases into a single test.

## Tests
The test passes locally. Triggering nightlies to verify it works properly.

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/workspace_conf_test.go
+++ b/internal/acceptance/workspace_conf_test.go
@@ -34,54 +34,46 @@ func TestAccWorkspaceConfFullLifecycle(t *testing.T) {
 			assertEnableIpAccessList(t, "true")
 			return nil
 		},
-	})
-}
-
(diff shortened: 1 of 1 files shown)


alexott (approved):

tanmay-db (approved):
Should we have retries for tests that failed automatically in nightly. It doesn't solve the root cause (tests update shared resource) but could prevent false alarms so less things for oncall to look into.

mgyucht:
Yeah, I'll add retries during the EE fixit week next week.

PR #2953, opened 2023-11-23
Add `webhooks_notifications` on task levels in `databricks_job` resource

### LINKED ISSUE #2951: [ISSUE] webhook_notifications are not supported at task level
Hey there,

webhook_notifications block is not allowed at task level when using terraform. Though this feature is already supported via databricks UI and the api(https://docs.databricks.com/api/workspace/jobs/create).

It would be good to have webhook notifications on individual tasks of a job.

alexott:
## Changes

Also, replaced email notifications with corresponding structs from Go SDK to make it matching to the API specification, and updated docs correspondingly

This fixes #2951

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -118,16 +118,6 @@ type RunJobTask struct {
 	JobParameters map[string]string `json:"job_parameters,omitempty"`
 }
 
-// EmailNotifications contains the information for email notifications after job or task run start or completion
-type EmailNotifications struct {
-	OnStart                            []string `json:"on_start,omitempty"`
-	OnSuccess                          []string `json:"on_success,omitempty"`
-	OnFailure                          []string `json:"on_failure,omitempty"`
-	OnDurationWarningThresholdExceeded []string `json:"on_duration_warning_threshold_exceeded,omitempty"`
-	NoAlertForSkippedRuns              bool     `json:"no_alert_for_skipped_runs,omitempty"`
-	AlertOnLastAttempt                 bool     `json:"alert_on_last_attempt,omitempty"`
-}
-
 // WebhookNotifications contains the information for webhook notifications sent after job start or completion.
 type WebhookNotifications struct {
 	OnStart                            []Webhook `json:"on_start,omitempty"`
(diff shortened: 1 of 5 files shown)


jobs/resource_job.go
```diff
@@ -118,16 +118,6 @@ type RunJobTask struct {
 	JobParameters map[string]string `json:"job_parameters,omitempty"`
 }
 
-// EmailNotifications contains the information for email notifications after job or task run start or completion
```
mgyucht, line 121: Woohoo! Thanks

jobs/resource_job.go
```diff
@@ -208,7 +198,8 @@ type JobTaskSettings struct {
 	// ConditionTask is in private preview
 	ConditionTask *jobs.ConditionTask `json:"condition_task,omitempty" tf:"group:task_type"`
 
-	EmailNotifications     *EmailNotifications            `json:"email_notifications,omitempty" tf:"suppress_diff"`
+	EmailNotifications     *jobs.TaskEmailNotifications   `json:"email_notifications,omitempty" tf:"suppress_diff"`
+	WebhookNotifications   *WebhookNotifications          `json:"webhook_notifications,omitempty" tf:"suppress_diff"`
```
mgyucht, line 202: Can we not use the SDK for Webhook Notifications? If you need the Sort() method, I was thinking something like
```go
type WebhookNotifications struct {
  *jobs.WebhookNotifications
}

func (w *WebhookNotifications) Sort() {
}
```
could do the trick

mgyucht (approved):
LGTM, one small question

jobs/resource_job.go
alexott: I'm thinking if we ever need sorting - I think that we need to ask Jobs team to check if they are sorting the submitted values?

alexott: @mgyucht we still need to sort, but I reimplemented it a bit differently - please look to updates

mgyucht (approved):
That's perfect, thanks!

PR #2958, opened 2023-11-26
[NOT MERGED] Migrate library delete to use go-sdk

edwardfeng-db:
## Changes

- Migrate delete handler of `resource_library` to use `go-sdk`

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_library_test.go
+++ b/clusters/resource_library_test.go
@@ -65,6 +65,14 @@ func TestLibraryDelete(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:       "GET",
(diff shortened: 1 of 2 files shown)


clusters/resource_library_test.go
```diff
@@ -71,6 +72,13 @@ func TestLibraryDelete(t *testing.T) {
 					State: ClusterStateRunning,
 				},
 			},
+			{
+				Method:   "POST",
```
mgyucht, line 76: Just double checking here: if the cluster is already running (as is returned in the last response), why do we try to start it again?

edwardfeng-db: Yeah good point let me change that

PR #2960, opened 2023-11-27
Migrate delete to use go-sdk for library resource

edwardfeng-db:
## Changes

- Add `clusters_api_sdk.go` to implement same functionalities as in `clusters_api.go` with the go-sdk
- Migrate library delete to use the go sdk implementation

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/clusters_api_sdk.go
+++ b/clusters/clusters_api_sdk.go
@@ -5,22 +5,12 @@ import (
 	"log"
 	"time"
 
+	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/service/compute"
-	"github.com/databricks/terraform-provider-databricks/common"
 )
 
-// Start a terminated Spark cluster given its ID and wait till it's running
-func StartCluster(ctx context.Context, c *common.DatabricksClient, clusterID string) error {
-	_, err := StartClusterAndGetInfo(ctx, c, clusterID)
-	return err
-}
-
 // StartAndGetInfo starts cluster and returns info
-func StartClusterAndGetInfo(ctx context.Context, c *common.DatabricksClient, clusterID string) (*compute.ClusterDetails, error) {
-	w, err := c.WorkspaceClient()
-	if err != nil {
-		return nil, err
-	}
+func StartClusterAndGetInfo(ctx context.Context, w *databricks.WorkspaceClient, clusterID string) (*compute.ClusterDetails, error) {
 	cluster, err := w.Clusters.GetByClusterId(ctx, clusterID)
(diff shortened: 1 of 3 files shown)


edwardfeng-db:
@alexott Sorry for the confusion, I closed that one

clusters/clusters_api_sdk_test.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/terraform-provider-databricks/qa"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
edwardfeng-db, line 12: These tests are the same as the ones in `clusters_api_test.go`, but with the `go-sdk`

clusters/clusters_api_sdk.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/terraform-provider-databricks/common"
+)
+
+// Start a terminated Spark cluster given its ID and wait till it's running
+func StartCluster(ctx context.Context, c *common.DatabricksClient, clusterID string) error {
+	_, err := StartClusterAndGetInfo(ctx, c, clusterID)
+	return err
+}
+
```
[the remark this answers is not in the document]
mgyucht: Is there really a need for both functions? Let's just have one; the caller can ignore the first returned value if it isn't relevant.

clusters/clusters_api_sdk.go
```diff
@@ -0,0 +1,47 @@
...
+// StartAndGetInfo starts cluster and returns info
+func StartClusterAndGetInfo(ctx context.Context, c *common.DatabricksClient, clusterID string) (*compute.ClusterDetails, error) {
```
mgyucht, line 19: Can we use WorkspaceClient here directly rather than `common.DatabricksClient`?

mgyucht (approved):
Couple small nits, otherwise LGTM!

clusters/clusters_api_sdk.go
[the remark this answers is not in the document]
edwardfeng-db: Yeah sounds good, we can start with just one. I created the wrapper to keep consistency with the http client code path but I guess it's not necessary for now

clusters/clusters_api_sdk.go
edwardfeng-db: Updated

PR #2962, opened 2023-11-27
refactor `databricks_mws_credentials` to Go SDK

nkvuong:
## Tests

- [x] `make test` run locally
- [x] using Go SDK


docs/resources/mws_credentials.md
```diff
@@ -49,7 +49,7 @@ resource "databricks_mws_credentials" "this" {
 
 The following arguments are required:
 
-* `account_id` - (Required) Account Id that could be found in the bottom left corner of [Accounts Console](https://accounts.cloud.databricks.com/)
+* `account_id` - (Optional) Account Id that could be found in the bottom left corner of [Accounts Console](https://accounts.cloud.databricks.com/)
```
alexott, line 52: It's now in the top right corner

docs/resources/mws_credentials.md
alexott, line 1: Should we remove this note? "This resource has an evolving API, which will change in the upcoming versions of the provider in order to simplify user experience."

mws/resource_mws_credentials.go
```diff
@@ -59,71 +29,81 @@ func (a CredentialsAPI) List(mwsAcctID string) ([]Credentials, error) {
 	return mwsCredsList, err
 }
 
+type CredentialInfo struct {
+	// The human-readable name of the credential configuration object.
+	AccountId string `json:"account_id,omitempty"`
```
alexott, line 34: do we need `suppress_diff` here?

mws/resource_mws_credentials.go
```diff
@@ -59,71 +29,81 @@ func (a CredentialsAPI) List(mwsAcctID string) ([]Credentials, error) {
...
+			})
 			if err != nil {
 				return err
 			}
-			d.Set("credentials_id", credentials.CredentialsID)
+			d.Set("credentials_id", credentials.CredentialsId)
+			_, hasAccountId := d.GetOk("account_id")
+			if !hasAccountId {
+				d.Set("account_id", c.Config.AccountID)
+			}
```
alexott, line 74: If AccountID is always returned, we can save some lines by setting it without condition

alexott (commented):
in general looks good, just small comments

docs/resources/mws_credentials.md
nkvuong: done

mws/resource_mws_credentials.go
nkvuong: yes, we do, added

mws/resource_mws_credentials.go
nkvuong: done

mws/resource_mws_credentials.go
```diff
@@ -59,71 +29,82 @@ func (a CredentialsAPI) List(mwsAcctID string) ([]Credentials, error) {
 	return mwsCredsList, err
 }
 
+type CredentialInfo struct {
```
hectorcast-db, line 32: I see that you are creating this because the one in the SDK does not match the TF. But this also means that we still need to maintain this one by hand (e.g., when a new field is added). Is there anyway we could improve that?

nkvuong: this would need a more generic function to combine the `CreateCredentialRequest` struct and `Credential` for the responses. 

Additionally, this was a very old resource, so the tf schema flattens the struct `AwsCredentials`. Any changes within this will need maintenance by hand.

hectorcast-db (approved):

PR #2963, opened 2023-11-28
Added search by `display_name` to `databricks_service_principal` data source

### LINKED ISSUE #2959: [FEATURE] Add support for searching by service principal name in `databricks_service_principal` data source

alexott:
## Changes

Allow searching not only by `application_id`, but also by the exact `display_name`.

This fixes #2959

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/data_service_principal.go
+++ b/scim/data_service_principal.go
@@ -37,7 +37,11 @@ func DataSourceServicePrincipal() *schema.Resource {
 			return err
 		}
 		if len(spList) == 0 {
-			return fmt.Errorf("cannot find SP with ID %s", response.ApplicationID)
+			if response.ApplicationID != "" {
+				return fmt.Errorf("cannot find SP with ID %s", response.ApplicationID)
+			} else {
+				return fmt.Errorf("cannot find SP with name %s", response.DisplayName)
+			}
 		} else if len(spList) > 1 {
 			return fmt.Errorf("there are more than 1 service principal with name %s", response.DisplayName)
 		}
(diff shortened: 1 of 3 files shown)


scim/data_service_principal.go
```diff
@@ -24,15 +24,27 @@ func DataSourceServicePrincipal() *schema.Resource {
 	return common.DataResource(spnData{}, func(ctx context.Context, e any, c *common.DatabricksClient) error {
 		response := e.(*spnData)
 		spnAPI := NewServicePrincipalsAPI(ctx, c)
```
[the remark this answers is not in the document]
alexott: I need to look at how much effort it will require, because it should work on both account & workspace level

scim/data_service_principal.go
```diff
@@ -24,15 +24,31 @@ func DataSourceServicePrincipal() *schema.Resource {
...
-		spList, err := spnAPI.Filter(fmt.Sprintf("applicationId eq '%s'", response.ApplicationID), true)
+		var spList []User
+		var err error
+		if response.ApplicationID != "" {
+			spList, err = spnAPI.Filter(fmt.Sprintf("applicationId eq '%s'", response.ApplicationID), true)
```
tanmay-db, line 30: In case both of them are used, then we prefer the application id. Should we specify this in docs? Also should we throw error if application id doesn't match with display name?

tanmay-db (approved):

scim/data_service_principal.go
alexott: ah, yes - we need to throw an error...

alexott: @tanmay-db fixed

tanmay-db: Thank you

PR #2983, opened 2023-12-03
Exporter: generate cluster libraries blocks directly in `databricks_cluster`, don't generate `databricks_library` resources

alexott:
## Changes

Instead of generating separate `databricks_library` resources, cluster libraries are now generated inside of the `databricks_cluster` definition.  This decreases the number of the generated resources that is important for big workspaces as it decreases the state size.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -199,7 +199,7 @@ func resourceClusterRead(ctx context.Context, d *schema.ResourceData, c *common.
 		return err
 	}
 	d.Set("url", c.FormatURL("#setting/clusters/", d.Id(), "/configuration"))
-	shouldSkipLibrariesRead := common.IsExporter(ctx)
+	shouldSkipLibrariesRead := !common.IsExporter(ctx)
 	if d.Get("library.#").(int) == 0 && shouldSkipLibrariesRead {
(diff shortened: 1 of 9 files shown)


clusters/resource_cluster.go
```diff
@@ -199,7 +199,14 @@ func resourceClusterRead(ctx context.Context, d *schema.ResourceData, c *common.
...
 	}
 	d.Set("url", c.FormatURL("#setting/clusters/", d.Id(), "/configuration"))
-	if d.Get("library.#").(int) == 0 {
+	var isExporter bool
+	p, ok := ctx.Value(common.Provider).(*schema.Provider)
+	if !ok {
+		isExporter = false
+	} else {
+		isExporter = p.TerraformVersion == "exporter"
+	}
```
mgyucht, line 208: Suggestion: let's make this a function in the `common` package so that we can reuse it in the future.

mgyucht (approved):
LGTM, one suggestion to refactor the isExporter logic to a common place.

clusters/resource_cluster.go
alexott: Done...

PR #2985, opened 2023-12-03
Migrate library read to use go-sdk

edwardfeng-db:
## Changes

- Migrate library read method to use the go-sdk
- Added a `WaitForLibrariesInstalledSdk` method to mirror the one currently used, but implemented with the go-sdk

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_library.go
+++ b/clusters/resource_library.go
@@ -35,7 +35,7 @@ func ResourceLibrary() *schema.Resource {
 		return split[0], split[1]
 	}
 	return common.Resource{
-		Schema: s,
+		Schema: libraySdkSchema,
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			clusterID := d.Get("cluster_id").(string)
 			err := NewClustersAPI(ctx, c).Start(clusterID)
--- a/clusters/resource_library.go
+++ b/clusters/resource_library.go
@@ -20,6 +20,13 @@ func ResourceLibrary() *schema.Resource {
 		}
 		return m
 	})
+	libraySdkSchema := common.StructToSchema(compute.Library{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		m["cluster_id"] = &schema.Schema{
+			Type:     schema.TypeString,
+			Required: true,
+		}
+		return m
+	})
 	parseId := func(id string) (string, string) {
 		split := strings.SplitN(id, "/", 2)
 		if len(split) != 2 {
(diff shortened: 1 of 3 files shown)


libraries/libraries_api_sdk.go
```diff
@@ -0,0 +1,76 @@
...
+				return resource.NonRetryableError(err)
+			}
+			if apiErr.StatusCode != 404 && strings.Contains(apiErr.Message,
+				fmt.Sprintf("Cluster %s does not exist", wait.ClusterID)) {
+				apiErr.StatusCode = 404
+			}
+			return resource.NonRetryableError(apiErr)
+		}
+		if !wait.IsRunning {
+			log.Printf("[INFO] Cluster %s is currently not running, so just returning list of %d libraries",
```
edwardfeng-db, line 33: I copied this from the existing funciton but this logging statement seems to be inaccurate? I think instead of saying "Cluster is currently not running", we should probably print something like "we don't have to wait for pending libraries, so just returning the list of libraries"

libraries/libraries_api_sdk.go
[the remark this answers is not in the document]
edwardfeng-db: Hmm, I think it's actually cleaner with these in a separate file, so it's easy to see which methods and structs belong to which category. 

After the migration we can remove the original file, and we can do some renaming (maybe remove the `_sdk` suffix). 

What do you think?

clusters/resource_library.go
```diff
@@ -20,6 +20,13 @@ func ResourceLibrary() *schema.Resource {
 		}
 		return m
 	})
+	libraySdkSchema := common.StructToSchema(compute.Library{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		m["cluster_id"] = &schema.Schema{
+			Type:     schema.TypeString,
+			Required: true,
+		}
+		return m
+	})
```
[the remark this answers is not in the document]
mgyucht: Can we use this schema in place of `s` on line 38?

clusters/resource_library.go
```diff
@@ -58,18 +65,21 @@ func ResourceLibrary() *schema.Resource {
 		},
 		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			clusterID, libraryRep := parseId(d.Id())
-			cll, err := libraries.NewLibrariesAPI(ctx, c).WaitForLibrariesInstalled(libraries.Wait{
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			cll, err := libraries.WaitForLibrariesInstalledSdk(ctx, w, compute.Wait{
```
mgyucht, line 72: Do we ever set `IsRunning` on this wait?

libraries/libraries_api_sdk.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+	"time"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func WaitForLibrariesInstalledSdk(ctx context.Context, w *databricks.WorkspaceClient, wait compute.Wait, timeout time.Duration) (result *compute.ClusterLibraryStatuses, err error) {
```
mgyucht, line 17: Let's add a doccomment here

libraries/libraries_api_sdk.go
```diff
@@ -0,0 +1,75 @@
...
+				log.Printf("[WARN] Removing failed library %s from %s", v.Library, wait.ClusterID)
+				cleanup.Libraries = append(cleanup.Libraries, *v.Library)
+				continue
+			}
+			installed = append(installed, v)
+		}
+		// and result contains only the libraries that were successfully installed
+		result.LibraryStatuses = installed
+		if len(cleanup.Libraries) > 0 {
+			w.Libraries.Uninstall(ctx, cleanup)
```
mgyucht, line 68: Wow, a read method will change libraries installed on a cluster... this is kind of wild...

mgyucht (commented):
Couple comments

clusters/resource_library.go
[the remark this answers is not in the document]
edwardfeng-db: Yep sounds good, I think this schema is exactly the same as `s` so there shouldn't be any problem updating that

libraries/libraries_api_sdk.go
edwardfeng-db: Yeah, not sure why we need this, and why this api is designed this way

clusters/resource_library.go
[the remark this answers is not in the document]
edwardfeng-db: Updated and confirmed that there's no schema diff

PR #2988, opened 2023-12-04
Added `filter` block to `databricks_instance_profiles` data source

ehab-eb:
## Changes

This PR adds filtering capabilities to the `databricks_instance_profiles` data source. A new optional block `filter` is introduced that exposes two arguments:

1. `name`: This takes the name of the instance profile attribute to filter on
2. `pattern`: This takes a regex pattern to use for filtering based on the field specified in the `name` argument

Only one `filter` block can be specified.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/aws/data_instance_profiles.go
+++ b/aws/data_instance_profiles.go
@@ -2,36 +2,81 @@ package aws
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/exp/slices"
 )
 
+func instanceProfileMatchesFilter(ip *instanceProfileData, filter *instanceProfileFilter) bool {
+	var ipMap map[string]interface{}
+	m, _ := json.Marshal(ip)
+	_ = json.Unmarshal(m, &ipMap)
+	val := ipMap[filter.Name]
+	stringVal := fmt.Sprint(val)
+	re := regexp.MustCompile(filter.Pattern)
+	return re.Match([]byte(regexp.QuoteMeta(stringVal)))
+}
+
+type instanceProfileData struct {
+	Name    string `json:"name"`
+	Arn     string `json:"arn"`
+	RoleArn string `json:"role_arn"`
+	IsMeta  bool   `json:"is_meta"`
+}
+
+type instanceProfileFilter struct {
+	Name    string `json:"name"`
+	Pattern string `json:"pattern"`
+}
+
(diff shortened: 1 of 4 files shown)


aws/data_instance_profiles.go
```diff
@@ -2,36 +2,89 @@ package aws
...
-		RoleArn string `json:"role_arn,omitempty" tf:"computed"`
-		IsMeta  bool   `json:"is_meta,omitempty" tf:"computed"`
+func contains[T comparable](s []T, e T) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
 	}
+	return false
+}
```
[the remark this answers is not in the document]
ehab-eb: I've removed the dependnecy on the custom function altogether based on @mgyucht's [suggestion](https://github.com/databricks/terraform-provider-databricks/pull/2990#pullrequestreview-1762738675).

aws/data_instance_profiles.go
```diff
@@ -2,36 +2,81 @@ package aws
...
+
+		if data.Filter != (instanceProfileFilter{}) {
+			if data.Filter.Pattern == "" {
+				return fmt.Errorf("field `pattern` cannot be empty")
+			}
+			var fieldNames []string
+			val := reflect.ValueOf(instanceProfileData{})
+			for i := 0; i < val.Type().NumField(); i++ {
+				fieldNames = append(fieldNames, val.Type().Field(i).Tag.Get("json"))
+			}
```
alexott, line 53: We can avoid having this if we expose `listAllFields` function implemented in #3044 - in this case it will handle inherited fields correctly

aws/data_instance_profiles.go
```diff
@@ -2,36 +2,81 @@ package aws
...
+			if !slices.Contains(fieldNames, data.Filter.Name) {
+				if data.Filter.Name == "" {
+					return fmt.Errorf("field `name` cannot be empty")
+				}
```
alexott, line 57: I think that we can move this condition after the line 48

aws/data_instance_profiles.go
```diff
@@ -2,36 +2,81 @@ package aws
...
+	"golang.org/x/exp/slices"
 )
 
+func instanceProfileMatchesFilter(ip *instanceProfileData, filter *instanceProfileFilter) bool {
+	var ipMap map[string]interface{}
+	m, _ := json.Marshal(ip)
+	_ = json.Unmarshal(m, &ipMap)
+	val := ipMap[filter.Name]
+	stringVal := fmt.Sprint(val)
+	re := regexp.MustCompile(filter.Pattern)
```
alexott, line 23: It's better to use `.Compile` and check for the error. Also, it's very inefficient - it's better to check validity of the regexp earlier, around line 46, compile regexp, and pass it explicitly. Otherwise we'll compile it on each function invocation.

alexott (changes requested):

tanmay-db (commented):

PR #2990, opened 2023-12-04
Moved `contains` function to common package

ehab-eb:
## Changes

Moving the `contains` function from the `catalog` module to `common` to be used across packages
## Tests

- [x] `make test` run locally
- ~[ ] relevant change in `docs/` folder~
- ~[ ] covered with integration tests in `internal/acceptance`~
- ~[ ] relevant acceptance tests are passing~
- ~[ ] using Go SDK~

### DIFF
--- a/catalog/update.go
+++ b/catalog/update.go
@@ -8,22 +8,13 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func contains[T comparable](s []T, e T) bool {
-	for _, a := range s {
-		if a == e {
-			return true
-		}
-	}
-	return false
-}
-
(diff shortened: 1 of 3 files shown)


mgyucht (commented):
Just curious, can we use https://pkg.go.dev/golang.org/x/exp/slices#Contains? I think it is almost exactly the same thing.

ehab-eb:
I see no reason not to. The repo is already part of the project so there will be no new dependencies and the functionality is exactly what's needed. I've gone ahead and replaced it.

I only opened this PR to aid with my other PR #2988 - If I'm omitting the custom function altogether then this PR is no longer a dependency, but maybe it's worth just dealing with it now since we're on the topic :)

PR #2991, opened 2023-12-04
Add integration test coverage for delta sharing parameters

mgyucht:
## Changes
As a follow up to a regression with delta sharing parameters in the Terraform provider, this PR adds more integration test coverage to the metastore resource to ensure that we submit `0` as a parameter explicitly for `delta_sharing_recipient_token_lifetime_in_seconds`.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/metastore_test.go
+++ b/internal/acceptance/metastore_test.go
@@ -1,73 +1,90 @@
 package acceptance
 
 import (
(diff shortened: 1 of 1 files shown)


mgyucht:
Passed nightlies. One failure unrelated to this test case.

PR #2997, opened 2023-12-06
Introduce Generic Settings Resource

mgyucht:
## Changes
This is an experimental PR for a generic settings resource in the TF provider. Because all Settings API settings have a consistent schema & mechanism for interaction, the definition of a setting resource in Terraform can also be made generic with several simple assumptions about each setting:

1. Each setting needs a name (in this case, the name would be "default_namespace", so the resource name would be "databricks_default_namespace_setting") and a field mask used to update the value for that setting.
2. Each setting needs a structure T containing an Etag field.
3. Each setting has an API to read the setting by Etag.
4. Each setting has an API to delete the setting by Etag.
5. Each setting has a single API to update the setting. The update request must have an AllowMissing boolean field, a Setting field of type *T, and a field mask.

These above requirements are encapsulated in the `genericSettingDefinition` interface. Because API calls require a specific client depending on whether the setting is an account-level or workspace-level setting, we define two interfaces, one for workspace-level settings where the workspace client is provided, and one for account-level settings where the account client is provided.

After this change, to add a new setting:
1. Create a new instance of either the `workspaceSetting` or `accountSetting` for your setting. The type parameter should be your setting resource.
```go
var defaultNamespaceSetting = workspaceSetting[settings.DefaultNamespaceSetting]{
	settingStruct: settings.DefaultNamespaceSetting{},
	readFunc: func(ctx context.Context, w *databricks.WorkspaceClient, etag string) (*settings.DefaultNamespaceSetting, error) {
		return w.Settings.ReadDefaultWorkspaceNamespace(ctx, settings.ReadDefaultWorkspaceNamespaceRequest{
			Etag: etag,
		})
	},
	updateFunc: func(ctx context.Context, w *databricks.WorkspaceClient, t settings.DefaultNamespaceSetting) (string, error) {
		t.SettingName = "default"
		res, err := w.Settings.UpdateDefaultWorkspaceNamespace(ctx, settings.UpdateDefaultWorkspaceNamespaceRequest{
			AllowMissing: true,
			Setting:      &t,
			FieldMask:    "namespace.value",
		})
		if err != nil {
			return "", err
		}
		return res.Etag, err
	},
	deleteFunc: func(ctx context.Context, w *databricks.WorkspaceClient, etag string) (string, error) {
		res, err := w.Settings.DeleteDefaultWorkspaceNamespace(ctx, settings.DeleteDefaultWorkspaceNamespaceRequest{
			Etag: etag,
		})
		if err != nil {
			return "", err
		}
		return res.Etag, err
	},
}
```
2. Add a new constant for the name of the setting and to the AllSettingsResources() function:
```go
func AllSettingsResources() map[string]*schema.Resource {
  return {
    ...
    "my_new_setting": makeSettingResource[settings.MyNewSetting, *databricks.WorkspaceClient](myNewSetting),
  }
}
```

Additionally, I migrated the tests for this resource to use the Go SDK.

## Tests

- [x] Unit tests verify that the generic setting 
- [x] Process for change is documented in this PR description as well as in `all_settings.go`.
- [x] using Go SDK


alexott (commented):
If the schema is consistent, I'm wondering if we need to have individual resources, because we may have dozens of settings and this may lead to the proliferation of resources and resource instances that may significantly slow down Terraform in the big installations.

Why not define two resources - one for workspace settings, and another - for account? Both will support patch method, not put, so it will be possible to define multiple instances if necessary.

deryasusm-db (commented):
This should be tested by running the TF integration test for Default Namespace against a test shard

mgyucht:
@alexott We could also do that, however, 1) it probably has less to do with Terraform than it does with the API itself, as there are still many API calls that need to be made to retrieve all values of all settings. There definitely will be TF overhead, but I suspect it would be comparable or maybe even smaller than the cost of making requests to the backend. 2) This keeps TF as close as possible to the REST API, which is generally a design goal. I think for consistency & as a general principle we should keep these as close as possible to one another.

settings/resource_default_namespace_setting.go
```diff
@@ -46,130 +45,193 @@ func (a DefaultNamespaceSettingsAPI) getEtagFromError(err error) (string, error)
...
-		return "", err
-	}
-	return response.Etag, nil
+type workspaceSettingDefinition[T any] genericSettingDefinition[T, *databricks.WorkspaceClient]
+type accountSettingDefinition[T any] genericSettingDefinition[T, *databricks.AccountClient]
+
+// Instructions for adding a new setting:
+//
+//  1. Add a new setting name constant. The resulting Terraform resource name will be "databricks_<your settings_name>_setting".
+//  2. Create a struct corresponding to your setting, and implement either the workspaceSettingDefinition or accountSettingDefinition interface.
```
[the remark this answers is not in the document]
mgyucht: The latter, I can clarify that.

deryasusm-db:
@hectorcast-db Can you review this one as you did the TF integration for Default Namespace setting? Thanks.

settings/generic_setting.go
```diff
@@ -0,0 +1,280 @@
...
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			var res *T
+			switch defn := defn.(type) {
+			case workspaceSettingDefinition[T]:
+				w, err := c.WorkspaceClient()
+				if err != nil {
+					return err
+				}
+				res, err = retryOnEtagError(func(etag string) (*T, error) { return defn.Read(ctx, w, etag) }, d.Id(), func(req *string, newEtag string) { *req = newEtag })
```
hectorcast-db, line 220: Can read fail on a EtagError?
From the docs:

The server will accept any etag and respond with a response which is at least as recent as the etag.

settings/generic_setting.go
```diff
@@ -0,0 +1,280 @@
...
+		},
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			var etag string
+			switch defn := defn.(type) {
+			case workspaceSettingDefinition[T]:
+				w, err := c.WorkspaceClient()
+				if err != nil {
+					return err
+				}
+				etag, err = retryOnEtagError(func(etag string) (string, error) { return defn.Delete(ctx, w, etag) }, d.Id(), func(req *string, newEtag string) { *req = newEtag })
```
alexott, line 260: Do we really need to retry on the NotFound error? If it's not found then this typically it means that it was removed by external agent, so we just need to record deletion...

mgyucht: Good point. Removed retry on `Not Found`, so we only retry on conflict.

settings/generic_setting.go
mgyucht: Good point, removed need to retry in reads.

hectorcast-db (approved):

PR #2998, opened 2023-12-06
Migrate library create to use go-sdk

edwardfeng-db:
## Changes

- Migrate library create to use go-sdk
- Remove the old schema `s` as it's no longer used

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_library.go
+++ b/clusters/resource_library.go
@@ -13,13 +13,6 @@ import (
 )
 
 func ResourceLibrary() *schema.Resource {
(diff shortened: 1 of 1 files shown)


mgyucht (approved):
Excellent! Are there still references to libraries.Library, or can we remove that as well?

edwardfeng-db:
> Are there still references to libraries.Library, or can we remove that as well?
Will take a look and remove if possible in a follow-up PR. I think there might be references from cluster resources.

PR #3006, opened 2023-12-08
Exporter: add emitting of libraries that are workspace files

alexott:
## Changes

We now support libraries that are uploaded into the Databricks workspace as files (`databricks_workspace_file` resource). We now support emitting them in jobs, clusters & cluster policies.

Also added the `workspace_path` attribute to the `databricks_repo` and `databricks_directory` resources.

## Tests

- [x] `make test` run locally
- [x] tested manually
- [ ] using Go SDK

### DIFF
--- a/workspace/resource_directory.go
+++ b/workspace/resource_directory.go
@@ -32,6 +32,10 @@ func ResourceDirectory() *schema.Resource {
 			Default:  false,
 			Optional: true,
 		},
+		"workspace_path": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
 	}
 
 	directoryRead := func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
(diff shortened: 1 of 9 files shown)


alexott:
This should be merged after #2983

workspace/resource_directory.go
```diff
@@ -44,7 +48,12 @@ func ResourceDirectory() *schema.Resource {
 			d.SetId("")
 			return fmt.Errorf("different object type, %s, on this path other than a directory", objectStatus.ObjectType)
 		}
-		return common.StructToData(objectStatus, s, d)
+		err = common.StructToData(objectStatus, s, d)
+		if err != nil {
+			return err
+		}
+		d.Set("workspace_path", "/Workspace"+d.Id())
```
mgyucht, line 55: Is there any scenario in which this is called before the Id is set for this resource? Maybe we just use `.Path` instead like we do for repos?

mgyucht (approved):
LGTM, one small thought but feel free to merge if you think this is ok>

workspace/resource_directory.go
alexott: No, for Read operation `d.Id()` is always set - see `objectStatus, err := notebooksAPI.Read(d.Id())` earlier... Also, for directories `path == d.Id()`

PR #3007, opened 2023-12-08
Documentation: consistent header levels and same doc structure

alexott:
## Changes

In a number of resources the header levels were inconsistent for documentation about blocks (some were on the 2nd level, some on the 3rd, while most of the resources use the 3rd level).  Also, changed the order of sections a bit to get a uniform order of sections.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -405,7 +405,7 @@ resource "databricks_job" "sql_aggregation_job" {
 }
 ```
 
-### Exported attributes
+## Attribute Reference
 
 In addition to all arguments above, the following attributes are exported:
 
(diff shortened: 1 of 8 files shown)


docs/resources/job.md
```diff
@@ -405,7 +405,7 @@ resource "databricks_job" "sql_aggregation_job" {
 }
 ```
 
-### Exported attributes
```
nkvuong, line 408: Have you replaced all the instances of this in the doc?

nkvuong (approved):
1 small comment, otherwise lgtm

docs/resources/job.md
alexott: yes, it was only one place with such a name - everything else uses "Attribute Reference"

PR #3024, opened 2023-12-12
Add resource `databricks_grant` for managing singular principal

### LINKED ISSUE #2704: [ISSUE] different databricks_grants inside different modules overwrite eachother
Hello,

I have a multi module setup and I'm applying multiple databricks_grants resource blocks (one or more in each module) 

### Configuration

Module A: 

```hcl
resource "databricks_grants" "use_catalog" {
  provider  = databricks.workspace
  catalog   = "strada"
  grant {
    principal  = "Group 1
    privileges = ["USE_CATALOG"]
  }
}
```

Module B:
```hcl
resource "databricks_grants" "use_catalog" {
  provider  = databricks.workspace
  catalog   = "strada"
  grant {
    principal  = "Group 2"
    privileges = ["USE_CATALOG"]
  }
}
```

### Expected Behavior
In this case, databricks_grants.use_catalog in Module A and databricks_grants.use_catalog in Module B have different addresses due to the module prefix. This should ensure that they are treated as separate resources and will not overwrite each other during the apply process.

### Actual Behavior
Every `terraform apply` i run, they keep changing from group 1 to group 2.

martin-walsh:
## Changes
Add resource "databricks_grant" for managing grants for a single principal on a single securable. This allows securing of resources in distributed terraform repos e.g. granting an api user SELECT on a table without wiping out centrally managed privileges such as human user access.

Opening up PR to discuss whether the approach is acceptable to you and will add documentation, and clean up if it is.

Fixes #2704

## Tests

- [x] `make test` run locally
- [x] using Go SDK


martin-walsh:
@Jonathan-Choi FYI

catalog/resource_grant.go
```diff
@@ -0,0 +1,187 @@
...
+	return common.Resource{
+		Schema: s,
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff) error {
+			if d.Id() == "" {
+				// unfortunately we cannot do validation before dependent resources exist with tfsdkv2
+				return nil
+			}
+			var grants permissions.UnityCatalogPermissionsList
+			common.DiffToStructPointer(d, s, &grants)
+			return permissions.Mappings.Validate(d, grants)
```
nkvuong, line 124: we should let server side validation to kick in

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,243 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// API
+func NewUnityCatalogPermissionsAPI(ctx context.Context, m any) UnityCatalogPermissionsAPI {
```
nkvuong, line 13: we should use the go sdk for this, instead of constructing API calls manually

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,243 @@
...
+	if securable == "model" {
+		return fmt.Sprintf("/unity-catalog/permissions/function/%s", name)
+	}
+	return fmt.Sprintf("/unity-catalog/permissions/%s/%s", securable, name)
+}
+
+//Responses
+
+// UnityCatalogPermissionsDiff is the inner structure of updatePermissions RPC
+type UnityCatalogPermissionsDiff struct {
```
nkvuong, line 42: all these structs are available in go sdk as well

catalog/resource_grant.go
```diff
@@ -0,0 +1,187 @@
...
+					},
+				},
+			}
+			securable, name := permissions.Mappings.KeyValue(d)
+			unityCatalogPermissionsAPI := permissions.NewUnityCatalogPermissionsAPI(ctx, c)
+			err := replacePermissionsForPrincipal(unityCatalogPermissionsAPI, securable, name, principal, grants)
+			if err != nil {
+				return err
+			}
+			d.SetId(permissions.Mappings.Id(d))
```
nkvuong, line 143: this id won't be unique - we should use the principal in the id

catalog/resource_grant.go
```diff
@@ -0,0 +1,187 @@
...
+			var grants = permissions.UnityCatalogPermissionsList{
+				Assignments: []permissions.UnityCatalogPrivilegeAssignment{
+					{
+						Principal:  principal,
+						Privileges: privileges,
+					},
+				},
+			}
+			securable, name := permissions.Mappings.KeyValue(d)
+			unityCatalogPermissionsAPI := permissions.NewUnityCatalogPermissionsAPI(ctx, c)
```
nkvuong, line 138: for the new resource, we should use go sdk to avoid tech debt. `databricks_grants` can be refactored later

hectorcast-db:
There is an issue with the current implementation that we are fixing here
https://github.com/databricks/terraform-provider-databricks/pull/3026

This will need to be adjusted here after we merged.

catalog/resource_grant.go
```diff
@@ -0,0 +1,187 @@
...
+	for _, v := range in.Assignments {
+		if v.Principal == principal {
+			out.Assignments = append(out.Assignments, v)
+		}
+	}
+	return
+}
+
+func ResourceGrant() *schema.Resource {
+	s := map[string]*schema.Schema{
```
hectorcast-db, line 87: We should use `UnityCatalogPrivilegeAssignment` and `common.StructToSchema` , like in the ResourceGrants() resource

catalog/resource_grants.go
```diff
@@ -7,62 +7,37 @@ import (
...
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-type PermissionsAPI struct {
-	client  *common.DatabricksClient
-	context context.Context
-}
-
 // permissionsDiff is the inner structure of updatePermissions RPC
```
hectorcast-db, line 16: Nit, move this description

catalog/resource_grant.go
martin-walsh: I matched the impl to that which exists in `resource_grants`. Should I remove it there also for consistency?

catalog/resource_grant.go
```diff
@@ -0,0 +1,187 @@
...
+			if err != nil {
+				return err
+			}
+			d.SetId(permissions.Mappings.Id(d))
+			return nil
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			principal := d.Get("principal").(string)
+			split := strings.SplitN(d.Id(), "/", 2)
+			if len(split) != 2 {
```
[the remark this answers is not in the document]
martin-walsh: Sure. Matched `resource_grants.go`. Should I adjust there also ?

catalog/resource_grant.go
martin-walsh: Sure, I'll take a look. I have a question though, how would I conditionally mark these fields as `Required` & `ForcesNew` if using the reflective method ?

catalog/resource_grant.go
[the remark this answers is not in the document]
martin-walsh: Also, having checked the docs, SplitN is at most n strings so it is an extra layer of safety i.e.
```
fmt.Println(strings.SplitN("abc", "/", 2))

Output:
[abc]
```

catalog/resource_grants.go
martin-walsh: Done

catalog/resource_grant.go
martin-walsh: Done

catalog/resource_grant.go
martin-walsh: Done.

catalog/resource_grant.go
martin-walsh: I'm not sure how I can address this in the scope of this PR. The `common.Resource` struct requires `common.DatabricksClient`

martin-walsh:
Added documentation for `databricks_grant` also.

catalog/resource_grant.go
nkvuong: we should leave it as is for `resource_grants` for backwards compatibility, but should avoid it for the new resource

catalog/resource_grant.go
nkvuong: so you can initialise the sdk client using

```go
w, err := c.WorkspaceClient()
if err != nil {
	return err
}

w.Grants.Update(ctx, catalog.UpdatePermissions{
	Changes:       permissionsChange,
	SecurableType: securable,
	FullName:      name,
})
```

martin-walsh: I made a start (see https://github.com/siteminder-au/terraform-provider-databricks/compare/resource_grant...siteminder-au:terraform-provider-databricks:sdk-update), but I have some concerns that I don't know enough about the backend to make this change. For example, the permissions for shares changes URL & using the new types from the sdk, the StructToSchema generates schema with different Required & Optional settings.

Using the existing `resource_grants.go` as a base, I was comfortable to make the change as it is amounted to little more than verifying that the content of the HTTP requests match what is expected.

I can proceed if you need me to, but I will need some direction

docs/resources/grants.md
```diff
@@ -4,7 +4,7 @@ subcategory: "Unity Catalog"
 # databricks_grants Resource
 
 -> **Note**
-  It is required to define all permissions for a securable in a single resource, otherwise Terraform cannot guarantee config drift prevention.
+  It is required to define all permissions for a securable in a single resource, otherwise Terraform cannot guarantee config drift prevention. See [databricks_grant](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/grant) for managing grants for a single principal.
```
nkvuong, line 7:
```suggestion
Two different resources help you manage your Unity Catalog grants for a securable. Each of these resources serves a different use case:

[databricks_grants](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/grants): Authoritative. Sets the grants of a securable and replaces any existing grants defined inside or outside of Terraform.
[databricks_grant](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/grant): Authoritative for a given principal. Updates the grants of a securable to a single principal. Other principals within the grants for the securables are preserved.
```

catalog/resource_grant.go
martin-walsh: See the diff above for the `resource_grant_test.go` for the share related tests

docs/resources/grant.md
nkvuong, line 1: the content of this page would be mostly duplicated from docs/resources/grants.md, so might make sense to just collapse them in a single page, highlighting the differences
wdyt @alexott

catalog/resource_grant.go
martin-walsh: Re the different Optional & Required behaviour, I narrowed it down to the difference of `omitempty` in jsontag
e.g. existing
```go
// PrivilegeAssignment reflects on `grant` block
type PrivilegeAssignment struct {
	Principal  string   `json:"principal"`
	Privileges []string `json:"privileges" tf:"slice_set"`
}
```

sdk
```go
type PrivilegeAssignment struct {
	// The principal (user email address or group name).
	Principal string `json:"principal,omitempty"`
	// The privileges assigned to the principal.
	Privileges []Privilege `json:"privileges,omitempty"`

	ForceSendFields []string `json:"-"`
}
```

nkvuong: sorry I should have been more specific, for `StructToSchema`, we can leave it to use the same custom struct as `resource_grants` - there are some outstanding work on how we can add additional tf annotation to those structs

it should be easier to swap these API calls
```go
unityCatalogPermissionsAPI := permissions.NewUnityCatalogGrantsAPI(ctx, c)
err := replacePermissionsForPrincipal(unityCatalogPermissionsAPI, securable, name, principal, grants)
```
with the appropriate sdk call (need adjustment for type), e.g.
```go
existing, err := w.Grants.GetBySecurableTypeAndFullName(ctx, catalog.SecurableType(securable), name)
if err != nil {
  return err
}

w.Grants.Update(ctx, catalog.UpdatePermissions{
  Changes:       diffPermissionsForPrincipal(principal, list, existing),
  SecurableType: securable,
  FullName:      name,
})
```

the two resources with special handling here are `shares` (different API call `w.Shares.UpdatePermissions(ctx, sharing.UpdateSharePermissions{})`) and `connection` (reserved TF keyword)

catalog/resource_grant.go
martin-walsh: Done

catalog/resource_grant.go
martin-walsh: Should I revert the changes to `resource_grants` as the code diverges now ?

martin-walsh: @nkvuong I reverted those `resource_grants` changes & refactored `resource_grant` to use the SDK. I kept the `UnityCatalogPermissionsAPI` to hide the difference between Shares & Grants APIs.

martin-walsh: I can bring back the refactor easily enough also & update to use sdk, let me know

martin-walsh: If needed: https://github.com/siteminder-au/terraform-provider-databricks/compare/resource_grant...siteminder-au:terraform-provider-databricks:resource_grants_refactor

nkvuong: you can raise another PR for the refactoring of `resource_grants` to Go SDK - makes it easier to review but also revert (if needed)

nkvuong (commented):
Thank you again for the contribution - the code looks good to me.

We will need to add some integration tests to this. Since there was no integration test for `databricks_grants`, I'll try to add it to a separate PR first and we can then replicate them for `databricks_grant` in this PR

martin-walsh:
@hectorcast-db I ported over the changes from #3026

nkvuong:
I've added some acceptance tests in #3034, if you can port it to this one to validate

martin-walsh:
> @martin-walsh I've added some acceptance tests in #3034, if you can port it to this one to validate

@nkvuong tests ported

docs/resources/grant.md
```diff
@@ -0,0 +1,329 @@
+---
```
nkvuong, line 1: @martin-walsh since this page is mostly duplicated from grants.md, we should collapse the 2 pages into one grants.md, and highlight the different usage of `databricks_grant` and `databricks_grants`

nkvuong (approved):
Integration test passed - looks good to me

One last comment about the doc

docs/resources/grant.md
alexott: Imho, two pages may help users to understand differences. I think that to avoid duplication we can just have links from another page to common things, like, syntax, etc.

martin-walsh: Do you guys need anything more from me here? I'm not really sure how you guys want your documentation structured.

nkvuong: we should put a link on this page for the content that are the same as `databricks_grants.md`, e.g. line 27 onwards. Something like "See [databricks_grant.md] for more usage details

I've highlighted the part that should be called out differently to `databricks_grants`

docs/resources/grant.md
```diff
@@ -0,0 +1,329 @@
...
+Securable objects are hierarchical and privileges are inherited downward. The highest level object that privileges are inherited from is the catalog. This means that granting a privilege on a catalog or schema automatically grants the privilege to all current and future objects within the catalog or schema. Privileges that are granted on a metastore are not inherited.
+
+Every `databricks_grant` resource must have exactly one securable identifier and the following arguments:
+
+- `principal` - User name, group name or service principal application ID.
+- `privileges` - One or more privileges that are specific to a securable type.
+
+For the latest list of privilege types that apply to each securable object in Unity Catalog, please refer to the [official documentation](https://docs.databricks.com/en/data-governance/unity-catalog/manage-privileges/privileges.html#privilege-types-by-securable-object-in-unity-catalog)
+
+Terraform will handle any configuration drift on every `terraform apply` run, even when grants are changed outside of Terraform state.
```
nkvuong, line 23:
```suggestion
Terraform will handle any configuration drift on every `terraform apply` run for the specified principal. even when grants are changed outside of Terraform state.
```

docs/resources/grant.md
martin-walsh: I can't see the highlight. Do you mean remove everything on `databricks_grants.md` from line 27 onwards ? If so, if it were me I would leave it. It is very user friendly to be able to copy valid syntax.

docs/resources/grant.md
```diff
@@ -0,0 +1,329 @@
...
+  name = "my_recipient"
+}
+
+resource "databricks_grant" "some" {
+  share = databricks_share.some.name
+
+  principal  = databricks_recipient.some.name
+  privileges = ["SELECT"]
+}
+```
```
nkvuong, line 325: I was suggesting to leave the HCL examples here, but defer to grants.md on what permissions can be applied to a resource (as future grants will be added)

docs/resources/grant.md
```diff
@@ -0,0 +1,329 @@
...
+Every `databricks_grant` resource must have exactly one securable identifier and the following arguments:
+
+- `principal` - User name, group name or service principal application ID.
+- `privileges` - One or more privileges that are specific to a securable type.
+
+For the latest list of privilege types that apply to each securable object in Unity Catalog, please refer to the [official documentation](https://docs.databricks.com/en/data-governance/unity-catalog/manage-privileges/privileges.html#privilege-types-by-securable-object-in-unity-catalog)
+
+Terraform will handle any configuration drift for the specified principal on every `terraform apply` run, even when grants are changed outside of Terraform state.
+
+Unlike the [SQL specification](https://docs.databricks.com/sql/language-manual/sql-ref-privileges.html#privilege-types), all privileges to be written with underscore instead of space, e.g. `CREATE_TABLE` and not `CREATE TABLE`. Below summarizes which privilege types apply to each securable object in the catalog:
```
nkvuong, line 25:
```suggestion
See [databricks_grants](grants.md) for the list of privilege types that apply to each securable object in the catalog.

## Examples
```

docs/resources/grant.md
martin-walsh: Done

nkvuong (approved):

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,152 @@
...
+func (sm SecurableMapping) Id(d *schema.ResourceData) string {
+	securable, name := sm.KeyValue(d)
+	return fmt.Sprintf("%s/%s", securable, name)
+}
+
+// Mappings
+// See https://docs.databricks.com/api/workspace/grants/update for full list
+// Omitting provider as a reserved keyword
+var Mappings = SecurableMapping{
+	"catalog":            catalog.SecurableType("catalog"),
```
mgyucht, line 106: Not for this PR, but a general remark: seems like it would be useful for the SDK itself to expose a function that returns all enum values for an enum type.

catalog/resource_grant.go
```diff
@@ -0,0 +1,179 @@
...
+			unityCatalogPermissionsAPI := permissions.NewUnityCatalogPermissionsAPI(ctx, c)
+			err := replacePermissionsForPrincipal(unityCatalogPermissionsAPI, securable, name, principal, grants)
+			if err != nil {
+				return err
+			}
+			d.SetId(fmt.Sprintf("%s/%s", permissions.Mappings.Id(d), principal))
+			return nil
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			principal := d.Get("principal").(string)
```
mgyucht, line 139: Suggestion: for readability, could we make helper methods to construct/deconstruct the ID? It's not obvious what each part of the ID ought to correspond to.

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func SetToSlice(set *schema.Set) (ss []catalog.Privilege) {
+	for _, v := range set.List() {
+		ss = append(ss, catalog.Privilege(v.(string)))
+	}
+	return
+}
+
+func ToPrivilegeSlice(in []interface{}) (out []catalog.Privilege) {
```
mgyucht, line 137: Is this actually needed? I think you can replace all calls to this function with SetToSlice() and just not calling `.List()` on the parameter.

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, v := range in {
+		out = append(out, catalog.Privilege(v.(string)))
+	}
+	return
+}
+
+func SliceWithoutString(in []string, without string) (out []string) {
```
mgyucht, line 144: This is the same as `slices.DeleteFunc(in, func(s string) { return s == without })`. Can we use the standard library function instead?

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,152 @@
...
+	"recipient":          catalog.SecurableType("recipient"),
+	"schema":             catalog.SecurableType("schema"),
+	"share":              catalog.SecurableType("share"),
+	"storage_credential": catalog.SecurableType("storage_credential"),
+	"table":              catalog.SecurableType("table"),
+	"volume":             catalog.SecurableType("volume"),
+}
+
+// Utils for Slice and Set
+func SliceToSet(in []catalog.Privilege) *schema.Set {
```
mgyucht, line 122: These are nice! I'm going to think about a way to generalize these so you don't have to implement them for every type. This is great for now though.

catalog/resource_grant.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/catalog/permissions"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// diffPermissionsForPrincipal returns UnityCatalogPermissionsDiff of this permissions list with `diff` privileges removed
+func diffPermissionsForPrincipal(principal string, pl catalog.PermissionsList, existing catalog.PermissionsList) (diff []catalog.PermissionsChange) {
+	// diffs change sets for principal
+	configured := map[string]*schema.Set{}
```
mgyucht, line 20: Maybe I'm misreading this, but because this function only ever operates on one principal, this map will only have one key defined? If so, we can just use `*schema.Set` as opposed to a map. This applies to the other data structures throughout this function.

catalog/permissions/permissions.go
```diff
@@ -0,0 +1,152 @@
...
+			return nil
+		}
+		return retry.RetryableError(
+			fmt.Errorf("permissions for %s-%s are %v, but have to be %v", securable.String(), name, current, desired),
+		)
+	})
+}
+
+// Terraform Schema
+type UnityCatalogPrivilegeAssignment struct {
```
mgyucht, line 69: Can you comment here that this is based off of `catalog.PermissionsList`? As @nkvuong mentioned, in the future we'll refactor these resources to inherit fields from the Go SDK. Knowing what this is derived from will help us track this down in the future.

catalog/resource_grant.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+			grants, err := permissions.NewUnityCatalogPermissionsAPI(ctx, c).GetPermissions(permissions.Mappings.GetSecurableType(split[0]), split[1])
+			grantsForPrincipal := filterPermissionsForPrincipal(*grants, principal)
+			if err != nil {
+				return err
+			}
+			if len(grantsForPrincipal.PrivilegeAssignments) == 0 {
+				return apierr.NotFound("got empty permissions list")
+			}
+			return common.StructToData(grantsForPrincipal, s, d)
```
mgyucht, line 152: Note: here we're calling StructToData on a different datatype than the one that was used in StructToSchema. This works because the catalog.PermissionsList struct contains a subset of the fields of the UnityCatalogPrivilegeAssignment struct, but this could be a bit confusing. Could we return UnityCatalogPrivilegeAssignment from filterPermissionsForPrincipal instead?

catalog/resource_grant_test.go
```diff
@@ -0,0 +1,732 @@
...
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/qa"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGrantPermissionsCornerCases(t *testing.T) {
+	qa.ResourceCornerCases(t, ResourceGrant(), qa.CornerCaseID("schema/sandbox/me"))
+}
+
+func TestResourceGrantCreate(t *testing.T) {
```
mgyucht, line 15: This is really nice that this resource works even if there are existing permissions for a user on the securable.

catalog/resource_grant_test.go
```diff
@@ -0,0 +1,732 @@
...
+		HCL: `
+		table = "foo.bar.baz"
+
+		principal = "me"
+		privileges = ["MODIFY", "SELECT"]
+		`,
+	}.ApplyNoError(t)
+}
+
+func TestResourceGrantUpdateWithChangedPrincipalForcesNewResource(t *testing.T) {
```
mgyucht, line 261: Nice!

catalog/resource_grant_test.go
```diff
@@ -0,0 +1,732 @@
...
+
+func TestResourceGrantCreateNoSecurable(t *testing.T) {
+	qa.ResourceFixture{
+		Resource: ResourceGrant(),
+		Create:   true,
+		HCL: `
+		principal = "me"
+		privileges = ["MODIFY", "SELECT"]
+		`,
+	}.ExpectError(t, "invalid config supplied. [catalog] Missing required argument. [external_location] Missing required argument. [foreign_connection] Missing required argument. [function] Missing required argument. [metastore] Missing required argument. [model] Missing required argument. [pipeline] Missing required argument. [recipient] Missing required argument. [schema] Missing required argument. [share] Missing required argument. [storage_credential] Missing required argument. [table] Missing required argument. [volume] Missing required argument")
```
mgyucht, line 361: Oof... We don't have any control over this error message do we?

catalog/resource_grant.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/catalog/permissions"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// diffPermissionsForPrincipal returns UnityCatalogPermissionsDiff of this permissions list with `diff` privileges removed
+func diffPermissionsForPrincipal(principal string, pl catalog.PermissionsList, existing catalog.PermissionsList) (diff []catalog.PermissionsChange) {
```
mgyucht, line 18: Could we rename `pl` to `desired`? Helps with understanding the intent.

mgyucht (commented):
Overall, this is a class PR and I support this change. Mainly a couple nits and other remarks but I think with one final pass this is good to merge.

catalog/resource_grant.go
martin-walsh: Matched the existing impl for `resource_grants` as closely as I could

catalog/resource_grant.go
martin-walsh: Done

catalog/permissions/permissions.go
martin-walsh: We could but, DeleteFunc modifies the original slice & the tests fail without cloning the slice first due to side effects. I can swap the code to the following if you prefer:
```suggestion
func SliceWithoutString(in []string, without string) (out []string) {
	out = slices.Clone(in)
	out = slices.DeleteFunc(out, func(s string) bool { return s == without })
	return
}

```

catalog/permissions/permissions.go
martin-walsh: Good catch, pushed

catalog/resource_grant.go
martin-walsh: Done & added extra defensive check.

catalog/resource_grant.go
martin-walsh: In any case, added the helper methods to encapsulate the string wrangling

catalog/resource_grant_test.go
martin-walsh: Unfortunately not. I did the best I could by using `ConflictsWith` to trim the error down for the `TestResourceGrantCreateOneSecurableOnly` case, but can't do anything here.

catalog/permissions/permissions.go
mgyucht: Ah, good point. You can leave this as-is then.

catalog/resource_grant.go
mgyucht: Ping here

mgyucht (commented):
This seems great, thanks for addressing my points. Just one question about diffPermissionsForPrincipal, which I think was kind of written to support multiple principals but only ever acts on one; can we clean this up?

martin-walsh:
> This seems great, thanks for addressing my points. Just one question about diffPermissionsForPrincipal, which I think was kind of written to support multiple principals but only ever acts on one; can we clean this up?

It returns an array as that's what the `UpdatePermissions` request takes downstream. It is also a near carbon copy of the impl from resource_grants which to me is a positive as it makes addressing potential bugs easier. I would suggest leaving it as is for those reasons. I have a subsequent PR to update that resource to use go sdk and may refactor further there.

mgyucht:
Gotcha, that makes a lot of sense.

mgyucht (approved):
Got clarification on the main point. Thanks!

mgyucht:
This failed one of our nightly tests. @nkvuong could you take a look at https://github.com/databricks/eng-dev-ecosystem/actions/runs/7399425920/job/20130815184 and see if you can debug this?

nkvuong:
@mgyucht basically `TestUcAccGrant` and `TestUcAccGrants` are overwriting each other permissions at metastore-level
```hcl
resource "databricks_grant" "metastore" {
	metastore = "{env.TEST_METASTORE_ID}"
	principal  = "%s"
	privileges = ["CREATE_STORAGE_CREDENTIAL"]
}
```
```hcl
resource "databricks_grants" "metastore" {
	metastore = "{env.TEST_METASTORE_ID}"
	grant {
		principal  = "%s"
		privileges = ["CREATE_STORAGE_CREDENTIAL"]
	}
}
```

so either we need to make these 2 tests run sequentially, or remove the metastore bit for `databricks_grant`

mgyucht:
Ah, that makes perfect sense. Can we comment out the databricks_grant test for metastore for now to merge? we need to fix this for metastore assignment for workspaces as well, which also races between two test cases. I can fix both of these at the same time.

internal/acceptance/grant_test.go
```diff
@@ -0,0 +1,106 @@
...
+	comment         = "Managed by TF"
+}
+
+resource "databricks_grant" "metastore" {
+	metastore = "{env.TEST_METASTORE_ID}"
+
+	principal  = "%s"
+	privileges = ["CREATE_STORAGE_CREDENTIAL"]
+}
+
```
nkvuong, line 63:
```suggestion
```

martin-walsh:
> Ah, that makes perfect sense. Can we comment out the databricks_grant test for metastore for now to merge? we need to fix this for metastore assignment for workspaces as well, which also races between two test cases. I can fix both of these at the same time.

@mgyucht done.

nkvuong:
@mgyucht integration tests now passed (failed tests are not related to this PR)

mgyucht:
All nightlies passed. Merging now!

@martin-walsh thank you so much for your work on this and your patience through this process! We and our other customers really appreciate contributions like these. This is one of the biggest pain points with permission management for the provider, so you've solved a major problem with this.

kvedes:
Using provider version 1.39, and it seems that `databricks_grant` overwrites current permissions:

```
resource "databricks_grant" "unity_catalog" {
  catalog = "dummy"
  principal  = "group_a"
  privileges = ["CREATE_SCHEMA"]
}

resource "databricks_grant" "unity_catalog2" {
  catalog = "dummy"
  principal  = "group_a"
  privileges = ["USE_CATALOG"]
}
```
This leaves me only with the USE_CATALOG permission.

PR #3028, opened 2023-12-13
Remove requirement to provide accountId in mws_credentials resource

hectorcast-db:
## Changes
Remove requirement to provide accountId in mws_credentials resource

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/mws/resource_mws_credentials.go
+++ b/mws/resource_mws_credentials.go
@@ -49,6 +49,10 @@ func ResourceMwsCredentials() *schema.Resource {
 	p := common.NewPairSeparatedID("account_id", "credentials_id", "/")
 	return common.Resource{
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			err := c.SetAccountId(d.Get("account_id").(string))
+			if err != nil {
+				return err
+			}
 			acc, err := c.AccountClient()
 			if err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


mws/resource_mws_credentials.go
```diff
@@ -105,6 +109,10 @@ func ResourceMwsCredentials() *schema.Resource {
 			}
 			return acc.Credentials.DeleteByCredentialsId(ctx, credsId)
 		},
-		Schema: common.StructToSchema(CredentialInfo{}, common.NoCustomize),
+		Schema: common.StructToSchema(CredentialInfo{}, func(s map[string]*schema.Schema) map[string]*schema.Schema {
+			// nolint
```
mgyucht, line 113: Just curious, why is `nolint` needed?

mws/resource_mws_credentials.go
```diff
@@ -105,6 +109,10 @@ func ResourceMwsCredentials() *schema.Resource {
...
+			s["account_id"].Deprecated = "`account_id` should be set as part of the Databricks Config, not in the resource."
```
mgyucht, line 114: Can we also update the docs to refer to this?

mgyucht (approved):
Code change looks good to me. I think we should also update the docs to note that it is deprecated to specify account ID in the resource itself and that support will be removed in a later version.

Issue #2152, opened 2023-03-24
[FEATURE] Storage Credential updates with 'Force'


ralphkootker:
### Use-cases
Updating storage credentials to match updates to the underlying infrastructure on AWS is not possible using Terraform, but also the UI and CLI. It requires "Force create" to be set according to the error messages whereas all possibilities do not provide that option.

### Attempted Solutions
Drop all linked objects such as storage locations, drop the credential, re-recreate the credential, and re-attach the storage locations to the credential

### Proposal
Allow FORCE to recreate the storage credential and ensure all linked objects remain intact

### References
https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/storage_credential

![image](https://user-images.githubusercontent.com/23193161/227473330-2170bbe3-bc88-4efb-b27e-d69f91f89eea.png)
![image](https://user-images.githubusercontent.com/23193161/227473511-d4fd6af5-60c3-4771-b312-f5ebb8d597d7.png)
![image](https://user-images.githubusercontent.com/23193161/227473816-5981f63d-2df4-413e-9b51-d7be13e9c186.png)

PR #3029, opened 2023-12-13
[Do Not Review] [WIP] Added support for ownership change for all UC resources

tanmay-db:
## Changes

Add support for ownership change with rollback. Update to latest Go SDK. 

## Tests

Added Integration and unit tests
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -61,6 +61,7 @@ func ResourceCatalog() *schema.Resource {
 
 			var createCatalogRequest catalog.CreateCatalog
 			common.DataToStructPointer(d, catalogSchema, &createCatalogRequest)
+			createCatalogRequest.Name = d.Get("name").(string)
 			ci, err := w.Catalogs.Create(ctx, createCatalogRequest)
 			if err != nil {
 				return err
@@ -124,22 +125,33 @@ func ResourceCatalog() *schema.Resource {
 			}
 			var updateCatalogRequest catalog.UpdateCatalog
 			common.DataToStructPointer(d, catalogSchema, &updateCatalogRequest)
-			ownerInRequest := updateCatalogRequest.Owner
-			updateCatalogRequest.Owner = ""
-			ci, err := w.Catalogs.Update(ctx, updateCatalogRequest)
-			if err != nil {
-				return err
-			}
-			if ownerInRequest != "" && d.HasChange("owner") {
-				var updateOwnerCatalogRequest catalog.UpdateCatalog
-				updateOwnerCatalogRequest.Owner = ownerInRequest
-				updateOwnerCatalogRequest.Name = updateCatalogRequest.Name
-				ci, err = w.Catalogs.Update(ctx, updateOwnerCatalogRequest)
+			updateCatalogRequest.Name = d.Id()
+			if d.HasChange("owner") {
+				_, err = w.Catalogs.Update(ctx, catalog.UpdateCatalog{
+					Name:  updateCatalogRequest.Name,
+					Owner: updateCatalogRequest.Owner,
+				})
 				if err != nil {
-					// roll back the previous changes
-
 					return err
 				}
+			}
+
+			updateCatalogRequest.Owner = ""
+			ci, err := w.Catalogs.Update(ctx, updateCatalogRequest)
+
+			if err != nil {
+				if d.HasChange("owner") {
+					// Rollback
+					old, _ := d.GetChange("owner")
+					_, err = w.Catalogs.Update(ctx, catalog.UpdateCatalog{
+						Name:  updateCatalogRequest.Name,
+						Owner: old.(string),
+					})
+					if err != nil {
+						return err
+					}
+				}
+				return err
 			}
 
 			// We need to update the resource data because Name is updatable
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -126,6 +126,7 @@ func ResourceCatalog() *schema.Resource {
 			var updateCatalogRequest catalog.UpdateCatalog
 			common.DataToStructPointer(d, catalogSchema, &updateCatalogRequest)
 			updateCatalogRequest.Name = d.Id()
+
 			if d.HasChange("owner") {
 				_, err = w.Catalogs.Update(ctx, catalog.UpdateCatalog{
 					Name:  updateCatalogRequest.Name,
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -124,9 +124,22 @@ func ResourceCatalog() *schema.Resource {
 			}
 			var updateCatalogRequest catalog.UpdateCatalog
 			common.DataToStructPointer(d, catalogSchema, &updateCatalogRequest)
+			ownerInRequest := updateCatalogRequest.Owner
+			updateCatalogRequest.Owner = ""
 			ci, err := w.Catalogs.Update(ctx, updateCatalogRequest)
 			if err != nil {
 				return err
+			}
+			if ownerInRequest != "" && d.HasChange("owner") {
+				var updateOwnerCatalogRequest catalog.UpdateCatalog
+				updateOwnerCatalogRequest.Owner = ownerInRequest
+				updateOwnerCatalogRequest.Name = updateCatalogRequest.Name
+				ci, err = w.Catalogs.Update(ctx, updateOwnerCatalogRequest)
+				if err != nil {
+					// roll back the previous changes
+
+					return err
+				}
 			}
 
 			// We need to update the resource data because Name is updatable
--- a/catalog/resource_catalog_test.go
+++ b/catalog/resource_catalog_test.go
@@ -149,7 +149,7 @@ func TestCatalogCreateWithOwnerAlsoDeletesDefaultSchema(t *testing.T) {
 			},
 			{
 				Method:   "PATCH",
-				Resource: "/api/2.1/unity-catalog/catalogs/a",
+				Resource: "/api/2.1/unity-catalog/catalogs/",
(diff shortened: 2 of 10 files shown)


tanmay-db:
I have kept this as not draft so I can easily see unit test failures.

catalog/resource_catalog_test.go
```diff
@@ -417,11 +496,194 @@ func TestUpdateCatalogIfMetastoreIdNotExplicitelySet(t *testing.T) {
...
+			"name":    "a",
+			"comment": "c",
+			"owner":   "administrators",
+		},
+		HCL: `
+		name = "a"
+		comment = "e"
+		owner = "updatedOwner"
+		`,
+	}.ApplyNoError(t)
```
mgyucht, line 626: This should be an error, right? The update failed.

permissions/resource_permissions_test.go
```diff
@@ -1464,11 +1462,10 @@ func TestPathPermissionsResourceIDFields(t *testing.T) {
 			m = x
 		}
 	}
-	_, err := m.idRetriever(context.Background(), databricks.Must(databricks.NewWorkspaceClient(
-		(*databricks.Config)(config.NewMockConfig(func(r *http.Request) error {
-			return fmt.Errorf("nope")
-		})))), "x")
-	assert.EqualError(t, err, "cannot load path x: nope")
+	w, err := databricks.NewWorkspaceClient(&databricks.Config{})
```
mgyucht, line 1465: What is this change?

internal/acceptance/external_location_test.go
```diff
@@ -60,3 +60,112 @@ func TestUcAccExternalLocationForceDestroy(t *testing.T) {
 		}`,
 	})
 }
+
+func TestUcAccExternalLocationUpdate(t *testing.T) {
+	unityWorkspaceLevel(t, step{
+		Template: `
+		resource "databricks_storage_credential" "external" {
```
mgyucht, line 67: It's very hard to see the differences between these steps. Could we refactor this a bit so that the differences are more obvious? E.g. a function (or a few functions) to construct the different resources involved in these templates.

internal/acceptance/external_location_test.go
```diff
@@ -60,3 +60,112 @@ func TestUcAccExternalLocationForceDestroy(t *testing.T) {
 		}`,
 	})
 }
+
+func TestUcAccExternalLocationUpdate(t *testing.T) {
```
mgyucht, line 64: Why are we testing storage credentials changes here as well?

internal/acceptance/metastore_test.go
```diff
@@ -88,3 +100,28 @@ func runMetastoreTest(t *testing.T, extraAttributes map[string]any) {
...
+			owner = "account users"
+			%s
+		}`, template),
+	}, step{
+		Template: fmt.Sprintf(`resource "databricks_metastore" "this" {
+			name = "{var.STICKY_RANDOM}"
+			force_destroy = true
+			owner = "{env.TEST_METASTORE_ADMIN_GROUP_NAME}"
+			%s
+		}`, template),
```
mgyucht, line 125: Do you need to test owner update twice? IIUC: we should test 1) owner update by itself, 2) both owner & other param update at the same time, to be 100% sure we cover the main cases. (Other param update by itself would be covered by existing integration tests.)

internal/acceptance/recipient_test.go
```diff
@@ -43,3 +43,57 @@ func TestUcAccCreateRecipientDb2DbAws(t *testing.T) {
 		}`,
 	})
 }
+
+func TestUcAccUpdateRecipientDb2Open(t *testing.T) {
+	unityWorkspaceLevel(t, step{
+		Template: `
+		resource "databricks_recipient" "db2open" {
```
mgyucht, line 50: ditto: let's refactor this to make it more obvious what the actual changes between each step.

internal/acceptance/share_test.go
```diff
@@ -93,3 +105,54 @@ func TestUcAccCreateShare(t *testing.T) {
 		`,
 	})
 }
+
+func TestUcAccCreateUpdate(t *testing.T) {
```
mgyucht, line 109: The test name must say something about which component is being tested.

mgyucht (changes requested):
A first pass. Overall, I have a concern that there is too much repetition between the different UC resources, which can cause variations in behavior to arise & increase the cost of maintaining this logic across all UC resources with an owner. However, behaviorally, I think this is correct.

My two main concerns:
1. Error propagation is not complete. When both the update of other fields and the owner rollback fail, the error from the first message is lost, which will make it very difficult for users to debug what happened.
2. Some unit tests assert that the rollback option doesn't return an error, but I think it should error. I think you're overwriting the error variable when rolling back the owner, so that error is lost.

access/resource_ip_access_list_test.go
```diff
@@ -182,7 +182,7 @@ func TestIPACLUpdate_Error(t *testing.T) {
 			{
 				Method:   http.MethodPatch,
 				Resource: "/api/2.0/ip-access-lists/" + TestingId,
-				ExpectedRequest: ipAccessListUpdateRequest{
```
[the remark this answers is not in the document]
tanmay-db: We use `settings.UpdateIpAccessList` in Update method in the resource so we should use that here as well. This probably started happening after Go SDK upgrade (I didn't look closely) but it was failing tests.

catalog/resource_metastore_data_access.go
```diff
@@ -48,6 +48,11 @@ func adjustDataAccessSchema(m map[string]*schema.Schema) map[string]*schema.Sche
 		Optional: true,
 	}
 
+	m["force_update"] = &schema.Schema{
```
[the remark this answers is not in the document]
tanmay-db: There is an issue https://github.com/databricks/terraform-provider-databricks/issues/2152 that needed to be fixed before we can do updates

permissions/resource_permissions_test.go
tanmay-db: This happened after upgrading to Go SDK, we no longer have config.NewMockConfig: https://github.com/databricks/databricks-sdk-go/pull/739/files#diff-3434249a481acd0afcf6d4786db5bcae154e942666f2082de7d3f3ff3e44ff01

internal/acceptance/share_test.go
tanmay-db: Done

internal/acceptance/metastore_test.go
tanmay-db: I added another param update by itself in between steps to be sure there isn't any behaviour change in between steps without non owner changes. I will remove it.

internal/acceptance/external_location_test.go
tanmay-db: We used to have these tests combined together before as well (eg: above test: `TestUcAccExternalLocationForceDestroy`), since external location needs storage credential so it removed having another tests that does that same thing but we remove code duplication

catalog/resource_share.go
```diff
@@ -220,10 +220,41 @@ func ResourceShare() *schema.Resource {
...
+					Name:  afterSi.Name,
+					Owner: afterSi.Owner,
+				})
+				if err != nil {
+					return err
+				}
+			}
+
+			err = NewSharesAPI(ctx, c).update(d.Id(), ShareUpdates{
+				Owner:   "",
```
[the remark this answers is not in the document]
tanmay-db: Doesn't it make code less readable? Looking at code now I know we send empty owner whereas I won't know what default is.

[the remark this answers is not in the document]
mgyucht: The default value in a struct is always the zero value in Go.

catalog/resource_catalog.go
```diff
@@ -135,8 +136,33 @@ func ResourceCatalog() *schema.Resource {
...
 			if err != nil {
+				if d.HasChange("owner") {
+					// Rollback
+					old, _ := d.GetChange("owner")
+					_, err = w.Catalogs.Update(ctx, catalog.UpdateCatalog{
+						Name:  updateCatalogRequest.Name,
+						Owner: old.(string),
+					})
+					if err != nil {
+						return err
```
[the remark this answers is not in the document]
tanmay-db: Good point, this will make error message more detailed. Added

catalog/resource_catalog_test.go
```diff
@@ -417,11 +496,194 @@ func TestUpdateCatalogIfMetastoreIdNotExplicitelySet(t *testing.T) {
...
+			"comment": "c",
+			"owner":   "administrators",
+		},
+		HCL: `
+		name = "a"
+		comment = "e"
+		owner = "updatedOwner"
+		`,
+	}.Apply(t)
+	qa.AssertErrorStartsWith(t, err, "Internal error happened")
```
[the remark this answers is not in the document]
tanmay-db: Agree, added two error checks, one for normal error and one during rollback.

internal/acceptance/external_location_test.go
tanmay-db: updated

tanmay-db:
@mgyucht Thanks for the review, addressed all comments.

catalog/resource_catalog.go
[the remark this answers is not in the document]
tanmay-db: Owner of this resource was updated but other fields couldn't be updated and owner couldn't be rolled back, as a result, the owner of this resource is updated to <> but no other attributes are, current state: <print>

Rollback the owner change manually to <old> through UI to unblock.

mgyucht (approved):
LGTM, but let's just append the error message to the error rather than printing in the log, will be more usable for customers.

PR #3035, opened 2023-12-15
Validate metastore_id field in worspace UC resources

hectorcast-db:
## Changes
Validate metastore_id field in worspace UC resources

## Tests

- [X] `make test` run locally
- [X] using Go SDK

### DIFF
--- a/catalog/resource_catalog_test.go
+++ b/catalog/resource_catalog_test.go
@@ -236,6 +236,13 @@ func TestUpdateCatalog(t *testing.T) {
 	qa.ResourceFixture{
 		Fixtures: []qa.HTTPFixture{
 			{
+				Method:   "GET",
+				Resource: "/api/2.1/unity-catalog/current-metastore-assignment",
+				Response: catalog.MetastoreAssignment{
+					MetastoreId: "d",
+				},
+			},
+			{
 				Method:   "PATCH",
 				Resource: "/api/2.1/unity-catalog/catalogs/a",
 				ExpectedRequest: catalog.UpdateCatalog{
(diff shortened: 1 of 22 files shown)


mws/resource_mws_credentials_test.go
```diff
@@ -56,7 +56,7 @@ func TestResourceCredentialsCreate(t *testing.T) {
 	})
 }
 
-func TestAccountIdOnlyInState(t *testing.T) {
+func TestCredentialsAccountIdOnlyInState(t *testing.T) {
```
hectorcast-db, line 59: Unrelated, but since the name starts with `TestAcc` it runs on the integration tests an fails due to the fixture.

docs/resources/catalog.md
```diff
@@ -25,6 +25,7 @@ resource "databricks_catalog" "sandbox" {
 The following arguments are required:
 
 * `name` - Name of Catalog relative to parent metastore.
+* `metastore_id` - (Required if changing the metastore assigned to a workspace) If set, it must match the ID of the metastore assigned to the worspace. If not set, the current ID is exported.
```
nkvuong, line 28: the statement `(Required if changing the metastore assigned to a workspace)` will likely cause confusion - let's just mention it in the [troubleshooting guide](docs/guides/troubleshooting.md)

catalog/resource_catalog_test.go
```diff
@@ -277,6 +284,140 @@ func TestUpdateCatalog(t *testing.T) {
...
+			"name":         "a",
+			"comment":      "c",
+		},
+		HCL: `
+		metastore_id = "correct_id"
+		name = "a"
+		comment = "c"
+		owner = "administrators"
+		`,
+	}.ApplyNoError(t)
```
mgyucht, line 366: For sanity, let's make sure the metastore ID is updated properly?

docs/resources/catalog.md
nkvuong, line 1: we should also remove `metastore_id = databricks_metastore.this.id` from the example usage (and any other docs)

catalog/utils.go
```diff
@@ -0,0 +1,23 @@
+package catalog
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/databricks/databricks-sdk-go"
+)
+
+func validateMetastoreId(ctx context.Context, w *databricks.WorkspaceClient, metastoreId string) error {
```
mgyucht, line 10: nit: add a small doccomment here explaining the purpose of this and where it should be added (e.g. all create/update/delete methods for UC resources that accept a metastore_id parameter).

nkvuong (approved):
just a couple of nits

docs/resources/catalog.md
mgyucht: +1. IMO we don't need to show this to customers explicitly; the happy path would be not to specify the metastore ID, I think. Once these resources are available at the account-level, we'll change the docs to match what you've done for storage credentials.

mgyucht (approved):
Awesome, thanks for fixing so quickly! One small test nit/question, and let's align on how we want to (not) document this (not) feature.

PR #3044, opened 2023-12-19
Migrate SQL Warehouse to Go SDK

mgyucht:
## Changes
This PR migrates the SQL Warehouse resource to the Go SDK. It also modifies all tests that used the original API client, which means appending `?` to resource URIs in fixtures.

To do this, some changes to supporting code are also made:

1. Introduced several helper methods in `common` to simplify customizing schemas computed by StructToSchema for several common cases:
* `SetSuppressDiff`: sets a diff suppressor on the field so that fields not populated by the user in their configuration are not treated as a diff, even if the API returned some value for that field.
* `SetDefault`: marks a schema as optional and sets a default value for it.
* `SetReadOnly`: marks a schema as not configurable by users but that can be referenced in Terraform.
2. Improved suppress diff logic to work with lists/sets. If two lists have a different number of elements, the size of the list is compared, rather than individual list items. The corresponding check should be that `new` is `"0"` (not in config) and `old` is not `"0"` (state captured something).
3. Improved all reflection-based methods to work with embedded structs. It is now possible to use `StructToSchema` on structs like
```go
type S struct {
  sdk.SdkType
  AnotherField string `json:"another_field,omitempty"`
}
```
The fields of the embedded field are expanded into the schema. This way, maintainers can reuse the Go type as-is without having to copy all of its fields into a new structure.

4. Bugfix in `collectionToMaps` that assumed that a data structure is a list iff `MaxItems` is 1. `MaxItems` can't be set on read-only fields. This now checks the underlying type is a slice or array.
5. Introduced `WorkspaceDataWithCustomAttrs` and `AccountDataWithCustomAttrs`. These functions are similar to `WorkspaceData` and `AccountData` except that they permit two type parameters. The first corresponds to the type used for the resource, ensuring that the schema of the data source and that of the resource are identical. The second corresponds to any additional data that a user might specify in the data source.

Side note: we should probably look into changing suppress diff from being a schema-level configuration to actually adjusting the returned value from the `Read` method. Otherwise, if the only change made by a user is e.g. to remove all tags, that won't show up as a diff.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


sql/data_sql_warehouses.go
```diff
@@ -14,20 +16,18 @@ func DataSourceWarehouses() *schema.Resource {
...
-		list, err := a.List()
+	return common.WorkspaceData(func(ctx context.Context, data *warehousesData, w *databricks.WorkspaceClient) error {
+		list, err := w.Warehouses.ListAll(ctx, sql.ListWarehousesRequest{})
 		if err != nil {
 			return err
 		}
 		name_contains := (*data).WarehouseNameContains
-		for _, e := range list.Endpoints {
+		for _, e := range list {
 			match_name := strings.Contains(strings.ToLower(e.Name), name_contains)
```
nkvuong, line 26: this looks like a bug with case sensitivity that we should fix

common/resource.go
```diff
@@ -256,9 +256,17 @@ func DataResource(sc any, read func(context.Context, any, *DatabricksClient) err
...
-	return genericDatabricksData(func(c *DatabricksClient) (*databricks.WorkspaceClient, error) {
+	return genericDatabricksData[T, struct{}, *databricks.WorkspaceClient](func(c *DatabricksClient) (*databricks.WorkspaceClient, error) {
 		return c.WorkspaceClient()
-	}, read)
+	}, func(ctx context.Context, s struct{}, t *T, wc *databricks.WorkspaceClient) error {
+		return read(ctx, t, wc)
+	}, false)
+}
+
+func WorkspaceData2[SdkType, OtherType any](read func(context.Context, OtherType, *SdkType, *databricks.WorkspaceClient) error) *schema.Resource {
```
nkvuong, line 266: should we name this `WorkspaceDataSdkType`?

sql/resource_sql_endpoint.go
```diff
@@ -214,33 +85,58 @@ func ResourceSqlEndpoint() *schema.Resource {
 			Create: schema.DefaultTimeout(30 * time.Minute),
 		},
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			var se SQLEndpoint
-			common.DataToStructPointer(d, s, &se)
-			if err := NewSQLEndpointsAPI(ctx, c).Create(&se, d.Timeout(schema.TimeoutCreate)); err != nil {
```
nkvuong, line 219: would customers lose the ability to set custom timeout during creation?

common/reflect_resource.go
```diff
@@ -260,7 +317,7 @@ func typeToSchema(v reflect.Value, t reflect.Type, path []string) map[string]*sc
 			scm[fieldName].Type = schema.TypeList
 			elem := typeField.Type.Elem()
 			sv := reflect.New(elem).Elem()
-			nestedSchema := typeToSchema(sv, elem, append(path, fieldName, "0"))
+			nestedSchema := typeToSchema(sv, append(path, fieldName, "0"))
 			if strings.Contains(tfTag, "suppress_diff") {
 				blockCount := strings.Join(append(path, fieldName, "#"), ".")
 				scm[fieldName].DiffSuppressFunc = makeEmptyBlockSuppressFunc(blockCount)
```
edwardfeng-db, line 323: With the new suffix checking in `diffSuppressor`, can we replace `makeEmptyBlockSuppressFunc` calls?

edwardfeng-db (commented):
Halfway through reviewing, need a bit more time understanding context on the WorkspaceData and AccountData

alexott (commented):
in general, looks good for me from the first pass...

common/reflect_resource_test.go
alexott, line 1: @mgyucht  I think that we need to add at least two new test cases:

1. For inherited structure - so we can see that it's flattened correctly into schema
2. For a structure with a field from Go SDK

edwardfeng-db (approved):

PR #3071, opened 2024-01-04
Fix diff schema on PRs

mgyucht:
## Changes
The diff-schema workflow doesn't work anymore because it fails to checkout master. This PR should fix this by always checking out the base & head commits and passing those commits via env vars to `diff-schema.sh`. With no variables set, it will show the diff between the current checked-out branch and master.

## Tests
- [x] Ran locally with no flags, worked.
- [ ] Ran locally with BASE_COMMIT=bb12b5a614f9c1357356b68d58c15ad87f6cfa08, worked
- [ ] Triggered on this PR
- [ ] Triggered via workflow dispatch

### DIFF
--- a/.github/workflows/schema.yml
+++ b/.github/workflows/schema.yml
@@ -7,12 +7,12 @@ on:
   workflow_dispatch:
     inputs:
       base:
-        description: 'Base ref'
-        default: 'master'
+        description: "Base ref"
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm

hectorcast-db (approved):

Issue #2966, opened 2023-11-28
[ISSUE] Issue with `databricks_cluster` resource


FORDBX4:
Hello,

databricks_cluster defines "id" in its schema as an "optional" and a "computed" but if I try and define the id, it throws an error "Invalid or unknown key"

### Configuration

```hcl
terraform {
	required_providers {
		databricks = {
			source = "databricks/databricks"
			version = "1.30.0"
		}
	}
}
```

### Expected Behavior
```json
. . .
"id": {
        "type": "string",
        "description_kind": "plain",
        "computed": true
},
. . .
```

### Actual Behavior
```json
. . .
"id": {
        "type": "string",
        "description_kind": "plain",
        "optional": true,
        "computed": true
},
. . .
```

### Steps to Reproduce

1. terraform init
2. terraform providers schema -json

### Terraform and provider versions

Terraform v1.5.5
on windows_amd64

Your version of Terraform is out of date! The latest version
is 1.6.4. You can update by downloading from https://www.terraform.io/downloads.html

FORDBX4:
> `id` is generated on the backend, you can't set it explicitly.

That is the case, but it is declared in the schema that `id` can optionally be set. As you have just pointed out, that is incorrect.

alexott:
closing as outdated

PR #3076, opened 2024-01-05
Migrate cluster schema to use the go-sdk struct

edwardfeng-db:
## Changes

- Before we are able to migrate the cluster resources to use the go-sdk schema, we need to figure out a way to represent the current `tf` tags on the manually maintained structs
- This PR introduces a new interface `ResourceProviderStruct` which contains a method for aliases and all of the customizations.
  - `Aliases() map[string]string`
  - `CustomizeSchema(map[string]*schema.Schema) map[string]*schema.Schema`
- Also added a series of customization functions such as `SetOptional`, etc so that we can easily write customization logics into the `CustomizeSchema` function to replace the current `tf` tags
- Migrated the cluster schema into the go-sdk schema with this new interface

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -60,8 +60,8 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 
 type ClusterResourceProvider struct{}
 
-func (ClusterResourceProvider) UnderlyingType() compute.ClusterDetails {
-	return compute.ClusterDetails{}
+func (ClusterResourceProvider) UnderlyingType() compute.ClusterSpec {
+	return compute.ClusterSpec{}
 }
 
 func (ClusterResourceProvider) Aliases() map[string]string {
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
+	"github.com/databricks/databricks-sdk-go/service/compute"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/databricks/terraform-provider-databricks/libraries"
 )
@@ -57,71 +58,116 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	return false
 }
 
-func resourceClusterSchema() map[string]*schema.Schema {
-	return common.StructToSchema(Cluster{}, func(s map[string]*schema.Schema) map[string]*schema.Schema {
-		s["spark_conf"].DiffSuppressFunc = SparkConfDiffSuppressFunc
-		common.MustSchemaPath(s, "aws_attributes", "zone_id").DiffSuppressFunc = ZoneDiffSuppress
-		common.MustSchemaPath(s, "gcp_attributes", "use_preemptible_executors").Deprecated = "Please use 'availability' instead."
+type ClusterResourceProvider struct{}
 
-		common.MustSchemaPath(s, "init_scripts", "dbfs").Deprecated = DbfsDeprecationWarning
+func (ClusterResourceProvider) UnderlyingType() compute.ClusterDetails {
+	return compute.ClusterDetails{}
+}
 
-		// adds `library` configuration block
-		s["library"] = common.StructToSchema(libraries.ClusterLibraryList{},
-			func(ss map[string]*schema.Schema) map[string]*schema.Schema {
-				ss["library"].Set = func(i any) int {
-					lib := libraries.NewLibraryFromInstanceState(i)
-					return schema.HashString(lib.String())
-				}
+func (ClusterResourceProvider) Aliases() map[string]string {
+	return map[string]string{"cluster_mount_infos": "cluster_mount_info"}
+}
+
+func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
+	common.CustomizeSchemaPath(s, "enable_elastic_disk").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "enable_local_disk_encryption").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "node_type_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
+	common.CustomizeSchemaPath(s, "driver_node_type_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
+	common.CustomizeSchemaPath(s, "driver_instance_pool_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_node_type_id", "node_type_id"})
+	common.CustomizeSchemaPath(s, "ssh_public_keys").SetMaxItems(10)
+	common.CustomizeSchemaPath(s, "init_scripts").SetMaxItems(10).AddNewField("abfss", common.StructToSchema(InitScriptStorageInfo{}, func(ss map[string]*schema.Schema) map[string]*schema.Schema {
+		return ss
+	})["abfss"]).AddNewField("gcs", common.StructToSchema(InitScriptStorageInfo{}, func(ss map[string]*schema.Schema) map[string]*schema.Schema {
+		return ss
+	})["gcs"])
+	common.CustomizeSchemaPath(s, "init_scripts", "dbfs").SetDeprecated(DbfsDeprecationWarning)
+	common.CustomizeSchemaPath(s, "init_scripts", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "init_scripts", "s3", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "workload_type", "clients").SetRequired()
+	common.CustomizeSchemaPath(s, "workload_type", "clients", "notebooks").SetOptional().SetDefault(true)
+	common.CustomizeSchemaPath(s, "workload_type", "clients", "jobs").SetOptional().SetDefault(true)
+	s["idempotency_token"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Optional: true,
+		ForceNew: true,
+	}
+	common.CustomizeSchemaPath(s, "data_security_mode").SetSuppressDiff()
+	common.CustomizeSchemaPath(s, "docker_image", "url").SetRequired()
+	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "password").SetRequired().SetSensitive()
+	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "username").SetRequired()
+	common.CustomizeSchemaPath(s, "spark_conf").SetCustomSuppressDiff(SparkConfDiffSuppressFunc)
+	common.CustomizeSchemaPath(s, "aws_attributes").SetSuppressDiff().SetConflictsWith([]string{"azure_attributes", "gcp_attributes"})
+	common.CustomizeSchemaPath(s, "aws_attributes", "zone_id").SetCustomSuppressDiff(ZoneDiffSuppress)
+	common.CustomizeSchemaPath(s, "azure_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "gcp_attributes"})
+	common.CustomizeSchemaPath(s, "gcp_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "azure_attributes"}).AddNewField(
+		"use_preemptible_executors",
+		&schema.Schema{
+			Type:       schema.TypeBool,
+			Optional:   true,
+			Deprecated: "Please use 'availability' instead.",
+		},
+	).AddNewField(
+		"zone_id",
+		&schema.Schema{
+			Type:     schema.TypeString,
+			Optional: true,
+		},
+	)
+	s["library"] = common.StructToSchema(libraries.ClusterLibraryList{},
+		func(ss map[string]*schema.Schema) map[string]*schema.Schema {
+			ss["library"].Set = func(i any) int {
+				lib := libraries.NewLibraryFromInstanceState(i)
+				return schema.HashString(lib.String())
+			}
+			return ss
+		})["library"]
+
+	common.CustomizeSchemaPath(s, "autotermination_minutes").SetDefault(60)
+	common.CustomizeSchemaPath(s, "autoscale", "max_workers").SetOptional()
+	common.CustomizeSchemaPath(s, "autoscale", "min_workers").SetOptional()
+	s["apply_policy_default_values"] = &schema.Schema{
+		Optional: true,
+		Type:     schema.TypeBool,
+	}
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "s3", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "spark_version").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_id").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "instance_pool_id").SetConflictsWith([]string{"driver_node_type_id", "node_type_id"})
+	common.CustomizeSchemaPath(s, "runtime_engine").SetValidateFunc(validation.StringInSlice([]string{"PHOTON", "STANDARD"}, false))
+	s["is_pinned"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			if old == "" && new == "false" {
+				return true
+			}
+			return old == new
+		},
+	}
+	common.CustomizeSchemaPath(s, "state").SetReadOnly()
+	s["url"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: true,
+	}
+	common.CustomizeSchemaPath(s, "default_tags").SetReadOnly()
+	common.CustomizeSchemaPath(s, "num_workers").SetDefault(0).SetValidateDiagFunc(validation.ToDiagFunc(validation.IntAtLeast(0)))
+	s["cluster_mount_info"] = &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: common.StructToSchema(MountInfo{}, func(ss map[string]*schema.Schema) map[string]*schema.Schema {
 				return ss
-			})["library"]
+			}),
+		},
+	}
 
-		s["autotermination_minutes"].Default = 60
-		s["cluster_id"] = &schema.Schema{
-			Type:     schema.TypeString,
-			Optional: true,
-			Computed: true,
-		}
-		s["aws_attributes"].ConflictsWith = []string{"azure_attributes", "gcp_attributes"}
-		s["azure_attributes"].ConflictsWith = []string{"aws_attributes", "gcp_attributes"}
-		s["gcp_attributes"].ConflictsWith = []string{"aws_attributes", "azure_attributes"}
-		s["instance_pool_id"].ConflictsWith = []string{"driver_node_type_id", "node_type_id"}
-		s["driver_instance_pool_id"].ConflictsWith = []string{"driver_node_type_id", "node_type_id"}
-		s["driver_node_type_id"].ConflictsWith = []string{"driver_instance_pool_id", "instance_pool_id"}
-		s["node_type_id"].ConflictsWith = []string{"driver_instance_pool_id", "instance_pool_id"}
+	return s
+}
 
-		s["runtime_engine"].ValidateFunc = validation.StringInSlice([]string{"PHOTON", "STANDARD"}, false)
-
-		s["is_pinned"] = &schema.Schema{
-			Type:     schema.TypeBool,
-			Optional: true,
-			Default:  false,
-			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-				if old == "" && new == "false" {
-					return true
-				}
-				return old == new
-			},
-		}
-		s["state"] = &schema.Schema{
-			Type:     schema.TypeString,
-			Computed: true,
-		}
-		s["default_tags"] = &schema.Schema{
-			Type:     schema.TypeMap,
-			Computed: true,
-		}
-		s["num_workers"] = &schema.Schema{
-			Type:             schema.TypeInt,
-			Optional:         true,
-			Default:          0,
-			ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(0)),
-		}
-		s["url"] = &schema.Schema{
-			Type:     schema.TypeString,
-			Computed: true,
-		}
-		return s
-	})
+func resourceClusterSchema() map[string]*schema.Schema {
+	return common.ResourceProviderStructToSchema[compute.ClusterDetails](ClusterResourceProvider{})
 }
 
 func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -69,6 +69,7 @@ func (ClusterResourceProvider) Aliases() map[string]string {
 }
 
 func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
+	common.CustomizeSchemaPath(s, "cluster_source").SetComputed()
 	common.CustomizeSchemaPath(s, "enable_elastic_disk").SetComputed()
 	common.CustomizeSchemaPath(s, "enable_local_disk_encryption").SetComputed()
 	common.CustomizeSchemaPath(s, "node_type_id").SetComputed().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -86,11 +86,11 @@ func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[
 	common.CustomizeSchemaPath(s, "workload_type", "clients").SetRequired()
 	common.CustomizeSchemaPath(s, "workload_type", "clients", "notebooks").SetOptional().SetDefault(true)
 	common.CustomizeSchemaPath(s, "workload_type", "clients", "jobs").SetOptional().SetDefault(true)
-	s["idempotency_token"] = &schema.Schema{
+	common.CustomizeSchemaPath(s).AddNewField("idempotency_token", &schema.Schema{
 		Type:     schema.TypeString,
 		Optional: true,
 		ForceNew: true,
-	}
+	})
 	common.CustomizeSchemaPath(s, "data_security_mode").SetSuppressDiff()
 	common.CustomizeSchemaPath(s, "docker_image", "url").SetRequired()
 	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "password").SetRequired().SetSensitive()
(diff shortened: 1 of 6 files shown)


common/reflect_resource.go
```diff
@@ -175,10 +418,19 @@ func diffSuppressor(zero string) func(k, old, new string, d *schema.ResourceData
 			log.Printf("[DEBUG] Suppressing diff for %v: platform=%#v config=%#v", k, old, new)
 			return true
 		}
+		if strings.HasSuffix(k, ".#") && new == "0" && old != "0" {
+			field := strings.TrimSuffix(k, ".#")
+			log.Printf("[DEBUG] Suppressing diff for list or set %v: no value configured but platform returned some value (likely {})", field)
+			return true
+		}
```
edwardfeng-db, line 425: This change is introduced in https://github.com/databricks/terraform-provider-databricks/pull/3044

common/reflect_resource.go
```diff
@@ -74,6 +218,111 @@ func MustSchemaPath(s map[string]*schema.Schema, path ...string) *schema.Schema
 	return sch
 }
```
edwardfeng-db, line 220: Will need to figure out a way to merge these with the code introduced in https://github.com/databricks/terraform-provider-databricks/pull/3044. Shouldn't be too hard

clusters/resource_cluster.go
```diff
@@ -57,71 +58,116 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	common.CustomizeSchemaPath(s, "autoscale", "max_workers").SetOptional()
+	common.CustomizeSchemaPath(s, "autoscale", "min_workers").SetOptional()
+	s["apply_policy_default_values"] = &schema.Schema{
+		Optional: true,
+		Type:     schema.TypeBool,
+	}
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "s3", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "spark_version").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_id").SetReadOnly()
```
edwardfeng-db, line 135: This will introduce a "breaking" diff, but it shouldn't be a problem since users are not expected to set this any way. And this will address https://github.com/databricks/terraform-provider-databricks/issues/2966

clusters/resource_cluster.go
```diff
@@ -57,71 +58,116 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
-				return ss
-			})["library"]
+func (ClusterResourceProvider) Aliases() map[string]string {
+	return map[string]string{"cluster_mount_infos": "cluster_mount_info"}
+}
 
-		s["autotermination_minutes"].Default = 60
-		s["cluster_id"] = &schema.Schema{
+func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
+	common.CustomizeSchemaPath(s, "enable_elastic_disk").SetOptional().SetComputed()
```
tanmay-db, line 72: This looks a bit difficult to maintain and I think something similar would be required for jobs resource as well? should we split this up so it's easier to read, for example, we could split on top level field, for example `init_scripts`, `workload_type`? A question - how do we check if customizations are complete?

clusters/resource_cluster.go
```diff
@@ -57,71 +58,116 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+		Optional: true,
+		Default:  false,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			if old == "" && new == "false" {
+				return true
+			}
+			return old == new
+		},
+	}
+	common.CustomizeSchemaPath(s, "state").SetReadOnly()
```
tanmay-db, line 149: just curious, is there a reason for this instead of:

> s["state"] = &schema.Schema{
			Type:     schema.TypeString,
			Computed: true,
		}

Asking since we didn't do this change for url.

tanmay-db (commented):
Thanks for the PR, left some comments. Just to confirm, the nightly did pass on the PR right?

clusters/resource_cluster.go
edwardfeng-db: Because `url` is not in the original schema but `state` is in there. Yeah I can probably add a comment but the interesting/confusing part is that if we are adding a top level field that's new, we need to do the `s["url"]` thing. If it's not top level, we can use `.AddNewField`

common/reflect_resource.go
```diff
@@ -38,6 +40,122 @@ var kindMap = map[reflect.Kind]string{
...
+		switch typeField.Type.Kind() {
+		case reflect.Int, reflect.Int32, reflect.Int64:
+			scm[fieldName].Type = schema.TypeInt
+		case reflect.Float64:
+			scm[fieldName].Type = schema.TypeFloat
+		case reflect.Bool:
+			scm[fieldName].Type = schema.TypeBool
+		case reflect.String:
+			scm[fieldName].Type = schema.TypeString
+		case reflect.Map:
```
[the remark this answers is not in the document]
edwardfeng-db: Yeah this is basically a copy of the existing TypeToSchema function but with some modifications on top to work with the new interface. So you probably don't have to look too deep into this giant function

common/reflect_resource.go
```diff
@@ -74,6 +218,111 @@ func MustSchemaPath(s map[string]*schema.Schema, path ...string) *schema.Schema
...
+	return s
+}
+
+// SetReadOnly sets the schema to be read-only (i.e. computed, non-optional).
+// This should be used for fields that are not user-configurable but are returned
+// by the platform.
+func (s *CustomizableSchema) SetReadOnly() *CustomizableSchema {
+	s.Schema.Optional = false
+	s.Schema.Required = false
+	s.Schema.MaxItems = 0
```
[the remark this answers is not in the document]
edwardfeng-db: Yeah good question, I was just trying to stay consistent with the function introduced by Miles [here](https://github.com/databricks/terraform-provider-databricks/pull/3044/files#diff-50f3dff318c929631c545c2a0286e77407a68f99cc0d99949e6151f1a4f45d80R104-R109)

clusters/resource_cluster.go
edwardfeng-db: I can probably make it work with the `AddNewField` interface as well, let me give it a try

common/reflect_resource.go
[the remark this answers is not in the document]
edwardfeng-db: @mgyucht Any idea about this?

clusters/resource_cluster.go
edwardfeng-db: Yeah for the jobs resource, we would just set whatever path that leads to the cluster reference to be the schema derived from `ClusterResourceProvider` so we don't have to duplicate it.

For making sure the costomizations are complete, we rely on the diff schema PR check and I verified that there's no breaking change (only added field, nothing removed)

common/reflect_resource.go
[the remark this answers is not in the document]
tanmay-db: One more thing is that this won't work correctly if we call `SetReadOnly` after `SetMaxItems` as then the value would be overwritten. We would also need a check to prevent this. 

```
func (s *CustomizableSchema) SetMaxItems(value int) *CustomizableSchema {
	s.Schema.MaxItems = value
	return s
}
```

clusters/resource_cluster.go
edwardfeng-db: Okay yeah I've updated it. Now everything follows the `common.CustomizeSchemaPath().AddNewField()` pattern. This should look much better

tanmay-db: Awesome! Thank you

clusters/resource_cluster.go
```diff
@@ -57,71 +58,117 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	})
+	common.CustomizeSchemaPath(s, "data_security_mode").SetSuppressDiff()
+	common.CustomizeSchemaPath(s, "docker_image", "url").SetRequired()
+	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "password").SetRequired().SetSensitive()
+	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "username").SetRequired()
+	common.CustomizeSchemaPath(s, "spark_conf").SetCustomSuppressDiff(SparkConfDiffSuppressFunc)
+	common.CustomizeSchemaPath(s, "aws_attributes").SetSuppressDiff().SetConflictsWith([]string{"azure_attributes", "gcp_attributes"})
+	common.CustomizeSchemaPath(s, "aws_attributes", "zone_id").SetCustomSuppressDiff(ZoneDiffSuppress)
+	common.CustomizeSchemaPath(s, "azure_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "gcp_attributes"})
+	common.CustomizeSchemaPath(s, "gcp_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "azure_attributes"}).AddNewField(
```
edwardfeng-db, line 102: Make AddNewField panic if the field already exists

clusters/resource_cluster.go
```diff
@@ -57,71 +58,117 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+		Type:     schema.TypeString,
+		Computed: true,
 	})
+	common.CustomizeSchemaPath(s, "default_tags").SetReadOnly()
+	common.CustomizeSchemaPath(s, "num_workers").SetDefault(0).SetValidateDiagFunc(validation.ToDiagFunc(validation.IntAtLeast(0)))
+	common.CustomizeSchemaPath(s).AddNewField("cluster_mount_info", &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: common.StructToSchema(MountInfo{}, func(ss map[string]*schema.Schema) map[string]*schema.Schema {
```
edwardfeng-db, line 161: Try to put in a list of MountInfo

common/reflect_resource.go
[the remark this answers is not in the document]
edwardfeng-db: After syncing with Miles, we decided to consolidate this copy with the existing `typeToSchema` function, so this part is no longer needed. Should be much cleaner now

common/reflect_resource.go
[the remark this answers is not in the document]
edwardfeng-db: Yeah synced with Miles, the idea is that when a field is `Computed` or read-only, terraform doesn't really want us to set any constraint to the value itself (max items is an example).

Regarding calling `SetMaxItems` after calling `SetReadOnly`, it should probably not be a problem because people should see the schema being undesirable in the diff schema check or just see some failures.

edwardfeng-db:
Okay thanks @alexott

alexott (changes requested):
See my comments about computed attributes, etc.

clusters/resource_cluster.go
```diff
@@ -57,71 +58,111 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
-		s["autotermination_minutes"].Default = 60
-		s["cluster_id"] = &schema.Schema{
+func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
+	common.CustomizeSchemaPath(s, "enable_elastic_disk").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "enable_local_disk_encryption").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "node_type_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
+	common.CustomizeSchemaPath(s, "driver_node_type_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
+	common.CustomizeSchemaPath(s, "driver_instance_pool_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_node_type_id", "node_type_id"})
+	common.CustomizeSchemaPath(s, "ssh_public_keys").SetMaxItems(10)
+	common.CustomizeSchemaPath(s, "init_scripts").SetMaxItems(10).AddNewField("abfss", common.StructToSchema(InitScriptStorageInfo{}, nil)["abfss"]).AddNewField("gcs", common.StructToSchema(InitScriptStorageInfo{}, nil)["gcs"])
```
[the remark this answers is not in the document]
mgyucht: +1. We will address this holistically by generating the API spec from Protobuf. However, in the meantime, we want to unblock this effort by manually correcting these here.

clusters/resource_cluster.go
```diff
@@ -57,71 +58,111 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
-func resourceClusterSchema() map[string]*schema.Schema {
-	return common.StructToSchema(Cluster{}, func(s map[string]*schema.Schema) map[string]*schema.Schema {
-		s["spark_conf"].DiffSuppressFunc = SparkConfDiffSuppressFunc
-		common.MustSchemaPath(s, "aws_attributes", "zone_id").DiffSuppressFunc = ZoneDiffSuppress
-		common.MustSchemaPath(s, "gcp_attributes", "use_preemptible_executors").Deprecated = "Please use 'availability' instead."
+type ClusterResourceProvider struct{}
 
-		common.MustSchemaPath(s, "init_scripts", "dbfs").Deprecated = DbfsDeprecationWarning
+func (ClusterResourceProvider) UnderlyingType() compute.ClusterDetails {
+	return compute.ClusterDetails{}
```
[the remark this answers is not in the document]
mgyucht: +1, I think it makes sense to use ClusterSpec if possible.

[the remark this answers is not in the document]
alexott: `ClusterDetails` contains a lot of computed values that aren't relevant for Terraform...

clusters/resource_cluster.go
```diff
@@ -57,71 +58,111 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	common.CustomizeSchemaPath(s, "autoscale", "max_workers").SetOptional()
+	common.CustomizeSchemaPath(s, "autoscale", "min_workers").SetOptional()
+	common.CustomizeSchemaPath(s).AddNewField("apply_policy_default_values", &schema.Schema{
+		Optional: true,
+		Type:     schema.TypeBool,
+	})
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "s3", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "spark_version").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_id").SetOptional().SetComputed()
```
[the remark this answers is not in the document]
edwardfeng-db: I see, yeah good catch.

I updated some additional items with the following set to `Computed`. Let me know if this is accurate.
```
"cluster_cores"
"cluster_memory_mb"
"creator_user_name"
"driver"
"executors"
"jdbc_port"
"last_restarted_time"
"last_state_loss_time"
"spark_context_id"
"start_time"
"state_message"
"terminated_time"
"termination_reason"
```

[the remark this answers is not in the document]
edwardfeng-db: Okay yeah actually after changing it to use `ClusterSpec` these problems went away

clusters/resource_cluster.go
```diff
@@ -57,71 +58,111 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	common.CustomizeSchemaPath(s, "enable_elastic_disk").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "enable_local_disk_encryption").SetOptional().SetComputed()
+	common.CustomizeSchemaPath(s, "node_type_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
+	common.CustomizeSchemaPath(s, "driver_node_type_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_instance_pool_id", "instance_pool_id"})
+	common.CustomizeSchemaPath(s, "driver_instance_pool_id").SetComputed().SetOptional().SetConflictsWith([]string{"driver_node_type_id", "node_type_id"})
+	common.CustomizeSchemaPath(s, "ssh_public_keys").SetMaxItems(10)
+	common.CustomizeSchemaPath(s, "init_scripts").SetMaxItems(10).AddNewField("abfss", common.StructToSchema(InitScriptStorageInfo{}, nil)["abfss"]).AddNewField("gcs", common.StructToSchema(InitScriptStorageInfo{}, nil)["gcs"])
+	common.CustomizeSchemaPath(s, "init_scripts", "dbfs").SetDeprecated(DbfsDeprecationWarning)
+	common.CustomizeSchemaPath(s, "init_scripts", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "init_scripts", "s3", "destination").SetRequired()
```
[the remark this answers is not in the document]
edwardfeng-db: Updated that, PTAL, thanks!

clusters/resource_cluster.go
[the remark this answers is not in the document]
edwardfeng-db: Updated to use `ClusterSpec` now

clusters/resource_cluster.go
mgyucht: Ultimately, these things need to be derived one way or another from the OpenAPI specification. From my read, the main classes of issues are:
1. Computed fields (not in the OpenAPI spec, but potentially could be derived from comparing get to create/update).
2. Optional fields (this should be in the OpenAPI spec).
3. Required fields (this should be in the OpenAPI spec).
4. Fields that conflict with other fields (not yet in the OpenAPI spec, but can be with `oneOf` support).
5. OpenAPI spec missing information (in the short term, we can file a ticket to the upstream team, @edwardfeng-db would be great if you can do this; in the long term, our OpenAPI generator from @hectorcast-db will address this).
6. MaxItems validation (this could also be in the OpenAPI spec).
7. Deprecated fields (this is in the OpenAPI spec but is lost in the translation to the Go SDK).
8. Enum values (these are in the OpenAPI spec, but there is no convenient way to get all enum values in the Go SDK).
9. Sensitive data (needs an extension, probably, but could be in the OpenAPI spec).
10. Suppress diff

Here's how I think we can handle this:
2: Fields are optional by default, so I'm curious whether/why we need to call `SetOptional()` for so many of them. @edwardfeng-db why is this?
3, 5: file a ticket with the Clusters team to update their OpenAPI spec so that it is complete, including missing fields and the correct annotations for required/optional.
8: We can add a method to enums to list all values; then, we could automatically compute a validator for enum fields in the resource. Let's file a ticket for this @edwardfeng-db.
1, 4, 6, 7, 9, 10: I think most of these will be most easily addressed as we try to autogenerate the provider implementation from the OpenAPI spec itself; otherwise, they require hacks to embed this metadata somehow in Go SDK struct field tags.

common/reflect_resource.go
```diff
@@ -46,6 +63,35 @@ func reflectKind(k reflect.Kind) string {
...
+
+	var fieldName string
+
+	if value, ok := aliases[aliasKey]; ok {
+		fieldName = value
+	} else {
+		fieldName = jsonFieldName
+	}
+
+	return fieldName
```
mgyucht, line 82:
```suggestion
	if value, ok := aliases[aliasKey]; ok {
		return value
	}
	return jsonFieldName
```

common/reflect_resource_test.go
```diff
@@ -448,7 +448,7 @@ func TestTypeToSchemaNoStruct(t *testing.T) {
 			fmt.Sprintf("%s", p))
 	}()
 	v := reflect.ValueOf(1)
-	typeToSchema(v, []string{})
+	typeToSchema(v, []string{}, map[string]string{})
```
mgyucht, line 451: We need to add a test for aliases.

common/reflect_resource.go
```diff
@@ -46,6 +63,35 @@ func reflectKind(k reflect.Kind) string {
 	return n
 }
 
+func chooseFieldNameWithAliases(typeField reflect.StructField, fieldNamePath []string, aliases map[string]string) string {
```
mgyucht, line 66: nit: what do you think about having one function for choosing the field name that does:
1. Return `getAlias()` if not empty, or else
2. Check for the alias in the map, or else
3. Return getJsonFieldName

That way, in case we need to get the alias elsewhere, we don't need to use the `if` to call the right function.

mgyucht (commented):
Looking great. I don't think we need to block on the OpenAPI changes for this PR, but I do think we should take that as an immediate follow-up.

@tanmay-db let's also take note of these elements as things that need to be addressed as part of the autogeneration of TF resources from OpenAPI spec.

common/reflect_resource.go
edwardfeng-db: yep, sounds good, let me update that

clusters/resource_cluster.go
edwardfeng-db: Thanks. This sounds good. I'll create the follow up tickets.

For 2, there were many calls to `SetOptional()` because I was following the previous override function but I realized that a lot of these can be taken out because `omit_empty` exists on the openapi spec. I just took those out, it should look cleaner now.

common/reflect_resource_test.go
edwardfeng-db: Added, PTAL, thanks!

clusters/resource_cluster.go
edwardfeng-db: Ticket for enums: https://databricks.atlassian.net/browse/DECO-8056

mgyucht (approved):
LGTM, thanks man!

clusters/resource_cluster.go
```diff
@@ -57,71 +58,119 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+		Optional: true,
+		ForceNew: true,
+	})
+	common.CustomizeSchemaPath(s, "data_security_mode").SetSuppressDiff()
+	common.CustomizeSchemaPath(s, "docker_image", "url").SetRequired()
+	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "password").SetRequired().SetSensitive()
+	common.CustomizeSchemaPath(s, "docker_image", "basic_auth", "username").SetRequired()
+	common.CustomizeSchemaPath(s, "spark_conf").SetCustomSuppressDiff(SparkConfDiffSuppressFunc)
+	common.CustomizeSchemaPath(s, "aws_attributes").SetSuppressDiff().SetConflictsWith([]string{"azure_attributes", "gcp_attributes"})
+	common.CustomizeSchemaPath(s, "aws_attributes", "zone_id").SetCustomSuppressDiff(ZoneDiffSuppress)
```
alexott, line 98: We have few more attributes in the new schema - we need to at least document them:

```
@ ["provider_schemas","registry.terraform.io/databricks/databricks","resource_schemas","databricks_cluster","block","attributes","cluster_source"]
+ {"description_kind":"plain","optional":true,"type":"string"}
@ ["provider_schemas","registry.terraform.io/databricks/databricks","resource_schemas","databricks_cluster","block","block_types","aws_attributes","block","attributes","ebs_volume_iops"]
+ {"description_kind":"plain","optional":true,"type":"number"}
@ ["provider_schemas","registry.terraform.io/databricks/databricks","resource_schemas","databricks_cluster","block","block_types","aws_attributes","block","attributes","ebs_volume_throughput"]
+ {"description_kind":"plain","optional":true,"type":"number"}
@ ["provider_schemas","registry.terraform.io/databricks/databricks","resource_schemas","databricks_cluster","block","block_types","azure_attributes","block","block_types"]
+ {"log_analytics_info":{"block":{"attributes":{"log_analytics_primary_key":{"description_kind":"plain","optional":true,"type":"string"},"log_analytics_workspace_id":{"description_kind":"plain","optional":true,"type":"string"}},"description_kind":"plain"},"max_items":1,"nesting_mode":"list"}}
```

clusters/resource_cluster.go
```diff
@@ -57,71 +58,119 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	common.CustomizeSchemaPath(s, "init_scripts").SetMaxItems(10).AddNewField("abfss", common.StructToSchema(InitScriptStorageInfo{}, nil)["abfss"]).AddNewField("gcs", common.StructToSchema(InitScriptStorageInfo{}, nil)["gcs"])
+	common.CustomizeSchemaPath(s, "init_scripts", "dbfs").SetDeprecated(DbfsDeprecationWarning)
+	common.CustomizeSchemaPath(s, "init_scripts", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "init_scripts", "s3", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "init_scripts", "volumes", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "init_scripts", "workspace", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "workload_type", "clients").SetRequired()
+	common.CustomizeSchemaPath(s, "workload_type", "clients", "notebooks").SetDefault(true)
+	common.CustomizeSchemaPath(s, "workload_type", "clients", "jobs").SetDefault(true)
+	common.CustomizeSchemaPath(s).AddNewField("idempotency_token", &schema.Schema{
```
alexott, line 87: If it's not part of the `ClusterSpec` and isn't returned back by it, then we may get a permanent drift? And because iit's `ForceNew`, then it will try to recreate a cluster.

clusters/resource_cluster.go
```diff
@@ -57,71 +58,119 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	common.CustomizeSchemaPath(s, "azure_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "gcp_attributes"})
+	common.CustomizeSchemaPath(s, "gcp_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "azure_attributes"}).AddNewField(
+		"use_preemptible_executors",
+		&schema.Schema{
+			Type:       schema.TypeBool,
+			Optional:   true,
+			Deprecated: "Please use 'availability' instead.",
+		},
+	).AddNewField(
+		"zone_id",
```
alexott, line 108: If it's not a part of `ClusterSpec`, how it will behave?

clusters/resource_cluster.go
```diff
@@ -57,71 +58,119 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
-		s["url"] = &schema.Schema{
-			Type:     schema.TypeString,
-			Computed: true,
-		}
-		return s
+	common.CustomizeSchemaPath(s, "autotermination_minutes").SetDefault(60)
+	common.CustomizeSchemaPath(s, "autoscale", "max_workers").SetOptional()
+	common.CustomizeSchemaPath(s, "autoscale", "min_workers").SetOptional()
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "dbfs", "destination").SetRequired()
+	common.CustomizeSchemaPath(s, "cluster_log_conf", "s3", "destination").SetRequired()
```
alexott, line 128: It's a duplicate of lines 80-81

clusters/resource_cluster.go
```diff
@@ -57,71 +58,119 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+		Computed: true,
+	})
+	common.CustomizeSchemaPath(s, "num_workers").SetDefault(0).SetValidateDiagFunc(validation.ToDiagFunc(validation.IntAtLeast(0)))
+	common.CustomizeSchemaPath(s).AddNewField("cluster_mount_info", &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: common.StructToSchema(MountInfo{}, nil),
+		},
+	})
```
alexott, line 167: Same comment about `ClusterSpec` - if it's not returned in the payload, how do we handle the drift? We need to make sure to include it into the API spec

alexott (changes requested):
There are still changes required in the cluster resource.

clusters/resource_cluster.go
edwardfeng-db: Yeah they are pointing to the same object but with different paths. If I take these two lines out then the `destination` on this path wouldn't be marked as required.

clusters/resource_cluster.go
alexott: Ah, sorry - misread...

alexott:
There is a permanent drift at least on Azure:

```
  # databricks_cluster.shared_autoscaling will be updated in-place
  ~ resource "databricks_cluster" "shared_autoscaling" {
        id                           = "0118-192555-7kqkpwml"

      ~ azure_attributes {
          - availability       = "ON_DEMAND_AZURE" -> null
          - first_on_demand    = 1 -> null
          - spot_bid_max_price = -1 -> null
        }

        # (1 unchanged block hidden)
    }

```

Code to reproduce:

```hcl
terraform {
  required_providers {
    databricks = {
      version = "1.34.0"
    }
  }
}

provider "databricks" {
  # Configuration options
}


  local_disk = true
}

  long_term_support = true
}

  autotermination_minutes = 20
  idempotency_token = "123"
  autoscale {
    min_workers = 1
    max_workers = 50
  }
}
```

clusters/resource_cluster.go
edwardfeng-db: This is interesting.

- `cluster_source` Feels like this shouldn't be specified as an input argument, I asked in the ticket I filed for the cluster team
- `ebs_volume_iops`, `ebs_volume_throughput` and `log_analytics_info` are not documented on the REST API documentation as well. I also asked in that ticket

clusters/resource_cluster.go
mgyucht: This won't be impacted by this PR, but it will be when @tanmay-db migrates the resource implementation to the Go SDK.

clusters/resource_cluster.go
mgyucht: Ditto to above.

edwardfeng-db:
@alexott Thanks! I fixed the problem with the permanent drift on `azure_attributes`. It was due to `suppress_diff` not set on those fields. Now we set those properly.

clusters/resource_cluster.go
mgyucht: We'll mark cluster_source as read-only/computed and make a follow-up to document the rest.

mgyucht:
Todo: check aws & gcp attributes don't drift.

mgyucht:
Checked other nightlies: they did fail previously before suppress diff was added to the cloud-specific resources.

alexott (approved):

PR #3078, opened 2024-01-05
DON'T MERGE: Added `IS_OWNER` permission and fixed support for service principals for SQL Alerts/Dashboards/Queries

alexott:
## Changes

Default permissions API for SQL Alerts/Dashboards/Queries doesn't have `IS_OWNER` permission. Instead, there is a separate [REST API](https://docs.databricks.com/api/workspace/dbsqlpermissions/transferownership) to change ownership of the objects.

This PR implements the following things:

* Adds `IS_OWNER` permission for SQL Alerts/Dashboards/Queries by using a separate REST API
* It also fixes the previously existing issue with setting permissions for SQL Alerts/Dashboards/Queries for service principals. The current implementation has a problem with permanent drift because the [corresponding API](https://docs.databricks.com/api/workspace/dbsqlpermissions/set) doesn't have a setting for service principals, and it uses `user_name` instead.  The code now correctly rewrites SP -> user when setting permissions and user -> SP when reading permissions back.
* 
## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/resources/permissions.md
+++ b/docs/resources/permissions.md
@@ -660,7 +660,11 @@ resource "databricks_permissions" "endpoint_usage" {
 
 ## SQL Dashboard usage
 
-[SQL dashboards](https://docs.databricks.com/sql/user/security/access-control/dashboard-acl.html) have three possible permissions: `CAN_VIEW`, `CAN_RUN` and `CAN_MANAGE`:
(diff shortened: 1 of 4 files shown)


alexott:
not required as new query/alert apis support setting the owner

PR #3090, opened 2024-01-09
catalog/resource_grants: Refactor to use Go SDK

martin-walsh:
## Changes

Follow on from #3024 and update resource_grants to use Go SDK.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_grants.go
+++ b/catalog/resource_grants.go
@@ -143,11 +143,6 @@ var mapping = securableMapping{
 		"APPLY_TAG":      true,
 		"BROWSE":         true,
 	},
-	"view": { // TODO: Is this supported as it is not documented ?
-		"SELECT":    true,
-		"APPLY_TAG": true,
-		"BROWSE":    true,
-	},
 	"catalog": {
 		"CREATE": true,
 		"USAGE":  true,
@@ -240,11 +235,6 @@ var mapping = securableMapping{
 		"ALL_PRIVILEGES": true,
 		"APPLY_TAG":      true,
 		"EXECUTE":        true,
-	},
-	"materialized_view": { // TODO: Is this supported as it is not documented ?
-		"ALL_PRIVILEGES": true,
-		"SELECT":         true,
-		"REFRESH":        true,
 	},
 	"share": {
 		"SELECT": true,
--- a/catalog/resource_grants.go
+++ b/catalog/resource_grants.go
@@ -26,7 +26,7 @@ type PermissionsList struct {
 	Assignments []PrivilegeAssignment `json:"privilege_assignments" tf:"slice_set,alias:grant"`
 }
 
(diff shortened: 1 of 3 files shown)


martin-walsh:
See comment re `materialized_view` & `view` securable as I cannot see these documented on https://docs.databricks.com/api/workspace/grants/update

catalog/resource_grants.go
```diff
@@ -308,7 +241,7 @@ var mapping = securableMapping{
 		"APPLY_TAG":      true,
 		"EXECUTE":        true,
 	},
-	"materialized_view": {
+	"materialized_view": { // TODO: Is this supported as it is not documented ?
```
nkvuong, line 244: you are correct - this is done via `table` securable

catalog/resource_grants.go
```diff
@@ -210,7 +143,7 @@ var mapping = securableMapping{
 		"APPLY_TAG":      true,
 		"BROWSE":         true,
 	},
-	"view": {
+	"view": { // TODO: Is this supported as it is not documented ?
```
nkvuong, line 146: you are correct - this is done via `table` securable

internal/acceptance/grant_test.go
```diff
@@ -54,6 +54,13 @@ resource "databricks_external_location" "some" {
 	comment         = "Managed by TF"
 }
 
+resource "databricks_grant" "metastore" {
```
nkvuong, line 57: could you comment this block out for now (we need to exclude this from running in parallel with the grants test first)

martin-walsh: Sure.

catalog/resource_grants.go
martin-walsh: Ok, so, should I keep the `view` for backwards compat and just pass it to `table` similarly to `model` --> `function`, or remove these args ?.

Ditto `materialized_view` above.

martin-walsh: Follow up question: The newly added `databricks_grant` does not support these as I followed the api docs, so if you choose to keep `view` as an arg here should I add it to `databricks_grant` also for consistency ?

internal/acceptance/grant_test.go
martin-walsh: Done

catalog/resource_grants.go
nkvuong: `view` and `materialized_view` have never worked (the API will error out), so should be safe to remove - you just need to add these permissions back to `table` mapping

martin-walsh: Removed

mgyucht (approved):
Thanks for this!

@nkvuong let's talk offline about that, I can address it as I'm on-call this week.

PR #3093, opened 2024-01-09
Update databricks_model_serving resource to match latest API spec/SDK

### LINKED ISSUE #3067: Update `databricks_model_serving` resource to match latest API spec/SDK
We need to make a number of changes in `databricks_model_serving` to make it matching to latest API spec:

- Document new configuration options: `rate_limits`, `tags`, `config.auto_capture_config`
- Add `ForceNew` to some fields of `config.auto_capture_config`
- Add deprecation message to `config.served_models` to use `config.served_entities` instead
- Use `config.served_entities` instead of `config.served_models` to support foundational models, etc.

### LINKED ISSUE #3230: [FEATURE] Ability to add tags in databricks_model_serving resource.
### Use-cases

When terraforming model serving , I need to tag the resource for invoicing. Tagging is availible in model serving Databricks UI.  Current Terraform databricks_model_serving resource does not support tags as arguments. 

### Proposal

Add possibility for tagging in databricks_model_serving resource.  

### References

None found

840:
## Changes
We need to make a number of changes in `databricks_model_serving` to make it matching to latest API spec:

- [x] Document new configuration options: `rate_limits`, `tags`, `config.auto_capture_config`
- [x] Add ForceNew to some fields of `config.auto_capture_config`
- [x] Add deprecation message to `config.served_models` to use `config.served_entities` instead
- [x] Use `config.served_entities` instead of `config.served_models` to support foundational models, etc.

Resolves #3067, resolves #3230

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


docs/resources/model_serving.md
```diff
@@ -11,17 +11,17 @@ This resource allows you to manage [Model Serving](https://docs.databricks.com/m
 resource "databricks_model_serving" "this" {
   name = "ads-serving-endpoint"
   config {
-    served_models {
+    served_entities {
```
tanmay-db, line 14: We cannot do this directly right? There has to be a deprecation pathway otherwise it would be a breaking change.

tanmay-db: + @nkvuong

serving/resource_model_serving.go
```diff
@@ -29,6 +28,16 @@ func ResourceModelServing() *schema.Resource {
 				return old == "" && new == "CPU"
 			}
 			common.MustSchemaPath(m, "config", "traffic_config").Computed = true
+			common.MustSchemaPath(m, "config", "served_models").Deprecated = "Please use 'config.served_entities' instead of 'config.served_models'."
```
alexott, line 31: We need to add `ConflictsWith` between `served_models` and `served_entities` otherwise people could try to specify both

alexott (changes requested):
we may need to think on what we'll do with `config.auto_capture_config.enabled` - do we need to have ForceSendFields for it?

Also, it's not clear how to handle options inside the `config.served_entities.external_model.config` - technically we should have one entry per different model type, similar to the task types in jobs

docs/resources/model_serving.md
nkvuong: you need a state migration - similar to this https://github.com/databricks/terraform-provider-databricks/blob/08f816dd93ae969a0e2820386af2d19bc8c5bfde/catalog/resource_metastore_data_access.go#L81

alexott: @nkvuong do we really need it if the people can just change from `served_models` to `served_entities`? It should work imho

840: I agree with Alex. Apart from the name changes:
`served_models` -> `served_entities`
`served_models.*.model_name` -> `served_entities.*.entity_name`
`served_models.*.model_version` -> `served_entities.*.entity_version`

And the addition of `served_entities.*.external_model`, the two resources are interchangeable.

We still provide the option to use `served_models`, but a deprecation message will be added stating that you should use `served_entities`.

nkvuong: the issue will be with the read method, as it will return both `served_models` and `served_entities` and Terraform will treat that as a diff, so we need a way to suppress that.

docs/resources/model_serving.md
```diff
@@ -45,32 +45,73 @@ The following arguments are supported:
...
+* `entity_version` - The version of the model in Databricks Model Registry to be served or empty if the entity is a `FEATURE_SPEC`.
+* `workload_size` - The workload size of the served entity. The workload size corresponds to a range of provisioned concurrency that the compute autoscales between. A single unit of provisioned concurrency can process one request at a time. Valid workload sizes are `Small` (4 - 4 provisioned concurrency), `Medium` (8 - 16 provisioned concurrency), and `Large` (16 - 64 provisioned concurrency). If `scale-to-zero` is enabled, the lower bound of the provisioned concurrency for each workload size is 0.
+* `workload_type` - The workload type of the served entity. The workload type selects which type of compute to use in the endpoint. The default value for this parameter is `CPU`. For deep learning workloads, GPU acceleration is available by selecting workload types like `GPU_SMALL` and others. See the available [GPU types](https://docs.databricks.com/machine-learning/model-serving/create-manage-serving-endpoints.html#gpu-workload-types).
```
ankit-db, line 97: Can you add `min_provisioned_throughput` and `max_provisioned_throughput` as per the specification here: https://docs.databricks.com/en/machine-learning/foundation-models/deploy-prov-throughput-foundation-model-apis.html

docs/resources/model_serving.md
arpitjasa-db: FWIW, we don't store separate state for `served_models` and `served_entities`, and only one can ever be populated at a time on an endpoint. So what we can do is when both are returned, we discard the one that the user did not specify and populate the one that the user did (is that possible in Terraform)? @nkvuong

docs/resources/model_serving.md
```diff
@@ -45,32 +45,75 @@ The following arguments are supported:
...
+        * `traffic_percentage` - (Required) The percentage of endpoint traffic to send to this route. It must be an integer between 0 and 100 inclusive.
+  * `auto_capture_config` - Configuration for Inference Tables which automatically logs requests and responses to Unity Catalog.
+    * `catalog_name` - The name of the catalog in Unity Catalog. NOTE: On update, you cannot change the catalog name if it was already set.
+    * `schema_name` - The name of the schema in Unity Catalog. NOTE: On update, you cannot change the schema name if it was already set.
+    * `table_name_prefix` - The prefix of the table in Unity Catalog. NOTE: On update, you cannot change the prefix name if it was already set.
+    * `enabled` - If inference tables are enabled or not. NOTE: If you have already disabled payload logging once, you cannot enable again.
+* `tags` - Tags to be attached to the serving endpoint and automatically propagated to billing logs.
+* `rate_limits` - Rate limits to be applied to the serving endpoint. NOTE: only external and foundation model endpoints are supported as of now.
```
arpitjasa-db, line 60: Can we add a block in the documents expanding on the structure for `rate_limits` as well?

docs/resources/model_serving.md
alexott: @arpitjasa-db That's what the latest changes are about - discarding what isn't specified by user.
One question - if I replace `served_models` with `served_entities` in the existing serving endpoint - it won't cause any redeployment, etc.?

alexott: @mgyucht ^

arpitjasa-db: Well it doesn't cause a real redeployment, but it behaves the same as when you call "UpdateEndpoint" with no diff in the config, where it go into an updating state briefly (~30 seconds) and increments the config version
Screenshot 2024-02-05 at 4 03 10 PM

compared to a real update which takes longer (~15 min)
Screenshot 2024-02-05 at 4 57 50 PM

alexott: Thank you! @840 We'll need to put a note saying about that

840:
> Also, it's not clear how to handle options inside the `config.served_entities.external_model.config` - technically we should have one entry per different model type, similar to the task types in jobs

@arpitjasa-db It looks like `config.served_entities.external_model.provider` is tightly coupled with `config.served_entities.external_model.<provider>_config`.

Is it a guarantee that the provider name will always be the prefix of the corresponding config? For example:

config.served_entities.external_model.provider = `ai21labs`
config.served_entities.external_model.`ai21labs`_config

arpitjasa-db:
> > Also, it's not clear how to handle options inside the `config.served_entities.external_model.config` - technically we should have one entry per different model type, similar to the task types in jobs
> 
> @arpitjasa-db It looks like `config.served_entities.external_model.provider` is tightly coupled with `config.served_entities.external_model.<provider>_config`.
> 
> Is it a guarantee that the provider name will always be the prefix of the corresponding config? For example:
> 
> config.served_entities.external_model.provider = `ai21labs` config.served_entities.external_model.`ai21labs`_config

@840 Yes but with a caveat, for multi-word models, the connector will be hyphens (e.g. `databricks-model-serving` or `aws-bedrock`), but the config will be underscores (e.g. `databricks_model_serving_config` or `aws_bedrock_config`)

serving/resource_model_serving.go
```diff
@@ -96,3 +137,31 @@ func ResourceModelServing() common.Resource {
 		},
 	}
 }
+
+func hasExactlyOneExternalModel(d *schema.ResourceDiff) error {
```
mgyucht, line 141: I don't think this is what this is checking. It seems to verify two things:
1. that you specified the appropriate config for the provider of the first external model, and
2. that you did not specify any other configs for the first external model.
Let's correct the name here. Maybe `validateExternalModelConfig` is good?

BTW: it might be easier to write this check if you do
```go
var req serving.CreateServingEndpoint
common.DataToStructPointer(schema, d, &req)
```
and then do your checks on the typed `req` rather than trying to get the right fields from TF.

serving/resource_model_serving.go
```diff
@@ -55,13 +83,26 @@ func ResourceModelServing() common.Resource {
...
+			var sOrig serving.ServingEndpointDetailed
+			common.DataToStructPointer(d, s, &sOrig)
 			if err != nil {
 				return err
 			}
 			endpoint, err := w.ServingEndpoints.GetByName(ctx, d.Id())
 			if err != nil {
 				return err
 			}
+			if d.IsNewResource() || sOrig.Config == nil {
```
mgyucht, line 95: Is `d.IsNewResource()` check needed? Even if it returned true, wouldn't we still want to apply the same logic as the else case?

serving/resource_model_serving.go
```diff
@@ -39,6 +56,17 @@ func ResourceModelServing() common.Resource {
 		})
 
 	return common.Resource{
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff) error {
+			old, new := d.GetChange("config.0.auto_capture_config.0.enabled")
+			if old != nil && old == false && new == true {
+				d.ForceNew("config.0.auto_capture_config.0.enabled")
```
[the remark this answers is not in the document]
arpitjasa-db: Enabling the first time does not require destroy and recreate, but [disabling and then trying to re-enable would](https://docs.databricks.com/en/machine-learning/model-serving/inference-tables.html#enable-and-disable-inference-tables)
Screenshot 2024-03-04 at 11 29 03 AM

serving/resource_model_serving.go
```diff
@@ -96,3 +137,31 @@ func ResourceModelServing() common.Resource {
 		},
 	}
 }
+
+func hasExactlyOneExternalModel(d *schema.ResourceDiff) error {
+	_, e := d.GetOk("config.0.served_entities.0.external_model")
+	provider, p := d.GetOk("config.0.served_entities.0.external_model.0.provider")
+	if e && p {
```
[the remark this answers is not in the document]
840: Thanks! Added the guard clause

docs/resources/model_serving.md
```diff
@@ -45,32 +47,78 @@ The following arguments are supported:
...
-
-* `served_models` - (Required) Each block represents a served model for the endpoint to serve. A model serving endpoint can have up to 10 served models.
-* `traffic_config` - A single block represents the traffic split configuration amongst the served models.
-
-### served_models Configuration Block
+  * `served_entities` - A list of served entities for the endpoint to serve. A serving endpoint can have up to 10 served entities.
+  * `served_models` - (Deprecated, use `served_entities` instead) Each block represents a served model for the endpoint to serve. A model serving endpoint can have up to 10 served models.
+  * `traffic_config` - A single block represents the traffic split configuration amongst the served models.
+    * `routes` - (Required) Each block represents a route that defines traffic to each served model. Each `served_models` block needs to have a corresponding `routes` block
+        * `served_model_name` - (Required) The name of the served model this route configures traffic for. This needs to match the name of a `served_models` block
```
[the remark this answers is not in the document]
840: Yes, looks like this indeed was changed to `served_entity_name`
image

We should update the [API docs](https://docs.databricks.com/api/workspace/servingendpoints/create#config-traffic_config-routes-served_model_name)

docs/resources/model_serving.md
```diff
@@ -45,32 +47,78 @@ The following arguments are supported:
...
+    * `schema_name` - The name of the schema in Unity Catalog. NOTE: On update, you cannot change the schema name if it was already set.
+    * `table_name_prefix` - The prefix of the table in Unity Catalog. NOTE: On update, you cannot change the prefix name if it was already set.
+    * `enabled` - If inference tables are enabled or not. NOTE: If you have already disabled payload logging once, you cannot enable again.
+* `tags` - Tags to be attached to the serving endpoint and automatically propagated to billing logs.
+* `rate_limits` - A list of rate limits to be applied to the serving endpoint. NOTE: only external and foundation model endpoints are supported as of now.
+  * `calls` - (Required) Used to specify how many calls are allowed for a key within the renewal_period.
+  * `key` - Key field for a serving endpoint rate limit. Currently, only `user` and `endpoint` are supported, with `endpoint` being the default if not specified.
+  * `renewal_period` - (Required) Renewal period field for a serving endpoint rate limit. Currently, only `minute` is supported.
```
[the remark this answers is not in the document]
840: Good point, done

serving/resource_model_serving.go
840: Decided to go with the current approach since using a typed approach would require reflection and string mutation for CamelCase & snake_case

mgyucht (approved):

PR #3112, opened 2024-01-13
WIP: Exporter: Fix exporting of `parent` attribute for SQL objects

alexott:
## Changes

There is an edge case when SQL object (dashboard/query/alert) is in the user home folder, not a subfolder - in this case instead of reference to a resource, we generated a reference to the data source, and because `parent` is marked as `ForceNew` it lead to SQL object recreation when importing existing SQL objects into the TF state.

The workaround is to check if the given parent folder is top-level user folder, and remove the `parent` because by default SQL object is put into home folder of the user.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -810,6 +810,10 @@ func (ic *importContext) ResourceName(r *resource) string {
 	return name
 }
 
+func (ic *importContext) isServiceEnabled(service string) bool {
+	return strings.Contains(ic.services, service)
+}
+
 func (ic *importContext) Emit(r *resource) {
 	// TODO: change into channels, if stack trace depth issues would surface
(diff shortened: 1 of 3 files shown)


alexott:
@mgyucht do we have a doc on this API: https://github.com/databricks/terraform-provider-databricks/pull/3052/files#diff-4e6c247f890143b35a00162638c63d0d18e0a596acb35ae7c03124a35a967c49R114 ?

Maybe we should force the creation of home folders in `databricks_user` and `databricks_service_principal`?

alexott:
After a bit of thinking, the logic of this PR isn't completely correct. Created #3113 to track the actual issue.

PR #3121, opened 2024-01-16
Support Go SDK Mocking Library in Terraform Unit Tests

mgyucht:
## Changes
The Databricks Go SDK recently introduced a mocking library to make it easier to write unit tests, and some improvements to this were included in version [0.29.1](https://github.com/databricks/databricks-sdk-go/releases/tag/v0.29.1). This PR adds support to our unit testing framework to mock the Go SDK interface rather than the network interface. This has the following advantages:

1. The SDK can make multiple requests for a single API call, for example, to list paginated resources. With mocking, you can mock the list of resources without having to emulate the actual API behavior.
2. The return types of the mocked APIs are type-checked: you need to return something with the correct type for the called method.
3. No more need to know the underlying URL and method when mocking.

To verify this, I moved two sets of tests (SQL warehouses & metastores) to this mocking approach.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott:
The existing framework supports two use cases:

* Mock a REST API call, and use that mock for all calls
* Mock a REST API call with the same request params but have different answers

Is it supported?

mgyucht:
1. The default behavior for the stretchr mocking library is to return the arguments specified for every call, so you don't need to do anything special for that.
2. If you want to have different answers, you mock the request a second time. You need to specify `.Once()` to ensure that that response is only used one time.
```
func TestMockReturnDifferentValues(t *testing.T) {
	w := mocks.NewMockWorkspaceClient(t)
	e := w.GetMockCatalogsAPI().EXPECT()
	e.GetByName(mock.Anything, "abc").Return(&catalog.CatalogInfo{
		Name: "a",
	}, nil).Once()
	e.GetByName(mock.Anything, "abc").Return(&catalog.CatalogInfo{
		Name: "b",
	}, nil).Once()
	ctx := context.Background()
	first, err := w.WorkspaceClient.Catalogs.GetByName(ctx, "abc")
	assert.NoError(t, err)
	second, err := w.WorkspaceClient.Catalogs.GetByName(ctx, "abc")
	assert.NoError(t, err)
	assert.Equal(t, "a", first.Name)
	assert.Equal(t, "b", second.Name)
}
```

qa/testing.go
```diff
@@ -190,34 +196,84 @@ func (f ResourceFixture) setDatabricksEnvironmentForTest(client *common.Databric
 	}
 }
 
-// Apply runs tests from fixture
-func (f ResourceFixture) Apply(t *testing.T) (*schema.ResourceData, error) {
+func (f ResourceFixture) validateMocks() error {
+	isMockConfigured := f.MockAccountClientFunc != nil || f.MockWorkspaceClientFunc != nil
+	isFixtureConfigured := f.Fixtures != nil
+	if isFixtureConfigured && isMockConfigured {
```
alexott, line 202: In some cases, we may not use API for everything - I don't know if it will be possible to have both mocks together?

alexott (approved):
lgtm

PR #3129, opened 2024-01-17
Bump TF provider to go 1.21

mgyucht:
## Changes
Let's use the latest Go version to take advantage of [all of the great things in go 1.21](https://tip.golang.org/doc/go1.21). Of note, we can move to `slices` in the standard library and improvements to type inference.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott (changes requested):
We need to update github actions as well:

* https://github.com/databricks/terraform-provider-databricks/blob/master/.github/workflows/push.yml#L22
* https://github.com/databricks/terraform-provider-databricks/blob/master/.github/workflows/release.yml#L22
* ...

alexott (approved):

PR #3132, opened 2024-01-18
Fix public_access_enabled property for mws_private_access_settings

hectorcast-db:
## Changes
Fix public_access_enabled property for mws_private_access_settings by adding 

## Tests

- [X] `make test` run locally
  - Tests for mws_private_access_settings are disabled
- [X] using Go SDK


docs/resources/mws_private_access_settings.md
```diff
@@ -73,7 +73,7 @@ The following arguments are available:
...
+* `public_access_enabled` (Boolean, Optional, `false` by default on AWS, `true` by default on GCP) - If `true`, the [databricks_mws_workspaces](mws_workspaces.md) can be accessed over the [databricks_mws_vpc_endpoint](mws_vpc_endpoint.md) as well as over the public network. In such a case, you could also configure an [databricks_ip_access_list](ip_access_list.md) for the workspace, to restrict the source networks that could be used to access it over the public network. If `false`, the workspace can be accessed only over VPC endpoints, and not over the public network. Once explicitely set, this field becomes mandatory.
```
alexott, line 76: `explicitely` -> `explicitly`

mws/resource_mws_private_access_settings_test.go
```diff
@@ -51,6 +51,51 @@ func TestResourcePASCreate(t *testing.T) {
...
+				Method:   "GET",
+				Resource: "/api/2.0/accounts/abc/private-access-settings/pas_id",
+
+				Response: PrivateAccessSettings{
+					AccountID:           "abc",
+					PasID:               "pas_id",
+					PublicAccessEnabled: false,
+					Region:              "ar",
+					PasName:             "pas_name",
+					ForceSendFields:     []string{"PublicAccessEnabled"},
```
alexott, line 82: is it really returned as answer?

alexott (commented):
looks good, just few comments

mws/resource_mws_private_access_settings_test.go
hectorcast-db: Do you mean the ForceSendFields?
It is not returned per se. The mocking library does not return the object itself (it's here just for convenience), but instead will marshall this object and "return" the json, which then is deserialised.
We need to add it here so that the returned json contains the relevant field.

alexott (approved):

mgyucht:
Ah, was gonna migrate this to the Go SDK at the same time...

PR #3135, opened 2024-01-19
Migrate `databricks_mws_private_access_settings` to Go SDK

mgyucht:
## Changes
Migrates this resource to the Go SDK. Also marks the `account_id` field at the resource level as deprecated, like what we did for `databricks_mws_credentials`.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


mws/resource_mws_private_access_settings.go
```diff
@@ -2,110 +2,77 @@ package mws
...
-			var pas PrivateAccessSettings
+			a, err := c.AccountClientWithAccountIdFromConfig(d)
+			if err != nil {
+				return err
+			}
+			var pas provisioning.UpsertPrivateAccessSettingsRequest
 			common.DataToStructPointer(d, s, &pas)
 			common.SetForceSendFields(&pas, d, []string{"public_access_enabled"})
-			if err := NewPrivateAccessSettingsAPI(ctx, c).Create(&pas); err != nil {
+			res, err := a.PrivateAccess.Create(ctx, provisioning.UpsertPrivateAccessSettingsRequest(pas))
```
hectorcast-db, line 40: Why do we create a new `provisioning.UpsertPrivateAccessSettingsRequest` ?

mgyucht: Good catch. This doesn't make a new one but just does a type conversion to itself. Removing this now.

hectorcast-db (approved):

PR #3136, opened 2024-01-19
Use double quotes with filters in SCIM

mgyucht:
## Changes
The SCIM API team is changing the library used to parse SCIM requests. This library is more strict with the interpretation of filters, specifically mandating that string values in filters, like for display name or application ID, must use double quotes instead of single quotes. This PR modifies SCIM APIs to use double quotes for strings in filters instead.

## Tests
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/data_group_test.go
+++ b/scim/data_group_test.go
@@ -18,7 +18,7 @@ func TestDataSourceGroup(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "GET",
-				Resource: "/api/2.0/preview/scim/v2/Groups?filter=displayName%20eq%20%27ds%27",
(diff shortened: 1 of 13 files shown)


scim/data_group_test.go
```diff
@@ -18,7 +18,7 @@ func TestDataSourceGroup(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "GET",
-				Resource: "/api/2.0/preview/scim/v2/Groups?filter=displayName%20eq%20%27ds%27",
+				Resource: `/api/2.0/preview/scim/v2/Groups?filter=displayName%20eq%20%22ds%22`,
```
hectorcast-db, line 21: nit, I don't think the string quotes need to be changed here

hectorcast-db (approved):

alexott (approved):

PR #3150, opened 2024-01-25
Adds `databricks_volumes` as data source 

### LINKED ISSUE #3133: [FEATURE] Data source for `databricks_volume`
### Use-cases
It's needed so that we could grant access to externally managed volumes with `databricks_grants`

karolusz:
## Changes
Adds  `databricks_volumes` data source.

A set of volumes ids (full names) becomes available as a data source, in a similar fashion to existing Unity Catalogs objects data sources (e.g. `databricks_schemas` or `databricks_tables`).

Closes #3133 

## Tests

- [x] `make test` run locally
- [x] using Go SDK


docs/data-sources/volumes.md
```diff
@@ -0,0 +1,39 @@
...
+## Argument Reference
+
+* `catalog_name` - (Required) Name of [databricks_catalog](../resources/catalog.md)
+* `schema_name` - (Required) Name of [databricks_schema](../resources/schema.md)
+
+## Attribute Reference
+
+This data source exports the following attributes:
+
+* `ids` - a set of [databricks_volume](../resources/volume.md) full names: *`catalog`.`schema`.`volume`*
```
tanmay-db, line 32: nit: Since we are mentioning above: `Retrieves a list of`, should we also mention list here instead of set for consistency?

docs/data-sources/volumes.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Related Resources
+
+The following resources are used in the same context:
+
+* [databricks_schema](../resources/schema.md) to manage schemas within Unity Catalog.
```
tanmay-db, line 38: also databricks volumes resource

tanmay-db (commented):
This looks good to me, however there is concern of usability of list of ids instead of getting information for each id. We should also have databricks_volume resource so we can query information for each id. 

In list of ids, maybe we should provide their respective names as well because getting a list of ids won't be very useful in their own. Let's discuss this more in terraform office hours today. 

We do this for other data sources as well for example: `databricks_share`, `databricks_service_principal`

karolusz:
> This looks good to me, however there is concern of usability of list of ids instead of getting information for each id. We should also have databricks_volume resource so we can query information for each id.
> 
> In list of ids, maybe we should provide their respective names as well because getting a list of ids won't be very useful in their own. Let's discuss this more in terraform office hours today.
> 
> We do this for other data sources as well for example: `databricks_share`, `databricks_service_principal`

The only viable use currently would be to loop over all volumes in the schema when creating `databricks_grants` resources for those volumes.

`databricks_shares` and `databricks_service_principals`  data sources return some list of identifiers, which then allow you to reference specific instances of those resources via `databricks_share` and `databricks_service_principals`.

Please let me know what is decided internally and I would be happy to make the adjustments accordingly.

docs/data-sources/volumes.md
karolusz: Changes made.

docs/data-sources/volumes.md
karolusz: Changes made.

tanmay-db (approved):
LGTM given we have similar data sources for schemas, metastores so this is consistent with that.

karolusz:
this will need a little rework after the latest changes

karolusz:
Small changes were  made to solve a merge conflict.

PR #3152, opened 2024-01-25
Update actions/checkout to v4

pietern:
## Changes

* Update the action to the latest version
* Remove unnecessary unshallow step on tests
* Fold unshallow step into actions/checkout on release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,10 +11,10 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v2
-      -
-        name: Unshallow
-        run: git fetch --prune --unshallow
+        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -11,10 +11,10 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v2
-      -
-        name: Unshallow
-        run: git fetch --prune --unshallow
+        uses: actions/checkout@v4
+        with:
```
hectorcast-db, line 15: Why do we need this here and not above?

pietern: The release action needs to know the current tag. This is not populated by default and needs the unshallow.

hectorcast-db (approved):

PR #3153, opened 2024-01-25
Test if tags are fetched with fetch-depth: 0

pietern:
This is a test only.


pietern:
https://github.com/databricks/terraform-provider-databricks/actions/runs/7654490587/job/20858587208?pr=3153

This confirms that `fetch-depth: 0` is enough.

PR #2529, opened 2023-07-28
Added READ_VOLUME and WRITE_VOLUME to schema/catalog level grants

jl178:
## Changes
According to the [Databricks documentation](https://docs.databricks.com/data-governance/unity-catalog/manage-privileges/privileges.html) & the [docs in this repo](https://github.com/databricks/terraform-provider-databricks/blob/6b7220ac4a5f8134fdaf4dc0ffdaf373bc370c00/docs/resources/grants.md?plain=1#L59), you should be able to have `READ_VOLUME` and `WRITE_VOLUME` at the schema level. However, this currently fails [here](https://github.com/databricks/terraform-provider-databricks/blob/master/catalog/resource_grants.go#L162).

Example resource block:
```hcl
resource "databricks_grants" "volumes_grant" {
  schema = databricks_schema.volumes.id
  grant {
    principal  = databricks_group.group.display_name
    privileges = ["READ_VOLUME", "WRITE_VOLUME"]
  }
}
```

Error:

```
Error: READ_VOLUME is not allowed on schema
```

It seems like this was missed during [this PR](https://github.com/databricks/terraform-provider-databricks/pull/2399), even though the docs were updated correctly.

This is my first contribution, so please let me know if I am missing something! Thanks!

## Tests

- [x] `make test` run locally
- [ ] ~~relevant change in `docs/` folder~~
- [ ] ~~covered with integration tests in `internal/acceptance`~~
- [ ] ~~relevant acceptance tests are passing~~
- [ ] ~~using Go SDK~~

PR #2399, opened 2023-06-13
Update `databricks_grant` resource with latest set of permissions

nkvuong:
## Changes
Update `databricks_grant` resource with latest set of permissions (mostly for `databricks_volume`)
Close #2384

## Tests

- [x] `make test` run locally
- [x] relevant change in `docs/` folder
- [ ] ~covered with integration tests in `internal/acceptance`~
- [x] relevant acceptance tests are passing
- [ ] ~using Go SDK~

PR #3163, opened 2024-01-26
Added support for `RawConfig()` in testing framework and fixed `databricks_grants` resource for "Error: unknown is not fully supported yet"

tanmay-db:
## Changes

`databricks_grants` resource marks all fields in secure_mapping as force_new and when these ids change (caused by change in name of other UC resources for example: `name` in `databricks_external_location`, this causes terraform to have 2 diff runs during the plan phase. From CustomizeDiff in schema.Resource:

> Existing resource, forced new: One run with state (before ForceNew), then one run without state (as if new resource)

We use the RawConfig() and for that support is added in unit test framework. 

## Tests

Added Unit test, Integration test, Manually tested E2E after making terraform binary to check the actual E2E flow without any testing framework mocks.  
- [ ] `make test` run locally
- [ ] using Go SDK


tanmay-db:
This is WIP -- not marking it as draft since I would like to see unit test if any failures on CI. Will add to review once tests are added.

catalog/resource_grants.go
```diff
@@ -316,6 +328,9 @@ func ResourceGrants() common.Resource {
 			}
 			var grants PermissionsList
 			common.DiffToStructPointer(d, s, &grants)
+			if firstPassOfForcedNew(d) {
```
tanmay-db, line 331: Testing for this is covered in integration test below. 

We need to add support for testing CustomizeDiff before we can unit test the 2phases flow of forced new as part of unit test.

catalog/resource_grants.go
```diff
@@ -374,7 +389,12 @@ func ResourceGrants() common.Resource {
 			var grants PermissionsList
 			common.DataToStructPointer(d, s, &grants)
 			unityCatalogPermissionsAPI := permissions.NewUnityCatalogPermissionsAPI(ctx, c)
-			return replaceAllPermissions(unityCatalogPermissionsAPI, securable, name, grants.toSdkPermissionsList())
+			err = replaceAllPermissions(unityCatalogPermissionsAPI, securable, name, grants.toSdkPermissionsList())
+			if err != nil {
+				return err
+			}
+			d.SetId(mapping.id(d))
```
tanmay-db, line 396: This was another bug, we don't update id when changing name of underlying resource

internal/acceptance/grant_test.go
```diff
@@ -54,13 +55,6 @@ resource "databricks_external_location" "some" {
 	comment         = "Managed by TF"
 }
 
-# resource "databricks_grant" "metastore" {
```
tanmay-db, line 57: Removing dead code

catalog/resource_grants.go
```diff
@@ -290,6 +290,18 @@ func parseId(d *schema.ResourceData) (string, string, error) {
 	return split[0], split[1], nil
 }
 
+// Check whether it's a first pass of forced new. From CustomizeDiff in schema.Resource:
+// "Existing resource, forced new: One run with state (before ForceNew), then one run without state (as if new resource)"
+func firstPassOfForcedNew(d *schema.ResourceDiff) bool {
+	fieldSetAsId := strings.Split(d.Id(), "/")[0]
```
mgyucht, line 296: Let's reuse existing code rather than copying this here. Also, instead of `fieldSetAsId`, how about something like `securableType`?
```suggestion
	fieldSetAsId, _, _, _ := parseSecurableId(d.Id())
```

catalog/resource_grants.go
```diff
@@ -290,6 +290,18 @@ func parseId(d *schema.ResourceData) (string, string, error) {
...
+	old, new := d.GetChange(fieldSetAsId)
+	if old.(string) != "" && new.(string) == "" {
+		// First pass of forced new
+		return true
+	}
+	return false
```
mgyucht, line 302: This logic is incorrect, I believe. In the first pass, the state will be present, so there will be a change. It passes unit tests because our synthesized InstanceState has a bug: it doesn't include the resource ID in the state. This causes CustomizeDiff to be [skipped earlier](https://github.com/databricks/terraform-provider-databricks/pull/3163/files#diff-b368dac73336f7c30e2ef9bf3258af896959ba0b75433759c6bb3499fad4feedR325).

In this PR, let's update `testing.go` to have
```go
	is := &terraform.InstanceState{
		ID:         f.ID,
		Attributes: f.InstanceState,
	}
```
on line 302. Then, you can see that this check returns `false` for existing resources.

internal/acceptance/grant_test.go
```diff
@@ -104,3 +98,30 @@ func TestUcAccGrant(t *testing.T) {
...
+	
+	resource "databricks_grant" "cred" {
+		storage_credential = databricks_storage_credential.external.id
+		principal  = "{env.TEST_DATA_ENG_GROUP}"
+		privileges = ["ALL_PRIVILEGES"]
+	}
+	`, suffix)
+}
+
+func TestUcAccGrantForIdChange(t *testing.T) {
```
mgyucht, line 120: I tried running this test in debug mode, and it didn't seem to hit the CustomizeDiff codepath.

catalog/resource_grants.go
```diff
@@ -290,6 +290,18 @@ func parseId(d *schema.ResourceData) (string, string, error) {
 	return split[0], split[1], nil
 }
 
+// Check whether it's a first pass of forced new. From CustomizeDiff in schema.Resource:
```
mgyucht, line 293: Is the same logic needed in `databricks_grant` as well?

catalog/resource_grants_test.go
```diff
@@ -600,6 +600,72 @@ func TestShareGrantUpdate(t *testing.T) {
...
+						{
+							Principal:  "me",
+							Privileges: []catalog.Privilege{"CREATE_EXTERNAL_TABLE"},
+						},
+					},
+				},
+			},
+		},
+		Resource:    ResourceGrants(),
+		Update:      true,
```
mgyucht, line 651: This is a very weird test. It verifies that the Update pathway works when the underlying external_location changes. However, in reality, the Update pathway is never taken when the external_location changes; instead, we delete and recreate. Obviously that can't be mocked with this test framework, but it is a bit weird.

catalog/resource_grants.go
mgyucht: Is this actually a bug? When do we need to change the ID? If either the securable type or securable itself changes, the resource will be recreated, so this pathway is not taken.

catalog/resource_grants.go
```diff
@@ -316,6 +328,9 @@ func ResourceGrants() common.Resource {
 			}
 			var grants PermissionsList
 			common.DiffToStructPointer(d, s, &grants)
+			if firstPassOfForcedNew(d) {
+				return nil
+			}
 			return mapping.validate(d, grants)
```
mgyucht, line 334: After looking at this code for 2.5 hours... I think I finally realized what the core issue is and that the actual bug is in `securableMapping.validate()`, specifically `kv()`. The issue is that the new value of the securable may not be known at plan time, but we still should know the type of the securable. That is sufficient for validation. This means that in `kv()`, instead of getting the value of the field, we can check the raw config to see which field is not nil.
grant_validation

mgyucht (commented):
Sorry this took so long. I spent a long time trying to understand TF internals to make sure I really understood this solution. While this solves the bug in a way, I think it does skip validation on the recreate pathway. In my comment, I suggest another approach: inspect the raw config during CustomizeDiff, at which point we definitely know what kind of grant we're dealing with, which is enough for validation purposes.

Also, doesn't this need to be applied to `databricks_grant` as well?

catalog/resource_grants.go
```diff
@@ -93,16 +94,27 @@ type securableMapping map[string]map[string]bool
...
 // reuse ResourceDiff and ResourceData
 type attributeGetter interface {
-	Get(key string) any
+	GetRawConfig() cty.Value
 }
 
 func (sm securableMapping) kv(d attributeGetter) (string, string) {
+	rawConfig := d.GetRawConfig()
+	if rawConfig.IsNull() {
+		return "unknown", "unknown"
```
mgyucht, line 103: We should return some kind of error here. We basically don't expect this case in the happy path. If the user hasn't specified any resource, we should say so ("Specify one of: ..."). If they have, it is definitely a provider bug.

catalog/resource_grants.go
```diff
@@ -93,16 +94,27 @@ type securableMapping map[string]map[string]bool
...
+		if value.IsNull() {
 			continue
 		}
-		return field, v
+		// We need this check because:
+		// "AsString returns the native string from a non-null, non-unknown cty.String value, or panics if called on any other value."
+		if !value.IsKnown() {
+			// This is the case when the new value isn't known but the field is required for validation
+			return field, "unknown"
+		}
```
mgyucht, line 116: Excellent comments!

catalog/resource_grants.go
```diff
@@ -93,16 +94,27 @@ type securableMapping map[string]map[string]bool
...
+		return field, value.AsString()
 	}
 	return "unknown", "unknown"
```
mgyucht, line 119: What should we do in this case?

internal/acceptance/grant_test.go
```diff
@@ -104,3 +98,30 @@ func TestUcAccGrant(t *testing.T) {
...
+	unityWorkspaceLevel(t, step{
+		Template: grantTemplateForNameChange("-old"),
+	},
+		step{
+			Template: grantTemplateForNameChange("-new"),
+		})
```
mgyucht, line 126: Can we add a step that validates that the template fails when an invalid permission is passed?

internal/acceptance/grant_test.go
```diff
@@ -104,3 +98,30 @@ func TestUcAccGrant(t *testing.T) {
...
+		principal  = "{env.TEST_DATA_ENG_GROUP}"
+		privileges = ["ALL_PRIVILEGES"]
+	}
+	`, suffix)
+}
+
+func TestUcAccGrantForIdChange(t *testing.T) {
+	unityWorkspaceLevel(t, step{
+		Template: grantTemplateForNameChange("-old"),
+	},
```
mgyucht, line 123: nit: remove the newline so all steps have same indenting

internal/acceptance/grants_test.go
```diff
@@ -110,3 +111,32 @@ func TestUcAccGrants(t *testing.T) {
...
+			privileges = ["ALL_PRIVILEGES"]
+		}
+	}
+	`, suffix)
+}
+
+func TestUcAccGrantsForIdChange(t *testing.T) {
+	unityWorkspaceLevel(t, step{
+		Template: grantsTemplateForNameChange("-old"),
+	},
```
mgyucht, line 138: ditto: remove newline so steps have same indenting

qa/testing.go
```diff
@@ -249,6 +249,49 @@ func (f ResourceFixture) setupClient(t *testing.T) (*common.DatabricksClient, se
...
 }
 
+// Quoting from: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/internal/configs/hcl2shim/values.go#L19
+// "UnknownVariableValue is a sentinel value that can be used
+// to denote that the value of a variable is unknown at this time.
+// RawConfig uses this information to build up data about
+// unknown keys.""
+const UnknownVariableValue = "74D93920-ED26-11E3-AC10-0800200C9A66"
+
+// Reference: HCL2ValueFromConfigValue from https://github.com/hashicorp/terraform-plugin-sdk/blob/main/internal/configs/hcl2shim/values.go#L199
```
mgyucht, line 259: We're able to use this exactly as is?

qa/testing.go
```diff
@@ -249,6 +249,49 @@ func (f ResourceFixture) setupClient(t *testing.T) (*common.DatabricksClient, se
...
+func GetRawConfig(v interface{}) cty.Value {
```
mgyucht, line 260: This isn't a good name choice, as this method has nothing to do with configs, only `cty.Value`. How about something like `ConvertToValue`?

qa/testing_test.go
```diff
@@ -332,3 +332,55 @@ func TestResourceCornerCases(t *testing.T) {
 func TestAssertErrorStartsWith(t *testing.T) {
 	AssertErrorStartsWith(t, fmt.Errorf("abc"), "a")
 }
+
+func TestGetRawConfig(t *testing.T) {
```
mgyucht, line 336: Can we use `t.Run()` for each of these test cases? Especially panic test cases: you won't be able to add new test cases after the last one as there will be a panic. Start with
```go
type convertToValueTestCase struct {
  input any
  expected cty.Value
}
```

mgyucht (commented):
Are we adding the validation to `resource_grant` as well, while we're at it?

qa/testing.go
tanmay-db: Yes if we change our tests to conform to the method. There were 2 options:
- Have a modified version that supports other types namely the one we use in our tests: `type data map[string]interface{}`. 
- Use the same version that is used in production (which the library calls) 

In option 1, we could make it more flexible by not calling panic and instead throw the error (since we are going to use this for unit tests only) but I think it would be good to have the same behaviour that is run in production.

catalog/resource_grants.go
tanmay-db: thank you

catalog/resource_grants.go
tanmay-db: In the case of "unknown", "unknown" we fail at plan which is expected right?

catalog/resource_grants.go
nkvuong: the way we define the schema of the resource means tf will fail if the user does not specify one of the securables

nkvuong:
> Are we adding the validation to `resource_grant` as well, while we're at it?
@mgyucht 
having this client side validation has created a lot of issues though, and we keep having to add them manually - #2529, #2399, #1998, etc.

internal/acceptance/grant_test.go
```diff
@@ -99,8 +94,36 @@ resource "databricks_grant" "some" {
 func TestUcAccGrant(t *testing.T) {
 	unityWorkspaceLevel(t, step{
 		Template: strings.ReplaceAll(grantTemplate, "%s", "{env.TEST_DATA_ENG_GROUP}"),
-	},
```
tanmay-db, line 102: Fixed indentification at other places that I noticed.

mgyucht:
@nkvuong that's fair. Also, it unblocks private preview customers who want to use new TF permissions.

tanmay-db:
closing in favour of: https://github.com/databricks/terraform-provider-databricks/pull/3290

PR #3164, opened 2024-01-26
Added support for case insensitivity for id fields in UC resources

tanmay-db:
## Changes

API responses return names in small cases, because of this the names which are used as `id` as stored in small case in the state. This leads to the following issue: 
1. name = "Abc"
2. terraform apply successful
3. terraform plan shows drift from abc -> Abc

Since we only need the small case names, we will suppress the diff on terraform side during terraform plan phase.

Currently, I have used securables list that we use in databricks_grants as reference for which we should have case sensitivity. There are probably some more resources for which we should do this or some resources for which we shouldn't (according to past decisions) -- started a thread for discussion. Will update once there is a conclusion. 

## Tests

Unit test
- [ ] `make test` run locally
- [ ] using Go SDK


tanmay-db:
This is WIP -- not marking it as draft since I would like to see unit test if any failures on CI. Will add to review once tests are added.

Will add this for other resources as well.

common/util.go
```diff
@@ -34,3 +34,7 @@ func SuppressDiffWhitespaceChange(k, old, new string, d *schema.ResourceData) bo
 	log.Printf("[DEBUG] Suppressing diff for %v: old=%#v new=%#v", k, old, new)
 	return strings.TrimSpace(old) == strings.TrimSpace(new)
 }
+
+func SupressCaseSensitivity(k, old, new string, d *schema.ResourceData) bool {
```
alexott: We already have similar function: https://github.com/databricks/terraform-provider-databricks/blob/main/common/resource.go#L427, so we need to unify them...

alexott (changes requested):

catalog/resource_schema.go
```diff
@@ -30,6 +30,7 @@ func ResourceSchema() common.Resource {
 				Default:  false,
 			}
 			m["storage_root"].DiffSuppressFunc = ucDirectoryPathSlashOnlySuppressDiff
+			m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 33: should we add it for `catalog_name` as well?

catalog/resource_share.go
```diff
@@ -175,6 +175,7 @@ func (beforeSi ShareInfo) Diff(afterSi ShareInfo) []ShareDataChange {
 
 func ResourceShare() common.Resource {
 	shareSchema := common.StructToSchema(ShareInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 178: Should we have it for names of objects in the share as well?

catalog/resource_table.go
```diff
@@ -65,7 +65,10 @@ func (a TablesAPI) deleteTable(name string) error {
 
 func ResourceTable() common.Resource {
 	tableSchema := common.StructToSchema(TableInfo{},
-		common.NoCustomize)
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 69: same here - what about `catalog_name` and `schema_name` ?

catalog/resource_volume.go
```diff
@@ -46,6 +46,7 @@ func ResourceVolume() common.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			}
+			m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 49: same as for table

pipelines/resource_pipeline.go
```diff
@@ -346,7 +346,7 @@ func adjustPipelineResourceSchema(m map[string]*schema.Schema) map[string]*schem
 	m["storage"].DiffSuppressFunc = suppressStorageDiff
 	m["storage"].ConflictsWith = []string{"catalog"}
 	m["catalog"].ConflictsWith = []string{"storage"}
-
+	m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 349: Hmm, was it checked with the DLT team?

serving/resource_model_serving.go
```diff
@@ -17,6 +17,7 @@ func ResourceModelServing() common.Resource {
 		serving.CreateServingEndpoint{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			m["name"].ForceNew = true
+			m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 20: Not sure about this as well

catalog/resource_table.go
alexott: What about `databricks_sql_table` ?

alexott (changes requested):

catalog/resource_share.go
tanmay-db: @alexott, are the share object names case insensitive? I didn't find the information here: https://docs.databricks.com/en/data-sharing/create-share.html

serving/resource_model_serving.go
tanmay-db: I have asked with the team.

pipelines/resource_pipeline.go
tanmay-db: I have asked with the team for confirmation.

tanmay-db: removed it for the time being

serving/resource_model_serving.go
tanmay-db: removed it for the time being.

catalog/resource_sql_table.go
```diff
@@ -439,18 +439,14 @@ func columnChangesCustomizeDiff(d *schema.ResourceDiff) error {
 func ResourceSqlTable() common.Resource {
 	tableSchema := common.StructToSchema(SqlTableInfo{},
 		func(s map[string]*schema.Schema) map[string]*schema.Schema {
-			s["data_source_format"].DiffSuppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
-				if new == "" {
```
tanmay-db, line 443: Hi @alexott, do you know why we return true if new is empty? 
`common.EqualFoldDiffSuppress` would return false if it's empty since it doesn't match with old string which should be the expected behaviour.

catalog/resource_share.go
alexott: I don't remember, but they could be if we refer to specific objects - tables, volumes and models.
@nkvuong do you remember?

catalog/resource_sql_table.go
alexott: it's optional field, so we may not specify it for managed tables, but backend will return DELTA for it (as I remember)

alexott: I suggest that we leave it for `data_source_format`

catalog/resource_volume.go
```diff
@@ -41,11 +41,16 @@ func getNameFromId(id string) (string, error) {
...
+			caseInsensitiveFields := []string{"name", "catalog_name", "schema_name"}
+			for _, field := range caseInsensitiveFields {
+				m[field].DiffSuppressFunc = common.EqualFoldDiffSuppress
+			}
 			m["storage_location"].DiffSuppressFunc = ucDirectoryPathSlashAndEmptySuppressDiff
 			m["volume_path"] = &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
 			}
+			m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
```
alexott, line 53: We don't need this line if we're doing iteration above

alexott (changes requested):
overall good, just small changes are required in the current implementation. But if you rebase to the latest `main` then we'll need to add such checks for following resources:

* `name` in registered model
* `name` and `source_table_full_name` for online tables
* `name` and `delta_sync_index_spec.source_table` for vector search index

tanmay-db:
@alexott regarding:
> name in registered model
also added this for catalog and schema.

catalog/resource_online_table.go
```diff
@@ -55,6 +55,8 @@ func waitForOnlineTableDeletion(w *databricks.WorkspaceClient, ctx context.Conte
 func ResourceOnlineTable() common.Resource {
 	s := common.StructToSchema(catalog.OnlineTable{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
+			common.CustomizeSchemaPath(m, "name", "source_table_full_name").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
```
alexott, line 59: it should be `spec` instead of `name`

alexott (changes requested):

catalog/resource_online_table.go
tanmay-db: Ahh, mistyped in hurry.

alexott (approved):

PR #3167, opened 2024-01-26
Exporter: performance improvements for big workspaces

alexott:
## Changes

Performance improvements for the exporter:

* Parallel generation of resources: for each identified resource we're generating its body separately from other resources using the `EXPORTER_RESOURCE_GENERATORS` (default is 50) goroutines - this helps with processing references to other resources.  Generated bodies are sent to dedicated channels that are responsible for writing the code into files.
* `stateApproximation` is reimplemented to avoid linear search, and also locking the whole structure.  We'll get a per-resource list of resource approximations together with the direct lookup structure that is used in the `Has` call.
* `importContext.Find` is optimized to perform the direct lookup for most of the match types.  Only prefix match and some cases of case-insensitive match will lead to iteration over approximations of the given type. 
* Dedicated channels are now created only for specific resources (SCIM-related right now) - could be overridden with the `EXPORTER_DEDICATED_RESOUSE_CHANNELS` environment variable.   The rest is handled by the shared channel, so we'll have better resource utilization.  The shared channel by default is handled by 15 goroutines, but this could be adjusted with the `EXPORTER_PARALLELISM_default` environment variable.
* Added the `-trace` command-line flag to enable trace logging level.
* For notebooks, we're not waiting for the listing to finish, but instead, emit from the visitor function (that is also done indirectly to avoid blocking on the SCIM API)

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


workspace/resource_notebook.go
```diff
@@ -171,7 +170,7 @@ const (
 )
 
 func (a NotebooksAPI) recursiveAddPathsParallel(directory directoryInfo, dirChannel chan directoryInfo,
-	answer *syncAnswer, wg *sync.WaitGroup, shouldIncludeDir func(ObjectStatus) bool) {
+	answer *syncAnswer, wg *sync.WaitGroup, shouldIncludeDir func(ObjectStatus) bool, visitor func([]ObjectStatus)) {
```
mgyucht, line 173: I think this API and `ListParallel` are only used in the exporter. Given that, maybe it makes sense to move them the `exporter` package? That way you can more or less freely make changes to them as needed for exporter, and I don't need to worry about them breaking components in the main provider.

mgyucht (approved):
Overall LGTM, but one small suggestion to refactor some methods to the `exporter` package to decrease friction when changing them.

workspace/resource_notebook.go
alexott: yep, I thought about it. But need to look how to make it correctly

alexott: I didn it already

mgyucht (approved):
Great, thanks!

Issue #3148, opened 2024-01-24
[FEATURE] Add a `databricks_sql_table` data source


jdavidheiser:
### Use-cases
We want to reference metadata from our Unity catalog tables in other Terraform configs - for example, we'd like to read the S3 paths where managed tables are stored, and the properties set on those tables.

### Attempted Solutions
We tried managing table creation in Terraform, so we could read the path from the resource, but this does not cover use cases where tables are dynamically generated outside of Terraform.

### Proposal
create a `databricks_sql_table` data source whose outputs mirror those in the existing `databricks_sql_table` resource.

nkvuong:
Closed by #3571

PR #3170, opened 2024-01-26
Add a `databricks_table` data source

jdavidheiser:
## Changes
This adds a new data source to describe a Databricks table, to address the request in https://github.com/databricks/terraform-provider-databricks/issues/3148. I am not particularly familiar with the Databricks Terraform Provider repo, and not sure which styles are best to mimic. I opted to re-use the struct from the Databricks SDK, rather than implementing a new one. I don't see this pattern a lot in the repo, but it seems like recent changes to the SQL data source may have taken this approach.

## Tests
I wanted to see if this approach of using the SDK structs is acceptable before writing tests. Go is not a language I normally work in, so it might take a bit of doing for me to figure out that part.

I did manually test it against a real Databricks workspace and confirm the return data shows up in my `terraform.tfstate`

- [ ] `make test` run locally
- [ ] using Go SDK


jdavidheiser:
I fixed a formatting error and added some basic unit testing. I am not really set up for running integration tests yet. Setting that up seems like a lot of overhead to commit a change that just passes an SDK result through unmodified, so if someone already set up for SDK tests could help me out that would be appreciated.

alexott:
Please add at least a unit test...

catalog/data_table.go
```diff
@@ -0,0 +1,24 @@
...
+	FullName string `json:"full_name" tf:"computed,optional`
+}
+
+func DataSourceTable() *schema.Resource {
+	return common.WorkspaceDataWithParams(func(ctx context.Context, data tableParams, w *databricks.WorkspaceClient) (*catalog.TableInfo, error) {
+		table_info, err := w.Tables.Get(ctx, catalog.GetTableRequest{FullName: data.FullName})
+		if err != nil {
+			return nil, err
+		}
+		return table_info, nil
```
alexott, line 22: you can simplify this to the

```
return w.Tables.Get(ctx, catalog.GetTableRequest{FullName: data.FullName})
```

jdavidheiser:
Sorry, I did and forgot to commit it.. Should be there now.

catalog/data_table.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type tableParams struct {
+	FullName string `json:"full_name" tf:"computed,optional`
```
alexott, line 13: Maybe it's better to have two params: catalog & schema?
We use this approach in other data sources: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/data_views.go#L14

alexott (changes requested):
Please add a

catalog/data_table.go
jdavidheiser: The Databricks API takes a [single parameter for the fully qualified table name](https://docs.databricks.com/api/workspace/tables/get). The views and tables data sources follow the interface of the API exactly already (there, the API takes two parameters), so it makes sense to me to do the same for the table call.

alexott: ah, Ok - I didn't check the API spec, sorry :-)

catalog/data_table.go
jdavidheiser: done, thanks

alexott:
Please rebase to the latest `main` branch to fix the conflict. The changes should be like this: https://github.com/databricks/terraform-provider-databricks/pull/3191/commits/e221fe7a142a2806bf702ff8a211d3caf325ec59

jdavidheiser:
conflict should be fixed and switched to `common.resource`

alexott:
you can always run `make fmt && make lint` before committing...

alexott:
Oh, I missed that documentation isn't a part of PR - please add the documentation as `doc/data-sources/table.md` file.

jdavidheiser:
I added a doc file - mostly copy/pasted from the Databricks SDK for Go docs.

alexott:
@jdavidheiser can you please resolve conflicts in the provider?

jdavidheiser:
fixed

docs/data-sources/table.md
```diff
@@ -0,0 +1,78 @@
...
+* `data_access_configuration_id` - Unique ID of the Data Access Configuration to use with the table data
+* `data_source_format` - Data source format
+* `deleted_at` - Time at which this table was deleted, in epoch milliseconds. Field is omitted if table is not deleted
+* `delta_runtime_properties_kvpairs` - Information pertaining to current state of the delta table
+* `effective_predictive_optimization_flag`
+* `enable_predictive_optimization` - Whether predictive optimization should be enabled for this object and objects under it
+* `encryption_details` - Encryption options that apply to clients connecting to cloud storage
+
+__catalog_name__.__schema_name__.__table_name__
+* `full_name` - Full name of table, in form of *`catalog`.`schema`.`table`*
```
nkvuong, line 53:
```suggestion
```

nkvuong (commented):
thank you for this PR. a couple more changes suggested

PR #3178, opened 2024-01-30
New `tf` tags for annotating resource schemas with `ExactlyOneOf`, `AtLeastOneOf`, `ConflictsWith`, and `RequiredWith`

karolusz:
## Changes
I am proposing to add additional supported `tf` tags values to enable annotating Terraform schema with `ExactlyOneOf`, `AtLeastOneOf`, `ConflictsWith`, `RequiredWith`. These would simplify schema validation and/or schema definition for a several resources and data sources by offloading the validation to Terraform SDK (https://github.com/hashicorp/terraform-plugin-sdk/blob/b7c8ee78f0ccdcdaa2a3fe4b26229203c7f5c3a3/helper/schema/schema.go)

Each of these attributes expects a slice of strings (`[]string`) representing attribute paths. In the `tf` tag, in the schema struct definition of resources/data, it is is proposed to separate attributes paths with `;` as a comma is already being used to separate multiple tags.

For example schema from `resource_entitlement.go` could be redefined from:
```go
	type entity struct {
		GroupId string `json:"group_id,omitempty" tf:"force_new"`
		UserId  string `json:"user_id,omitempty" tf:"force_new"`
		SpnId   string `json:"service_principal_id,omitempty" tf:"force_new"`
	}
	entitlementSchema := common.StructToSchema(entity{},
			addEntitlementsToSchema(&m)
			alof := []string{"group_id", "user_id", "service_principal_id"}
			for _, field := range alof {
				m[field].AtLeastOneOf = alof
			}
			return m
		})
```
to:
```go
	type entity struct {
		GroupId string `json:"group_id,omitempty" tf:"force_new,at_least_one_of:user_id;service_principal_id"`
		UserId  string `json:"user_id,omitempty" tf:"force_new,at_least_one_of:group_id;service_principal_id"`
		SpnId   string `json:"service_principal_id,omitempty" tf:"force_new,at_least_one_of:user_id;service_principal_id"`
	}
	entitlementSchema := common.StructToSchema(entity{},
			addEntitlementsToSchema(&m)
			return m
		})
```
## Tests
Unit tests added.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/reflect_resource.go
+++ b/common/reflect_resource.go
@@ -218,6 +218,46 @@ func handleSuppressDiff(typeField reflect.StructField, v *schema.Schema) {
 	}
 }
 
+func handleExactlyOneOf(typeField reflect.StructField, schema *schema.Schema) {
+	tfTags := strings.Split(typeField.Tag.Get("tf"), ",")
+	prefix := "exactly_one_of:"
+	for _, tag := range tfTags {
+		if strings.HasPrefix(tag, prefix) {
+			schema.ExactlyOneOf = strings.Split(strings.TrimPrefix(tag, prefix), ";")
+		}
+	}
+}
+
+func handleAtLeastOneOf(typeField reflect.StructField, schema *schema.Schema) {
+	tfTags := strings.Split(typeField.Tag.Get("tf"), ",")
+	prefix := "at_least_one_of:"
+	for _, tag := range tfTags {
+		if strings.HasPrefix(tag, prefix) {
+			schema.AtLeastOneOf = strings.Split(strings.TrimPrefix(tag, prefix), ";")
+		}
+	}
+}
+
+func handleConflictsWith(typeField reflect.StructField, schema *schema.Schema) {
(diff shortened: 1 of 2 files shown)


alexott:
This should be discussed - we're switching to Go SDK and such annotations doesn't exist there, so schema tuning needs to be done separately. See latest PR about migrating cluster resource to the Go SDK

karolusz:
> This should be discussed - we're switching to Go SDK and such annotations doesn't exist there, so schema tuning needs to be done separately. See latest PR about migrating cluster resource to the Go SDK

I see!  I think it was merged while I was opening this one :)
It seems that `customizable_schema.go` handles only `ConflictsWith` .  As the Go SDK is the preferrable approach, I will abandon this PR and extend `customizable_schema.go` with `RequiredWith`, `AtLeastOneOf` and `ExactlyOneOf`.

alexott:
Yep, thank you!

Cc: @mgyucht @tanmay-db

Issue #3018, opened 2023-12-11
[ISSUE] Issue with Unity Catalog resources that are on Workspace API


zshao9:
I have a list of Unity Catalogs (one for each region) whose catalog/schemas/grants need to be created with Terraform.

Since those resources need to go to the Workspace-API providers, I need a list of providers one for each workspace (associated with the Unity Catalog).

However, Terraform doesn't support a dynamic list of providers (for for_each).  See:
* 2019 issue: https://github.com/hashicorp/terraform/issues/19932
* 2020 issue https://github.com/hashicorp/terraform/issues/24476
* 2022 discussions: https://stackoverflow.com/questions/72541598/using-for-each-with-provider-setting

Can we consider to use "metastore_id" properly on the catalog/schema/grants resources?
Then I just need to use a single provider with whatever workspace for all the different metastores.

alexott:
No, it's not possible because such functionality isn't available in the UC APIs.

mgyucht:
Just to expand on this a bit: workspace-level UC APIs only allow interacting with the metastore assigned to that workspace. Even if the TF provider were to accept a metastore ID, it wouldn't be able to interact with that metastore without a running workspace that is assigned to said metastore. We're working with the UC team to align on exposing these APIs at the account-level, in which case you would be able to interact with them purely when configured at the account level.

bobdoah:
This would be incredibly useful

sreedhanyagn:
We have more than 100 workspaces our organization and defining provider for each workspace is not feasible. Is there any alternative solution for this ? 
Can we use a single workspace as provider and create Catalogs for other workspaces using this provider ?

zshao9:
As long as they share the same unity catalog (e.g. in the same region) then
this is possible.

Curious what is the use case of 100+ workspaces?  One per user?

sreedhanyagn:
Hey @zshao9 , thanks for your reply. 
We are supporting many teams and each has team has multiple workspaces in different environments.

zshao9:
Got it.  Are they all in the same region?  If so, then they use the same
unity catalog.  You can just set up an "admin workspace" where all these
Terraform operations are done.

sreedhanyagn:
Yes, they all are in the same region.

Issue #2610, opened 2023-08-24
[ISSUE] Account-level auth is broken: provider is not picking up account-level OAuth token from Databricks CLI


nfx:
It's not possible to fully automate a fresh Databricks Account in one apply using Databricks CLI authentication.

### Expected Behavior

After running `databricks auth login --host https://accounts.cloud.databricks.com --account-id XXXX`, I expect the workspace provider to pick up the token it used to authenticate to account console.

```
provider "databricks" {
  alias      = "account"
  host       = module.defaults.aws_account_console
  account_id = module.defaults.aws_databricks_account_id
}

...

resource "databricks_mws_workspaces" "this" {
  provider       = databricks.account
  account_id     = module.defaults.aws_databricks_account_id
  aws_region     = module.defaults.aws_region
  workspace_name = "${local.prefix}-ucws"

  credentials_id           = databricks_mws_credentials.this.credentials_id
  storage_configuration_id = databricks_mws_storage_configurations.this.storage_configuration_id
}

...

provider "databricks" {
  alias = "workspace"
  host  = databricks_mws_workspaces.this.workspace_url
}

```

### Actual Behavior

I received `cannot configure default credentials. Config: host=<host from databricks_mws_workspaces.this.workspace_url>`.

## Attempt 2

SPN auth doesn't work as well:

```
resource "databricks_service_principal" "owner" {
  provider     = databricks.account
  display_name = "owner-sp"
}

resource "databricks_service_principal_role" "account_admin" {
  provider             = databricks.account
  service_principal_id = databricks_service_principal.owner.id
  role                 = "account_admin"
}

resource "databricks_service_principal_secret" "ower" {
  provider             = databricks.account
  service_principal_id = databricks_service_principal.owner.id
}

...

provider "databricks" {
  alias         = "workspace"
  host          = databricks_mws_workspaces.this.workspace_url
  client_id     = databricks_service_principal.owner.application_id
  client_secret = databricks_service_principal_secret.ower.secret
}
```

results in `│ Error: inner token: oauth2: "invalid_client" "Client authentication failed"`

## Attempt 3

Assigning workspace access permission doesn't work either: `│ Error: cannot create mws permission assignment: Permission assignment APIs are not available for this workspace.`, until the UC is enabled. and UC requires a workspace to execute APIs.

```
resource "databricks_mws_permission_assignment" "owner" {
  provider     = databricks.account
  principal_id = databricks_service_principal.owner.id
[...]

PR #3188, opened 2024-01-31
Support account-level provider with workspace-level resources

mgyucht:
## Changes
Partially addresses #2610, #3018.

This change will add an optional `workspace_id` field to all workspace-level resources. When using an account-level provider, if the workspace_id is specified, a workspace client will be constructed reusing that account-level provider's configuration. This means that users of the TF provider will only need to define a single `provider` block per account and will no longer need to create separate provider blocks per resource. The main downside is that the workspace ID needs to be specified for every single resource. This is slightly worse than today, where users need to specify `depends_on` for all workspace resources that don't have any dependencies.

For example:
```
data "databricks_spark_version" "latest" {
	workspace_id = <WSID>
}
	workspace_id = <WSID>
	cluster_name = "singlenode-{var.RANDOM}"
	spark_version = data.databricks_spark_version.latest.id
	instance_pool_id = "<ID>"
	num_workers = 0
	autotermination_minutes = 10
	spark_conf = {
		"spark.databricks.cluster.profile" = "singleNode"
		"spark.master" = "local[*]"
	}
	custom_tags = {
		"ResourceClass" = "SingleNode"
	}
}
```

will work with an account-level provider, given that that workspace with ID `<WSID>` belongs to the account.

### Supported resources

All workspace-level resources that do not have a workspace_id field will be supported for this customization. The only resources with a workspace_id field are databricks_catalog_workspace_binding and databricks_metastore_assignment. The latter is already supported at the account-level, so you would be able to switch to using an account-level provider for this resource by importing it into your provider configuration.

### Migration

To migrate from the current provider to this mechanism:

1. Add the `workspace_id` field to all resources managed at the workspace level.
2. Use the account-level provider instead of the workspace-level provider.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


common/resource.go
```diff
@@ -87,7 +87,10 @@ func (r Resource) ToResource() *schema.Resource {
 	if r.Update != nil {
 		update = func(ctx context.Context, d *schema.ResourceData,
 			m any) diag.Diagnostics {
-			c := m.(*DatabricksClient)
+			c, err := m.(*DatabricksClient).InConfiguredWorkspace(ctx, d)
```
nfx, line 90: it's not enough, you also need to prepend workspace_id to `d.Id()` on a read side and on a write side - e.g. `fmt.Sprintf("%d:%s", d.Get("workspace_id"), d.Id())`. otherwise there's too much risk involved.

clusters/data_spark_version.go
```diff
@@ -131,15 +131,20 @@ func DataSourceSparkVersion() *schema.Resource {
 
 		s["photon"].Deprecated = "Specify runtime_engine=\"PHOTON\" in the cluster configuration"
 		s["graviton"].Deprecated = "Not required anymore - it's automatically enabled on the Graviton-based node types"
+		common.AddWorkspaceIdField(s)
```
nfx, line 134: this won't scale, add it on the level of `common.Resource` and `common.DataSource`

nfx (changes requested):

clusters/data_spark_version.go
mgyucht: I was thinking about how to do this. We don't know which resources are workspace-only or also at the account level. We need some explicit mapping to make sure this isn't forgotten about. I will probably add this at the provider-level to make sure there is an explicit opt-in/out for each resource type.

common/resource.go
mgyucht: This field is marked as `ForceNew`, so changes to workspace ID would not trigger update, only delete/recreate. However, we do need to store the old ID somewhere in the state to be able to appropriately move from one workspace to another (as there is no way to get the "old" workspace ID during the delete call).

common/client.go
```diff
@@ -252,44 +388,6 @@ func (c *DatabricksClient) FormatURL(strs ...string) string {
 	return strings.Join(data, "")
 }
 
-// ClientForHost creates a new DatabricksClient instance with the same auth parameters,
```
mgyucht, line 255: This is replaced by `InWorkspace()`.

cfunkhouser:
This change would make (almost) all of the problems I'm facing using the Databricks provider go away. Is there any sort of timeline on which we might expect these changes?

vavison:
Is this still being worked on? It would solve several issues we're currently experiencing with the provider so it would be great to know if / when this is coming.

hectorcast-db:
This feature has been developed and released separately as part of the Unified Provider story. Closing this PR as duplicate.

PR #3189, opened 2024-01-31
[WIP] Fix updating owners

tanmay-db:
## Changes

We make additional update request even when there are no other updates other than owner. We return earlier before proceeding further. 

## Tests

ToDo: To add unit tests
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_catalog.go
+++ b/catalog/resource_catalog.go
@@ -148,6 +148,10 @@ func ResourceCatalog() *schema.Resource {
 				}
 			}
 
+			if !d.HasChangeExcept("owner") {
+				return nil
+			}
+
 			updateCatalogRequest.Owner = ""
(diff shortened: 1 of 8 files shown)


mgyucht:
Let's make sure this is integration tested before merging.

nkvuong (approved):
looks good to me

mgyucht (approved):
LGTM, let's make sure integration tests pass.

Can we update the docs to describe the behavior for UC resources? I.e. that when changing owner & other fields, we first update the owner, then other fields, and that other fields are updated using the new owner's permissions, so if your principal is not an owner, you can't change those fields

tanmay-db:
Integration tests passed and added doc.

PR #3193, opened 2024-01-31
Use common.Resource consistently throughout the provider

mgyucht:
## Changes
Purely a refactor of existing resources to consistently use `common.Resource` throughout the provider, instead of a mixture of `*schema.Resource` and `common.Resource`. This is needed for https://github.com/databricks/terraform-provider-databricks/pull/3188, in which I'll add a field in `common.Resource` indicating whether or not the `workspace_id` attribute needs to be added to the schema. 

## Tests
Passes integration tests (with one exception, which is being fixed in https://github.com/databricks/terraform-provider-databricks/pull/3200).

- [ ] `make test` run locally
- [ ] using Go SDK


common/reflect_resource.go
```diff
@@ -654,8 +654,8 @@ func DataToStructPointer(d *schema.ResourceData, scm map[string]*schema.Schema,
 }
 
 // DataToReflectValue reads reflect value from data
-func DataToReflectValue(d *schema.ResourceData, r *schema.Resource, rv reflect.Value) error {
-	return readReflectValueFromData([]string{}, d, rv, r.Schema)
+func DataToReflectValue(d *schema.ResourceData, s map[string]*schema.Schema, rv reflect.Value) error {
```
mgyucht, line 657: The only places where this is used construct a resource in-line purely to pass the schema, so I changed it to accept the schema instead.

common/reflect_resource_test.go
```diff
@@ -609,15 +609,12 @@ func TestDataResource(t *testing.T) {
...
 			}
 			return nil
-		})
+		}).ToResource()
 	}()
 	d := r.TestResourceData()
 	d.Set("in", "test")
 
-	diags := r.ReadContext(context.Background(), d, nil)
-	assert.Len(t, diags, 1)
```
mgyucht, line 618: This is an unnecessary test case, as the provider will ensure that the `meta` argument is always set to a non-nil `*DatabricksClient`.

common/resource.go
```diff
@@ -16,15 +16,17 @@ import (
...
+	Read               func(ctx context.Context, d *schema.ResourceData, c *DatabricksClient) error
+	Update             func(ctx context.Context, d *schema.ResourceData, c *DatabricksClient) error
+	Delete             func(ctx context.Context, d *schema.ResourceData, c *DatabricksClient) error
+	CustomizeDiff      func(ctx context.Context, d *schema.ResourceDiff) error
+	StateUpgraders     []schema.StateUpgrader
+	Schema             map[string]*schema.Schema
+	SchemaVersion      int
+	Timeouts           *schema.ResourceTimeout
+	DeprecationMessage string
+	Importer           *schema.ResourceImporter
```
mgyucht, line 29: These are the only fields from schema.Resource that need to be added to this Resource structure.

storage/mounts.go
```diff
@@ -161,6 +161,10 @@ func getCommonClusterObject(clustersAPI clusters.ClustersAPI, clusterName string
 func getOrCreateMountingCluster(clustersAPI clusters.ClustersAPI) (string, error) {
 	cluster, err := clustersAPI.GetOrCreateRunningCluster("terraform-mount", getCommonClusterObject(clustersAPI, "terraform-mount"))
 	if err != nil {
+		// Do not treat missing cluster like a missing resource.
```
mgyucht, line 164: Needed because `common.Resource` has a blanket catch-all for missing responses from APIs, but in this case, we need to prevent that from being triggered.

storage/mounts_test.go
```diff
@@ -333,21 +333,23 @@ func TestMountCRD(t *testing.T) {
 			Response:     apierr.NotFound("nope"),
 		},
 	}, func(ctx context.Context, client *common.DatabricksClient) {
-		r := ResourceMount()
+		r := ResourceMount().ToResource()
 		d := r.TestResourceData()
+		d.Set("name", "a name")
```
mgyucht, line 338: The actual resource definition adds a "preprocessing" step that checks whether the `name` and `uri` attributes are set.

nfx (approved):
lgtm

PR #3202, opened 2024-02-01
Exporter: Add retries for `Search`, `ReadContext` and `Import` operations when importing the resource

alexott:
## Changes

Under a very high load, Databricks backend may not answer on time, or return specific errors, so it makes sense to retry operation few times.

This PR uses "naive" implementation, I need to play a bit more with `retries` package before adopting it

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -758,7 +758,7 @@ func (ic *importContext) generateAndWriteResources(sh *os.File) {
 	for i, r := range resources {
 		ic.waitGroup.Add(1)
 		resourcesChan <- r
-		if i%50 == 0 {
+		if i%500 == 0 {
 			log.Printf("[INFO] Submitted %d of %d resources", i+1, scopeSize)
 		}
 	}
--- a/exporter/model.go
+++ b/exporter/model.go
@@ -245,8 +245,8 @@ func (r *resource) ImportCommand(ic *importContext) string {
 }
 
 var (
-	maxRetries = 5
-	retryDelay = 2
+	maxRetries        = 5
+	retryDelaySeconds = 2
 )
 
 func isRetryableError(err string, i int) bool {
@@ -285,7 +285,7 @@ func (r *resource) ImportResource(ic *importContext) {
 				return
 			}
 			log.Printf("[INFO] next retry (%d) for searching of %v", (i + 1), r)
-			time.Sleep(time.Duration(retryDelay) * time.Second)
+			time.Sleep(time.Duration(retryDelaySeconds) * time.Second)
 		}
 		if r.ID == "" {
 			log.Printf("[WARN] Cannot find %s", r)
@@ -317,7 +317,7 @@ func (r *resource) ImportResource(ic *importContext) {
 				return
 			}
 			log.Printf("[INFO] next retry (%d) for reading of %s#%s", (i + 1), r.Resource, r.ID)
-			time.Sleep(time.Duration(retryDelay) * time.Second)
+			time.Sleep(time.Duration(retryDelaySeconds) * time.Second)
 		}
 		if r.Data.Id() == "" {
 			r.Data.SetId(r.ID)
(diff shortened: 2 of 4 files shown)


alexott:
@mgyucht @tanmay-db It would be really useful to get this merged before the release...

tanmay-db:
looking

exporter/model.go
```diff
@@ -243,6 +244,15 @@ func (r *resource) ImportCommand(ic *importContext) string {
 	return fmt.Sprintf(`terraform import %s%s.%s "%s"`, m, r.Resource, r.Name, r.ID)
 }
 
+var (
+	maxRetries = 5
+	retryDelay = 2
```
tanmay-db, line 249: is this in seconds? 
nit: A small comment regarding time unit would be helpful for readability.

exporter/model.go
```diff
@@ -243,6 +244,15 @@ func (r *resource) ImportCommand(ic *importContext) string {
...
+)
+
+func isRetryableError(err string, i int) bool {
```
tanmay-db, line 252: we could have a list of errors on which we would like to retry. Then we can check the list. It would be easier to extend for future use cases otherwise there would be many `||` statements

exporter/model.go
```diff
@@ -288,19 +306,36 @@ func (r *resource) ImportResource(ic *importContext) {
 		if apiVersion != "" {
 			ctx = context.WithValue(ctx, common.Api, apiVersion)
 		}
-		if dia := pr.ReadContext(ctx, r.Data, ic.Client); dia != nil {
-			log.Printf("[ERROR] Error reading %s#%s: %v", r.Resource, r.ID, dia)
-			return
+		// TODO: rewrite to retries package...
```
tanmay-db, line 309: Is this something we plan to do in future? Just curious if retries package would require refactoring to support this.

exporter/model.go
alexott: In this case we really need a small number of error retried - the rest should be handled by Go SDK

exporter/model.go
alexott: Miles recommends it, but I need to play with it more

exporter/model.go
tanmay-db: Okay sounds good.

alexott: I've reimplemented it as I found more error types that we need to handle 🤦

exporter/context.go
```diff
@@ -758,7 +758,7 @@ func (ic *importContext) generateAndWriteResources(sh *os.File) {
 	for i, r := range resources {
 		ic.waitGroup.Add(1)
 		resourcesChan <- r
-		if i%50 == 0 {
+		if i%500 == 0 {
```
tanmay-db, line 761: I am guessing this is because there are too many items in logs so we want to minimise them?

exporter/util.go
```diff
@@ -1166,3 +1168,40 @@ func ListParallel(a workspace.NotebooksAPI, path string, shouldIncludeDir func(w
...
+		for _, msg := range retriableErrors {
+			if strings.Contains(err, msg) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func runWithRetries[ERR any](runFunc func() ERR, msg string) ERR {
```
tanmay-db, line 1189: Rest looks good to me, only concern is that we can use https://github.com/databricks/databricks-sdk-go/blob/main/retries/retries.go  here otherwise, we would have to maintain this retry mechanism separately.

If the change required is urgent and using retries would require major refactoring, then we can go ahead with this. What do you think? @mgyucht

exporter/context.go
alexott: yes, it's just too much

exporter/util.go
mgyucht: `retries` package is a convenience, not a core part of the SDK. The retries package does not expose anything to e.g. print when a retry is taking place or print a specific message after including the total number of retries that have happened, so if @alexott needs these, it's perfectly fine for him to implement his own retry logic.

mgyucht: Though @alexott curious if we added things like `WithMaxRetries` or `WithRetryOnErrorSubstrings` if that would satisfy your need.

mgyucht (approved):
No problem

exporter/util.go
alexott: yep, let discuss if we can make something generic...

PR #3207, opened 2024-02-02
Make IterFields take in aliases

edwardfeng-db:
## Changes

- Make aliases information from `ResourceProviderStruct` take effect in `iterFields` function so that `StructToData` and `DataToStructPointer` will take information about the aliases
- Taking out `UnderlyingType` method because it is no longer needed
- Adding an anonymous field for the `ClusterResourceProvider` struct so that reflection will work without `UnderlyingType`
- Make `iterField` check alias information
- Adding unit test case

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -169,7 +169,7 @@ func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[
 }
 
 func resourceClusterSchema() map[string]*schema.Schema {
-	return common.ResourceProviderStructToSchema(ClusterResourceProvider{})
+	return common.StructToSchema(ClusterResourceProvider{}, nil)
 }
 
 func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -58,15 +58,15 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	return false
 }
 
-type ClusterResourceProvider struct {
+type ClusterSpec struct {
 	compute.ClusterSpec
 }
 
-func (ClusterResourceProvider) Aliases() map[string]string {
+func (ClusterSpec) Aliases() map[string]string {
 	return map[string]string{"cluster_mount_infos": "cluster_mount_info"}
 }
 
-func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
+func (ClusterSpec) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
 	common.CustomizeSchemaPath(s, "cluster_source").SetReadOnly()
 	common.CustomizeSchemaPath(s, "enable_elastic_disk").SetComputed()
 	common.CustomizeSchemaPath(s, "enable_local_disk_encryption").SetComputed()
@@ -169,7 +169,7 @@ func (ClusterResourceProvider) CustomizeSchema(s map[string]*schema.Schema) map[
 }
 
 func resourceClusterSchema() map[string]*schema.Schema {
-	return common.StructToSchema(ClusterResourceProvider{}, nil)
+	return common.StructToSchema(ClusterSpec{}, nil)
 }
 
 func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -58,7 +58,9 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	return false
 }
 
-type ClusterResourceProvider struct{}
+type ClusterResourceProvider struct {
+	compute.ClusterSpec
+}
 
 func (ClusterResourceProvider) UnderlyingType() compute.ClusterSpec {
 	return compute.ClusterSpec{}
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -62,10 +62,6 @@ type ClusterResourceProvider struct {
 	compute.ClusterSpec
 }
 
-func (ClusterResourceProvider) UnderlyingType() compute.ClusterSpec {
-	return compute.ClusterSpec{}
-}
-
 func (ClusterResourceProvider) Aliases() map[string]string {
 	return map[string]string{"cluster_mount_infos": "cluster_mount_info"}
 }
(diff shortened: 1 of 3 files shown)


mgyucht (commented):
Making good progress. Let's see if there is a way we can continue to use StructToSchema as the main entrypoint so that developers can easily switch to using this custom implementation without having to rewrite their calls to StructToSchema. This makes library code like WorkspaceData/AccountData more portable to the new model (and would immediately solve problems like https://github.com/databricks/terraform-provider-databricks/pull/3211 is trying to solve)

common/reflect_resource.go
```diff
@@ -40,16 +40,14 @@ var kindMap = map[reflect.Kind]string{
 
 // Generic interface for resource provider struct. Using CustomizeSchema and Aliases functions to keep track of additional information
 // on top of the generated go-sdk struct. This is used to replace manually maintained structs with `tf` tags.
-type ResourceProviderStruct[T any] interface {
-	UnderlyingType() T
+type ResourceProviderStruct interface {
```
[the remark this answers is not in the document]
edwardfeng-db: Yeah good point, we should update that

common/reflect_resource_test.go
```diff
@@ -327,23 +365,102 @@ func TestIterFields(t *testing.T) {
...
 	assert.Equal(t, []any{"a", "b"}, v)
 
 	_, err = collectionToMaps([]int{1, 2}, &schema.Schema{
 		Elem: schema.TypeBool,
-	})
+	}, map[string]string{})
 	assert.EqualError(t, err, "not resource")
 }
 
+func TestStructToDataWithResourceProviderStruct(t *testing.T) {
```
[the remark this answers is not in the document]
edwardfeng-db: Done

common/reflect_resource_test.go
```diff
@@ -327,23 +365,102 @@ func TestIterFields(t *testing.T) {
...
+	assert.Equal(t, false, d.Get("enabled_alias"))
+	assert.Equal(t, 2, d.Get("addresses.#"))
+
+	assert.NotNil(t, s["home"].DiffSuppressFunc)
+	assert.True(t, s["home"].DiffSuppressFunc("home.#", "1", "0", d))
+	assert.False(t, s["home"].DiffSuppressFunc("home.#", "1", "1", d))
+
+	{
+		//lint:ignore SA1019 Empty optional string should not be set.
+		_, ok := d.GetOkExists("addresses.0.optional_string")
```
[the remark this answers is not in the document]
edwardfeng-db: I tried that but it doesn't seem to work, also tried replacing the existing test case on line 542, the test will fail with `GetOk`

common/reflect_resource.go
```diff
@@ -633,36 +661,51 @@ func DiffToStructPointer(d attributeGetter, scm map[string]*schema.Schema, resul
...
+}
+
+func getAliasesMapFromStruct(s any) map[string]string {
+	var aliases map[string]string
+	if v, ok := s.(ResourceProvider); ok {
+		aliases = v.Aliases()
+	} else {
+		aliases = map[string]string{}
+	}
+	return aliases
```
mgyucht, line 699: Cleaner.
```suggestion
	if v, ok := s.(ResourceProvider); ok {
		return v.Aliases()
	}
	return map[string]string{}
```

Also, let's document that this doesn't return aliases defined on `tf` tags.

common/reflect_resource.go
```diff
@@ -463,7 +485,11 @@ func iterFields(rv reflect.Value, path []string, s map[string]*schema.Schema,
 	fields := listAllFields(rv)
 	for _, field := range fields {
 		typeField := field.sf
-		fieldName := chooseFieldName(typeField)
+		// Always pass in an empty list for path because we unwrap the aliases map on each call.
+		// We do not use the `path` variable in this function because it sometimes contains indices
+		// and sometimes it is empty. Thus we rely on the caller of iterFields to unwrap the keys
+		// of the aliases map.
+		fieldName := chooseFieldNameWithAliases(typeField, []string{}, aliases)
```
mgyucht, line 492: Why do we need both the path and aliases map here? With the alias unwrapping, we should need one or the other, not both. Wherever we recurse into a nested field, we can just unwrap the aliases one more time, so the path parameter should be unnecessary.

common/reflect_resource_test.go
```diff
@@ -301,15 +364,15 @@ func TestPrimitiveReflectValueFromInterface(t *testing.T) {
 
 func TestIterFields(t *testing.T) {
 	v := reflect.ValueOf("x")
-	err := iterFields(v, []string{"x"}, scm, nil)
+	err := iterFields(v, []string{"x"}, scm, map[string]string{}, nil)
```
mgyucht, line 367: Since we're only reading from this map, we can pass `nil` instead of `map[string]string{}` everywhere.

common/reflect_resource.go
```diff
@@ -117,6 +120,10 @@ func MustSchemaPath(s map[string]*schema.Schema, path ...string) *schema.Schema
 
 // StructToSchema makes schema from a struct type & applies customizations from callback given
 func StructToSchema(v any, customize func(map[string]*schema.Schema) map[string]*schema.Schema) map[string]*schema.Schema {
+	// If the input 'v' is an instance of ResourceProvider, call resourceProviderStructToSchema instead.
+	if rp, ok := v.(ResourceProvider); ok {
+		return resourceProviderStructToSchema(rp)
```
mgyucht, line 125: We should panic if customize is not nil in this case.

mgyucht (commented):
Couple more comments.

clusters/resource_cluster.go
```diff
@@ -58,10 +58,8 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	return false
 }
 
-type ClusterResourceProvider struct{}
-
-func (ClusterResourceProvider) UnderlyingType() compute.ClusterSpec {
-	return compute.ClusterSpec{}
+type ClusterResourceProvider struct {
```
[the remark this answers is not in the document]
edwardfeng-db: I think `Cluster` is already defined, so how about name it `ClusterSpec`?

common/reflect_resource.go
edwardfeng-db: Yeah you're right, I guess this is because originally I wanted to keep the same map and update the path, but it's not always possible that way with `iterFields`. Let me update this

clusters/resource_cluster.go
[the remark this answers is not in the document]
mgyucht: Either is better than `ClusterResourceProvider` :)

common/reflect_resource.go
```diff
@@ -122,10 +120,13 @@ func MustSchemaPath(s map[string]*schema.Schema, path ...string) *schema.Schema
 func StructToSchema(v any, customize func(map[string]*schema.Schema) map[string]*schema.Schema) map[string]*schema.Schema {
 	// If the input 'v' is an instance of ResourceProvider, call resourceProviderStructToSchema instead.
 	if rp, ok := v.(ResourceProvider); ok {
+		if customize != nil {
+			panic("Customize should be nil if the input struct is an instance of ResourceProvider, use CustomizeSchema of ResourceProvider instead.")
```
mgyucht, line 124:
```suggestion
			panic("customize should be nil if the input implements the ResourceProvider interface; use CustomizeSchema of ResourceProvider instead")
```

mgyucht (approved):
LGTM!

Issue #3195, opened 2024-02-01
Exporter: Investigate effect of using `/Workspace` path everywhere


alexott:
We need to investigate how [this](https://learn.microsoft.com/en-us/azure/databricks/release-notes/product/2024/january#workspace-path-update) will affect the exporter:

> Historically, users were required to include the `/Workspace` path prefix for some Databricks APIs (`%sh`) but not for others (`%run`, REST API inputs). Now, users can provide workspace paths with the `/Workspace` prefix everywhere in Databricks. Old references to paths without the `/Workspace` prefix are redirected and will continue to work. Databricks recommends that all workspace paths use the `/Workspace` prefix to differentiate them from Volume and DBFS paths.

PR #3208, opened 2024-02-02
Add `source` attribute to `dbt_task` and `sql_task.file` tasks to support files from workspace

alexott:
## Changes

Also added support for it in TF exporter

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -323,7 +323,7 @@ You can invoke Spark submit tasks only on new clusters. **In the `new_cluster` s
 ### dbt_task Configuration Block
 
 * `commands` - (Required) (Array) Series of dbt commands to execute in sequence. Every command must start with "dbt".
-* `source` - (Optional) What is the source of this project. Possible values are `WORKSPACE` and `GIT`.  Default value depends on if `git_source` block presents in the job definition or not - if it's present, then the default value is `GIT`.
+* `source` - (Optional) The source of the project. Possible values are `WORKSPACE` and `GIT`.  Defaults to `GIT` if a `git_source` block is present in the job definition.
 * `project_directory` - (Optional  if `source` is `GIT`, Required when `source` is `WORKSPACE`) If `source` is `GIT`, then it should be a relative path to the directory in the repository specified in `git_source` where dbt should look in for the `dbt_project.yml` file. If not specified, defaults to the repository's root directory. Equivalent to passing `--project-dir` to a dbt command.  If `source` is `WORKSPACE`, then it should be absolute path to the folder in the workspace.
 * `profiles_directory` - (Optional) The relative path to the directory in the repository specified by `git_source` where dbt should look in for the `profiles.yml` file. If not specified, defaults to the repository's root directory. Equivalent to passing `--profile-dir` to a dbt command.
 * `catalog` - (Optional) The name of the catalog to use inside Unity Catalog.
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -367,7 +367,7 @@ One of the `query`, `dashboard` or `alert` needs to be provided.
   * `pause_subscriptions` - (Optional) flag that specifies if subscriptions are paused or not.
 * `file` - (Optional) block consisting of single string fields: 
   * `source` - (Optional) The source of the project. Possible values are `WORKSPACE` and `GIT`.
-  * `path` - If `source` is `GIT`, then it a relative path to the file (inside the Git repository) with SQL commands to execute.  If `source` is `WORKSPACE` then it should be absolute path to the file in the workspace.
+  * `path` - If `source` is `GIT`: Relative path to the file in the repository specified in the `git_source` block with SQL commands to execute. If `source` is `WORKSPACE`: Absolute path to the file in the workspace with SQL commands to execute.
 
 Example
 
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -323,7 +323,8 @@ You can invoke Spark submit tasks only on new clusters. **In the `new_cluster` s
 ### dbt_task Configuration Block
 
 * `commands` - (Required) (Array) Series of dbt commands to execute in sequence. Every command must start with "dbt".
-* `project_directory` - (Optional) The relative path to the directory in the repository specified in `git_source` where dbt should look in for the `dbt_project.yml` file. If not specified, defaults to the repository's root directory. Equivalent to passing `--project-dir` to a dbt command.
+* `source` - (Optional) What is the source of this project. Possible values are `WORKSPACE` and `GIT`.  Default value depends on if `git_source` block presents in the job definition or not - if it's present, then the default value is `GIT`.
+* `project_directory` - (Optional  if `source` is `GIT`, Required when `source` is `WORKSPACE`) If `source` is `GIT`, then it should be a relative path to the directory in the repository specified in `git_source` where dbt should look in for the `dbt_project.yml` file. If not specified, defaults to the repository's root directory. Equivalent to passing `--project-dir` to a dbt command.  If `source` is `WORKSPACE`, then it should be absolute path to the folder in the workspace.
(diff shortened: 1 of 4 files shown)


docs/resources/job.md
```diff
@@ -323,7 +323,8 @@ You can invoke Spark submit tasks only on new clusters. **In the `new_cluster` s
 ### dbt_task Configuration Block
 
 * `commands` - (Required) (Array) Series of dbt commands to execute in sequence. Every command must start with "dbt".
-* `project_directory` - (Optional) The relative path to the directory in the repository specified in `git_source` where dbt should look in for the `dbt_project.yml` file. If not specified, defaults to the repository's root directory. Equivalent to passing `--project-dir` to a dbt command.
+* `source` - (Optional) What is the source of this project. Possible values are `WORKSPACE` and `GIT`.  Default value depends on if `git_source` block presents in the job definition or not - if it's present, then the default value is `GIT`.
```
gaborratky-db, line 326:
```suggestion
* `source` - (Optional) The source of the project. Possible values are `WORKSPACE` and `GIT`.  Defaults to `GIT` if a `git_source` block is present in the job definition.
```

docs/resources/job.md
```diff
@@ -323,7 +323,8 @@ You can invoke Spark submit tasks only on new clusters. **In the `new_cluster` s
...
+* `project_directory` - (Optional  if `source` is `GIT`, Required when `source` is `WORKSPACE`) If `source` is `GIT`, then it should be a relative path to the directory in the repository specified in `git_source` where dbt should look in for the `dbt_project.yml` file. If not specified, defaults to the repository's root directory. Equivalent to passing `--project-dir` to a dbt command.  If `source` is `WORKSPACE`, then it should be absolute path to the folder in the workspace.
```
gaborratky-db, line 327:
```suggestion
* `project_directory` - (Required when `source` is `WORKSPACE`) The path where dbt should look for `dbt_project.yml`. Equivalent to passing `--project-dir` to the dbt CLI. 
  * If `source` is `GIT`: Relative path to the directory in the repository specified in the `git_source` block. Defaults to the repository's root directory when not specified.
  * If `source` is `WORKSPACE`: Absolute path to the folder in the workspace.
```

docs/resources/job.md
```diff
@@ -362,7 +363,9 @@ One of the `query`, `dashboard` or `alert` needs to be provided.
   * `alert_id` - (Required) (String) identifier of the Databricks SQL Alert.
   * `subscriptions` - (Required) a list of subscription blocks consisting out of one of the required fields: `user_name` for user emails or `destination_id` - for Alert destination's identifier.
   * `pause_subscriptions` - (Optional) flag that specifies if subscriptions are paused or not.
-* `file` - (Optional) block consisting of single string field: `path` - a relative path to the file (inside the Git repository) with SQL commands to execute.  *Requires `git_source` configuration block*.
+* `file` - (Optional) block consisting of single string fields: 
+  * `source` - (Optional) What is the source of this project. Possible values are `WORKSPACE` and `GIT`.
```
gaborratky-db, line 367:
```suggestion
  * `source` - (Optional) The source of the project. Possible values are `WORKSPACE` and `GIT`.
```

docs/resources/job.md
```diff
@@ -362,7 +363,9 @@ One of the `query`, `dashboard` or `alert` needs to be provided.
...
+  * `path` - If `source` is `GIT`, then it a relative path to the file (inside the Git repository) with SQL commands to execute.  If `source` is `WORKSPACE` then it should be absolute path to the file in the workspace.
```
gaborratky-db, line 368:
```suggestion
  * `path` - If `source` is `GIT`: Relative path to the file in the repository specified in the `git_source` block with SQL commands to execute. If `source` is `WORKSPACE`: Absolute path to the file in the workspace with SQL commands to execute.
```

exporter/util.go
```diff
@@ -218,17 +218,24 @@ func (ic *importContext) IsUserOrServicePrincipalDirectory(path, prefix string,
 }
 
 func (ic *importContext) emitRepoByPath(path string) {
-	ic.Emit(&resource{
-		Resource:  "databricks_repo",
-		Attribute: "path",
-		Value:     strings.Join(strings.SplitN(path, "/", 5)[:4], "/"),
-	})
+	parts := strings.SplitN(path, "/", 5)
```
gaborratky-db, line 221: This code was cryptic before and it is still cryptic. Can you add a comment with an example of why this string manipulation is necessary?

exporter/importables.go
```diff
@@ -499,6 +506,25 @@ var resourcesMap map[string]importable = map[string]importable{
 							ID:       task.DbtTask.WarehouseId,
 						})
 					}
+					if task.DbtTask.Source == "WORKSPACE" {
+						directory := task.DbtTask.ProjectDirectory
+						if strings.HasPrefix(directory, "/Repos") {
+							ic.emitRepoByPath(directory)
+						} else {
+							nbAPI := workspace.NewNotebooksAPI(ic.Context, ic.Client)
+							objects, err := nbAPI.List(directory, true, true)
```
gaborratky-db, line 515: Does the `true` mean here that it is recursive? A comment would be nice here as well (e.g., `// Traverse the dbt project directory and emit all objects found`)

gaborratky-db (changes requested):
Added suggestions for docs and a few questions/requests to clarify what and why the code does what it does.

Is there test coverage for `Import` the PR could extend? Logic like `len(parts) >= 4` feel brittle and the assumptions could break about how a path looks like in the future.

exporter/util.go
alexott: It's to create a base repos path... This part is covered by tests, but it will change when we start to support Git folders.

exporter/util.go
```diff
@@ -240,6 +247,7 @@ func (ic *importContext) emitNotebookOrRepo(path string) {
 	if strings.HasPrefix(path, "/Repos") {
 		ic.emitRepoByPath(path)
 	} else {
+		// TODO: strip /Workspace prefix if it's provided
```
tanmay-db, line 250: Question: Is this TODO not required currently? what happens if /Workspace prefix is provided?

exporter/util.go
```diff
@@ -218,17 +218,24 @@ func (ic *importContext) IsUserOrServicePrincipalDirectory(path, prefix string,
...
-	})
+	parts := strings.SplitN(path, "/", 5)
+	if len(parts) >= 4 {
+		ic.Emit(&resource{
+			Resource:  "databricks_repo",
+			Attribute: "path",
+			Value:     strings.Join(parts[:4], "/"),
+		})
+	} else {
+		log.Printf("[WARN] Incorrect Repos path")
```
tanmay-db, line 229: instead of warn, shouldn't we Fail here?

exporter/util.go
tanmay-db: I see, it's not immediately obvious to me why we are splitting by 5, can you also mention an example of before and after split in the comments as well. Do we always stop splitting after 4th `/`? This seems a bit fragile if there is change in how path is represented on backend but I guess there isn't a better way to do this?

tanmay-db (approved):
LGTM, have some questions, logic overall looks good.

exporter/util.go
alexott: it's related to this issue: https://github.com/databricks/terraform-provider-databricks/issues/3195 - we're introduced ability to refer to objects with `/Workspace`  and without, but notebooks/workspace files right now are using path without `/Workspace`, so I just left a comment for myself to handle this case in the near future

exporter/util.go
alexott: Unfortunately, our APIs aren't strictly consistent, so it's possible to create jobs pointing to non-existing paths, clusters, etc., but we should tolerate that :-( Otherwise, a lot of exports will fail - it's easier to fix TF code sometimes than fix objects in the workspace

PR #3219, opened 2024-02-06
UC exporter databricks storage credential feature

ukannikanti:
## Changes
UC exporter databricks storage credential feature.

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -2333,12 +2333,10 @@ var resourcesMap map[string]importable = map[string]importable{
 			}
 
 			for _, v := range objList {
-				if v.Name != "" {
-					ic.Emit(&resource{
-						Resource: "databricks_storage_credential",
-						ID:       v.Name,
-					})
-				}
+				ic.Emit(&resource{
+					Resource: "databricks_storage_credential",
+					ID:       v.Name,
+				})
 			}
 			return nil
 		},
@@ -2351,7 +2349,7 @@ var resourcesMap map[string]importable = map[string]importable{
 	},
 	"databricks_external_location": {
 		WorkspaceLevel: true,
-		Service:        "uc-external-location",
+		Service:        "uc-external-locations",
 		Name: func(ic *importContext, d *schema.ResourceData) string {
 			name := d.Get("name").(string)
 			if name == "" {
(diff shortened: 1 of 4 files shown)


exporter/importables.go
```diff
@@ -2294,4 +2294,40 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			if ic.meAdmin {
+				ic.Emit(&resource{
+					Resource: "databricks_grants",
+					ID:       fmt.Sprintf("storage_credential/%s", r.ID),
+				})
+			}
+			return nil
+		},
+		List: func(ic *importContext) error {
+			objList, err := catalog.NewStorageCredentials(ic.Client.DatabricksClient).ListAll(ic.Context, catalog.ListStorageCredentialsRequest{})
```
alexott, line 2317: it would be simpler to use already existing workspace (or account) client:

```
ic.workspaceClient.StorageCredentials.ListAll(...)
```

exporter/importables.go
```diff
@@ -2294,4 +2294,40 @@ var resourcesMap map[string]importable = map[string]importable{
 		},
 		// TODO: add Depends & Import to emit corresponding UC Volumes when support for them is added
 	},
+	"databricks_storage_credential": {
+		WorkspaceLevel: true,
+		Service:        "storage_credentials",
```
alexott, line 2299: let's make it explicit that it's uc-related and use something like `uc-storage-credentials`, like here: https://registry.terraform.io/providers/databricks/databricks/latest/docs/guides/experimental-exporter#uc-artifact-allowlist

exporter/importables.go
```diff
@@ -2294,4 +2294,40 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			objList, err := catalog.NewStorageCredentials(ic.Client.DatabricksClient).ListAll(ic.Context, catalog.ListStorageCredentialsRequest{})
+			if err != nil {
+				return err
+			}
+			for _, v := range objList {
+				if v.Name != "" {
+					id := v.Name
+					ic.Emit(&resource{
+						Resource: "databricks_storage_credential",
+						ID:       id,
```
alexott, line 2326: Just do `ID: v.Name` instead of having separate line `id := v.Name`

exporter/importables.go
```diff
@@ -2294,4 +2294,40 @@ var resourcesMap map[string]importable = map[string]importable{
 		},
 		// TODO: add Depends & Import to emit corresponding UC Volumes when support for them is added
 	},
+	"databricks_storage_credential": {
+		WorkspaceLevel: true,
```
alexott, line 2298: We also may need to define `ShouldOmitField` otherwise some fields marked as `computed` (i.e., `owner`, ...) will be skipped when generating the code.

alexott (changes requested):
we also need to update `guides/experimental-exporter.md` with the new service and update the support matrix

alexott (approved):

PR #3221, opened 2024-02-06
[Feature] Add support for Databricks Account Console IP Access Lists

### LINKED ISSUE #3054: [ISSUE] `databricks_ip_access_list` doesn't work on account level
### Configuration

```hcl
resource "databricks_ip_access_list" "vpn_and_test_infra" {
    provider = databricks.account
    for_each = module.defaults.cidrs
    label     = each.key
    list_type = "ALLOW"
    ip_addresses = each.value
}
```

### Expected Behavior
Account IP ACLs are there for 1+ year
image

### Actual Behavior
```
╷
│ Error: cannot create ip access list: invalid Databricks Workspace configuration
│ 
│   with databricks_ip_access_list.cidrs["Databricks VPN"],
│   on vpn.tf line 1, in resource "databricks_ip_access_list" "cidrs":
│    1: resource "databricks_ip_access_list" "cidrs" {
```

nkvuong:
> [!WARNING]  
> This couldn't be done right now due to the backend behavior that prevents Terraform implementation from working correctly

## Changes
- Add support for `databricks_ip_access_list` at account-level 
- Refactor `databricks_ip_access_list` tests to use Mock

Resolves #3054

## Tests

- [x] `make test` run locally
- [x] using Go SDK


alexott:
@nkvuong will we avoid the problem with the new "skip read after write" functionality?

nkvuong:
@alexott let me test this out

access/resource_ip_access_list.go
```diff
@@ -30,48 +33,86 @@ func ResourceIPAccessList() common.Resource {
...
+					var apiErr *apierr.APIError
+					if !errors.As(err, &apiErr) {
+						return retry.NonRetryableError(err)
+					}
+					if apiErr.StatusCode == 404 {
+						return retry.RetryableError(err)
+					} else {
+						return retry.NonRetryableError(err)
+					}
+				})
```
alexott, line 63: Should we remove this workaround?

access/resource_ip_access_list.go
```diff
@@ -30,48 +33,86 @@ func ResourceIPAccessList() common.Resource {
...
+				//need to enable the IP Access List with update
+				if d.Get("enabled").(bool) {
+					updateIacl.IpAccessListId = ipAclId
+					err = acc.IpAccessLists.Update(ctx, updateIacl)
+					if err != nil {
+						return err
+					}
+				}
+				d.SetId(ipAclId)
```
alexott, line 72: I think that we need to add `StructToData` on the result of the `acc.IpAccessLists.Create` here (and same for workspace-level create)

access/resource_ip_access_list.go
```diff
@@ -30,48 +33,86 @@ func ResourceIPAccessList() common.Resource {
 	})
 	return common.Resource{
 		Schema: s,
+		CanSkipReadAfterCreateAndUpdate: func(d *schema.ResourceData) bool {
+			//only skip read after create
+			return d.IsNewResource()
```
alexott, line 38: But if we'll have Read after Update, then we may get inconsistent state - API docs explicitly say that it may take a few minutes to propagate.

access/resource_ip_access_list.go
nkvuong: so the issue, we can only enable the Account-level IP ACL after it has been created, so we still need to wait until it has been created. Otherwise the call to enable the ACL will fail

pandabrun:
any plan on merging this one in a foreseeable future folks?

tanmay-db (commented):

Issue #3001, opened 2023-12-07
[ISSUE] Issue with `databricks_sql_table` resource during table schema evolution


vsluc:
When a new column is added to the column configuration in the `end`, the databricks_sql_table resource recreates a table instead of altering the table and adding the new column.

The table was first created using the same table configuration provided below except that `name` column was not there.

### Configuration

```hcl
resource "databricks_sql_table" "test_managed" {
  name               = "test_managed"
  catalog_name       = "my_catalog"
  schema_name        = "my_schema"
  table_type         = "MANAGED"
  data_source_format = "DELTA"

  column {
    name = "id"
    type = "int"
  }
  column {
    name = "sid"
    type = "string"
  }
  column {
    name = "name"
    type = "string"
  }  

  properties = {
    "delta.autoOptimize.autoCompact" = true,
    "delta.autoOptimize.optimizeWrite" = true,    
    "delta.enableChangeDataFeed" = true
  }
}
```

### Expected Behavior

Any new column added in the end, should be treated as an ALTER to the table and the new column is added without destroying & recreating the table with the new column. The recreation causes the data in the table to be deleted, which is undesirable. When using SQL interface (on a SQL warehouse) to do the same operation of adding a new column in the end, it just ALTERs the table and adds the new column in the end. The data is retained as well. We should have the same behavior in terraform as well.

### Actual Behavior

The table gets destroyed and recreated. The data in the table is `lost`.

### Steps to Reproduce

1. Create a table with 2 columns `id, sid`:

```hcl
resource "databricks_sql_table" "test_managed" {
  name               = "test_managed"
  catalog_name       = "my_catalog"
  schema_name        = "my_schema"
  table_type         = "MANAGED"
  data_source_format = "DELTA"

  column {
    name = "id"
    type = "int"
  }
  column {
    name = "sid"
    type = "string"
  }

  properties = {
    "delta.autoOptimize.autoCompact" = true,
    "delta.autoOptimize.optimizeWrite" = true,    
    "delta.enableChangeDataFeed" = true
  }
}
```
2. Insert sample rows:

```sql
insert into table my_catalog.my_schema.test_managed
values
(1,"sid_1")
,(2,"sid_2");
```

3. Alter the table by adding a new column `name` at the end:

```hcl
resource "databricks_sql_table" "test_managed" {
  name               = "test_managed"
  catalog_name       = "my_catalog"
  schema_name        = "my_schema"
  table_type         = "MANAGED"
[...]

PR #3222, opened 2024-02-06
Fixed `SqlColumnInfo` for `force_new`

tanmay-db:
## Changes

We need to mark the fields as force_new since they are part of structure. As updating column isn't supported, we mark all the fields as force_new.

Relevant code ref:
````
ColumnInfos []SqlColumnInfo `json:"columns,omitempty" tf:"alias:column,computed,force_new"`
````

This doesn't seem to be regression on TF side, this test fails on 1.35.0 commit as well. 

## Tests

`TestUcAccResourceSqlTable_Managed` passes after change
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -20,10 +20,10 @@ import (
 var MaxSqlExecWaitTimeout = 50
 
 type SqlColumnInfo struct {
-	Name     string `json:"name"`
-	Type     string `json:"type_text,omitempty" tf:"suppress_diff,alias:type"`
-	Comment  string `json:"comment,omitempty"`
-	Nullable bool   `json:"nullable,omitempty" tf:"default:true"`
+	Name     string `json:"name" tf:"force_new"`
+	Type     string `json:"type_text,omitempty" tf:"suppress_diff,alias:type,force_new"`
+	Comment  string `json:"comment,omitempty" tf:"force_new"`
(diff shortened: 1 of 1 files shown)


hectorcast-db (approved):

alexott:
Imho it's too dangerous... We may need something like this: https://github.com/databricks/terraform-provider-databricks/pull/2424

mgyucht:
@alexott @nkvuong Do you know how https://github.com/databricks/terraform-provider-databricks/blob/main/internal/acceptance/sql_table_test.go#L39-L66 succeeded in the past? Updating the type of the table would not cause the table to be deleted/recreated, but as a result, there should still have been a diff in the type of this column.

I believe this was the intention from the start, as SqlColumnInfos is annotated with `ForceNew`. However, in the SDKv2, my understanding is that ForceNew on TypeList attributes will only cause the resource to be deleted/recreated when the number of entries in the list changes. In this case, the table would be recreated when columns are added and removed, but not if the number of columns remains unchanged. 

In any case: lack of update support has been raised a number of times: https://github.com/databricks/terraform-provider-databricks/issues/3001. We can patch this to prevent diffs from appearing when users change the types of a column, but this is really just covering up an underlying lack of support of updating table schema. Rather than patch this like this, can we address the root cause by simply supporting updates to tables & issuing `ALTER TABLE` statements to modify columns?

PR #3225, opened 2024-02-07
Exporter: detect & handle deleted workspace objects (notebooks/files/directories) when running in incremental mode

alexott:
## Changes

The initial version of code for detection & handling of removed workspace objects (notebooks, workspace files, and directories) and associated permissions... Renames are handled as (delete/create new)...

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


exporter/context.go
```diff
@@ -499,6 +529,110 @@ func (ic *importContext) Run() error {
...
+	var rtype string
+	switch obj.ObjectType {
+	case workspace.Directory:
+		rtype = "databricks_directory"
+	case workspace.File:
+		rtype = "databricks_workspace_file"
+	case workspace.Notebook:
+		rtype = "databricks_notebook"
+	default:
+		log.Printf("[WARN] Unsupported WS object type: %s in obj %v", obj.ObjectType, obj)
```
tanmay-db, line 546: Should we fail here?

exporter/context.go
```diff
@@ -649,7 +796,17 @@ func (ic *importContext) writeImports(sh *os.File, importChan importWriteChannel
 	}
 	if sh != nil {
 		log.Printf("[DEBUG] Writing the rest of import commands. len=%d", len(ic.shImports))
+		// TODO: remove deleted imports...
```
tanmay-db, line 799: What happens if we don't do it now?

exporter/util.go
```diff
@@ -963,9 +963,13 @@ func (ic *importContext) shouldSkipWorkspaceObject(object workspace.ObjectStatus
 	}
 	modifiedAt := wsObjectGetModifiedAt(object)
 	if ic.incremental && modifiedAt < updatedSinceMs {
-		log.Printf("[DEBUG] skipping '%s' that was modified at %d (last active=%d)",
-			object.Path, modifiedAt, updatedSinceMs)
-		return true
+		p := ic.oldWorkspaceObjectMapping[object.ObjectID]
+		if p == "" || p == object.Path {
+			log.Printf("[DEBUG] skipping '%s' that was modified at %d (last active=%d)",
+				object.Path, modifiedAt, updatedSinceMs)
```
tanmay-db, line 969: is last active same as last updated?

tanmay-db (commented):
Hi @alexott, I have some comments, mentioned them. Right now, we don't have E2E tests for exporter but just to check, has this been tested E2E?

exporter/util.go
alexott: No. Last Active - when the exporter ran last time...

exporter/context.go
alexott: it's too late to fail - it's better to don't some missing code that could be added manually, than fail multi-hour work. Although, this block is just defensive piece - because we have filters earlier we don't reach it.

exporter/context.go
alexott: forgot to remove the comment

alexott:
Most of the tests in `exporter_test` are running this tool as a command-line app. I'm testing all things manually with manual checks for generated code. Setting up fully automated testing is a significant task on its own (especially on huge workspaces), and will consume a lot of resources that I don't have.

tanmay-db (approved):

PR #3227, opened 2024-02-07
Add CustomDiffFunc for health in sql_endpoint resources

### LINKED ISSUE #3204: [ISSUE] Issue with `databricks_sql_endpoint` resource
### Configuration
Any databricks_sql_endpoint resource updates the plan every single time a plan/apply is called to update the 'health'.  This started happening sometime after 1.31.0 but not sure which specific version.  We have half a dozen SQL endpoints created through each terraform deployment and every single time a plan or apply runs it shows all of them getting updated with "health = (known after apply)". It pollutes the plan and makes it very annoying to approve pull requests.

### Expected Behavior
terraform plan should never show any updates to anything that shouldn't/hasn't changed.

### Actual Behavior
Every single databricks_sql_endpoint resource shows a change to the 'health' attribute every time a plan or apply runs, regardless if nothing has changed (we can run 10 applies in a row with 10 seconds between each one and each of them show all SQL endpoint health attributes changing.

### Steps to Reproduce
1. define a databricks_sql_endpoint resource
2. run terraform apply
3. run terraform plan or terraform apply as many times as you want and watch the health attribute always update

I've tried both 1.34.0 and 1.35.0 - the behavior does not occur on our previous version (1.31.0)
Terraform version is 1.6.5

### Is it a regression?
Yes, worked fine on 1.31.0 and all previous versions

### Debug Output
I cannot easily get this information, if it's really needed please comment, otherwise its going to take a fairly large amount of effort to produce it.

edwardfeng-db:
## Changes

- Add `CustomDiffFunc` so that we ignore the diff if `health` is the only field with the diff
- Add `ApplyCheckDiff` function in `ResourceFixture` and added a test case to make sure `health` diff is ignored

Fixes #3204

## Tests

- [x] `make test` run locally
- [x] using Go SDK


alexott:
Why not simply make `Optional` to `true`?

tanmay-db:
> Why not simply make `Optional` to `true`?

Then users would be able to specify the `health` field in their resource which they shouldn't be able to right? cc: @mgyucht

mgyucht (commented):
This looks correct to me. Let's add a unit test that there is no diff, even if read returns a difference for `health`.

qa/testing.go
```diff
@@ -344,6 +351,11 @@ func (f ResourceFixture) Apply(t *testing.T) (*schema.ResourceData, error) {
 	return resourceData, err
 }
 
+// ApplyCheckDiff not only applies the test case from fixture, but also takes in a diffChecker function and asserts the output of the diffChecker function to be true.
+func (f ResourceFixture) ApplyCheckDiff(t *testing.T, diffChceker func(*terraform.InstanceDiff) bool) (*schema.ResourceData, error) {
+	return f.applyInternal(t, diffChceker)
```
mgyucht, line 356: Did you consider adding DiffChecker to ResourceFixture? New `Apply*` methods diverge the behavior between different cases. Suppose we want to do some other check after `Apply()`, like verify the final state. Would we want `ApplyCheckDiffAndState` and `ApplyCheckState` methods? Similarly, note that there aren't `ApplyRead/ApplyUpdate/ApplyCreate/ApplyDelete` methods. Almost all logic is in `Apply` itself. 

A better design would be to add this as a field to `ResourceFixture` itself. This is backwards compatible (no current test specifies this) and can be combined with current functionality easily.

sql/resource_sql_endpoint_test.go
```diff
@@ -239,6 +240,45 @@ func TestResourceSQLEndpointUpdate(t *testing.T) {
...
+		},
+		Update: true,
+		HCL: `
+		name = "foo"
+  		cluster_size = "Small"
+		`,
+	}.ApplyCheckDiff(t, func(diff *terraform.InstanceDiff) bool {
+		m := diff.Attributes
+		_, healthDiffExists := m["health"]
+		return !healthDiffExists
```
mgyucht, line 278: This is weird. This test executes the update pathway, but we don't actually care about that. To check the diff, we don't need anything except for the config and instance state. Is this really the optimal interface for testing diff checking?

Here's what I think would be a better example of a customize diff check:
```go
qa.ResourceFixture{
  Resource: ResourceSqlEndpoint(),
  ID: "abc",
  InstanceState: ...,
  HCL: ...,
  CustomizeDiff: true,
  ExpectedDiff: map[string]*terraform.ResourceAttrDiff // Or func(*terraform.InstanceDiff) error for flexibility
}.ApplyNoError(t)
```

We don't need to test any execution when unit testing customize diff.

mgyucht (changes requested):
Let's make sure we're happy with the interface. I don't want to take shortcuts when it comes to core interfaces, whether in testing or production code.

qa/testing.go
edwardfeng-db: This makes sense to me, I can update this

sql/resource_sql_endpoint_test.go
edwardfeng-db: A bit confused by `we don't need to test any execution` and the example you put above. I think running `ApplyNoError` still does the execution?

For the `ExpectedDiff` interface, I thought about that but I decided to put in a lambda that asserts certain properties of the diff to be true because the diff can be fairly large sometimes and specifying the whole diff manually can be less elegant than just asserting certain diffs exists

edwardfeng-db: Discussed offline, for `ExpectedDiff`, it is giving us better control over the diffs if we specify the full ResourceAttrDiff object, so going with that.

Issue #3231, opened 2024-02-08
[ISSUE] Issue with `databricks_cluster` resource while upgrading provider from 1.35.0 to 1.36.1


donatasm:
Trying to upgrade provider from version `1.35.0` to `1.36.1` and all `databricks_cluster` resources having `init_scripts` return errors when running terraform apply:

```
Error: cannot read cluster: init_scripts: init_scripts: workspace: [workspace]: inconsistency: destination has omitempty, but is not optional
```

Cluster definition:

```
resource "databricks_cluster" "this" {
  cluster_name = "my_cluster"

   ...

  init_scripts {
    workspace {
      destination = "/path/to/script"
    }
  }
}
```

I cannot confirm it works for clusters without `init_scripts`.

### Debug Output

```
2024-02-08T10:23:09.182Z [DEBUG] provider.terraform-provider-databricks_v1.36.1: GET /api/2.0/clusters/get?cluster_id=0222-123456-abcd
< HTTP/2.0 200 OK
< {
<   "autoscale": {
<     "max_workers": 2,
<     "min_workers": 1,
<     "target_workers": 1
<   },
<   "autotermination_minutes": 40,
<   "aws_attributes": {
<     "availability": "SPOT_WITH_FALLBACK",
<     "first_on_demand": 1,
<     "instance_profile_arn": "arn:aws:iam::123456789000:instance-profile/databricks-profile",
<     "spot_bid_price_percent": 100,
<     "zone_id": "auto"
<   },
<   "cluster_id": "0222-123456-abcd",
<   "cluster_name": "my_cluster",
<   "cluster_source": "API",
<   "creator_user_name": "email@company.org",
<   "default_tags": {
<     "ClusterId": "0222-123456-abcd",
<     "ClusterName": "my_cluster",
<     "Creator": "email@company.org",
<     "Vendor": "Databricks"
<   },
<   "disk_spec": {},
<   "driver_healthy": true,
<   "driver_instance_source": {
<     "node_type_id": "m5a.large"
<   },
<   "driver_node_type_id": "m5a.large",
<   "effective_spark_version": "14.3.x-cpu-ml-scala2.12",
<   "enable_elastic_disk": true,
<   "enable_local_disk_encryption": true,
<   "init_scripts": [
<     {
<       "workspace": {
<         "destination": "/path/to/script"
<       }
<     }
<   ],
<   "init_scripts_safe_mode": false,
<   "instance_source": {
[...]

PR #3232, opened 2024-02-08
Remove `omitempty` in `destination` fields in clustes_api

edwardfeng-db:
## Changes

- Addressing issue raised in https://github.com/databricks/terraform-provider-databricks/issues/3231

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -194,22 +194,22 @@ type S3StorageInfo struct {
 
 // GcsStorageInfo contains the struct for when storing files in GCS
 type GcsStorageInfo struct {
-	Destination string `json:"destination,omitempty"`
+	Destination string `json:"destination"`
 }
 
 // AbfssStorageInfo contains the struct for when storing files in ADLS
 type AbfssStorageInfo struct {
-	Destination string `json:"destination,omitempty"`
+	Destination string `json:"destination"`
 }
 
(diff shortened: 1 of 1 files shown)


clusters/clusters_api.go
```diff
@@ -194,22 +194,22 @@ type S3StorageInfo struct {
...
 
 // LocalFileInfo represents a local file on disk, e.g. in a customer's container.
 type LocalFileInfo struct {
-	Destination string `json:"destination,omitempty"`
+	Destination string `json:"destination"`
 }
 
 // WorkspaceFileInfo represents a file in the Databricks workspace.
 type WorkspaceFileInfo struct {
-	Destination string `json:"destination,omitempty"`
```
alexott, line 212: It shouldn't be used in the cluster resource...

hectorcast-db (approved):

alexott (commented):
I'm not sure that it will address the problem - `workspace` block comes from `compute.ClusterSpec`

edwardfeng-db:
Yeah @alexott it is partially migrated, the `create` `update` and `read` functions are still not using the SDK to make API calls. So it is using the `clusters_api.go` structs for API calls with the go sdk schema as the registered schema for this resource. If that makes sense.

alexott (approved):

PR #3238, opened 2024-02-09
[MLOPS-596] Add Lakehouse Monitor Resource

aravind-segu:
## Changes
Adds Lakehouse Monitor Resource to the Terraform Provider

## Tests

- [x] `make test` run locally
- [x] using Go SDK


monitoring/resourse_lakehouse_monitor.go
```diff
@@ -0,0 +1,139 @@
...
+	for k, v := range originalSchema {
+		tmpSchema[k] = v
+	}
+	delete(tmpSchema, "snapshot")
+	return tmpSchema
+}
+
+type Snapshot struct {
+}
+type Monitor struct {
```
alexott, line 22: Why do we need this struct? Why can't we use `catalog.MonitorInfo` directly?

monitoring/resourse_lakehouse_monitor.go
```diff
@@ -0,0 +1,139 @@
...
+	SlicingExprs             []string                                 `json:"slicing_exprs,omitempty"`
+	TimeSeries               *catalog.MonitorTimeSeriesProfileType    `json:"time_series,omitempty"`
+	WarehouseId              string                                   `json:"warehouse_id,omitempty"`
+}
+
+type MonitorInfo struct {
+	Monitor
+	DriftMetricsTableName   string                    `json:"drift_metrics_table_name,omitempty"`
+	ProfileMetricsTableName string                    `json:"profile_metrics_table_name,omitempty"`
+	Status                  catalog.MonitorInfoStatus `json:"status,omitempty"`
```
alexott, line 43: All of these should be marked as `computed` when customizing the schema...

monitoring/resourse_lakehouse_monitor.go
```diff
@@ -0,0 +1,139 @@
...
+			common.MustSchemaPath(m, "inference_log", "timestamp_col").Required = true
+			common.MustSchemaPath(m, "inference_log", "timestamp_col").Optional = false
+
+			common.MustSchemaPath(m, "inference_log", "prediction_col").Default = "prediction"
+
+			common.MustSchemaPath(m, "time_series", "granularities").Required = true
+			common.MustSchemaPath(m, "time_series", "granularities").Optional = false
+
+			common.MustSchemaPath(m, "time_series", "timestamp_col").Required = true
+			common.MustSchemaPath(m, "time_series", "timestamp_col").Optional = false
```
alexott, line 68: These customizations aren't reflected in the Terraform schema because the `Schema` field is set to `readSchema`

monitoring/resourse_lakehouse_monitor.go
```diff
@@ -0,0 +1,139 @@
...
+	AssetsDir                string                                   `json:"assets_dir"`
+	BaselineTableName        string                                   `json:"baseline_table_name,omitempty"`
+	CustomMetrics            []catalog.MonitorCustomMetric            `json:"custom_metrics,omitempty"`
+	DataClassificationConfig *catalog.MonitorDataClassificationConfig `json:"data_classification_config,omitempty"`
+	FullName                 string                                   `json:"table_name"`
+	InferenceLog             *catalog.MonitorInferenceLogProfileType  `json:"inference_log,omitempty"`
+	Notifications            []catalog.MonitorNotificationsConfig     `json:"notifications,omitempty"`
+	OutputSchemaName         string                                   `json:"output_schema_name"`
+	Schedule                 *catalog.MonitorCronSchedule             `json:"schedule,omitempty"`
+	Snapshot                 Snapshot                                 `json:"snapshot,omitempty"`
```
alexott, line 32: If snapshot isn't used right now, we can remove it completely for now... This could be done in schema customization function as well

alexott (changes requested):

monitoring/resourse_lakehouse_monitor.go
alexott: The lack of `computed` leads to failure of the integration tests:

```
          ~ resource "databricks_lakehouse_monitor" "testMonitor" {
              - drift_metrics_table_name   = "sandbox$tblbheiblgbgc.things$tblbheiblgbgc.bar$tblbheiblgbgc_drift_metrics" -> null
                id                         = "sandbox$tblbheiblgbgc.things$tblbheiblgbgc.bar$tblbheiblgbgc"
              - profile_metrics_table_name = "sandbox$tblbheiblgbgc.things$tblbheiblgbgc.bar$tblbheiblgbgc_profile_metrics" -> null
              - status                     = "MONITOR_STATUS_PENDING" -> null
                # (3 unchanged attributes hidden)
        
                # (1 unchanged block hidden)
            }
        
```

monitoring/resourse_lakehouse_monitor.go
aravind-segu: We need this struct because of the `Snapshot` field. Snapshot is of type `any` and it is not compatible with the schema. Therefore this custom struct uses an empty struct for snapshot, and then initializes the snapshot field later in the create and update below.

monitoring/resourse_lakehouse_monitor.go
aravind-segu: Snapshot is being used. There are three types of monitors, Snapshot, Timeseries and Inference Log. For a snapshot there are no extra parameters, so the user just needs to instantiate the Snapshot field in order to create a snapshot monitor.

alexott:
@aravind-segu integration tests are passing (at least on AWS)

monitoring/resourse_lakehouse_monitor.go
mgyucht: Why is `Snapshot` `any`? What is its actual type? Is it mean to be `struct{}`, and this is a bug with the types in the Go SDK? Or can users actually pass any value for this field?

mgyucht (changes requested):
Thank you for contributing this resource! We do need to make some adjustments to improve the long-term maintainability story, and they may necessitate deeper changes. If it isn't possible to do this right from the start, we need to prioritize the work needed for this to be properly supported.

monitoring/resourse_lakehouse_monitor.go
```diff
@@ -0,0 +1,139 @@
...
+			if _, ok := d.GetOk("snapshot"); ok {
+				create.Snapshot = struct{}{}
+			}
+			create.FullName = d.Get("table_name").(string)
+
+			endpoint, err := w.LakehouseMonitors.Create(ctx, create)
+			if err != nil {
+				return err
+			}
+			d.SetId(endpoint.TableName)
```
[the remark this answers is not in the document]
BoopBoopBeepBoop: the ID of the monitor should not be changed

monitoring/resourse_lakehouse_monitor.go
```diff
@@ -0,0 +1,139 @@
...
+				return err
+			}
+			tmpSchema := removeSnapshotField(monitorSchema)
+
+			var create catalog.CreateMonitor
+			common.DataToStructPointer(d, tmpSchema, &create)
+			if _, ok := d.GetOk("snapshot"); ok {
+				create.Snapshot = struct{}{}
+			}
+			create.FullName = d.Get("table_name").(string)
```
alexott, line 91: Why do we need this? `DataToStructPointer` deserialize things - you already have `FullName` annotated as `table_name`...

monitoring/resourse_lakehouse_monitor.go
alexott, line 32: If `Snapshot` is optional, then it should be a pointer, not struct: 

```
Snapshot                 *Snapshot                                 `json:"snapshot,omitempty"`
```

monitoring/resourse_lakehouse_monitor.go
alexott, line 43: This also should be a pointer

monitoring/resourse_lakehouse_monitor.go
mgyucht: This is a bug in the Go SDK. The Go SDK currently treats empty structures and components that can take on any value in the same way. I am opening a PR to fix this here: https://github.com/databricks/databricks-sdk-go/pull/821

monitoring/resourse_lakehouse_monitor.go
aravind-segu: MonitorInfoStatus is just a string enum. If I make it a pointer, the `StructToSchema` method expects it to be a struct.

monitoring/resourse_lakehouse_monitor.go
aravind-segu: In the `catalog.CreateMonitor` struct we define the `FullName` with the json path `json:"-" url:"-"`. Therefore we have to set it manually as there is no json path that the schema can set it to.

alexott:
Integration tests are failing - I fixed incorrect template (please pull changes), but even after that, it looks like `Update` method doesn't send the correct payload:

```
2024-02-24T11:02:07.276Z [DEBUG] databricks: PUT /api/2.1/unity-catalog/tables/sandbox$tdeaabhaalaji.things$tabikdcljiild.bar$tabikdcljiild_inference/monitor
> {
>   "inference_log": {
>     "granularities": [
>       "1 hour"
>     ],
>     "model_id_col": "model_id",
>     "prediction_col": "prediction",
>     "problem_type": "PROBLEM_TYPE_REGRESSION",
>     "timestamp_col": "timestamp"
>   },
>   "output_schema_name": "sandbox$tdeaabhaalaji.things$tabikdcljiild",
>   "snapshot": {}
> }
< HTTP/2.0 400 Bad Request
< {
<   "message": "The given `info` is invalid for the following reason(s):\n - `assets_dir` is required but missing... (154 more bytes)"
< }: tf_req_id=e6f4373c-29d5-83ef-c7fd-ea412a3c369d tf_provider_addr=registry.terraform.io/hashicorp/databricks tf_rpc=ApplyResourceChange tf_resource_type=databricks_lakehouse_monitor
Error: -24T11:02:07.276Z [ERROR] sdk.proto: Response contains error diagnostic: diagnostic_detail="" tf_req_id=e6f4373c-29d5-83ef-c7fd-ea412a3c369d tf_provider_addr=registry.terraform.io/hashicorp/databricks tf_proto_version=5.4 tf_rpc=ApplyResourceChange tf_resource_type=databricks_lakehouse_monitor diagnostic_severity=ERROR
  diagnostic_summary=
  | cannot update lakehouse monitor: The given `info` is invalid for the following reason(s):
  |  - `assets_dir` is required but missing.
  |  - `table_name` is required but missing.
  |  - `profile_metrics_table_name` is required but missing.
  |  - `drift_metrics_table_name` is required but missing.
```

catalog/resource_lakehouse_monitor.go
```diff
@@ -0,0 +1,87 @@
...
+func ResourceLakehouseMonitor() common.Resource {
+	monitorSchema := common.StructToSchema(
+		Monitor{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			common.CustomizeSchemaPath(m, "output_schema_name").SetRequired()
+
+			common.CustomizeSchemaPath(m, "inference_log", "granularities").SetOptional()
+			common.CustomizeSchemaPath(m, "inference_log", "problem_type").SetOptional()
+			common.CustomizeSchemaPath(m, "inference_log", "timestamp_col").SetOptional()
+			common.CustomizeSchemaPath(m, "inference_log", "prediction_col").SetDefault("prediction")
```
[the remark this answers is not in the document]
mgyucht: +1, let's not have defaults in the provider that aren't already in the API. If we want this behavior, it should be consistent with the SDKs and other API users. Could we change the REST API itself to provide a default value if not specified (i.e. make this a computed field)?

catalog/resource_lakehouse_monitor.go
```diff
@@ -0,0 +1,87 @@
...
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/catalog"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type Monitor struct {
+	catalog.CreateMonitor
+	FullName string `json:"table_name"`
```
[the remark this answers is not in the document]
mgyucht: +1 for matching the path parameter exactly. `table_name` is also kind of confusing because it might suggest that you only need the table name itself for this request, but I think you want the fully qualified name including catalog/schema, right?

Hm... @alexott we should probably handle this automatically in `StructToSchema`, right? Do you know why we ignore fields that are missing the `json` tag or have `json:"-"`?

catalog/resource_lakehouse_monitor_test.go
```diff
@@ -0,0 +1,225 @@
...
+				TimeSeries: &catalog.MonitorTimeSeriesProfileType{
+					Granularities: []string{"1 day"},
+					TimestampCol:  "timestamp",
+				},
+			}).Return(&catalog.MonitorInfo{
+				AssetsDir:        "sample.dir",
+				OutputSchemaName: "output.schema",
+				TableName:        "test_table",
+				Status:           catalog.MonitorInfoStatusMonitorStatusActive,
+			}, nil)
```
mgyucht, line 47: This is duplicate? Why do we need to call CreateMonitor twice? Maybe this is a bug in the unit-testing framework?

mgyucht (commented):
+1 on @alexott's comments. Otherwise, this is a much nicer approach for us to maintain.

@edwardfeng-db it might be good to look at this w.r.t. my comment on StructToSchema.

catalog/resource_lakehouse_monitor.go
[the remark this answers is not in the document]
aravind-segu: I think its because this field is part of the query  parameters in the URL rather than the request body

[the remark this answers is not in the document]
mgyucht: Yes, but this structure is not used for sending requests, only for computing the TF schema.

catalog/resource_lakehouse_monitor_test.go
aravind-segu: removed it.

alexott:
Integration test is still failing:

```
    init_test.go:236: Step 1/2 error: After applying this test step, the plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # databricks_lakehouse_monitor.testMonitorInference will be updated in-place
          ~ resource "databricks_lakehouse_monitor" "testMonitorInference" {
              - drift_metrics_table_name   = "sandbox$tdkkfakkkfadj.things$tdkkfakkkfadj.bar$tdkkfakkkfadj_inference_drift_metrics" -> null
                id                         = "sandbox$tdkkfakkkfadj.things$tdkkfakkkfadj.bar$tdkkfakkkfadj_inference"
              - monitor_version            = "0" -> null
              - profile_metrics_table_name = "sandbox$tdkkfakkkfadj.things$tdkkfakkkfadj.bar$tdkkfakkkfadj_inference_profile_metrics" -> null
              - status                     = "MONITOR_STATUS_PENDING" -> null
                # (3 unchanged attributes hidden)
        
                # (1 unchanged block hidden)
            }
        
```

alexott:
after fixing computed fields, another error in update  test:

```
    init_test.go:236: Step 2/2 error: Error running apply: exit status 1
        
        Error: cannot update lakehouse monitor: Data Monitor 'sandbox$tijgeledgkhhb.things$tejcfjihaeblg.bar$tejcfjihaeblg_inference' does not exist.
        
          with databricks_lakehouse_monitor.testMonitorInference,
          on terraform_plugin_test.tf line 36, in resource "databricks_lakehouse_monitor" "testMonitorInference":
          36:         resource "databricks_lakehouse_monitor" "testMonitorInference" {
        

```

aravind-segu:
Integration tests are passing: https://github.com/databricks-eng/eng-dev-ecosystem/actions/runs/8147735527

alexott (changes requested):
Small changes are still required in the code. 

Plus there is no documentation yet

catalog/resource_lakehouse_monitor.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const DefaultProvisionTimeout = 15 * time.Minute
+
+func WaitForMonitor(w *databricks.WorkspaceClient, ctx context.Context, monitorName string) error {
+	return retry.RetryContext(ctx, DefaultProvisionTimeout, func() *retry.RetryError {
+		endpoint, err := w.LakehouseMonitors.GetByFullName(ctx, monitorName)
```
mgyucht, line 19: The SDK supports long-running operations automatically through the `x-databricks-wait` annotation. Will send you the link offline. This will let you use the following standard pattern for long-running operations:
```go
wait, err := w.LakehouseMonitors.Create(ctx, req)
if err != nil {
  return err
}
wait.Get()
```

catalog/resource_lakehouse_monitor.go
```diff
@@ -0,0 +1,120 @@
...
+		return retry.RetryableError(fmt.Errorf("monitor %s is still pending", monitorName))
+	})
+}
+
+func ResourceLakehouseMonitor() common.Resource {
+	monitorSchema := common.StructToSchema(
+		catalog.MonitorInfo{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			common.CustomizeSchemaPath(m, "assets_dir").SetRequired()
+			common.CustomizeSchemaPath(m, "output_schema_name").SetRequired()
```
mgyucht, line 39: These can be removed, based on the specification: these fields are not marked `omitempty`, so they will be required in the Terraform schema.

catalog/resource_lakehouse_monitor.go
```diff
@@ -0,0 +1,120 @@
...
+			common.CustomizeSchemaPath(m, "table_name").SetRequired()
```
mgyucht, line 40: This does not seem to match any field in the create request. What field does this correspond to? I think this should be `SetReadOnly()`.

catalog/resource_lakehouse_monitor.go
```diff
@@ -0,0 +1,120 @@
...
+	return common.Resource{
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+
+			var create catalog.CreateMonitor
+			common.DataToStructPointer(d, monitorSchema, &create)
+			create.FullName = d.Get("table_name").(string)
```
mgyucht, line 69: Isn't this name part of the Create response? How does the user know this ahead of time?

mgyucht (commented):
Two main comments:
1. Waiters are supported in the SDK by default, you just need to annotate your API appropriately. I'll send you the link to this offline.
2. How exactly is table_name supposed to work? It seems like it is a required field but it is not in the Create or Update requests.

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,130 @@
...
+* `schedule` - The schedule for automatically updating and refreshing metric tables.
+* `skip_builtin_dashboard` - Whether to skip creating a default dashboard summarizing data quality metrics.
+* `slicing_exprs` - List of column expressions to slice data with for targeted analysis. The data is grouped by each expression independently, resulting in a separate slice for each predicate and its complements. For high-cardinality columns, only the top 100 unique values by frequency will generate slices.
+* `warehouse_id` - Optional argument to specify the warehouse for dashboard creation. If not specified, the first running warehouse will be used.
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` -  ID of this monitor is the same as the full table name of the format `{catalog}.{schema_name}.{table_name}`
```
alexott, line 122: We should document the computed fields: `monitor_version`, `drift_metrics_table_name`, `profile_metrics_table_name`, `status`, `dashboard_id`

alexott (commented):
In general looks good - we need to decide about readiness waiting - should we wait for new Go SDK, or implement it later

catalog/resource_lakehouse_monitor.go
aravind-segu: The struct we are using here is the `MonitorInfo` and not `CreateMonitor`. `MonitorInfo` has additional fields such as `drift_metrics_table_name`, `profile_metrics_table_name` etc that are not present in `CreateSchema`. Therefore if we use `CreateMonitor`, the GET request is not translating all the fields correctly. 

In the `MonitorInfo` struct, assets_dir, output_schema_name and table_name are not set to `omitempty`. So this forces them to be required.

catalog/resource_lakehouse_monitor.go
aravind-segu: We are using `MonitorInfo` as the base struct and this is part of that struct

catalog/resource_lakehouse_monitor.go
aravind-segu: Yes this is the `FullName` which is a path parameter. The name here is the same as the table name that the user wants to attach the monitor to. Therefore they do know this ahead of time.

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,137 @@
...
+* `skip_builtin_dashboard` - Whether to skip creating a default dashboard summarizing data quality metrics.
+* `slicing_exprs` - List of column expressions to slice data with for targeted analysis. The data is grouped by each expression independently, resulting in a separate slice for each predicate and its complements. For high-cardinality columns, only the top 100 unique values by frequency will generate slices.
+* `warehouse_id` - Optional argument to specify the warehouse for dashboard creation. If not specified, the first running warehouse will be used.
+
+### Computed Fields
+* `monitor_version` - The version of the monitor config (e.g. 1,2,3). If negative, the monitor may be corrupted
+* `drift_metrics_table_name` - The full name of the drift metrics table. Format: __catalog_name__.__schema_name__.__table_name__.
+* `profile_metrics_table_name` - The full name of the profile metrics table. Format: __catalog_name__.__schema_name__.__table_name__.
+* `status` - Status of the Monitor 
+* `dashboard_id` - The ID of the generated dashboard.
```
alexott, line 123: small nit - it's better to move it to the `Attribute Reference` section

alexott (approved):
in general good, pending decision on if we should wait for OpenAPI spec changes for wait command

aravind-segu:
> in general good, pending decision on if we should wait for OpenAPI spec changes for wait command

I checked with Miles, and he is ok to push this for now. I will wait for his approval as well

catalog/resource_lakehouse_monitor.go
mgyucht: Thanks for clarifying. This is fine then.

catalog/resource_lakehouse_monitor.go
mgyucht: Alright. Sounds reasonable to me.

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,137 @@
...
+# databricks_lakehouse_monitor Resource
+
+This resource allows you to manage [Lakehouse Monitors](https://docs.databricks.com/en/lakehouse-monitoring/index.html) in Databricks. 
+
+A `databricks_lakehouse_monitor` is attached to a [databricks_sql_table](sql_table.md) and can be of type timeseries, snapshot or inference. 
+
+## Example Usage
+
+```hcl
+
```
mgyucht, line 13:
```suggestion
```

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,137 @@
...
+resource "databricks_schema" "things" {
+    catalog_name = databricks_catalog.sandbox.id
+    name         = "things"
+    comment      = "this database is managed by terraform"
+    properties = {
+      kind = "various"
+    }
+  }
+  
+resource "databricks_table" "myTestTable" {
```
mgyucht, line 31:
```suggestion
resource "databricks_sql_table" "myTestTable" {
```

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,137 @@
...
+    name = "bar"
+    table_type = "MANAGED"
+    data_source_format = "DELTA"
+
+    column {
+        name      = "timestamp"
+        position  = 1
+        type_name = "INT"
+        type_text = "int"
+        type_json = "{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}"
```
mgyucht, line 43: Is this right? I think this has extra information.
```suggestion
        type = "int"
```

mgyucht (approved):
LGTM, provided you address @alexott's comment to move computed fields under the Attribute Reference section of the doc.

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,135 @@
+---
+subcategory: "Unity Catalog"
+---
+# databricks_lakehouse_monitor Resource
+
+This resource allows you to manage [Lakehouse Monitors](https://docs.databricks.com/en/lakehouse-monitoring/index.html) in Databricks. 
+
+A `databricks_lakehouse_monitor` is attached to a [databricks_sql_table](sql_table.md) and can be of type timeseries, snapshot or inference.
```
arpitjasa-db, line 8: It can be attached to any UC table right, not just this specific SQL table which is a workaround to initialize tables via Terraform? Maybe we should call out that it can be used with the table data source as well

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,135 @@
...
+        name      = "timestamp"
+        position  = 1
+        type_name = "INT"
+        type_text = "int"
+        type_json = "{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}"
+    }
+}
+
+resource "databricks_lakehouse_monitor" "testTimeseriesMonitor" {
+    table_name = "${databricks_catalog.sandbox.name}.${databricks_schema.things.name}.${databricks_table.myTestTable.name}"
```
arpitjasa-db, line 48: Hmm I wonder if this is a bit too much, we ended up separating this in a UC Model into 3 parameters for this reason: https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/registered_model

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,135 @@
...
+    } 
+}
+```
+### Snapshot Monitor
+```hcl
+resource "databricks_lakehouse_monitor" "testMonitorInference" {
+    table_name = "${databricks_catalog.sandbox.name}.${databricks_schema.things.name}.${databricks_table.myTestTable.name}"
+    assets_dir = "/Shared/provider-test/databricks_lakehouse_monitoring/${databricks_table.myTestTable.name}"
+    output_schema_name = "${databricks_catalog.sandbox.name}.${databricks_schema.things.name}"
+    snapshot  {}
```
arpitjasa-db, line 80: Looks like a bit weird syntax, would it be better to expose this as a boolean instead and then convert in the resource implementation?

docs/resources/lakehouse_monitor.md
```diff
@@ -0,0 +1,135 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `table_name` - (Required) - The full name of the table to attach the monitor too. Its of the format {catalog}.{schema}.{tableName}
+* `assets_dir` - (Required) - The directory to store the monitoring assets (Eg. Dashboard and Metric Tables)
+* `output_schema_name` - (Required) - Schema where output metric tables are created
```
arpitjasa-db, line 90: Let's clarify that it needs to be catalog.schema

docs/resources/lakehouse_monitor.md
```diff
@@ -38,15 +37,13 @@ resource "databricks_table" "myTestTable" {
...
-        type_text = "int"
-        type_json = "{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}"
+        type = "int"
     }
 }
 
 resource "databricks_lakehouse_monitor" "testTimeseriesMonitor" {
-    table_name = "${databricks_catalog.sandbox.name}.${databricks_schema.things.name}.${databricks_table.myTestTable.name}"
-    assets_dir = "/Shared/provider-test/databricks_lakehouse_monitoring/${databricks_table.myTestTable.name}"
+    table_name = "${databricks_catalog.sandbox.name}.${databricks_schema.things.name}.${databricks_sql_table.myTestTable.name}"
```
alexott, line 45: This could be simplified to `databricks_sql_table.myTestTable.id` now: https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/sql_table#id

alexott (approved):

docs/resources/lakehouse_monitor.md
aravind-segu: According to their python api docs (https://api-docs.databricks.com/python/lakehouse-monitoring/latest/databricks.lakehouse_monitoring.html#databricks.lakehouse_monitoring.create_monitor) , it can be `{catalog}.{schema}.{table}` or `{schema}.{table}` or `{table}` and the api fills in the current catalog or schema. So we technically dont need to split it up and force users to fill in all three fields. I think this should be the recommendation to fill in all three, but if the user reads other docs, they should be able to just use all three of the options

arpitjasa-db: Got it, so if someone doesn't specify schema and catalog via Terraform though, what would be the "current" catalog and schema it would infer?

docs/resources/lakehouse_monitor.md
aravind-segu: We debated about this. We changed the Go SDK to use an empty struct in place of `any`. This is also the the example expected in the Go SDK, so thought it would be clear. Miles also did not want too many changes between Go SDK structs and the terraform input structs. 

Also in the future if there are any snapshot relevant parameters the team introduces, we dont need additional changes in the terraform provider as we are already using the struct.

docs/resources/lakehouse_monitor.md
alexott: No, in for `databricks_sql_table` you always specify catalog & schema and `id` will be three-level name

samuhepp:
Hey! Will this be included in the next release? If so, when is the approximate target date for that? Keen to use this instead of the notebook approach.

PR #3241, opened 2024-02-09
Exporter: rework handling of listings and interactive prompting

alexott:
## Changes

When configuring services to list, the old code was relying on the substring match, and this lead to problems when we have overlapping matches, like, `uc` service matched `uc-allowlist`, etc.  This PR changed to use a map to store services to list so we don't have a problem with overlapping services.

Changes also were made to the interactive prompting - instead of asking for each resource type separately, we now ask about service and show which resources belong to that service.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/command.go
+++ b/exporter/command.go
@@ -6,12 +6,14 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


nkvuong (approved):
lgtm

PR #3246, opened 2024-02-10
Explicitly set securable field when reading `databricks_grants` or `databricks_grant`

### LINKED ISSUE #3245: [ISSUE] `databricks_grants` and `databricks_grant` resources aren't imported correctly
The read operation of `databricks_grants` and `databricks_grant` doesn't fill fields like `catalog`, `schema`, etc. and even import is successful, then doing plan shows that grants will be replaced because corresponding field isn't filled correctly:

```

  # databricks_grants.catalog_cybersecurity must be replaced
-/+ resource "databricks_grants" "catalog_cybersecurity" {
      + catalog = "cybersecurity" # forces replacement
      ~ id      = "catalog/cybersecurity" -> (known after apply)

        # (1 unchanged block hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

This also prevents exporter from being able to generate corresponding fields

alexott:
## Changes

The read method of corresponding resources was reading only a list of permissions, but it didn't set the securable field, so when doing the import, that field was empty and this lead to the recreation of the resource after import.  This PR explicitly sets that field on read.

This fixes #3245

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_grant.go
+++ b/catalog/resource_grant.go
@@ -184,7 +184,11 @@ func ResourceGrant() common.Resource {
 			if err != nil {
 				return err
 			}
-			return common.StructToData(*grantsForPrincipal, s, d)
+			err = common.StructToData(*grantsForPrincipal, s, d)
+			if err != nil {
+				return err
+			}
+			return d.Set(securable, name)
 		},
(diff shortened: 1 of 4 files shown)


nkvuong (approved):
lgtm

Issue #3019, opened 2023-12-11
[ISSUE] Issue with `databricks_sql_table` resource. When we pass SQL using a here-document, differences appear even though there have been no changes.


case-k-git:
issue with `databricks_sql_table` resource. When we pass SQL using a here-document, differences appear even though there have been no changes. if we does not pass by here-document it will not detect the changes.

To add a view in Terraform without modifying the tf file. The SQL file used for the view is managed as a separate file(.sql) and passed as follows configuration. At this time, the SQL file is passed as a here-document and is detected as a change even though it has not been modified. I want to resolve this issue as it decreases deployment performance.

### Configuration

```hcl
# Copy-paste your Terraform configuration here
resource "databricks_sql_table" "views" {
  for_each = local.sql-association-map-read-from-json

  catalog_name    = each.value.catalog_name
  name            = each.value.name
  schema_name     = each.value.schema_name
  table_type      = "VIEW"
  view_definition =  each.value.sql
}
```

### Expected Behavior

Not detected as a difference in 'terraform plan'. Because not changing SQL

### Actual Behavior

 differences appear even though there have been no changes. if we does not pass by here-document it will not detect the changes

here-document
```
  # databricks_sql_table.views["table"] will be updated in-place
  ~ resource "databricks_sql_table" "views" {
        id              = "catalog.schema.table"
        name            = "table"
      ~ view_definition = <<-EOT
            SELECT * FROM catalog.schema.table
        EOT
        # (4 unchanged attributes hidden)

        # (10 unchanged blocks hidden)
    }
```

### Terraform and provider versions

- terraform 1.5.2
- databricks 1.29.0
### Debug Output

```
  # databricks_sql_table.views["table"] will be updated in-place
  ~ resource "databricks_sql_table" "views" {
        id              = "catalog.schema.table"
        name            = "table"
      ~ view_definition = <<-EOT
            SELECT * FROM catalog.schema.table
        EOT
        # (4 unchanged attributes hidden)

        # (10 unchanged blocks hidden)
    }
```

case-k-git:
May be related to this issue
https://github.com/hashicorp/terraform-provider-aws/issues/9042

anineharto:
I'm also experiencing the same issue ➕

case-k-git:
I have found workaround using trimspace can avoid this issue. 

```
resource "databricks_sql_table" "views" {
  for_each = local.sql-association-map-read-from-json

[...]

PR #3250, opened 2024-02-13
Fix ViewDefinition Unexpected Diff Detection When Created By Here-document

case-k-git:
## Changes

trim space from view definition not to detect change when created by here-document

fixed this issue
https://github.com/databricks/terraform-provider-databricks/issues/3019

issue with `databricks_sql_table` resource. When we pass SQL using a here-document, differences appear even though there have been no changes. if we does not pass by here-document it will not detect the changes.

To add a view in Terraform without modifying the tf file. The SQL file used for the view is managed as a separate file(.sql) and passed as follows configuration. At this time, the SQL file is passed as a here-document and is detected as a change even though it has not been modified. I want to resolve this issue as it decreases deployment performance.

### Configuration

```hcl
resource "databricks_sql_table" "views" {
  for_each = local.sql-association-map-read-from-json

  catalog_name    = each.value.catalog_name
  name            = each.value.name
  schema_name     = each.value.schema_name
  table_type      = "VIEW"
  view_definition =  each.value.sql
}
```
There are work around to avoid this using terraform `trimspace` function. This PR fix this issue without using `trimspace` by terraform user side.
```
resource "databricks_sql_table" "views" {
  for_each = local.sql-association-map-read-from-json

  catalog_name    = each.value.catalog_name
  name            = each.value.name
  schema_name     = each.value.schema_name
  table_type      = "VIEW"
  view_definition =  trimspace(each.value.sql)
}
```
## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -299,7 +299,7 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
 
 	if ti.TableType == "VIEW" {
 		// View only attributes
-		if ti.ViewDefinition != oldti.ViewDefinition {
+		if strings.TrimSpace(ti.ViewDefinition) != strings.TrimSpace(oldti.ViewDefinition) {
 			statements = append(statements, fmt.Sprintf("ALTER VIEW %s AS %s", ti.SQLFullName(), ti.ViewDefinition))
 		}
 	} else {
--- a/catalog/resource_sql_table_test.go
+++ b/catalog/resource_sql_table_test.go
@@ -423,6 +423,64 @@ func TestResourceSqlTableUpdateView(t *testing.T) {
 	assert.Equal(t, "bar", d.Get("name"))
 }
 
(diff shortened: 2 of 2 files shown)


catalog/resource_sql_table.go
```diff
@@ -299,7 +299,7 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
 
 	if ti.TableType == "VIEW" {
 		// View only attributes
-		if ti.ViewDefinition != oldti.ViewDefinition {
+		if strings.TrimSpace(ti.ViewDefinition) != strings.TrimSpace(oldti.ViewDefinition) {
```
alexott, line 302: This shouldn't fix the drift, it will just detect the changes later when the update function is invoked. I've merged #3251 which fixes the configuration drift.

case-k-git: I see. Thank you for your comment and fix the view creation issue.

PR #3252, opened 2024-02-13
[JOBS-16324] Terraform support for Foreach tasks (private preview)

geekswaroop:
## Changes
Add terraform support for foreach task (going private preview soon). 

Terraform does not support cyclic dependencies, hence we need to limit the nesting depth manually. Implemented using separate structs for the time being. 

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -116,6 +116,7 @@ This block describes individual tasks:
   * `spark_python_task`
   * `spark_submit_task`
   * `sql_task`
+  * `for_each_task`
 * `library` - (Optional) (Set) An optional list of libraries to be installed on the cluster that will execute the job. Please consult [libraries section](cluster.md#libraries) for [databricks_cluster](cluster.md) resource.
 * `depends_on` - (Optional) block specifying dependency(-ies) for a given task.
 * `run_if` - (Optional) An optional value indicating the condition that determines whether the task should be run once its dependencies have been completed. When omitted, defaults to `ALL_SUCCESS`.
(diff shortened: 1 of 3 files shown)


jobs/resource_job.go
```diff
@@ -118,6 +118,46 @@ type RunJobTask struct {
 	JobParameters map[string]string `json:"job_parameters,omitempty"`
 }
 
+// TODO: migrate to go sdk
```
hectorcast-db, line 121: nit, add a note stating that TF does not support recursive nesting and that we need to limit the depth. Maybe add a link to an example:

https://github.com/hashicorp/terraform-provider-aws/blob/b4a9f93a2b7323202c8904e86cff03d3f2cb006b/internal/service/wafv2/rule_group.go#L110

hectorcast-db (approved):

docs/resources/job.md
```diff
@@ -346,6 +347,12 @@ The `condition_task` specifies a condition with an outcome that can be used to c
 
 This task does not require a cluster to execute and does not support retries or notifications.
 
+### for_each_task Configuration Block
+
+* `concurrency` - (Optional) Controls the number of active iteration task runs. Default is 100 (maximal value).
```
gaborratky-db, line 352: @geekswaroop we changed the default to 20. It's OK if it is only fixed in the next regular Terraform release.

geekswaroop: I think the doc [fix](https://github.com/databricks/terraform-provider-databricks/pull/3271) is part of this release.

gaborratky-db: Awesome, well done!

PR #3265, opened 2024-02-15
Add `databricks_file` resource

hectorcast-db:
## Changes
Add `databricks_file` resource

## Tests

- [X] `make test` run locally
- [ ] (running) relevant acceptance tests are passing
- [X] using Go SDK

### DIFF
--- a/docs/resources/file.md
+++ b/docs/resources/file.md
@@ -28,25 +28,11 @@ resource "databricks_schema" "things" {
   }
 }
 
-resource "databricks_storage_credential" "external" {
-  name = "creds"
-  aws_iam_role {
-    role_arn = aws_iam_role.external_data_access.arn
-  }
-}
-
-resource "databricks_external_location" "some" {
-  name            = "external-location"
-  url             = "s3://${aws_s3_bucket.external.id}/some"
-  credential_name = databricks_storage_credential.external.name
-}
-
 resource "databricks_volume" "this" {
   name             = "quickstart_volume"
   catalog_name     = databricks_catalog.sandbox.name
   schema_name      = databricks_schema.things.name
-  volume_type      = "EXTERNAL"
-  storage_location = databricks_external_location.some.url
+  volume_type      = "MANAGED"
   comment          = "this volume is managed by terraform"
 }
 
--- a/docs/resources/file.md
+++ b/docs/resources/file.md
@@ -1,0 +1,104 @@
+---
+subcategory: "Storage"
+---
+# databricks_file Resource
+
+This resource allows uploading and downloading files in [databricks_volume](volume.md) up to 2GB in octet-stream.
+
+## Example Usage
+
+In order to manage a file on Unity Catalog Volumes with Terraform, you must specify the `source` attribute containing the full path to the file on the local filesystem.
+
+```hcl
+resource "databricks_catalog" "sandbox" {
+  metastore_id = databricks_metastore.this.id
+  name         = "sandbox"
+  comment      = "this catalog is managed by terraform"
+  properties = {
+    purpose = "testing"
+  }
+}
+
+resource "databricks_schema" "things" {
+  catalog_name = databricks_catalog.sandbox.name
+  name         = "things"
+  comment      = "this schema is managed by terraform"
+  properties = {
+    kind = "various"
+  }
+}
+
+resource "databricks_storage_credential" "external" {
+  name = "creds"
+  aws_iam_role {
+    role_arn = aws_iam_role.external_data_access.arn
+  }
+}
+
+resource "databricks_external_location" "some" {
+  name            = "external-location"
+  url             = "s3://${aws_s3_bucket.external.id}/some"
+  credential_name = databricks_storage_credential.external.name
+}
+
+resource "databricks_volume" "this" {
+  name             = "quickstart_volume"
+  catalog_name     = databricks_catalog.sandbox.name
+  schema_name      = databricks_schema.things.name
+  volume_type      = "EXTERNAL"
+  storage_location = databricks_external_location.some.url
+  comment          = "this volume is managed by terraform"
+}
+
+resource "databricks_file" "this" {
+  source = "/full/path/on/local/system"
+  path   = "/Volumes/${databricks_volume.this.catalog_name}/${databricks_volume.this.schema_name}/{databricks_volume.this.name}/fileName"
+}
+```
+
+You can also inline sources through `content_base64`  attribute.
+
+```hcl
+resource "databricks_file" "init_script" {
+  content_base64 = base64encode(<<-EOT
+    #!/bin/bash
+    echo "Hello World"
+    EOT
+  )
+  path   = "/Volumes/${databricks_volume.this.catalog_name}/${databricks_volume.this.schema_name}/{databricks_volume.this.name}/fileName"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `source` - The full absolute path to the file. Conflicts with `content_base64`.
+* `content_base64` - Contents in base 64 format. Conflicts with `source`.
+* `path` - The path of the file in which you wish to save. Should start with `/Volumes`
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - Same as `path`.
+* `file_size` - The file size of the file that is being tracked by this resource in bytes.
+* `modification_time` - The last time stamp when the file was modified
+
+
+## Import
+
+The resource `databricks_file` can be imported using the path of the file:
+
+```bash
+$ terraform import databricks_file.this <path>
+```
+
+## Related Resources
+
+The following resources are often used in the same context:
+
+* [databricks_workspace_file](./workspace_file.md)
+* [End to end workspace management](../guides/workspace-management.md) guide.
+* [databricks_dbfs_file](../data-sources/dbfs_file.md) data to get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_dbfs_file_paths](../data-sources/dbfs_file_paths.md) data to get list of file names from get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
--- a/docs/resources/file.md
+++ b/docs/resources/file.md
@@ -3,7 +3,11 @@ subcategory: "Storage"
 ---
 # databricks_file Resource
 
-This resource allows uploading and downloading files in [databricks_volume](volume.md) up to 5GiB in octet-stream.
+This resource allows uploading and downloading files in [databricks_volume](volume.md). 
+
+Notes: 
+* Currently the limit is 5GiB in octet-stream.
+* Currently, only UC volumes are supported. The list of destinations may change.
 
 ## Example Usage
 
@@ -38,7 +42,7 @@ resource "databricks_volume" "this" {
 
 resource "databricks_file" "this" {
   source = "/full/path/on/local/system"
-  path   = "/Volumes/${databricks_volume.this.catalog_name}/${databricks_volume.this.schema_name}/{databricks_volume.this.name}/fileName"
+  path   = "${databricks_volume.this.volume_path}/fileName"
 }
 ```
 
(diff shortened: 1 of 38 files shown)


docs/resources/file.md
```diff
@@ -0,0 +1,104 @@
+---
+subcategory: "Storage"
+---
+# databricks_file Resource
+
+This resource allows uploading and downloading files in [databricks_volume](volume.md) up to 2GB in octet-stream.
```
bonnetn, line 6:
```suggestion
This resource allows uploading and downloading files in [databricks_volume](volume.md) up to 5GiB in octet-stream.
```

bonnetn, line 6: Eventually this should support more than just uploading to volumes.
Should we update the doc now?

docs/resources/file.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `source` - The full absolute path to the file. Conflicts with `content_base64`.
+* `content_base64` - Contents in base 64 format. Conflicts with `source`.
+* `path` - The path of the file in which you wish to save. Should start with `/Volumes`
```
bonnetn, line 78: Same, eventually it may not need to start with /Volumes.

@seansmith-db do you know how we should refer to single path namespace in public docs?

hectorcast-db: Right now it's only volumes, right? I would leave it like this until other are supported to avoid customer confusion.

seansmith-db: Let's say "For example, `/Volumes/main/default/volume1/file.txt`" instead of trying to specify the full set of allowed paths here. We don't have a better reference at this time.

hectorcast-db: Updated

alexott:
I'm not sure about the name - it's too generic if it allows to work only with Volumes... Or do we plan to use the same resource for all file types - DBFS, WSFS, etc.? But documentation mentions only Volumes, and if it's true, then it should be called like `databricks_volume_file`, not generic `databricks_file`

hectorcast-db:
@alexott  For now, it only supports Volumes, but they plan to expand to other resources. We are discussing possible changes to the docs.

storage/resource_file.go
```diff
@@ -0,0 +1,135 @@
...
+			if err != nil {
+				return err
+			}
+			data.Set("modification_time", metadata.LastModified)
+			data.Set("file_size", metadata.ContentLength)
+			data.Set("has_been_modified", false)
+			data.SetId(path)
+			return nil
+		},
+		Read: func(ctx context.Context, data *schema.ResourceData, c *common.DatabricksClient) error {
```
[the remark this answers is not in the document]
hectorcast-db: Correct. Read is used by TF to determine if an update is needed. It is not used by users to actually retrieve the data. By only retrieving metadata, we can safe a lot of bandwith for both users and servers.

edwardfeng-db (approved):
For the naming part, might be good to add in documentation explaining why it is called databricks_file now and `volume` only is temporary

docs/resources/file.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `source` - The full absolute path to the file. Conflicts with `content_base64`.
+* `content_base64` - Contents in base 64 format. Conflicts with `source`.
+* `path` - The path of the file in which you wish to save. For example, g`/Volumes/main/default/volume1/file.txt`.
```
dstrodtman-db, line 78:
```suggestion
* `path` - The path of the file in which you wish to save. For example, `/Volumes/main/default/volume1/file.txt`.
```

Typo or is that formatting?

docs/resources/file.md
```diff
@@ -0,0 +1,104 @@
...
+  url             = "s3://${aws_s3_bucket.external.id}/some"
+  credential_name = databricks_storage_credential.external.name
+}
+
+resource "databricks_volume" "this" {
+  name             = "quickstart_volume"
+  catalog_name     = databricks_catalog.sandbox.name
+  schema_name      = databricks_schema.things.name
+  volume_type      = "EXTERNAL"
+  storage_location = databricks_external_location.some.url
```
adrianaisp, line 49: I would not use the term "external_location" here since it has a very specific meaning in Dtabricks, which is different from Volume

docs/resources/file.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+## Related Resources
+
+The following resources are often used in the same context:
+
+* [databricks_workspace_file](./workspace_file.md)
+* [End to end workspace management](../guides/workspace-management.md) guide.
+* [databricks_dbfs_file](../data-sources/dbfs_file.md) data to get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
+* [databricks_dbfs_file_paths](../data-sources/dbfs_file_paths.md) data to get list of file names from get file content from [Databricks File System (DBFS)](https://docs.databricks.com/data/databricks-file-system.html).
```
adrianaisp, line 104: We should rather add here links to Volumes docs, like https://docs.databricks.com/en/files/index.html or https://docs.databricks.com/en/connect/unity-catalog/volumes.html

I would not add references to DBFS in the current state of the docs

internal/acceptance/file_test.go
```diff
@@ -0,0 +1,478 @@
...
+			credential_name = databricks_storage_credential.external.id
+			comment         = "Managed by TF"
+		}
+
+		resource "databricks_volume" "this" {
+			name = "name-abc"
+			comment = "comment-abc"
+			catalog_name = "main"
+			schema_name = databricks_schema.this.name 
+			volume_type = "EXTERNAL"
```
adrianaisp, line 245: Same comment about EXTERNAL ....

internal/acceptance/file_test.go
```diff
@@ -0,0 +1,478 @@
...
```
adrianaisp, line 302: Same comment about EXTERNAL ....

internal/acceptance/file_test.go
```diff
@@ -0,0 +1,478 @@
...
```
adrianaisp, line 360: Same comment about EXTERNAL ....

internal/acceptance/file_test.go
```diff
@@ -0,0 +1,478 @@
...
```
adrianaisp, line 434: Same comment about EXTERNAL ....

internal/acceptance/file_test.go
```diff
@@ -0,0 +1,478 @@
...
```
adrianaisp, line 469: Same comment about EXTERNAL ....

docs/resources/file.md
bogdanrdc: external has to be used here. it references the previously created `databricks_external_location` resource, which is named `some`.

docs/resources/file.md
```diff
@@ -0,0 +1,104 @@
...
+  name            = "external-location"
+  url             = "s3://${aws_s3_bucket.external.id}/some"
+  credential_name = databricks_storage_credential.external.name
+}
+
+resource "databricks_volume" "this" {
+  name             = "quickstart_volume"
+  catalog_name     = databricks_catalog.sandbox.name
+  schema_name      = databricks_schema.things.name
+  volume_type      = "EXTERNAL"
```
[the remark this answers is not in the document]
adrian-ionescu: Makes sense, but can we show an example with a MANAGED volume instead? It's simpler and I don't see any downsides?

docs/resources/file.md
```diff
@@ -0,0 +1,89 @@
+---
+subcategory: "Storage"
+---
+# databricks_file Resource
+
+This resource allows uploading and downloading files in [databricks_volume](volume.md) up to 5GiB in octet-stream.
```
alexott, line 6: I would put `up to 5GiB in octet-stream` as a Note as the size may change over time.

alexott: Also, add a note that UC Volumes is only current implementation and that the list of supported destinations may change

storage/resource_file.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			path := data.Id()
+			metadata, err := w.Files.GetMetadata(ctx, files.GetMetadataRequest{FilePath: path})
+			if err != nil {
+				return err
+			}
+			storedModificationTime := data.Get("modification_time").(string)
+
+			data.Set("has_been_modified", storedModificationTime != metadata.LastModified)
```
alexott, line 95: But it will detect local updates?

docs/resources/file.md
```diff
@@ -0,0 +1,89 @@
...
+  name             = "quickstart_volume"
+  catalog_name     = databricks_catalog.sandbox.name
+  schema_name      = databricks_schema.things.name
+  volume_type      = "MANAGED"
+  comment          = "this volume is managed by terraform"
+}
+
+resource "databricks_file" "this" {
+  source = "/full/path/on/local/system"
+  path   = "/Volumes/${databricks_volume.this.catalog_name}/${databricks_volume.this.schema_name}/{databricks_volume.this.name}/fileName"
```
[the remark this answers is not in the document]
alexott: @hectorcast-db Can you rebase on the latest `main` so you can use the `volume_path` attribute instead?

storage/resource_file.go
hectorcast-db: Yes, we extend `FileContentSchema` here:
https://github.com/databricks/terraform-provider-databricks/pull/3265/files#diff-efed3c373bbceb3e2cb308c29e6871991a5e7d1b42dbe605f77ffc7afaee183dR41
which is already configured to detect local changes.

hectorcast-db: I renamed `has_been_modified` to `remote_file_modified` to make it more clear in the diff shown to the user what this flag means. For local changes, we have the `md5` provided by the link before.

storage/resource_file.go
```diff
@@ -0,0 +1,135 @@
...
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"file_size": {
+			Type:     schema.TypeInt,
+			Computed: true,
+		},
+		"has_been_modified": {
+			Type:     schema.TypeBool,
+			Optional: true,
```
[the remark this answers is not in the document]
hectorcast-db: Computed fields cannot be used for the diff. We do the same for the MDE5
https://github.com/databricks/terraform-provider-databricks/blob/main/workspace/file_resource.go#L88-L91

storage/resource_file.go
```diff
@@ -0,0 +1,135 @@
...
+			path := data.Id()
+			metadata, err := w.Files.GetMetadata(ctx, files.GetMetadataRequest{FilePath: path})
+			if err != nil {
+				return err
+			}
+			storedModificationTime := data.Get("modification_time").(string)
+
+			data.Set("remote_file_modified", storedModificationTime != metadata.LastModified)
+
+			// Do not store here the modification time. If the update fails, we will keep the wrong one in the state.
```
tanmay-db, line 106: if update fails we can rollback the modification time back right? i.e. try update and if error then update the state with previous value.

tanmay-db (approved):
Overall LGTM, left one comment.

workspace/file_resource.go
```diff
@@ -146,6 +146,15 @@ func FileContentSchema(extra map[string]*schema.Schema) map[string]*schema.Schem
 						},
 					}
 				}
+				if strings.HasPrefix(v, "Volume") {
```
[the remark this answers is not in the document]
tanmay-db: People can specify "Volume..." by mistake which isn't a valid path. We fail early in that case by having this check.

mgyucht (approved):
Great, thanks! LGTM provided integration tests pass.

docs/resources/file.md
```diff
@@ -0,0 +1,93 @@
+---
+subcategory: "Storage"
+---
+# databricks_file Resource
+
+This resource allows uploading and downloading files in [databricks_volume](volume.md). 
+
+Notes: 
+* Currently the limit is 5GiB in octet-stream.
+* Currently, only UC volumes are supported. The list of destinations may change.
```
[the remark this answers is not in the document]
alexott: We can do it later

PR #3272, opened 2024-02-19
Add computed `volume_path` attribute to `databricks_volume` resource

alexott:
## Changes

This attribute will expose path in form of `/Volumes/<catalog>/<schema>/<name>` for use with other resources, such as, `databricks_file`.  Plus it's necessary for Terraform Exporter to detect references between resources.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_volume.go
+++ b/catalog/resource_volume.go
@@ -42,6 +42,11 @@ func ResourceVolume() common.Resource {
 	s := common.StructToSchema(VolumeInfo{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			m["storage_location"].DiffSuppressFunc = ucDirectoryPathSlashAndEmptySuppressDiff
+			m["volume_path"] = &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+			}
 			return m
 		})
 	return common.Resource{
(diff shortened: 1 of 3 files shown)


catalog/resource_volume.go
```diff
@@ -82,7 +87,11 @@ func ResourceVolume() common.Resource {
 			if err != nil {
 				return err
 			}
-			return common.StructToData(v, s, d)
+			err = common.StructToData(v, s, d)
+			if err != nil {
+				return err
+			}
+			return d.Set("volume_path", "/Volumes/"+strings.ReplaceAll(v.FullName, ".", "/"))
```
nkvuong, line 94: we also need to update this in Update, as volume name could change

alexott: done, thank you!

nkvuong (approved):

catalog/resource_volume.go
```diff
@@ -42,6 +42,11 @@ func ResourceVolume() common.Resource {
 	s := common.StructToSchema(VolumeInfo{},
 		func(m map[string]*schema.Schema) map[string]*schema.Schema {
 			m["storage_location"].DiffSuppressFunc = ucDirectoryPathSlashAndEmptySuppressDiff
+			m["volume_path"] = &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
mgyucht, line 48: Only question: if this is optional, we shouldn't set this if a user specifies it. I think it probably shouldn't be optional, right?

mgyucht (commented):
One question, but otherwise LGTM.

catalog/resource_volume.go
alexott: yeah, that's correct - it was more kind of copy/paste - we have that code in multiple places. Corrected to be computed only

PR #3273, opened 2024-02-19
fix: properly propagate auth_type to the databricks client

guillaumelecerf:
## Changes
We want to use basic creds for databricks, while using GOOGLE_CREDENTIALS for another provider.
Specifying `auth_type = "basic"` in the provider configuration did not resolve the `more than one authorization method configured` error, only setting `DATABRICKS_AUTH_TYPE=basic` envvar did.

After analysis, it looks like the databricks-sdk-go client is not receiving the auth_type value.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


thomasisaacsibille (approved):

mgyucht (approved):
Thanks, nice catch!

PR #3278, opened 2024-02-20
Use stable IDs for settings resources

### LINKED ISSUE #3270: [ISSUE] Problems with the settings resources generation
Right now, there are a few problems with the settings resources, such as `databricks_default_namespace_setting`:

1. Because ETag is used as a resource ID, it's not possible to import the resource easily - we need to have predictable ID for each of the settings
2. When ETag is changing on update, the resource ID is also updated, leading to warnings from the Terraform, like this:

```
2024-02-16T11:10:43.113+0100 [WARN]  Provider "provider[\"registry.terraform.io/databricks/databricks\"]" produced an unexpected new value for databricks_default_namespace_setting.this, but we are tolerating it because it is using the legacy plugin SDK.
      - .etag: was cty.StringVal("ACi1L_0ASHABAAosCh0KCXdvcmtzcGFjZRIQNjY4ODgwODEzMDU2MjMxNxIJbmFtZXNwYWNlGAQBAAA="), but now cty.StringVal("ACi1L_0ASHABAAosCh0KCXdvcmtzcGFjZRIQNjY4ODgwODEzMDU2MjMxNxIJbmFtZXNwYWNlGAUBAAA=")
      - .id: was cty.StringVal("ACi1L_0ASHABAAosCh0KCXdvcmtzcGFjZRIQNjY4ODgwODEzMDU2MjMxNxIJbmFtZXNwYWNlGAQBAAA="), but now cty.StringVal("ACi1L_0ASHABAAosCh0KCXdvcmtzcGFjZRIQNjY4ODgwODEzMDU2MjMxNxIJbmFtZXNwYWNlGAUBAAA=")
```

Proposal: 
* Generate predictable resource ID, like, `setting_<name>` or `settings|name`, or just a constant, like we do for some resources, like [databricks_sql_global_config](https://github.com/databricks/terraform-provider-databricks/blob/main/sql/resource_sql_global_config.go#L127)

alexott:
## Changes

Before this change, generated settings resource were using `etag` value as resource ID. This lead to a situation that importing of these resources was problematic because ID was changing all the time.

This PR introduces stable IDs for generated resources. By default it uses a constant ID: `global`, similar to other resources, like, `databricks_sql_global_config`, but the resource implementor may provide a function that will generate a resource ID from the settings' data.

This fixes #3270

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/settings/generic_setting.go
+++ b/settings/generic_setting.go
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	defaultSettingName = "global"
+	defaultSettingId = "global"
 )
 
 func retryOnEtagError[Req, Resp any](f func(req Req) (Resp, error), firstReq Req, updateReq func(req *Req, newEtag string), retriableErrors []error) (Resp, error) {
@@ -133,7 +133,7 @@ func (w workspaceSetting[T]) GetETag(t *T) string {
 }
 
 func (w workspaceSetting[T]) GetId(t *T) string {
-	id := defaultSettingName
+	id := defaultSettingId
 	if w.generateIdFunc != nil {
 		id = w.generateIdFunc(t)
 	}
@@ -189,7 +189,7 @@ func (w accountSetting[T]) SetETag(t *T, newEtag string) {
 }
 
 func (w accountSetting[T]) GetId(t *T) string {
-	id := defaultSettingName
+	id := defaultSettingId
 	if w.generateIdFunc != nil {
 		id = w.generateIdFunc(t)
 	}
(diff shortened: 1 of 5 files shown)


settings/generic_setting.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var (
+	defaultSettingName = "global"
```
hectorcast-db, line 16: nit, I would call it `defaultSettingId`. It seems more clear.

hectorcast-db (approved):

PR #3279, opened 2024-02-20
Use workspace-level default for `enable_serverless_compute` if not specified in `databricks_sql_global_config`

### LINKED ISSUE #3009: [ISSUE] enable_serverless_compute needs to be 'true' by default
### Expected Behavior
By default `enable_serverless_compute` needs to be true

Currently by default `enable_serverless_compute` is false. Please set this to true by default

// GlobalConfig used to generate Terraform resource schema and bind to resource data
GoogleServiceAccount    string            `json:"google_service_account,omitempty"`
	EnableServerlessCompute bool              `json:"enable_serverless_compute,omitempty" tf:"default:false"`
	SqlConfigParams         map[string]string `json:"sql_config_params,omitempty"`
}

### Actual Behavior
Customer is getting SQL Serverless connection fails to connect and something changed the flag to serverless enabled = false?

mgyucht:
## Changes
Serverless SQL is enabled by default for workspaces created after March 28, 2022 (https://docs.databricks.com/en/compute/sql-warehouse/serverless.html#enable-serverless-compute-in-your-account). However, the TF resource `databricks_sql_global_config` _disables_ serverless SQL if not specified. This is a confusing behavior for many customers and limits serverless adoption.

This change removes the default value for `enable_serverless_compute`. The impact is:
1. No impact for users who explicitly set `enable_serverless_compute` in their Terraform configuration.
2. For users who already use this resource and don't set `enable_serverless_compute`, the existing state will be preserved (i.e. if it was implicitly disabled like with the previous default behavior, it will remain disabled).
3. For users who start using this resource and don't set `enable_serverless_compute`, the resource will initially fetch the value of `enable_serverless_compute` and set it in the state.

This resolves #3009.

This needs to be announced in the release notes for the next Terraform release, but as this field is not documented, there are no docs changes for this.
 
## Tests

- [x] `make test` run locally
- [ ] using Go SDK


sql/resource_sql_global_config.go
```diff
@@ -28,7 +29,7 @@ type GlobalConfig struct {
 	DataAccessConfig        map[string]string `json:"data_access_config,omitempty"`
 	InstanceProfileARN      string            `json:"instance_profile_arn,omitempty"`
 	GoogleServiceAccount    string            `json:"google_service_account,omitempty"`
-	EnableServerlessCompute bool              `json:"enable_serverless_compute,omitempty" tf:"default:false"`
+	EnableServerlessCompute bool              `json:"enable_serverless_compute,omitempty"`
```
alexott, line 32: Should we mark it `tf:computed` to prevent configuration drift?

alexott (commented):
overall looks good. We'll need to put a "behaviour change" section into the changelog...

mgyucht:
This actually fails integration tests because AWS apparently requires `enable_serverless_compute` to be set... following up with the SQL gateway team.

sql/resource_sql_global_config.go
```diff
@@ -28,7 +29,7 @@ type GlobalConfig struct {
 	DataAccessConfig        map[string]string `json:"data_access_config,omitempty"`
 	InstanceProfileARN      string            `json:"instance_profile_arn,omitempty"`
 	GoogleServiceAccount    string            `json:"google_service_account,omitempty"`
-	EnableServerlessCompute bool              `json:"enable_serverless_compute,omitempty" tf:"default:false"`
+	EnableServerlessCompute bool              `json:"enable_serverless_compute,omitempty" tf:"computed,optional"`
```
alexott, line 32: `optional` isn't required here. It's used only when we don't have `omitempty` on `json` tag...

alexott (approved):
lgtm

PR #3290, opened 2024-02-22
Removed `CustomizeDiff` and Client Side Validation for `databricks_grants`

tanmay-db:
## Changes

We remove client side validation for `databricks_grants`

## Tests

Integration tests, updated unit tests, manually tested binary. 
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_grants.go
+++ b/catalog/resource_grants.go
@@ -89,150 +89,11 @@ func replaceAllPermissions(a permissions.UnityCatalogPermissionsAPI, securable s
 	})
 }
 
-type securableMapping map[string]map[string]bool
+type securablesList []catalog.SecurableType
 
 // reuse ResourceDiff and ResourceData
 type attributeGetter interface {
 	Get(key string) any
-}
-
-func (sm securableMapping) kv(d attributeGetter) (string, string) {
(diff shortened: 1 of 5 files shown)


catalog/resource_grants.go
```diff
@@ -104,6 +104,7 @@ func (sm securableMapping) kv(d attributeGetter) (string, string) {
 		}
 		return field, v
 	}
+	fmt.Printf("[WARN] Unexpected resource or permissions. Please proceed at your own risk.")
```
mgyucht, line 107: We should use `tflog.Warn` here rather than printing out information: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-log/tflog#Warn

mgyucht (approved):
Small nit, otherwise LGTM.

catalog/resource_grants.go
tanmay-db: Yes, sorry about that, didn't push the latest commits. I have also removed the map because we can use the one in permissions.

mgyucht (approved):

nkvuong (approved):

PR #3292, opened 2024-02-22
Handle UC Resource Grants with Spaces

ledbutter:
## Changes
In our (private) Terraform module we were trying to configure grants using the `databricks_grants` resource, but kept running into trouble on apply because it detected duplicate grants to add/remove (snippet from our CI/CD pipeline):

> {
"add": [
"CREATE EXTERNAL VOLUME",
"CREATE EXTERNAL TABLE",
"READ FILES",
"WRITE FILES"
],
"principal": "58264056-137c-468c-850f-fff3042498d5",
"remove": [
"CREATE_EXTERNAL_VOLUME",
"READ_FILES",
"CREATE_EXTERNAL_TABLE",
"WRITE_FILES"
 ]
},

This was because we were missing the `_` characters there (notice the format of the privs to add). I've confirmed that the Databricks CLI/API silently accepts this:

```shell
databricks grants update external_location my_loc -p dev --json '{"changes": [{"principal": "user", "add": ["ALL PRIVILEGES"]}]}'
```

So here I add some tests showing the gap, and then added some invocations of the already-existing `validate` method on `CREATE`/`UPDATE`. I certainly am open to other approaches, just wanted to get this out here for discussion.

I'd also be fine with instead updating the "diff" logic here to also look for and replace spaces with underscores, but the existing `validate` tests seem to already account for that, but I'm not sure where--if at all--`validate` is called on the mapping in full system calls.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


catalog/resource_grants_test.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func TestPermissionsCornerCases(t *testing.T) {
-	qa.ResourceCornerCases(t, ResourceGrants(), qa.CornerCaseID("schema/sandbox"))
+	qa.ResourceCornerCases(t, ResourceGrants(), qa.CornerCaseSkipCRUD("Validation prevents CRUD"))
```
ledbutter, line 12: Had to change this since the corner case logic won't provide valid priv values

ledbutter:
Since the underlying "mapping" has gone away since I started on this, I've switched this behavior to match the API and just silently treat `" "` as `_` so that diff'ing can treat `"ALL PRIVILEGES"` as `"ALL_PRIVILEGES"` and react accordingly

alexott:
Does it work correctly? No configuration drift after apply?

ledbutter:
> Does it work correctly? No configuration drift after apply?

Sorry, I should have provided a little more detail originally. No, it does not work.

If I define a resource like this:

```hcl
resource "databricks_grants" "external_location" {
  external_location = "my_location"

  grant {
    principal  = "user"
    privileges = ["ALL PRIVILEGES"] # this should have an underscore between words
  }
}
```

The initial attempt fails with an error like this:
> cannot create grants: permissions for external_location-dev_my_location &{[{user [ALL_PRIVILEGES] [Principal]}] but have to be {[{user [ALL PRIVILEGES] []}]}

Retries fail with:
> │ Error: cannot create grants: Duplicate privileges to add and delete.
│ 
│   with databricks_grants.external_location["my_location"],
│   on externalLocations.tf line 85, in resource "databricks_grants" "external_location":
│   85: resource "databricks_grants" "external_location" {

Which matches the behavior in the CLI:
> databricks grants update external_location my_location -p my_profile --json '{"changes": [{"principal": "user", "add": ["ALL PRIVILEGES"]},{"principal": "user", "remove": ["ALL_PRIVILEGES"]} ]}'
Error: Duplicate privileges to add and delete.

The diff'ing logic doesn't take into consideration that "ALL PRIVILEGES" is the same as "ALL_PRIVILEGES" in the databricks API/CLI (it silently just accepts it without the underscore).

mgyucht:
@ledbutter Thanks for this contribution! We've discussed this problem and your approach seems to solve one half of the problem. The other half is making sure that there is no diff from TF perspective at plan time if the only difference is underscores. Can you add logic to `CustomizeDiff` for these resources to ensure there is no diff is the only difference between config & state is an underscore in the permission level?

ledbutter:
I've pushed changes for just the singular grant resource so far, but I'm still trying to figure out how to test this properly: I can't find prior art for a scenario like this with the CustomizeDiff being tested like this. The closest I've found is validating the "health" attribute is cleared on [sql endpoint](https://github.com/databricks/terraform-provider-databricks/blob/a0f2fe271396626fccd09122fa887b0e737fbf28/sql/resource_sql_endpoint_test.go#L244), but that doesn't require any actual setup. I'll keep grinding on this as I get the time.

catalog/resource_grant_test.go
```diff
@@ -834,3 +857,23 @@ func TestResourceGrantModelGrantCreate(t *testing.T) {
...
+
+func TestResourceGrantUpdateSpacesNoUpdate(t *testing.T) {
+	qa.ResourceFixture{
+		Resource: ResourceGrant(),
+		ID:       "table/foo.bar.baz/me",
+		InstanceState: map[string]string{
+			"table":                 "foo.bar.baz",
+			"principal":             "me",
+			"privileges.#":          "1",
+			"privileges.3182106578": "ALL_PRIVILEGES", // this indexer seems like a huge hack
```
ledbutter, line 869: I want to also call out this indexer logic that doesn't make sense to me, but this seemingly "magic" value didn't change between test runs; I would love to know how to configure this in a more stable fashion

catalog/resource_grant.go
```diff
@@ -231,5 +243,34 @@ func ResourceGrant() common.Resource {
...
+					oldString := old.(string)
+					newString := new.(string)
+
+					log.Printf("[DEBUG] CustomizeDiff: oldString: %s newString: %s", oldString, newString)
+
+					// if the only difference is spaces, remove the change
+					if newString != oldString && strings.ReplaceAll(newString, " ", "_") == oldString {
+						log.Printf("[DEBUG] CustomizeDiff: removing change: %s", k)
+						// this doesn't work since these privileges are not computed...
+						err := d.Clear(k)
```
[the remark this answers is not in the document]
ledbutter: Just recently, @tanmay-db did [remove](#3290) a variation of this behavior where it would throw if not found in an internal mapping of privileges

ledbutter:
> @ledbutter Thanks for this contribution! We've discussed this problem and your approach seems to solve one half of the problem. The other half is making sure that there is no diff from TF perspective at plan time if the only difference is underscores. Can you add logic to `CustomizeDiff` for these resources to ensure there is no diff is the only difference between config & state is an underscore in the permission level?

After more research, it seems like [`DiffSuppressFunc`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc) is what we actually want to use in this case. I'm having a hard time validating it is working correctly however.

nkvuong:
@ledbutter the only test I've seen that test this behaviour is `TestCatalogSuppressCaseSensitivity` under `catalog/resource_catalog_test.go`

CONTRIBUTING.md
```diff
@@ -353,7 +353,7 @@ So please run `make lint` instead.
 
 NOTE: This use of devcontainers for terraform-provider-databricks development is **experimental** and not officially supported by Databricks
 
-This project has configuration for working with [Visual Studio Code Devcontainers](https://code.visualstudio.com/docs/remote/containers) - this allows you to containerise your development rerequisites (e.g. golang, terraform). To use this you will need [Visual Studio Code](https://code.visualstudio.com/) and [Docker](https://www.docker.com/products/docker-desktop).
+This project has configuration for working with [Visual Studio Code Devcontainers](https://code.visualstudio.com/docs/remote/containers) - this allows you to containerize your development prerequisites (e.g. golang, terraform). To use this you will need [Visual Studio Code](https://code.visualstudio.com/) and [Docker](https://www.docker.com/products/docker-desktop).
```
nkvuong, line 356: could this change be separated in a different PR?

.devcontainer/devcontainer.json
```diff
@@ -8,7 +8,7 @@
 			// Update the VARIANT arg to pick a version of Go: 1, 1.18, 1.17
 			// Append -bullseye or -buster to pin to an OS version.
 			// Use -bullseye variants on local arm64/Apple Silicon.
-			"VARIANT": "1.19-bullseye",
+			"VARIANT": "1.21-bullseye",
```
nkvuong, line 11: same here

catalog/resource_grant.go
```diff
@@ -119,10 +130,14 @@ func parseSecurableId(d *schema.ResourceData) (string, string, string, error) {
 	return split[0], split[1], split[2], nil
 }
 
+// func privilegesSuppressWhitespaceChange(k, old, new string, d *schema.ResourceData) bool {
+// 	log.Printf("[INFO] privilegesSuppressWhitespaceChange: k: %s old: %s new: %s", k, old, new)
+// 	return strings.ReplaceAll(old, " ", "_") == strings.ReplaceAll(new, " ", "_")
+// }
+
```
nkvuong, line 137: could you remove commented out code

common/resource.go
```diff
@@ -63,7 +63,7 @@ func (r Resource) saferCustomizeDiff() schema.CustomizeDiffFunc {
 	return func(ctx context.Context, rd *schema.ResourceDiff, _ any) (err error) {
 		defer func() {
 			// this is deliberate decision to convert a panic into error,
-			// so that any unforeseen bug would we visible to end-user
+			// so that any unforeseen bug would be visible to end-user
```
nkvuong, line 66: same

qa/testing.go
```diff
@@ -124,7 +124,7 @@ type ResourceFixture struct {
 	New bool
 }
 
-// wrapper type for calling resource methords
+// wrapper type for calling resource methods
```
nkvuong, line 127: same

nkvuong (changes requested):

nkvuong:
> TestCatalogSuppressCaseSensitivity


> > @ledbutter the only test I've seen that test this behaviour is `TestCatalogSuppressCaseSensitivity` under `catalog/resource_catalog_test.go`
> 
> Right, and this is so trivial I don't see much value in it, but I will certainly add those tests if desired. What I'd really like to validate is the actual `plan/apply` suppresses this as expected, especially for the grants case where I'm a little less sure of the approach.

For unit testing, this is the way to test as we should validate that the `ExpectedDiff` is calculated correctly by Terraform. For integration test, you could simply create a test that creates a grant resource with space in privileges, e.g. similar to `TestUcAccGrantsForIdChange`. The integration test fixture will create the resources, validate that the tf plan is clean (which should confirm that the suppressdiff works as expected)

ledbutter:
> > > @ledbutter the only test I've seen that test this behaviour is `TestCatalogSuppressCaseSensitivity` under `catalog/resource_catalog_test.go`
> > 
> > 
> > Right, and this is so trivial I don't see much value in it, but I will certainly add those tests if desired. What I'd really like to validate is the actual `plan/apply` suppresses this as expected, especially for the grants case where I'm a little less sure of the approach.
> 
> For unit testing, this is the way to test as we should validate that the `ExpectedDiff` is calculated correctly by Terraform. For integration test, you could simply create a test that creates a grant resource with space in privileges, e.g. similar to `TestUcAccGrantsForIdChange`. The integration test fixture will create the resources, validate that the tf plan is clean (which should confirm that the suppressdiff works as expected)

I think I've done this with my latest commits, but I'm uncertain about how I'm trying to assert the integration tests...

internal/acceptance/grant_test.go
```diff
@@ -134,3 +137,33 @@ func TestUcAccGrantForIdChange(t *testing.T) {
...
+		storage_credential = databricks_storage_credential.external.id
+		principal  = "{env.TEST_DATA_ENG_GROUP}"
+		privileges = ["%s"]
+	}
+	`, permission)
+}
+
+func TestUcAccGrantForUnderscoreChange(t *testing.T) {
+	unityWorkspaceLevel(t, step{
+		Template: grantTemplateForUnderscoreChange("ALL_PRIVILEGES"),
```
nkvuong, line 161: it should suffice to just have a single step with `ALL PRIVILEGES`, the test would fail if there is still a diff shown in the plan

ledbutter: Ok, updated

catalog/resource_grant.go
```diff
@@ -138,6 +148,9 @@ func ResourceGrant() common.Resource {
 					ConflictsWith: permissions.SliceWithoutString(allFields, field),
 				}
 			}
+
+			m["privileges"].DiffSuppressFunc = permissions.PrivilegesSuppressWhitespaceChange
```
nkvuong, line 152: @ledbutter  integration test failed, so I looked at this in a bit more details. `DiffSuppressFunc` only works for primitive, i.e. string-like attributes. For lists, we need to look at `CustomizeDiff` instead.

@tanmay-db we probably need to re-add a lightweight version of `CustomizeDiff` to handle this scenario, as it is causing permanent drifts

CONTRIBUTING.md
ledbutter: Yep, handled this in #3512

ledbutter:
@nkvuong @mgyucht So I have half of the problem solved, but I'm not sure how--if at all, given the current tooling--to solve the other half: the suppression of any diffs during a plan operation. Thoughts? 

I see the options as:

- Roll forward with just the "normalization" and leave the phantom diff as-is
- Abandon this
- Figure out a way to handle customizing diffs in this way (I'd need help on this as it seems to be tied to limitations in the overall tooling)

ledbutter:
@nkvuong Thanks for fixing this in a better way! I'll close this now since it is no longer needed

PR #3302, opened 2024-02-25
Make StructToSchema work for recursive data types

edwardfeng-db:
## Changes

- This is a prerequisite PR for onboarding jobs foreach task to terraform because it contains recursive reference
- Added `recursion_tracking` utility functions and `recursionTrackingContext` struct to track recursion depth
- Added `MaxDepthForTypes` into `ResourceProvider` to limit recursion depth per type
- Added unit test
- Updated `SchemaPath` helper function to fix a bug that could make this function return early

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -155,6 +155,10 @@ func (ClusterSpec) CustomizeSchema(s map[string]*schema.Schema) map[string]*sche
 	return s
 }
 
+func (ClusterSpec) MaxDepthForTypes() map[string]int {
+	return map[string]int{}
+}
+
 func resourceClusterSchema() map[string]*schema.Schema {
 	return common.StructToSchema(ClusterSpec{}, nil)
 }
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -155,10 +155,6 @@ func (ClusterSpec) CustomizeSchema(s map[string]*schema.Schema) map[string]*sche
 	return s
 }
 
-func (ClusterSpec) MaxDepthForTypes() map[string]int {
-	return map[string]int{}
-}
-
 func resourceClusterSchema() map[string]*schema.Schema {
 	return common.StructToSchema(ClusterSpec{}, nil)
 }
--- a/common/recursion_tracking.go
+++ b/common/recursion_tracking.go
@@ -48,7 +48,7 @@ func getEmptyRecursionTrackingContext() recursionTrackingContext {
 	}
 }
 
-func getRecursionTrackingContext(rp ResourceProvider) recursionTrackingContext {
+func getRecursionTrackingContext(rp RecursiveResourceProvider) recursionTrackingContext {
 	return recursionTrackingContext{
 		map[string]int{},
 		rp.MaxDepthForTypes(),
--- a/common/recursion_tracking.go
+++ b/common/recursion_tracking.go
@@ -1,0 +1,56 @@
+package common
+
+import (
+	"maps"
+	"reflect"
+	"strings"
+)
+
+type recursionTrackingContext struct {
+	timesVisited     map[string]int
+	maxDepthForTypes map[string]int
+}
+
+func (rt recursionTrackingContext) depthExceeded(typeField reflect.StructField) bool {
+	typeName := rt.getNameForTypeField(typeField)
+	if maxDepth, ok := rt.maxDepthForTypes[typeName]; ok {
+		return rt.timesVisited[typeName]+1 > maxDepth
+	}
+	return false
+}
+
+func (rt recursionTrackingContext) getNameForTypeField(typeField reflect.StructField) string {
+	return strings.TrimPrefix(typeField.Type.String(), "*")
+}
+
+func (rt recursionTrackingContext) getMaxDepthForTypeField(typeField reflect.StructField) int {
+	typeName := rt.getNameForTypeField(typeField)
+	return rt.maxDepthForTypes[typeName]
+}
+
+func (rt recursionTrackingContext) copy() recursionTrackingContext {
+	newTimesVisited := map[string]int{}
+	maps.Copy(newTimesVisited, rt.timesVisited)
+	return recursionTrackingContext{
+		timesVisited:     newTimesVisited,
+		maxDepthForTypes: rt.maxDepthForTypes,
+	}
+}
+
+func (rt recursionTrackingContext) visit(v reflect.Value) {
+	rt.timesVisited[strings.TrimPrefix(v.Type().String(), "*")] += 1
+}
+
+func getEmptyRecursionTrackingContext() recursionTrackingContext {
+	return recursionTrackingContext{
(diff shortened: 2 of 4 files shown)


common/reflect_resource.go
```diff
@@ -91,12 +93,13 @@ func getJsonFieldName(typeField reflect.StructField) string {
...
 	cs := s
-	for _, p := range path {
+	for i := 0; i < len(path); i++ {
+		p := path[i]
 		v, ok := cs[p]
 		if !ok {
 			return nil, fmt.Errorf("missing key %s", p)
 		}
-		if p == path[len(path)-1] {
+		if i == len(path)-1 {
```
edwardfeng-db, line 102: There was a bug with the previous way of implementation because you could have an item in the path that is the same string as the last item, which would make this function return early.

common/reflect_resource.go
```diff
@@ -91,12 +93,13 @@ func getJsonFieldName(typeField reflect.StructField) string {
 // SchemaPath helps to navigate
 func SchemaPath(s map[string]*schema.Schema, path ...string) (*schema.Schema, error) {
 	cs := s
-	for _, p := range path {
+	for i := 0; i < len(path); i++ {
```
nfx, line 96:
```suggestion
	for i, p := range path {
```
this is simpler

common/reflect_resource.go
```diff
@@ -419,6 +428,11 @@ func typeToSchema(v reflect.Value, aliases map[string]string) map[string]*schema
 	return scm
 }
 
+func maxRecursionDepthExceeded(typeField reflect.StructField, timesVisited map[string]int) bool {
```
nfx, line 431: implementation is spread over multiple places, making it more fragile. add a small new type to contain all the functionality:

```go
type recursionTracking struct { 
  visited map[string]int
  maxDepth int 
}
func (rt recursionTracking) DepthExceeded(typeField reflect.StructField) bool
func (rt recursionTracking) Copy() recursionTracking
func (rt recursionTracking) name(typeField reflect.StructField) string
```

common/reflect_resource.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var maxRecursionDepth = 2
```
nfx, line 13: shared mutable state we have to avoid.

nfx (dismissed):

common/reflect_resource.go
edwardfeng-db: Nice, thanks, let me update it

common/recursion_tracking.go
```diff
@@ -0,0 +1,33 @@
+package common
+
+import (
+	"reflect"
+	"strings"
+)
+
+const DefaultMaxRecursionDepth = 2
```
edwardfeng-db, line 8: For now we'll keep a global shared default max, in the future if necessary we can add a field in `ResourceProvider` such that a specific resource can have custom limit

mgyucht, line 8: Let's not export anything that isn't necessary.

common/recursion_tracking.go
```diff
@@ -0,0 +1,33 @@
...
+
+type RecursionTracking struct {
```
mgyucht, line 10: also doesn't need to be exported.

common/recursion_tracking.go
```diff
@@ -0,0 +1,33 @@
...
+	timesVisited map[string]int
+	maxDepth     int
+}
+
+func (rt RecursionTracking) DepthExceeded(typeField reflect.StructField) bool {
```
mgyucht, line 15: Same for these methods.

common/util.go
```diff
@@ -34,3 +34,12 @@ func SuppressDiffWhitespaceChange(k, old, new string, d *schema.ResourceData) bo
 	log.Printf("[DEBUG] Suppressing diff for %v: old=%#v new=%#v", k, old, new)
 	return strings.TrimSpace(old) == strings.TrimSpace(new)
 }
+
+// Creates a deep copy of the given map.
+func DeepCopyMap[K comparable, V any](originalMap map[K]V) map[K]V {
```
mgyucht, line 39: use `maps.Copy`

common/reflect_resource.go
```diff
@@ -292,9 +292,14 @@ func typeToSchema(v reflect.Value, aliases map[string]string) map[string]*schema
 	if rk != reflect.Struct {
 		panic(fmt.Errorf("Schema value of Struct is expected, but got %s: %#v", reflectKind(rk), v))
 	}
+	rt.Visit(v)
 	fields := listAllFields(v)
 	for _, field := range fields {
 		typeField := field.sf
+		if rt.DepthExceeded(typeField) {
+			// Skip the field if recursion depth is over the limit.
```
mgyucht, line 300: trace log

common/reflect_resource_test.go
```diff
@@ -78,6 +78,33 @@ type testStruct struct {
...
+
+type testJobTask struct {
+	ForEachTask *testForEachTask `json:"for_each_task,omitempty"`
+	Extra       string           `json:"extra,omitempty"`
+}
+
+type testForEachTask struct {
+	Task  *testJobTask `json:"task,omitempty"`
+	Extra string       `json:"extra,omitempty"`
+}
```
edwardfeng-db, line 94: These structs above basically mimic the jobs `for_each_task` recursive structure that we are going to introduce

common/reflect_resource.go
```diff
@@ -43,12 +43,13 @@ var kindMap = map[reflect.Kind]string{
 type ResourceProvider interface {
 	Aliases() map[string]string
 	CustomizeSchema(map[string]*schema.Schema) map[string]*schema.Schema
+	MaxDepthForTypes() map[string]int
```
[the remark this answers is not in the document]
edwardfeng-db: Ah this is a great idea, let me update this

alexott (commented):
in general looks good

clusters/resource_cluster.go
```diff
@@ -155,6 +155,10 @@ func (ClusterSpec) CustomizeSchema(s map[string]*schema.Schema) map[string]*sche
 	return s
 }
 
+func (ClusterSpec) MaxDepthForTypes() map[string]int {
```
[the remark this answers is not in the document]
edwardfeng-db: Updated, this method is no longer needed. But I'm a bit confused why we need the `New` prefix, can you explain your rationale?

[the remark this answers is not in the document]
alexott: Because it was returning empty map

edwardfeng-db:
Stacked PR https://github.com/databricks/terraform-provider-databricks/pull/3320 to test the generated schema. See `compute_diff` to check the generaetd output for for_each_task

mgyucht (approved):
LGTM, thanks for addressing these comments systematically!

PR #3309, opened 2024-02-28
Fix webhook_notifications diff suppression for `databricks_job`

pietern:
## Changes
Currently, it is not possible to remove all webhook notifications of a certain type in a `databricks_job`. When removing all webhook notifications of a certain kind, `common.DataToStructPointer()` constructs a malformed request, where all of the webhooks of that kind have no ID set.

The reason for this is quite complex. Today, when a field is labeled with `suppress_diff`, the following things happen:
1. `DiffSuppressFunc` is set for that `*schema.Schema` to `diffSuppressor`.
2. If the field is a struct or pointer to a struct, `DiffSuppressFunc` is set for all of the fields within the struct to the same `diffSuppressor`.

`diffSuppressor` attempts to suppress a diff for a given schema in Terraform based on the type of the schema. The goal is to ignore fields returned by the Databricks API that are persisted into state but that are not marked as computed but still shouldn't result in any diff for users. It first computes the string representation of the zero value of that type. Then it performs two checks:
1. If the "new" value is that zero value and the old value is something else, this is ignored. For primitive types, this is sufficient. 
2. For nested types, like structures, the Schema type is `TypeList`. When `DiffSuppressFunc` is called for `TypeList` attributes, if the number of elements in the list changes, it is first called with a key ending in `.#` to show the difference in the number of elements. It is then called for every diff within the list that has not yet been suppressed. To ensure `suppress_diff` hides diffs in these kinds of empty responses, the provider checks whether the old value is nonzero and the new value is `0`, indicating that the user didn't specify a value for the list but it was persisted in the state and thus should be ignored.

In the case of webhook notifications, a DiffSuppressFunc is registered by 2. above on each field of webhook notifications because of the `suppress_diff` annotation on the parent `WebhookNotifications` field in `JobSettings`. Specifically for the list fields, if all elements are removed from the list, the second check of `diffSuppressor` causes the diff to be suppressed.

The ultimate result of all of this is that, when calling `d.GetOk("webhook_notifications")`, Terraform uses the original value of the `.#` attribute for the list fields of webhook notifications by referring to the union of config and state. As the webhooks existed in state and the diff was suppressed, the value from the state is used (non-zero). To contrast, if the diff was not suppressed, the new value would be used (zero).

For now, we will patch this behavior by making two minimal changes:
1. Modify `diffSuppressor` behavior 2 to only suppress diffs for changes in the current list, not in nested lists. We do that by verifying that the key of the diff ends with the name of the field for which the diff suppressor is registered plus `.#`. This way, the parent diff suppression on Webhook Notifications doesn't accidentally suppress the diff of a nested list field.
2. Remove the `diffSuppressor` for the nested fields of WebhookNotifications entirely. This way, removing all entries from the list is not diff-suppressed.

Ultimately, I think this diff suppression logic may not be needed and can be replaced by marking such fields as `computed`. Related, we should probably remove the recursive diff suppression where possible and either manually mark fields with diff suppression as needed or use tools like `CustomizeSchemaPath` to suppress diffs in other fields. However, as either of these would be a much more invasive change, I'm holding off on doing this for now.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -27,14 +27,14 @@ import (
 // NotebookTask contains the information for notebook jobs
 type NotebookTask struct {
 	NotebookPath   string            `json:"notebook_path"`
-	Source         string            `json:"source,omitempty" tf:"suppress_diff"`
-	BaseParameters map[string]string `json:"base_parameters,omitempty"`
+	Source         string            `json:"source,omitempty" tf:"computed"`
+	BaseParameters map[string]string `json:"base_parameters,omitempty" tf:"computed"`
 }
 
 // SparkPythonTask contains the information for python jobs
 type SparkPythonTask struct {
 	PythonFile string   `json:"python_file"`
-	Source     string   `json:"source,omitempty" tf:"suppress_diff"`
+	Source     string   `json:"source,omitempty" tf:"computed"`
 	Parameters []string `json:"parameters,omitempty"`
 }
 
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -27,14 +27,14 @@ import (
 // NotebookTask contains the information for notebook jobs
 type NotebookTask struct {
 	NotebookPath   string            `json:"notebook_path"`
-	Source         string            `json:"source,omitempty" tf:"computed"`
-	BaseParameters map[string]string `json:"base_parameters,omitempty" tf:"computed"`
+	Source         string            `json:"source,omitempty" tf:"suppress_diff"`
+	BaseParameters map[string]string `json:"base_parameters,omitempty"`
 }
 
 // SparkPythonTask contains the information for python jobs
 type SparkPythonTask struct {
 	PythonFile string   `json:"python_file"`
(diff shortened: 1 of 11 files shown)


alexott (approved):
In general, looks good to me if integration tests as passing.

jobs/resource_job.go
```diff
@@ -752,6 +752,11 @@ var jobSchema = common.StructToSchema(JobSettings{},
 		common.MustSchemaPath(s, "run_as", "user_name").ExactlyOneOf = run_as_eoo
 		common.MustSchemaPath(s, "run_as", "service_principal_name").ExactlyOneOf = run_as_eoo
 
+		// Clear the implied diff suppression for the webhook notification lists
+		for _, n := range []string{"on_start", "on_failure", "on_success", "on_duration_warning_threshold_exceeded"} {
+			common.MustSchemaPath(s, "webhook_notifications", n).DiffSuppressFunc = nil
+		}
+
```
pietern, line 759: @mgyucht Was this still needed?

PR #3310, opened 2024-02-28
Set `Computed` field to `true` for `common.Resource` `account_id` attribute

### LINKED ISSUE #3223: [ISSUE] Issue with `databricks_mws_private_access_settings` resource
As of provider version 1.35.0 I get a deprecated warning that account_id should not be specified in databricks_mws_private_access_settings resource and to set it in the provider block instead.  But, if I remove it from the resource block I get a plan that updates every time it runs to 'reset' account_id to null, and the only way to get rid of this is to set account_id in the resource.  account_id is set in the provider block and doesn't seem to make a difference.

### Configuration
```hcl
resource "databricks_mws_private_access_settings" "pas" {
  private_access_settings_name = "${local.workspace_name} Private Access Settings"
  region                       = var.region
  public_access_enabled        = true
  private_access_level         = "ACCOUNT" 
} 
```

### Expected Behavior
No plan update without account_id specified in resource

### Actual Behavior
Every time apply runs it sets account_id in the resource to null, then next time plan/apply is run it detects the value there again and tries to remove it again.

Databricks: 1.35.0
Terraform: 1.6.5

### Is it a regression?
The account_id setting in this resource was only deprecated with this latest version, behavior was fine in 1.34.0 and earlier.

840:
In #3135, the `account_id` resource field is deprecated in favour of the provider level `account_id` attribute. However, it was reported that for the `databricks_mws_private_access_settings` resource, this could lead to a diff for every apply since the API returns the `account_id`. 

The API returns `account_id`, putting the attribute back in the state. But since it is removed in the resource, it will try to set it back to null.

Setting the `account_id` field to `Computed` will use the API provided `account_id` which should not result in diffs.

Resolves #3223 

## Changes
Set `Computed` field to `true` for `common.Resource` `account_id` attribute.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/resource.go
+++ b/common/resource.go
@@ -443,6 +443,7 @@ var NoAuth string = "default auth: cannot configure default credentials, " +
 func AddAccountIdField(s map[string]*schema.Schema) map[string]*schema.Schema {
 	s["account_id"] = &schema.Schema{
 		Type:       schema.TypeString,
+		Computed:   true,
 		Optional:   true,
 		Deprecated: "Configuring `account_id` at the resource-level is deprecated; please specify it in the `provider {}` configuration block instead",
 	}
--- a/mws/resource_mws_private_access_settings_test.go
+++ b/mws/resource_mws_private_access_settings_test.go
@@ -3,6 +3,8 @@ package mws
 import (
 	"testing"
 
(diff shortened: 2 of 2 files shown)


mgyucht (approved):
LGTM!

PR #3325, opened 2024-03-03
Add support for ESC (Enhanced Security and Compliance) settings

harshshah-db:
## Changes
Adding support for ESC features - compliance security profile and enhanced security monitoring

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/settings/all_settings.go
+++ b/settings/all_settings.go
@@ -19,7 +19,7 @@ func AllSettingsResources() map[string]common.Resource {
 		"restrict_workspace_admins": makeSettingResource[settings.RestrictWorkspaceAdminsSetting, *databricks.WorkspaceClient](restrictWsAdminsSetting),
 		"compliance_security_profile_account": makeSettingResource[settings.CSPEnablementAccountSetting, *databricks.AccountClient](cspEnablementAccountSetting),
 		"enhanced_security_monitoring_account": makeSettingResource[settings.ESMEnablementAccountSetting, *databricks.AccountClient](esmEnablementAccountSetting),
-		"compliance_security_profile_workspace": makeSettingResource[settings.CSPEnablementSetting, *databricks.AccountClient](cspEnablementWorkspaceSetting),
-		"enhanced_security_monitoring_workspace": makeSettingResource[settings.ESMEnablementSetting, *databricks.AccountClient](esmEnablementWorkspaceSetting),
+		"compliance_security_profile_workspace": makeSettingResource[settings.CSPEnablementSetting, *databricks.WorkspaceClient](cspEnablementWorkspaceSetting),
+		"enhanced_security_monitoring_workspace": makeSettingResource[settings.ESMEnablementSetting, *databricks.WorkspaceClient](esmEnablementWorkspaceSetting),
 	}
 }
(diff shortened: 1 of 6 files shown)


alexott:
@harshshah-db please fix tests

settings/all_settings.go
```diff
@@ -17,5 +17,9 @@ func AllSettingsResources() map[string]common.Resource {
 	return map[string]common.Resource{
 		"default_namespace":         makeSettingResource[settings.DefaultNamespaceSetting, *databricks.WorkspaceClient](defaultNamespaceSetting),
 		"restrict_workspace_admins": makeSettingResource[settings.RestrictWorkspaceAdminsSetting, *databricks.WorkspaceClient](restrictWsAdminsSetting),
+		"compliance_security_profile_account": makeSettingResource[settings.CSPEnablementAccountSetting, *databricks.AccountClient](cspEnablementAccountSetting),
+		"enhanced_security_monitoring_account": makeSettingResource[settings.ESMEnablementAccountSetting, *databricks.AccountClient](esmEnablementAccountSetting),
+		"compliance_security_profile_workspace": makeSettingResource[settings.CSPEnablementSetting, *databricks.AccountClient](cspEnablementWorkspaceSetting),
+		"enhanced_security_monitoring_workspace": makeSettingResource[settings.ESMEnablementSetting, *databricks.AccountClient](esmEnablementWorkspaceSetting),
```
alexott, line 23: @mgyucht We need to think about how to avoid this proliferation of resources... Can we have a single resource that will work on both workspace & account level...

alexott (changes requested):
Documentation is missing. Tests are missing as well. The code is not compilable (forgot to commit a file?)

settings/all_settings.go
mgyucht, line 23: I think these won't work at all, as they require a WorkspaceClient, not an AccountClient.

mgyucht (changes requested):
I think we need to introduce a new kind of setting called a `workspaceAndAccountSetting` that has the following semantics:
* When using an account-level provider, configures the setting at the account level.
* When using a workspace-level provider, configures the setting at the workspace level.

Otherwise, I agree with @alexott that this will result in a lot more duplicate resources (and more opportunities for customers to use the wrong resource).

settings/all_settings.go
mgyucht: Sorry, took another look at this one. It seems like there are two separate settings, one about enabling these features at the workspace level, and one about enforcing them at the account level.

alexott:
@harshshah-db please rebase, and run `make fmt` on the code...

alexott:
@harshshah-db new SDK uses a slightly different structures, see this PR to see how to fix these things: https://github.com/databricks/terraform-provider-databricks/pull/3333/files

alexott:
@harshshah-db I rebased to the latest `main` and fixed API changes

alexott:
We need to add tests, documentation & maybe integration tests (I'm not sure how ESM/CSP enablement will affect the setup on the workspace level - can it be reverted after enablement?)

Cc: @mgyucht

mgyucht:
If not, then we can create new workspaces in those integration tests as needed. We'll probably need to anyways to make sure the "enforce" settings take effect as expected.

harshshah-db:
Closing as this is replaced by #3563

PR #3330, opened 2024-03-05
add `IF NOT EXISTS` option to sql_table

alvaroqueiroz:
## Changes
Hello!
I was building a pipeline that manages views using the terraform `databricks_sql_table` resource and felt the need to use IF NOT EXISTS clause.

This pull request adds create_if_not_exists option to the `databricks_sql_table` resource.

#### Reasons:

- Running terraform apply will fail if the resource exists, which is not the expected behavior
- Adding the IF NOT EXISTS clause makes the operation idempotent, which is desirable
- Using  IF NOT EXISTS clause will cause less overhead of handling errors and less pollution of logging

My organization currently uses the  `databricks_sql_table` resource to manage view creation, and it would be nice to have the option to use the IF NOT EXISTS clause for the reasons above and to stop our CICD pipelines from failing.

#### Terraform State:
Because this is an optional input and defaults to false, it would not make any changes to the current state because it would not make any changes to the DDL, so this change should be backward-compatible

#### Testing:
I have added one case for testing the changes with `create_if_not_exists` and run `make test` and individual tests as well

image

image

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -43,6 +43,7 @@ type SqlTableInfo struct {
 	Options               map[string]string `json:"options,omitempty" tf:"force_new"`
 	ClusterID             string            `json:"cluster_id,omitempty" tf:"computed"`
 	WarehouseID           string            `json:"warehouse_id,omitempty"`
+    CreateIfNotExists     bool              `json:"create_if_not_exists,omitempty"`
 
 	exec    common.CommandExecutor
 	sqlExec sql.StatementExecutionInterface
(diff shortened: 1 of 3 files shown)


alvaroqueiroz:
hello @hectorcast-db
do you think this is ready for review?

alvaroqueiroz:
Hello @hectorcast-db,
We are using terraform to manage the creation of views.
We have a file with the list of views that should exist and we would just add another one to the list when needed.

The problem we are having is: the terraform will fail to apply to for the views that already exist, that's why adding the `IF NOT EXISTS` clouse would fix our problem.

meaning: the TF controls the full cycle of our implementation, but it tries to run the create table DDL's anyway

alvaroqueiroz:
When running TF PLAN, it plans to change all the views, even if the code hasn't changed.
I think this might be related with how the strings of the view definition are compared

image

image

alvaroqueiroz:
@hectorcast-db after digging a bit more, I have found that the problem is in the section of the code:

https://github.com/databricks/terraform-provider-databricks/blob/30e7b600bc21f518ba53f927a7f9942352865f71/catalog/resource_sql_table.go#L302

I think there is some formatting going on that makes this comparison fail. I can fix it if you agree that this is the problem

PR #3338, opened 2024-03-06
Update Aliases function for ResourceProvider to handle recursive structures more gracefully

edwardfeng-db:
## Changes

- This is a sub PR of https://github.com/databricks/terraform-provider-databricks/pull/3320
- Make `Aliases` function return a map of map which can derive aliases from the parent type name and field name.
- The goal is to avoid redundantly specifying the aliases in recursive data types such as the jobs for_each_task
- Also updating the `recursion_tracking.go` functions to make `getNameForType` more sharable 

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/common/recursion_tracking.go
+++ b/common/recursion_tracking.go
@@ -14,7 +14,6 @@ type recursionTrackingContext struct {
 func (rt recursionTrackingContext) depthExceeded(typeField reflect.StructField) bool {
 	typeName := getNameForType(typeField.Type)
 	if maxDepth, ok := rt.maxDepthForTypes[typeName]; ok {
-		println("in if!!")
 		return rt.timesVisited[typeName]+1 > maxDepth
 	}
 	return false
--- a/common/recursion_tracking.go
+++ b/common/recursion_tracking.go
@@ -34,7 +34,7 @@ func (rt recursionTrackingContext) copy() recursionTrackingContext {
 }
 
 func (rt recursionTrackingContext) visit(v reflect.Value) {
-	rt.timesVisited[strings.TrimPrefix(v.Type().String(), "*")] += 1
+	rt.timesVisited[getNameForType(v.Type())] += 1
 }
 
 func getEmptyRecursionTrackingContext() recursionTrackingContext {
--- a/common/recursion_tracking.go
+++ b/common/recursion_tracking.go
@@ -13,8 +13,6 @@ type recursionTrackingContext struct {
 
 func (rt recursionTrackingContext) depthExceeded(typeField reflect.StructField) bool {
 	typeName := getNameForType(typeField.Type)
-	println("========")
-	println(typeName)
 	if maxDepth, ok := rt.maxDepthForTypes[typeName]; ok {
 		println("in if!!")
 		return rt.timesVisited[typeName]+1 > maxDepth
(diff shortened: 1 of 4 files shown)


common/reflect_resource.go
```diff
@@ -41,7 +41,7 @@ var kindMap = map[reflect.Kind]string{
 // Generic interface for ResourceProvider. Using CustomizeSchema and Aliases functions to keep track of additional information
 // on top of the generated go-sdk struct. This is used to replace manually maintained structs with `tf` tags.
 type ResourceProvider interface {
-	Aliases() map[string]string
+	Aliases() map[string]map[string]string
```
mgyucht, line 44: Can you add the doccomment here? What are the keys and values?

By the way, we may want to make `Aliases` optional for ResourceProvider. If a resource has no aliases, we shouldn't be forced to add an empty Aliases function, right?

common/reflect_resource.go
```diff
@@ -79,21 +79,24 @@ func reflectKind(k reflect.Kind) string {
 	return n
 }
 
-func chooseFieldNameWithAliases(typeField reflect.StructField, aliases map[string]string) string {
+func chooseFieldNameWithAliases(typeField reflect.StructField, parentTypeName string, aliases map[string]map[string]string) string {
```
mgyucht, line 82: Suggestion: let's make `parentTypeName` type `reflect.Type` rather than `string`. String is too broad as a useful interface, putting burden on the caller to pass the value of the right type.

mgyucht (commented):
Seems reasonable, small suggestions but we're almost there.

common/reflect_resource.go
edwardfeng-db: Added comments with an example.

For making `Aliases` optional, I played around with the idea of adding another interface without `Aliases()` but I think it's a bit more complicated since we also have `RecursiveResourceProvider`. I think I'm probably going to leave it as it is and try to resolve this later on. For now I think let's just merge this to unblock the next steps for the jobs migration

common/reflect_resource.go
edwardfeng-db: Good point, updated

common/recursion_tracking.go
```diff
@@ -12,19 +12,16 @@ type recursionTrackingContext struct {
 }
 
 func (rt recursionTrackingContext) depthExceeded(typeField reflect.StructField) bool {
-	typeName := rt.getNameForTypeField(typeField)
+	typeName := getNameForType(typeField.Type)
 	if maxDepth, ok := rt.maxDepthForTypes[typeName]; ok {
+		println("in if!!")
```
mgyucht, line 17:
```suggestion
```

common/reflect_resource.go
mgyucht: Discussed offline. We'll do this.

common/reflect_resource.go
```diff
@@ -38,13 +38,28 @@ var kindMap = map[reflect.Kind]string{
...
-// on top of the generated go-sdk struct. This is used to replace manually maintained structs with `tf` tags.
+// Generic interface for ResourceProvider. Using CustomizeSchema function to keep track of additional information
+// on top of the generated go-sdk struct.
 type ResourceProvider interface {
-	Aliases() map[string]string
 	CustomizeSchema(map[string]*schema.Schema) map[string]*schema.Schema
 }
 
+// Interface for ResourceProvider instances that need aliases for fields.
+// The function MaxDepthForTypes allows us to specify the max number of recursive depth for a specific field
```
mgyucht, line 48:
```suggestion
```

mgyucht (approved):

PR #3347, opened 2024-03-07
Add resource provider registry

edwardfeng-db:
## Changes

- Part of https://github.com/databricks/terraform-provider-databricks/pull/3320
- Add resource provider registry to avoid redundant alias and customization definitions
- Update `Library` resource to follow the ResourceProvider interface

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_library.go
+++ b/clusters/resource_library.go
@@ -16,10 +16,6 @@ type LibraryResource struct {
 	compute.Library
 }
 
-func (LibraryResource) Aliases() map[string]map[string]string {
-	return map[string]map[string]string{}
-}
-
 func (LibraryResource) CustomizeSchema(s map[string]*schema.Schema) map[string]*schema.Schema {
 	common.CustomizeSchemaPath(s).AddNewField("cluster_id", &schema.Schema{
 		Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


common/reflect_resource.go
```diff
@@ -38,6 +38,25 @@ var kindMap = map[reflect.Kind]string{
...
+// Example:
+//
+//	func init() {
+//		 common.RegisterResourceProvider(jobs.JobSettings{}) = JobSettingsResource{}
+//	}
+func RegisterResourceProvider(v any, r ResourceProvider) {
+	if resourceProviderRegistry == nil {
+		resourceProviderRegistry = make(map[string]ResourceProvider)
+	}
+	resourceProviderRegistry[getNameForType(reflect.ValueOf(v).Type())] = r
```
mgyucht, line 57: nit: let's panic if the same type is registered more than once?

clusters/resource_library.go
```diff
@@ -12,14 +12,24 @@ import (
...
-		}
-		return m
+type LibraryResource struct {
+	compute.Library
+}
+
+func (LibraryResource) Aliases() map[string]map[string]string {
+	return map[string]map[string]string{}
+}
+
```
mgyucht, line 22: Is this still needed? I thought we made aliases optional?

mgyucht (approved):
Two small nits, otherwise LGTM. Very excited to see the Jobs migration.

clusters/resource_library.go
edwardfeng-db: ah, yes, good catch

PR #3348, opened 2024-03-07
Delete Vector Search Endpoint when we get an error while waiting for creation

alexott:
## Changes

If we get an error while waiting for the endpoint creation we simply return that error.  But when we got a timeout, we didn't clean the endpoint and this led to the resource not being destroyed in the tests as the backend continued to provision it.

This PR deletes the vector search endpoint when we get a creation error during the wait phase. This behaviour is similar to the current behaviour of clusters.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/vectorsearch/resource_vector_search_endpoint.go
+++ b/vectorsearch/resource_vector_search_endpoint.go
@@ -2,6 +2,7 @@ package vectorsearch
 
 import (
 	"context"
+	"log"
 	"time"
 
 	"github.com/databricks/terraform-provider-databricks/common"
@@ -47,6 +48,13 @@ func ResourceVectorSearchEndpoint() common.Resource {
 			}
 			endpoint, err := wait.GetWithTimeout(d.Timeout(schema.TimeoutCreate))
 			if err != nil {
+				ctx2, cancel := context.WithDeadline(context.Background(), time.Now().Add(10*time.Second))
+				log.Printf("[ERROR] Error waiting for endpoint to be created: %s", err.Error())
+				err2 := w.VectorSearchEndpoints.DeleteEndpointByEndpointName(ctx2, req.Name)
+				if err2 != nil {
+					log.Printf("[ERROR] Error cleaning up endpoint: %s", err2.Error())
+				}
+				cancel()
 				return err
 			}
 			d.SetId(endpoint.Name)
--- a/vectorsearch/resource_vector_search_endpoint.go
+++ b/vectorsearch/resource_vector_search_endpoint.go
@@ -47,15 +47,13 @@ func ResourceVectorSearchEndpoint() common.Resource {
 			if err != nil {
 				return err
 			}
-			endpoint, err := wait.GetWithTimeout(d.Timeout(schema.TimeoutCreate))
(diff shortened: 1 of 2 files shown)


vectorsearch/resource_vector_search_endpoint.go
```diff
@@ -47,6 +48,13 @@ func ResourceVectorSearchEndpoint() common.Resource {
 			}
 			endpoint, err := wait.GetWithTimeout(d.Timeout(schema.TimeoutCreate))
 			if err != nil {
+				ctx2, cancel := context.WithDeadline(context.Background(), time.Now().Add(10*time.Second))
```
tanmay-db, line 51: 10*time.Second for this, let's have another variable.

vectorsearch/resource_vector_search_endpoint.go
```diff
@@ -47,6 +48,13 @@ func ResourceVectorSearchEndpoint() common.Resource {
...
+				log.Printf("[ERROR] Error waiting for endpoint to be created: %s", err.Error())
+				err2 := w.VectorSearchEndpoints.DeleteEndpointByEndpointName(ctx2, req.Name)
```
tanmay-db, line 53: Is this behaviour consistent? i.e. deleting works even if creation fails? This sounds strange and probably should be fixed on backend. Maybe it is similar to cancelling while creation is in progress. I think we should confirm this with the API owning team.

vectorsearch/resource_vector_search_endpoint.go
```diff
@@ -47,6 +48,13 @@ func ResourceVectorSearchEndpoint() common.Resource {
...
+				if err2 != nil {
```
tanmay-db, line 54: instead of err2, we could probably use nestedErr

vectorsearch/resource_vector_search_endpoint.go
tanmay-db, line 51: instead of ctx2, we could probably use another variable, maybe: `ctxWithDeadline`

tanmay-db (approved):
Some comments but overall LGTM, deleting end point when it's not successfully created sounds strange and probably should be fixed at backend. If this behaviour is consistent and expected from backend team (and won't change in future) then this sounds good. cc: @mgyucht

Also, does the integration test pass now? 

Let's discuss this once in TF office hours before merging.

vectorsearch/resource_vector_search_endpoint.go
```diff
@@ -45,8 +47,13 @@ func ResourceVectorSearchEndpoint() common.Resource {
 			if err != nil {
 				return err
 			}
-			endpoint, err := wait.GetWithTimeout(d.Timeout(schema.TimeoutCreate))
+			endpoint, err := wait.GetWithTimeout(d.Timeout(schema.TimeoutCreate) - deleteCallTimeout)
```
mgyucht, line 50: Technically, we should also subtract the time it took to make the create call. Otherwise, if it is longer than `deleteCallTimeout`, we will wait past the parent deadline.

mgyucht (approved):
One small nit, otherwise LGTM. I'll try to work on the abstraction tomorrow.

Issue #3057, opened 2023-12-27
Exporter: Add support for `depends_on` between resources/data sources.


alexott:
This could be required for #3056

PR #3357, opened 2024-03-10
Exporter: modify grants for some UC objects to allow handling of other owners

alexott:
## Changes

If UC objects, such as metastore, catalogs, schemas, volumes, etc. have an owner other than the current user, then it's impossible to create child objects during apply.  To mitigate this, the list of grants is expanded/changed to allow the creation of child objects by the identity who does the export (it's done only on the workspace level, and the same identity must be used when doing import).

TODOs:
- we need to discuss if we should use `ALL_PRIVILEGES` for catalogs/schemas/... vs. specific permissions like, `CREATE_SCHEMA` for catalogs, etc. @nkvuong What do you think?
- investigate if we can get the owner of the object, and don't add permissions if it's the current user - this may lead to increased API calls

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -21,7 +21,7 @@ import (
 	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/databricks-sdk-go/service/sharing"
 	"github.com/databricks/databricks-sdk-go/service/sql"
-	tfuc "github.com/databricks/terraform-provider-databricks/catalog"
+	tfcatalog "github.com/databricks/terraform-provider-databricks/catalog"
 	"github.com/databricks/terraform-provider-databricks/clusters"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/databricks/terraform-provider-databricks/jobs"
@@ -68,6 +68,10 @@ var (
 		"DBC":        ".dbc",
 		"R_MARKDOWN": ".Rmd",
 	}
+	metastorePrivilegesToAdd = []string{`CREATE_CATALOG`, `CREATE_CONNECTION`,
+		`CREATE_EXTERNAL_LOCATION`, `CREATE_PROVIDER`, `CREATE_RECIPIENT`, `CREATE_SCHEMA`,
+		`CREATE_STORAGE_CREDENTIAL`, `SET_SHARE_PERMISSION`}
+	defaultPrivilegesToAdd = []string{`ALL_PRIVILEGES`}
 )
 
 func generateMountBody(ic *importContext, body *hclwrite.Body, r *resource) error {
@@ -2351,7 +2355,7 @@ var resourcesMap map[string]importable = map[string]importable{
 			return nil
 		},
 		Import: func(ic *importContext, r *resource) error {
-			var cat tfuc.CatalogInfo
+			var cat tfcatalog.CatalogInfo
 			s := ic.Resources["databricks_catalog"].Schema
 			common.DataToStructPointer(r.Data, s, &cat)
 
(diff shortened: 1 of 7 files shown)


exporter/importables.go
```diff
@@ -68,6 +68,10 @@ var (
 		"DBC":        ".dbc",
 		"R_MARKDOWN": ".Rmd",
 	}
+	metastorePrivilegesToAdd = []string{`CREATE_CATALOG`, `CREATE_CONNECTION`,
```
nkvuong, line 71: only metastore admins can grant these permissions, so if you need to grant yourself these permissions prior to assigning away the ownership

nkvuong (commented):
this is where it gets complex in my opinion
- you're only able to create grants for objects that you own, or ones with parents objects that you own - so this would assume metastore admin permissions
- you need to apply these grants *before* creating any child objects, which would require an explicit `depends_on`

exporter/importables.go
```diff
@@ -68,6 +68,10 @@ var (
...
+		`CREATE_EXTERNAL_LOCATION`, `CREATE_PROVIDER`, `CREATE_RECIPIENT`, `CREATE_SCHEMA`,
+		`CREATE_STORAGE_CREDENTIAL`, `SET_SHARE_PERMISSION`}
+	defaultPrivilegesToAdd = []string{`ALL_PRIVILEGES`}
```
nkvuong, line 74: maintaining a separate list of `CREATE_X` privileges mapping would be sensible, as we do not add new securables that often

exporter/importables.go
```diff
@@ -2583,6 +2587,43 @@ var resourcesMap map[string]importable = map[string]importable{
...
+		Import: func(ic *importContext, r *resource) error {
+			if ic.meUserName == "" {
+				return nil
+			}
+			// TODO: discuss if we need to give ALL_PRIVILEGES vs. giving only specific CREATE permissions, ...
+			// https://docs.databricks.com/en/data-governance/unity-catalog/manage-privileges/privileges.html#privilege-types-by-securable-object-in-unity-catalog
+			var newPrivileges []string
+			if r.Data.Get("metastore").(string) != "" {
+				newPrivileges = append(newPrivileges, metastorePrivilegesToAdd...)
+			} else if r.Data.Get("catalog").(string) != "" || r.Data.Get("schema").(string) != "" ||
```
nkvuong, line 2599: volume & share don't have any child object, so don't need additional privileges granted

exporter/importables.go
alexott: this is an interesting case really, maybe we don't need to assign explicit permissions for a metastore... The problem that was reported is that when we create catalogs, schemas, etc., and the owner of an existing metastore is different than the metastore admin who runs the migration code, then we need to have explicit permissions to be able to create these objects...

exporter/importables.go
alexott: share yes - agree. But for Volume, should I give `VOLUME_WRITE` if I'm not the owner of the volume and want to write a `databricks_file` ?

alexott: We emit only Volume files explicitly referenced in jobs/clusters/libraries, not everything.

nkvuong:
this code makes sense, and looks good to me. my only concern is that the terraform template won't apply cleanly in the below scenario:

- a catalog with a different owner is created. after that, the following resources are created in parallel
    - grants on the catalog, including create schema for the exporter user
    - a schema is created inside that catalog - this would fail unless it starts after the grants are applied. unless we emit a depends_on here?

alexott:
Yes, I'm planning to implement support for `depends_on`: #3057 - after that, we'll add explicit dependencies for such grants

exporter/importables.go
```diff
@@ -2884,10 +2916,7 @@ var resourcesMap map[string]importable = map[string]importable{
 			return nil
 		},
 		Import: func(ic *importContext, r *resource) error {
-			ic.Emit(&resource{
-				Resource: "databricks_grants",
-				ID:       "metastore/" + r.ID,
-			})
+			ic.emitUCGrantsWithOwner("metastore/"+r.ID, r)
```
nkvuong, line 2919: I would export the metastore without the owner, otherwise you cannot create grants on objects you do not own

nkvuong (approved):
one more thing about metastore owner, otherwise lgtm

exporter/importables.go
alexott: Really, on the workspace level we won't emit metastores at all (the next PR is coming for it).

alexott:
https://github.com/databricks/terraform-provider-databricks/pull/3377 added generation of dependencies on grants

Issue #3341, opened 2024-03-06
[FEATURE] Support table column updates in databricks_sql_table resource


donatasm:
### Use-cases

Changing column type in `databricks_sql_table` resource has no effect, though Terraform detects and announces about successfully applied changes:

```
Terraform will perform the following actions:

  # databricks_sql_table.delivery_test_table will be updated in-place
  ~ resource "databricks_sql_table" "delivery_test_table" {
        id                      = "staging.delivery.test_table"
        name                    = "test_table"
        # (7 unchanged attributes hidden)

      ~ column {
            name     = "b"
          ~ type     = "int" -> "string"
            # (1 unchanged attribute hidden)
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
databricks_sql_table.delivery_test_table: Modifying... [id=staging.delivery.test_table]
databricks_sql_table.delivery_test_table: Modifications complete after 4s [id=staging.delivery.test_table]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

Registering issue as a feature request, since it looks like it is a known limitation at the moment https://github.com/databricks/terraform-provider-databricks/pull/3222#issuecomment-1929847313

Provider version `1.38.0`. Example table is external Delta table.

PR #3359, opened 2024-03-10
Added support for updating columns in `databricks_sql_table` resource

edwardfeng-db:
## Changes

- Addresses https://github.com/databricks/terraform-provider-databricks/issues/3341
- Add support for update `ColumnInfo` fields by issuing `ALTER` commands

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -293,6 +293,33 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
 	return strings.Join(statements, "")
 }
 
+// Wrapping the column name with backtiks to avoid special character messing things up.
+func getWrappedColumnName(ci SqlColumnInfo) string {
+	return fmt.Sprintf("`%s`", ci.Name)
+}
+
+func (ti *SqlTableInfo) getStatementsForColumnDiffs(oldti *SqlTableInfo, statements []string, typestring string) {
+	// TODO: take out "force_new" in `column` and add case to handle addition and removal of columns.
+	for i, ci := range ti.ColumnInfos {
+		oldCi := oldti.ColumnInfos[i]
+		if ci.Name != oldCi.Name {
+			statements = append(statements, fmt.Sprintf("ALTER %s %s RENAME COLUMN %s to %s", typestring, ti.SQLFullName(), getWrappedColumnName(oldCi), getWrappedColumnName(ci)))
+		}
+		if ci.Comment != oldCi.Comment {
+			statements = append(statements, fmt.Sprintf("ALTER %s %s ALTER COLUMN %s COMMENT %s", typestring, ti.SQLFullName(), getWrappedColumnName(ci), parseComment(ci.Comment)))
+		}
+		if ci.Nullable != oldCi.Nullable {
+			var keyWord string
+			if ci.Nullable {
+				keyWord = "SET"
+			} else {
+				keyWord = "DROP"
+			}
+			statements = append(statements, fmt.Sprintf("ALTER %s %s ALTER COLUMN %s %s NULLABLE", typestring, ti.SQLFullName(), getWrappedColumnName(ci), keyWord))
+		}
+	}
+}
+
 func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
 	statements := make([]string, 0)
 	typestring := ti.getTableTypeString()
@@ -331,6 +358,8 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
 		// Next handle property changes and additions
 		statements = append(statements, fmt.Sprintf("ALTER %s %s SET TBLPROPERTIES (%s)", typestring, ti.SQLFullName(), ti.serializeProperties()))
 	}
+
+	ti.getStatementsForColumnDiffs(oldti, statements, typestring)
 
 	return statements, nil
 }
@@ -384,6 +413,27 @@ func (ti *SqlTableInfo) applySql(sqlQuery string) error {
 	return nil
 }
 
+func columnChangesCustomizeDiff(d *schema.ResourceDiff) error {
+	if d.HasChange("column") {
+		old, new := d.GetChange("column")
+		oldCols := old.([]interface{})
+		newCols := new.([]interface{})
+
+		// Only handling same number of columns for now, will address different number of columns as a follow-up.
+		if len(oldCols) == len(newCols) {
+			for i, oldCol := range oldCols {
+				oldColMap := oldCol.(map[string]interface{})
+				newColMap := newCols[i].(map[string]interface{})
+
+				if oldColMap["type"] != newColMap["type"] {
+					return fmt.Errorf("changing the 'type' of an existing column is not supported")
+				}
+			}
+		}
+	}
+	return nil
+}
+
 func ResourceSqlTable() common.Resource {
 	tableSchema := common.StructToSchema(SqlTableInfo{},
 		func(s map[string]*schema.Schema) map[string]*schema.Schema {
@@ -406,6 +456,10 @@ func ResourceSqlTable() common.Resource {
 	return common.Resource{
 		Schema: tableSchema,
 		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff) error {
+			err := columnChangesCustomizeDiff(d)
+			if err != nil {
+				return err
+			}
 			if d.HasChange("properties") {
 				old, new := d.GetChange("properties")
 				oldProps := old.(map[string]any)
--- a/catalog/resource_sql_table_test.go
+++ b/catalog/resource_sql_table_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/databricks/terraform-provider-databricks/qa"
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/exp/slices"
 )
 
 func TestResourceSqlTableCreateStatement_External(t *testing.T) {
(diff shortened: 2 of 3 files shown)


mgyucht (commented):
Great that we were able to implement this. We should look into changing the set of columns as a follow-up. There will obviously be some challenges there. For example, we may not be able to allow users to modify columns and add/remove columns in the same plan; otherwise, it could become ambiguous if a change indicates that a column should be dropped and then added or an existing column should simply be renamed. However, even if it takes two applies to make an arbitrary change rather than one, it is likely sufficient for customers.

Please add integration tests covering all of these changes to verify the behavior is as expected.

catalog/resource_sql_table.go
```diff
@@ -332,6 +332,35 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
...
+		for i, ci := range ti.ColumnInfos {
+			oldCi := oldti.ColumnInfos[i]
+			if ci.Name != oldCi.Name {
+				println("name not equal!")
+				statements = append(statements, fmt.Sprintf("ALTER %s %s RENAME COLUMN %s to %s", typestring, ti.SQLFullName(), oldCi.Name, ci.Name))
+			}
+			if ci.Type != oldCi.Type {
+				println("type not equal!")
+
+				statements = append(statements, fmt.Sprintf("ALTER %s %s ALTER COLUMN %s TYPE %s", typestring, ti.SQLFullName(), ci.Name, ci.Type))
```
nkvuong, line 346: does not look like this syntax is [supported](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-alter-table-manage-column.html#alter-column-clause) - most of the time it requires rewriting the full table to change the type of a column

catalog/resource_sql_table.go
```diff
@@ -332,6 +332,35 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
...
+				statements = append(statements, fmt.Sprintf("ALTER %s %s ALTER COLUMN %s COMMENT %s", typestring, ti.SQLFullName(), ci.Name, parseComment(ci.Comment)))
+			}
+			if ci.Nullable != oldCi.Nullable {
+				var keyWord string
+				if ci.Nullable {
+					keyWord = "SET"
+				} else {
+					keyWord = "UNSET"
+				}
+				statements = append(statements, fmt.Sprintf("ALTER %s %s ALTER COLUMN %s %s NULLABLE", typestring, ti.SQLFullName(), ci.Name, keyWord))
```
nkvuong, line 359: the syntax is `{ SET | DROP } NOT NULL`

catalog/resource_sql_table.go
```diff
@@ -332,6 +332,35 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
 		statements = append(statements, fmt.Sprintf("ALTER %s %s SET TBLPROPERTIES (%s)", typestring, ti.SQLFullName(), ti.serializeProperties()))
 	}
 
+	// Since we marked `column_info` with `force_new`, this should only include the case where the size of the list is unchanged.
+	if !reflect.DeepEqual(ti.ColumnInfos, oldti.ColumnInfos) {
+		for i, ci := range ti.ColumnInfos {
+			oldCi := oldti.ColumnInfos[i]
+			if ci.Name != oldCi.Name {
+				println("name not equal!")
+				statements = append(statements, fmt.Sprintf("ALTER %s %s RENAME COLUMN %s to %s", typestring, ti.SQLFullName(), oldCi.Name, ci.Name))
```
nkvuong, line 341: the column name needs to be wrapped inside backticks, in case it contains conflicting characters

nkvuong (changes requested):
the requests we often receive are to add new columns to a table, so need to plan how to support it.

similar to @mgyucht , I think we should have 2 scenarios
- number of columns unchanged - treat this as rename operations based on position
- number of columns changed - treat this as add/remove operations

catalog/resource_sql_table.go
```diff
@@ -332,6 +332,35 @@ func (ti *SqlTableInfo) diff(oldti *SqlTableInfo) ([]string, error) {
 		statements = append(statements, fmt.Sprintf("ALTER %s %s SET TBLPROPERTIES (%s)", typestring, ti.SQLFullName(), ti.serializeProperties()))
 	}
 
+	// Since we marked `column_info` with `force_new`, this should only include the case where the size of the list is unchanged.
+	if !reflect.DeepEqual(ti.ColumnInfos, oldti.ColumnInfos) {
```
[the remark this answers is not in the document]
edwardfeng-db: Yeah, not necessary, took it out

catalog/resource_sql_table.go
```diff
@@ -406,6 +430,23 @@ func ResourceSqlTable() common.Resource {
 	return common.Resource{
 		Schema: tableSchema,
 		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff) error {
+			if d.HasChange("column") {
```
mgyucht, line 433: nit: to prevent CustomizeDiff from getting too long, let's split this out into a separate function, esp. given that we will be adding additional verification after this for adding/removing columns.

catalog/resource_sql_table.go
```diff
@@ -406,6 +430,23 @@ func ResourceSqlTable() common.Resource {
...
+				old, new := d.GetChange("column")
+				oldCols := old.([]interface{})
+				newCols := new.([]interface{})
```
mgyucht, line 436: I think these are `[]map[string]any`, let's use that instead, saving you the second typecheck below.
```suggestion
				oldCols := old.([]map[string]any)
				newCols := new.([]map[string]any)
```

catalog/resource_sql_table.go
```diff
@@ -406,6 +430,23 @@ func ResourceSqlTable() common.Resource {
...
+
+				// Only handling same number of columns for now, will address different number of columns as a follow-up.
+				if len(oldCols) == len(newCols) {
+					for i, oldCol := range oldCols {
+						oldColMap := oldCol.(map[string]interface{})
+						newColMap := newCols[i].(map[string]interface{})
+
+						if oldColMap["type"] != newColMap["type"] {
+							return fmt.Errorf("changing the 'type' of an existing column is not allowed, please apply twice by dropping and adding a new column")
```
mgyucht, line 445: This recommendation will cause the table to be dropped and recreated today. I would leave out the recommendation for now until we actually support removing and adding columns, and just say that changing the type is not supported.

catalog/resource_sql_table_test.go
```diff
@@ -587,6 +588,176 @@ func TestResourceSqlTableUpdateView_Comments(t *testing.T) {
 	assert.Equal(t, "barview", d.Get("name"))
 }
 
+func TestResourceSqlTableUpdateTable_Columns(t *testing.T) {
```
mgyucht, line 591: Please add a unit test covering change in nullability.
I'd also like to see an example where multiple changes in a single column are requested.

internal/acceptance/sql_table_test.go
```diff
@@ -264,3 +266,81 @@ func TestUcAccResourceSqlTable_Liquid(t *testing.T) {
...
+				nullable  = %t
+				comment   = "%s"
+			}
+			comment = "this table is managed by terraform"
+		}`, columnName, typeName, nullable, comment)
+}
+
+func TestUcAccResourceSqlTable_RenameColumn(t *testing.T) {
+	if os.Getenv("GOOGLE_CREDENTIALS") != "" {
+		t.Skipf("databricks_sql_table resource not available on GCP")
```
mgyucht, line 299: nit: use `skipf(t)(msg)` which will cause debuggers to report the test as a failure.

internal/acceptance/sql_table_test.go
```diff
@@ -264,3 +266,81 @@ func TestUcAccResourceSqlTable_Liquid(t *testing.T) {
 		}`,
 	})
 }
+
+func constructManagedSqlTableTemplate(columnName string, typeName string, nullable bool, comment string) string {
```
mgyucht, line 270: Suggestion: this will be much more flexible if we define it as
```go
func constructManagedSqlTableTemplate(columns []catalog.SqlColumnInfo) string {
  ...
}
```
It will extend to the next set of tests that cover changing the number of columns.

internal/acceptance/sql_table_test.go
```diff
@@ -264,3 +266,81 @@ func TestUcAccResourceSqlTable_Liquid(t *testing.T) {
...
+
+func TestUcAccResourceSqlTable_ChangeColumnTypeThrows(t *testing.T) {
+	if os.Getenv("GOOGLE_CREDENTIALS") != "" {
+		t.Skipf("databricks_sql_table resource not available on GCP")
+	}
+	pattern := "changing the 'type' of an existing column is not allowed, please apply twice by dropping and adding a new column"
+	r, err := regexp.Compile(pattern)
+	if err != nil {
+		panic(err)
+	}
```
mgyucht, line 338:
```suggestion
	r := regexp.MustCompile(pattern)
```

nkvuong (commented):
a few minor comments, but looks good to me

catalog/resource_sql_table_test.go
```diff
@@ -587,6 +588,176 @@ func TestResourceSqlTableUpdateView_Comments(t *testing.T) {
...
+						{
+							Name:     "one",
+							Type:     "string",
+							Comment:  "old comment",
+							Nullable: false,
+						},
+					},
+				},
+			},
+			{
```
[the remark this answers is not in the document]
edwardfeng-db: Removed, good point

catalog/resource_sql_table_test.go
```diff
@@ -587,6 +588,176 @@ func TestResourceSqlTableUpdateView_Comments(t *testing.T) {
...
+			"column.#":           "2",
+			"column.0.name":      "one",
+			"column.0.type":      "string",
+			"column.0.comment":   "old comment",
+			"column.0.nullable":  "false",
+			"column.1.name":      "two",
+			"column.1.type":      "string",
+			"column.1.nullable":  "true",
+		},
+		Fixtures: append([]qa.HTTPFixture{
```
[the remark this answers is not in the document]
edwardfeng-db: I can do that in a follow up because this also references `createClusterForSql` which is shared by a lot of other test cases

catalog/resource_sql_table.go
edwardfeng-db: Tried this but ran into some errors doing type casting, so going back to my old approach

internal/acceptance/sql_table_test.go
```diff
@@ -264,3 +268,100 @@ func TestUcAccResourceSqlTable_Liquid(t *testing.T) {
...
+	unityWorkspaceLevel(t, step{
+		Template: constructManagedSqlTableTemplate([]catalog.SqlColumnInfo{{Name: "name", Type: "string", Nullable: true, Comment: "comment"}}),
+	}, step{
+		Template: constructManagedSqlTableTemplate([]catalog.SqlColumnInfo{{Name: "new_name", Type: "string", Nullable: true, Comment: "comment"}}),
+	})
+}
+
+func TestUcAccResourceSqlTable_AddColumnComment(t *testing.T) {
+	if os.Getenv("GOOGLE_CREDENTIALS") != "" {
+		skipf(t)("databricks_sql_table resource not available on GCP")
```
nkvuong, line 323: is this true?

edwardfeng-db: Following the patterns for all other tests for now, we can address them all as a follow up

catalog/resource_sql_table.go
```diff
@@ -293,6 +293,33 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
 	return strings.Join(statements, "")
 }
 
+// Wrapping the column name with backtiks to avoid special character messing things up.
```
mgyucht, line 296:
```suggestion
// Wrapping the column name with backticks to avoid special character messing things up.
```

mgyucht (approved):
LGTM! Thanks for this @edwardfeng-db!

Issue #3351, opened 2024-03-08
[ISSUE] `StructToData` doesn't persist empty structs in the state leading to the configuration drift


alexott:
For online tables resource I've got a problem when the `spec.run_triggered` block is defined as an empty struct, and although it's returned in the response, it's not persisted in the state.

Response from backend:

```
< HTTP/2.0 200 OK
< {
<   "name": "main.default.online_table2",
<   "spec": {
<     "destination_table_full_name": "main.default.online_table2",
<     "pipeline_id": "34231f9e-ac77-4003-b66a-5521a6f3e495",
<     "primary_key_columns": [
<       "id"
<     ],
<     "run_triggered": {},
<     "source_table_full_name": "main.default.vs_test"
<   },
<   "status": {
<     "detailed_state": "ONLINE_TRIGGERED_UPDATE",
<     "message": "Online Table is online but is currently is in the process of re-syncing initial data. Check latest status in Delta Live Tables: https://adb-aaaa.17.azuredatabricks.net#joblist/pipelines/34231f9e-ac77-4003-b66a-5521a6f3e495/updates/f6d3f087-69b4-4fd0-b511-6c8811b9e17c",
<     "triggered_update_status": {
<       "last_processed_commit_version": 3,
<       "timestamp": "2024-03-08T13:18:38Z",
<       "triggered_update_progress": {
<         "latest_version_currently_processing": 3,
<         "sync_progress_completion": 1,
<         "synced_row_count": 200,
<         "total_row_count": 200
<       }
<     }
<   }
< }: tf_req_id=8fc59289-1e43-fd48-5755-73ab751a57bc tf_resource_type=databricks_online_table tf_rpc=ApplyReso
```

Data isn't persisted in the state:

```
      - .spec[0].run_triggered: block count changed from 1 to 0
```

Leads to recreation (there is no Update API):

```
         + run_triggered {} # forces replacement
```

PR #3362, opened 2024-03-11
Support persisting empty fields in StructToData

mgyucht:
## Changes
Some APIs legitimately return empty objects in fields, but StructToData does not persist these objects (e.g. #3351). This is caused by a filter in collectionToMaps that removes data structures that don't have any non-optional fields or optional fields set to a non-empty/non-nil value.

This PR removes this check. I need to do a bit more work to verify the impact of removing this on the rest of the provider.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/reflect_resource.go
+++ b/common/reflect_resource.go
@@ -613,9 +613,6 @@ func collectionToMaps(v any, s *schema.Schema, aliases map[string]map[string]str
 		if err != nil {
 			return nil, err
 		}
-		if len(data) == 0 {
-			continue
-		}
 		resultList = append(resultList, data)
 	}
 	return resultList, nil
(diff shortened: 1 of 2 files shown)


hectorcast-db:
We should be careful of APIs which "default" to empty structs / do not differentiate between not-set and empty struct. 
If a user does not set the value, and the API returns an empty struct, there will be always a diff.

alexott:
@hectorcast-db can we merge it, or we need to work on it further?

hectorcast-db (approved):

hectorcast-db:
@alexott @mgyucht 
We discussed offline. It looks good to me, but I think that @mgyucht still wanted to run some tests.

mgyucht:
I was concerned that this might exacerbate some issues with suppress diff, as it is will add empty dictionaries into state which might need diff suppression. However, I realized that that would only be needed if the API returned non-nil/empty values, which would always end up in the state anyways. So I think that isn't a concern, so we can merge this.

PR #3373, opened 2024-03-15
Port `databricks_secret_scope`, `databricks_secret` and `databricks_secret_acl` to Go SDK

alexott:
## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/secret_scope_test.go
+++ b/internal/acceptance/secret_scope_test.go
@@ -39,7 +39,7 @@ func TestAccSecretScopeResource(t *testing.T) {
 					assert.Equal(t, 1, len(acls))
 					assert.Equal(t, me.UserName, acls[0].Principal)
 
-					err = w.Secrets.DeleteScope(context.Background(), workspace.DeleteScope{Scope: scope})
+					err = w.Secrets.DeleteScope(ctx, workspace.DeleteScope{Scope: scope})
 					assert.NoError(t, err)
 					return nil
 				}),
--- a/internal/acceptance/secret_scope_test.go
+++ b/internal/acceptance/secret_scope_test.go
@@ -6,8 +6,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/databricks/databricks-sdk-go/service/workspace"
 	"github.com/databricks/terraform-provider-databricks/qa"
-	"github.com/databricks/terraform-provider-databricks/secrets"
 
 	"github.com/databricks/terraform-provider-databricks/common"
 
@@ -29,19 +29,17 @@ func TestAccSecretScopeResource(t *testing.T) {
 			resource.TestCheckResourceAttr("databricks_secret_scope.my_scope", "backend_type", "DATABRICKS"),
 			resourceCheck("databricks_secret_scope.my_scope",
 				func(ctx context.Context, client *common.DatabricksClient, id string) error {
-					secretACLAPI := secrets.NewSecretAclsAPI(ctx, client)
-					acls, err := secretACLAPI.List(id)
-					require.NoError(t, err)
-
 					w, err := client.WorkspaceClient()
 					require.NoError(t, err)
-
+					acls_resp, err := w.Secrets.ListAclsByScope(ctx, id)
+					require.NoError(t, err)
+					acls := acls_resp.Items
 					me, err := w.CurrentUser.Me(ctx)
 					require.NoError(t, err)
 					assert.Equal(t, 1, len(acls))
 					assert.Equal(t, me.UserName, acls[0].Principal)
 
-					err = secrets.NewSecretScopesAPI(context.Background(), client).Delete(scope)
+					err = w.Secrets.DeleteScope(context.Background(), workspace.DeleteScope{Scope: scope})
 					assert.NoError(t, err)
 					return nil
 				}),
--- a/secrets/resource_secret.go
+++ b/secrets/resource_secret.go
@@ -35,46 +35,45 @@ func readSecret(ctx context.Context, w *databricks.WorkspaceClient, scope string
 // ResourceSecret manages secrets
 func ResourceSecret() common.Resource {
 	p := common.NewPairSeparatedID("scope", "key", "|||")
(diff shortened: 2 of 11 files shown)


internal/acceptance/secret_scope_test.go
```diff
@@ -29,19 +29,17 @@ func TestAccSecretScopeResource(t *testing.T) {
...
+					acls_resp, err := w.Secrets.ListAclsByScope(ctx, id)
+					require.NoError(t, err)
+					acls := acls_resp.Items
 					me, err := w.CurrentUser.Me(ctx)
 					require.NoError(t, err)
 					assert.Equal(t, 1, len(acls))
 					assert.Equal(t, me.UserName, acls[0].Principal)
 
-					err = secrets.NewSecretScopesAPI(context.Background(), client).Delete(scope)
+					err = w.Secrets.DeleteScope(context.Background(), workspace.DeleteScope{Scope: scope})
```
mgyucht, line 42: Can you not use `ctx` here?

secrets/resource_secret.go
```diff
@@ -147,8 +66,16 @@ func ResourceSecret() common.Resource {
...
-				d.Get("key").(string)); err != nil {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			err = w.Secrets.PutSecret(ctx, workspace.PutSecret{
+				Scope:       d.Get("scope").(string),
+				Key:         d.Get("key").(string),
+				StringValue: d.Get("string_value").(string),
+			})
```
mgyucht, line 77: I think this is fine, but most of our other resources do this as:
```go
var putSecretReq workspace.PutSecret
common.DataToStructPointer(d, s, &putSecretReq)
err = w.Secrets.PutSecret(ctx, putSecretReq)
```
Does this not work?

secrets/resource_secret_acl.go
```diff
@@ -78,9 +33,16 @@ func ResourceSecretACL() common.Resource {
...
 		},
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			if err := NewSecretAclsAPI(ctx, c).Create(
-				d.Get("scope").(string), d.Get("principal").(string),
-				ACLPermission(d.Get("permission").(string))); err != nil {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			err = w.Secrets.PutAcl(ctx, workspace.PutAcl{
```
mgyucht, line 40: Ditto, can we use `DataToStructPointer`?

secrets/resource_secret_acl_test.go
```diff
@@ -14,8 +15,8 @@ func TestResourceSecretACLRead(t *testing.T) {
 			{
 				Method:   "GET",
 				Resource: "/api/2.0/secrets/acls/get?principal=something&scope=global",
-				Response: ACLItem{
-					Permission: "CAN_MANAGE",
+				Response: workspace.AclItem{
+					Permission: "MANAGE",
```
mgyucht, line 19: Was this a typo in the permission?

secrets/resource_secret_scope.go
```diff
@@ -5,99 +5,39 @@ import (
...
-		"scope": scope,
-	}, nil)
-}
-
-// List lists all secret scopes available in the workspace
-func (a SecretScopesAPI) List() ([]SecretScope, error) {
-	var listSecretScopesResponse SecretScopeList
-	err := a.client.Get(a.context, "/secrets/scopes/list", nil, &listSecretScopesResponse)
-	return listSecretScopesResponse.Scopes, err
+	Name                   string                                      `json:"name" tf:"force_new"`
```
mgyucht, line 19: I think this is fine, but did you consider doing something like
```
type SecretScope workspace.CreateScope

func (s SecretScope) CustomizeSchema(m map[string]string) map[string]string {
  s["name"].ValidateFunc = validScope
  common.CustomizeSchemaPath(m, "backend_type").SetComputed()
  common.CustomizeSchemaPath(m, "initial_manage_principal").SetForceNew()
  common.CustomizeSchemaPath(m, "keyvault_metadata").SetForceNew()
}

func (s SecretScope) Aliases() map[string]string {
  return map[string]string{
    "scope": "name",
    "backend_azure_keyvault": "keyvault_metadata",
  }
}
```

This way, you don't need to adapt one structure into another, and your main structure is based off of the Go SDK, so it inherits any changes from the Go SDK automatically.

secrets/resource_secret_acl_test.go
alexott: Yes, it was bug in tests

secrets/resource_secret_scope.go
alexott: Hmmm, I can't make it work - I'm getting empty values.

secrets/resource_secret_scope.go
```diff
@@ -49,13 +60,7 @@ var validScope = validation.StringMatch(regexp.MustCompile(`^[\w\.@_/-]{1,128}$`
 
 // ResourceSecretScope manages secret scopes
 func ResourceSecretScope() common.Resource {
-	s := common.StructToSchema(SecretScope{}, func(s map[string]*schema.Schema) map[string]*schema.Schema {
-		// TODO: DiffSuppressFunc for initial_manage_principal & importing
```
mgyucht, line 113: I don't think this is needed because you can't update this resource (is that intentional?)

mgyucht (approved):
LGTM. Pushed a couple small changes with what I was thinking, LMK what you think. If you're happy merge, if not feel free to modify what you need and we can revisit Monday.

secrets/resource_secret_scope.go
alexott: basically any change in the configuration makes it a different resource, plus we don't have Update API...

alexott:
Thank you for making these changes! I really had identical changes, but for me, for unknow reason, the `scope` <-> `name` translation didn't work correctly

PR #3374, opened 2024-03-15
SCIM patch to remove resources must not contain a value field

mgyucht:
## Changes
Integration tests `TestMwsAccGroupMemberResource` and `TestAccGroupMemberResource` started failing recently with error
```
       Error: cannot delete group member: Unrecognized field "value" (class com.unboundid.scim2.common.messages.PatchOperation$RemoveOperation), not marked as ignorable (one known property: "path"])
         at [Source: (String)"{"schemas":["urn:ietf:params:scim:api:messages:2.0:PatchOp"],"Operations":[{"op":"remove","path":"members[value eq \"268458960998283\"]","value":""}]}"; line: 1, column: 147] (through reference chain: com.unboundid.scim2.common.messages.PatchRequest["Operations"]->java.util.ArrayList[0]->com.unboundid.scim2.common.messages.PatchOperation$RemoveOperation["value"])
```
It seems the SCIM API has stricter validation requirements than in the past. We currently include `"value": ""` in the request to remove a user from a group, but this is no longer permitted.

This PR removes the `value` attribute from the patch operation for SCIM API calls used during remove. There are two functions:
* `PatchRequest`: used for patch requests without a value, like removes.
* `PatchRequestWithValue`: used for patch requests with a value.
Everywhere that value was `""` is now changed to use `PatchRequest`, ensuring that no value is included in the request.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott (commented):
lgtm

tanmay-db (approved):

PR #3376, opened 2024-03-18
Migrated `databricks_cluster` resource to Go SDK 

tanmay-db:
## Changes

Migrated cluster resource to Go SDK along with migrating major references of libraries api and method across clusters, jobs, exporter etc.

Note: Many resources use old clusters API, methods -- migration for these would be part of when we will be migrating those resources to Go SDK to account for resource specific OpenAPI changes, extra methods at that time like we did with the libraries resource. 

## Tests

Unit tests, Nightly tests are running ... 
- [x] `make test` run locally
- [x] using Go SDK


libraries/libraries_api.go
```diff
@@ -1,164 +1,12 @@
...
-	Coordinates string   `json:"coordinates"`
-	Repo        string   `json:"repo,omitempty"`
-	Exclusions  []string `json:"exclusions,omitempty"`
-}
-
-// Cran is a R library hosted on Maven
-type Cran struct {
-	Package string `json:"package"`
-	Repo    string `json:"repo,omitempty"`
+type LibraryList struct {
```
tanmay-db, line 7: tofix

clusters/resource_cluster.go
```diff
@@ -94,7 +149,8 @@ func (ClusterSpec) CustomizeSchema(s map[string]*schema.Schema) map[string]*sche
 	common.CustomizeSchemaPath(s, "azure_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "gcp_attributes"})
 	common.CustomizeSchemaPath(s, "gcp_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "azure_attributes"})
 
-	common.CustomizeSchemaPath(s).AddNewField("library", common.StructToSchema(libraries.ClusterLibraryList{},
+	// TODO: use compute.InstallLibraries -- fails
```
tanmay-db, line 152: same as below

clusters/resource_cluster.go
```diff
@@ -58,6 +60,59 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	return false
 }
 
+func Validate(cluster compute.CreateCluster) error {
```
mgyucht, line 63: nit: for the duplicated functions, leave a comment here and on the original functions to link them to one another while both exist, so that people know that changes need to be made in both.

clusters/resource_cluster.go
```diff
@@ -156,93 +211,110 @@ func resourceClusterSchema() map[string]*schema.Schema {
...
-	clusters := NewClustersAPI(ctx, c)
+	w, err := c.WorkspaceClient()
+	if err != nil {
+		return err
+	}
+	clusters := w.Clusters
+	var cluster compute.CreateCluster
 	common.DataToStructPointer(d, clusterSchema, &cluster)
-	if err := cluster.Validate(); err != nil {
+	if err := Validate(cluster); err != nil {
```
mgyucht, line 223: Not your issue, but shouldn't this be in CustomizeDiff? Cluster spec should be validated there so we fail before applying.

clusters/resource_cluster.go
```diff
@@ -323,36 +399,36 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
...
 		}
 		if err != nil {
 			return err
 		}
-
-	} else {
-		clusterInfo, err = clusters.Get(clusterID)
-		if err != nil {
-			return err
-		}
```
mgyucht, line 347: Did you mean to remove this branch?

clusters/resource_cluster.go
```diff
@@ -363,38 +439,54 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
...
-			if _, err = clusters.StartAndGetInfo(clusterID); err != nil {
+			if _, err = clusters.StartByClusterIdAndWait(ctx, clusterId); err != nil {
 				return err
 			}
 		}
 		// clusters.StartAndGetInfo() always returns a running cluster
 		// or errors out, so we just know the cluster is active.
-		err = librariesAPI.UpdateLibraries(clusterID, libsToInstall, libsToUninstall,
-			d.Timeout(schema.TimeoutUpdate))
+		// TODO: Wait needed here?
```
mgyucht, line 463: This seems like a nice improvement, but I would like to decouple behavioral changes from this PR if possible. Otherwise, if we decide we actually don't want to wait on library installation, we won't have to revert all of this work.

clusters/resource_cluster_test.go
```diff
@@ -531,7 +532,8 @@ func TestResourceClusterRead_NotFound(t *testing.T) {
 		Read:     true,
 		Removed:  true,
 		ID:       "abc",
-	}.ApplyNoError(t)
+	}.Apply(t)
+	qa.AssertErrorStartsWith(t, err, "Cluster abc does not exist") // TODO: Why was this passing before?
```
mgyucht, line 536: This shouldn't fail now, right?

clusters/resource_cluster_test.go
```diff
@@ -1379,7 +1383,7 @@ func TestResourceClusterDelete_Error(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "POST",
-				Resource: "/api/2.0/clusters/delete",
+				Resource: "/api/2.0/clusters/permanent-delete",
```
mgyucht, line 1386: What is this change?

clusters/resource_cluster.go
```diff
@@ -363,38 +439,54 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
...
+	libsClusterStatus, err := w.Libraries.ClusterStatusByClusterId(ctx, clusterId)
+	if err != nil {
+		return err
+	}
+	libsToInstall, libsToUninstall := libraries.GetLibrariesToInstallAndUninstall(libraryList, libsClusterStatus)
+	clusterInfo, err = clusters.GetByClusterId(ctx, clusterId)
+	err = wrapMissingClusterError(err, d.Id())
 	if err != nil {
 		return err
 	}
```
mgyucht, line 454: It looks very wrong to not have `if err != nil` immediately after it is returned.
```suggestion
	if err != nil {
		return wrapMissingClusterError(err, d.Id())
	}
```

libraries/libraries_api.go
```diff
@@ -177,85 +25,42 @@ func NewLibraryFromInstanceState(i any) (lib Library) {
...
-	inState := map[string]Library{}
+	inState := map[string]compute.Library{}
 	for _, status := range cls.LibraryStatuses {
 		lib := *status.Library
 		inState[lib.String()] = lib
 	}
-	toInstall := ClusterLibraryList{ClusterID: cll.ClusterID}
-	toUninstall := ClusterLibraryList{ClusterID: cll.ClusterID}
+	toInstall := compute.InstallLibraries{ClusterId: cll.ClusterId}
+	// TODO: UninstallLibraries doesn't have sort() method
```
mgyucht, line 62: You don't need a sort method to be defined on the type. You can use `slices.SortFunc()`. But why not just return `([]Library, []Library)`?

clusters/resource_cluster.go
```diff
@@ -363,38 +439,54 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
...
+			Uninstall: libsToUninstall.Libraries,
+		})
 		if err != nil {
 			return err
 		}
 		if clusterInfo.State == ClusterStateTerminated {
-			log.Printf("[INFO] %s was in TERMINATED state, so terminating it again", clusterID)
-			if err = clusters.Terminate(clusterID); err != nil {
+			log.Printf("[INFO] %s was in TERMINATED state, so terminating it again", clusterId)
+			if err = clusters.DeleteByClusterId(ctx, clusterId); err != nil {
```
mgyucht, line 474: Are delete & terminate the same?

mgyucht (commented):
Some comments, but otherwise looking good!

clusters/resource_cluster.go
tanmay-db: Yes you are right, it's has been marked as a TODO since long time back: https://github.com/databricks/terraform-provider-databricks/blob/main/clusters/clusters_api.go#L437. I have created a followup ticket to do this so we don't risk introducing behaviour changes in this PR.

clusters/resource_cluster_test.go
tanmay-db: Earlier what we used to do is explicitly terminate the cluster and then call permanent delete (permanent delete automatically terminates the clusters so we don't need to do it explicitly now). This pathway is hit because this test is about returning the error while deletion happens which hits the delete (terminate) end point. As now we don't do it, we change it to permanent-delete.

clusters/resource_cluster.go
tanmay-db: Yes, https://docs.databricks.com/api/workspace/clusters/delete
image

clusters/resource_cluster_test.go
tanmay-db: Yes

clusters/resource_cluster.go
tanmay-db: Yes, I moved it down: `clusterInfo, err = clusters.GetByClusterId(ctx, clusterId)` because we don't use clusterInfo before it. But we would want to check if we are able to successfully get the cluster before pinning / unpinning the cluster. Move it back.

clusters/resource_cluster.go
tanmay-db: This won't be a behaviour change as the inner functionality is same. Uninstall -> Install -> Wait

clusters/resource_cluster.go
```diff
@@ -58,8 +59,72 @@ func ZoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
...
+	oldInstancePool, newInstancePool := d.GetChange("instance_pool_id")
+	oldDriverPool, newDriverPool := d.GetChange("driver_instance_pool_id")
+	if oldInstancePool != newInstancePool &&
+		oldDriverPool == oldInstancePool &&
+		oldDriverPool == newDriverPool {
+		cluster.DriverInstancePoolId = cluster.InstancePoolId
+	}
+}
+
+type LibraryList struct {
```
tanmay-db, line 121: We need this type (with struct annotations) below, so we have to explicitly declare it. 
`````
var libraryList LibraryList
common.DataToStructPointer(d, clusterSchema, &libraryList)
`````
Given it's only used in clusters and to not encourage wide use, I think we should keep it here as compared to libraries_api.

Comment (tanmay-db) [reply clusters/resource_cluster.go:121]: cc: @miles, I am not sure if there is a better / easier way to do this, we could probably get the type from ClusterSpec.Libraries using reflection but I think having a 2 line struct would be simpler.

Comment (tanmay-db) [reply clusters/resource_cluster.go]: Updated to use `ClusterSpec`

File: clusters/resource_cluster.go [hunk 13]
```diff
@@ -156,93 +206,105 @@ func resourceClusterSchema() map[string]*schema.Schema {
...
-		err = clusters.Pin(clusterInfo.ClusterID)
+		err = clusters.PinByClusterId(ctx, d.Id())
 		if err != nil {
 			return err
 		}
 	}
-	var libraryList libraries.ClusterLibraryList
+
+	var libraryList compute.InstallLibraries
 	common.DataToStructPointer(d, clusterSchema, &libraryList)
```
mgyucht, line 244: The point of adding libraries to the schema was that we don't need to call `DataToStructPointer` multiple times, right?
```suggestion
	libraryList := compute.InstallLibraries{
	    ClusterId: clusterInfo.ClusterId,
	    Libraries: cluster.Libraries,
	}
```

clusters/resource_cluster.go
```diff
@@ -363,38 +433,64 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
...
 	if err != nil {
 		return err
 	}
-	libraryList.ClusterID = clusterID
-	libsToInstall, libsToUninstall := libraryList.Diff(libsClusterStatus)
-	if len(libsToUninstall.Libraries) > 0 || len(libsToInstall.Libraries) > 0 {
+
+	var libraryList ClusterSpec
+	common.DataToStructPointer(d, clusterSchema, &libraryList)
+	libsToInstall, libsToUninstall := libraries.GetLibrariesToInstallAndUninstall(libraryList.Libraries, libsClusterStatus)
```
mgyucht, line 443: This as well?
```suggestion
	libsToInstall, libsToUninstall := libraries.GetLibrariesToInstallAndUninstall(cluster.Libraries, libsClusterStatus)
```

mgyucht (approved):
Two very small nits, but otherwise this LGTM!

tanmay-db:
Going ahead with merge for now, need to check issues arising due to differences between ClusterSpec and other Cluster structs.

PR #3381, opened 2024-03-19
Support adding and removing columns in sql table

edwardfeng-db:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


catalog/resource_sql_table.go
```diff
@@ -293,29 +293,55 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
...
+				// Remove old column if old column is no longer found in the config.
+				statements = append(statements, fmt.Sprintf("ALTER %s %s DROP COLUMN IF EXISTS %s", typestring, ti.SQLFullName(), getWrappedColumnName(oldCi)))
+			}
+		}
+
+		for name, newCi := range nameToNewColumn {
+			if _, exists := nameToOldColumn[name]; !exists {
+				// Add new column if new column is detected.
+				newCiStatement := ti.serializeColumnInfo(newCi)
+				statements = append(statements, fmt.Sprintf("ALTER %s %s ADD COLUMN %s ", typestring, ti.SQLFullName(), newCiStatement))
```
nkvuong, line 323: it needs to be `ALTER TABLE table ADD COLUMN column data_type [COMMENT comment]`

nkvuong: and we could batch multiple columns in a single alter statement
`ALTER TABLE table ADD COLUMNS (columnA data_type, columnB data_type)`

catalog/resource_sql_table.go
```diff
@@ -293,29 +293,55 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
...
-			var keyWord string
-			if ci.Nullable {
-				keyWord = "SET"
-			} else {
-				keyWord = "DROP"
+
+		for name, oldCi := range nameToOldColumn {
+			if _, exists := nameToNewColumn[name]; !exists {
+				// Remove old column if old column is no longer found in the config.
+				statements = append(statements, fmt.Sprintf("ALTER %s %s DROP COLUMN IF EXISTS %s", typestring, ti.SQLFullName(), getWrappedColumnName(oldCi)))
```
nkvuong, line 315: same here `DROP COLUMNS IF EXISTS (columnA, columnB)`

catalog/resource_sql_table.go
```diff
@@ -32,7 +32,7 @@ type SqlTableInfo struct {
 	SchemaName            string            `json:"schema_name" tf:"force_new"`
 	TableType             string            `json:"table_type" tf:"force_new"`
 	DataSourceFormat      string            `json:"data_source_format,omitempty" tf:"force_new"`
-	ColumnInfos           []SqlColumnInfo   `json:"columns,omitempty" tf:"alias:column,computed,force_new"`
+	ColumnInfos           []SqlColumnInfo   `json:"columns,omitempty" tf:"alias:column,computed"`
```
mgyucht, line 35: 🎉

catalog/resource_sql_table.go
```diff
@@ -189,7 +189,7 @@ func (ti *SqlTableInfo) serializeColumnInfo(col SqlColumnInfo) string {
 	if col.Comment != "" {
 		comment = fmt.Sprintf(" COMMENT '%s'", parseComment(col.Comment))
 	}
-	return fmt.Sprintf("%s %s%s%s", col.Name, col.Type, notNull, comment) // id INT NOT NULL COMMENT 'something'
+	return fmt.Sprintf("%s %s%s%s", getWrappedColumnName(col), col.Type, notNull, comment) // id INT NOT NULL COMMENT 'something'
```
mgyucht, line 192: FYI: probably a more consistent way to do this is either 1) make a constructor for `ColumnInfo` that escapes the column name, or 2) define a method on `ColumnInfo` that returns the escaped name.

catalog/resource_sql_table.go
```diff
@@ -294,29 +294,66 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
...
 }
 
 func (ti *SqlTableInfo) getStatementsForColumnDiffs(oldti *SqlTableInfo, statements []string, typestring string) []string {
-	// TODO: take out "force_new" in `column` and add case to handle addition and removal of columns.
-	for i, ci := range ti.ColumnInfos {
-		oldCi := oldti.ColumnInfos[i]
-		if ci.Name != oldCi.Name {
-			statements = append(statements, fmt.Sprintf("ALTER %s %s RENAME COLUMN %s to %s", typestring, ti.SQLFullName(), getWrappedColumnName(oldCi), getWrappedColumnName(ci)))
+	if len(ti.ColumnInfos) != len(oldti.ColumnInfos) {
+		nameToOldColumn := make(map[string]SqlColumnInfo)
```
mgyucht, line 304: This method is too long & too deeply nested (code smell). Let's split this into a couple separate methods for readability. I feel like the core logic is:
```go
if len(ti.ColumnInfos) != len(oldti.ColumnInfos) {
  return addOrRemoveColumnStatements()
}
return alterExistingColumnStatements()
```

Much easier to read and review + decoupling independent logical components from one another.

catalog/resource_sql_table.go
```diff
@@ -294,29 +294,66 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
...
+		for name, newCi := range nameToNewColumn {
+			if _, exists := nameToOldColumn[name]; !exists {
+				// Add new column if new column is detected.
+				newCiStatement := ti.serializeColumnInfo(newCi)
+				addColumnStatements = append(addColumnStatements, newCiStatement)
+			}
+		}
+		if len(addColumnStatements) > 0 {
+			addColumnStatementsStr := strings.Join(addColumnStatements, ", ")
+			statements = append(statements, fmt.Sprintf("ALTER %s %s ADD COLUMN (%s)", typestring, ti.SQLFullName(), addColumnStatementsStr))
```
mgyucht, line 337: Should we try to add it at the right position (i.e. `AFTER` the column that immediately precedes it)?

catalog/resource_sql_table.go
```diff
@@ -431,6 +467,24 @@ func columnChangesCustomizeDiff(d *schema.ResourceDiff) error {
 					return fmt.Errorf("changing the 'type' of an existing column is not supported")
 				}
 			}
+		} else {
+			oldColsNameToMap := make(map[string]map[string]interface{})
+			newColsNameToMap := make(map[string]map[string]interface{})
```
mgyucht, line 472: At least for the `newColsNameToMap`, can we use `DiffToStructPointer`? I know our infra doesn't support getting the old version yet.

catalog/resource_sql_table.go
```diff
@@ -431,6 +467,24 @@ func columnChangesCustomizeDiff(d *schema.ResourceDiff) error {
 					return fmt.Errorf("changing the 'type' of an existing column is not supported")
 				}
 			}
+		} else {
+			oldColsNameToMap := make(map[string]map[string]interface{})
```
mgyucht, line 471: Ditto: this is really nested. Rule of thumb is 2 levels of indentation, with a 3rd allowed for `if err != nil {}` checks because Golang. More than that means that the function complexity is probably too high, or it is difficult to review.

catalog/resource_sql_table.go
```diff
@@ -431,6 +467,24 @@ func columnChangesCustomizeDiff(d *schema.ResourceDiff) error {
...
+				oldColMap := oldCol.(map[string]interface{})
+				oldColsNameToMap[oldColMap["name"].(string)] = oldColMap
+			}
+			for _, newCol := range newCols {
+				newColMap := newCol.(map[string]interface{})
+				newColsNameToMap[newColMap["name"].(string)] = newColMap
+			}
+			for name, oldColMap := range oldColsNameToMap {
+				if newColMap, exists := newColsNameToMap[name]; exists {
+					if oldColMap["type"] != newColMap["type"] || oldColMap["nullable"] != newColMap["nullable"] || oldColMap["comment"] != newColMap["comment"] {
```
mgyucht, line 483: This seems fragile.

catalog/resource_sql_table_test.go
```diff
@@ -726,6 +726,484 @@ func TestResourceSqlTableUpdateTable_ColumnsTypeThrowsError(t *testing.T) {
 	assert.EqualError(t, err, "changing the 'type' of an existing column is not supported")
 }
 
+func TestResourceSqlTableUpdateTable_ColumnsAdditionAndUpdateThrowsError(t *testing.T) {
```
mgyucht, line 729: This is 500 lines of test code for 5 tests, but what is actually being tested is around half of that. What can we do to reduce boilerplate in these tests, so they are easier to write/extend in the future and easier to review now? I think there are a lot of opportunities here.

mgyucht (commented):
This seems correct, but I think there is room for improvement in terms of code quality.

catalog/resource_sql_table.go
edwardfeng-db: Yeah, I agree, next step could be to make it possible to directly access data through the go struct instead of the map, but I'll probably deal with this in the future PRs

catalog/resource_sql_table.go
edwardfeng-db: Yeah good point, we should probably preserve the ordering

catalog/resource_sql_table.go
```diff
@@ -294,20 +294,68 @@ func (ti *SqlTableInfo) buildTableCreateStatement() string {
...
+	if len(removeColumnStatements) > 0 {
+		removeColumnStatementsStr := strings.Join(removeColumnStatements, ", ")
+		statements = append(statements, fmt.Sprintf("ALTER %s %s DROP COLUMN IF EXISTS (%s)", typestring, ti.SQLFullName(), removeColumnStatementsStr))
+	}
+
+	for i, newCi := range ti.ColumnInfos {
+		if _, exists := nameToOldColumn[newCi.Name]; !exists {
+			// Add new column if new column is detected.
+			newCiStatement := ti.serializeColumnInfo(newCi)
+			if i == 0 {
```
mgyucht, line 338: Very nice!

PR #3384, opened 2024-03-19
Fix integration test for sql column update

edwardfeng-db:
## Changes

- These test cases shouldn't be using the same table name so that we can avoid race conditions

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/acceptance/sql_table_test.go
+++ b/internal/acceptance/sql_table_test.go
@@ -300,6 +300,9 @@ func constructManagedSqlTableTemplate(tableName string, columnInfos []catalog.Sq
 			properties         = {
 				this      = "that"
 				something = "else"
+				"delta.minReaderVersion" = 2
+				"delta.minWriterVersion" = 5
+				"delta.columnMapping.mode" = "name"
 			}
 
 			%s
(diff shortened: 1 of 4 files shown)


internal/acceptance/sql_table_test.go
```diff
@@ -304,64 +304,69 @@ func constructManagedSqlTableTemplate(columnInfos []catalog.SqlColumnInfo) strin
...
 			comment = "this table is managed by terraform"
-		}`, columnsTemplate)
+		}`, tableName, columnsTemplate)
 }
 
 func TestUcAccResourceSqlTable_RenameColumn(t *testing.T) {
 	if os.Getenv("GOOGLE_CREDENTIALS") != "" {
 		skipf(t)("databricks_sql_table resource not available on GCP")
 	}
+	tableName := "rename"
```
tanmay-db, line 314: We should use random names, if in rare chances the test fails ( and the resource isn't destroyed) then this table will exists and the next run will have issue.

tanmay-db (approved):
LGTM, need to check if nightly run pass

nkvuong:
integration tests passed (failed tests are not related)

nkvuong (approved):

Issue #3383, opened 2024-03-19
[ISSUE] Can't set `config.served_entities.scale_to_zero_enabled = false` in `databricks_model_serving`


alexott:
Filing to track it...

```hcl
resource "databricks_catalog" "databricks_e5_v2_models" {
  share_name    = "e5_v2_share"
  provider_name = "databricks"
  name          = "databricks_e5_v2_models"
}

resource "databricks_model_serving" "this" {
  name = "e5_${local.STICKY_RANDOM}"
  config {
    served_entities {
      name           = "e5_small_v2"
      entity_name    = "${databricks_catalog.databricks_e5_v2_models.id}.models.e5_small_v2"
      entity_version = "2"
      workload_size  = "Small"
      scale_to_zero_enabled = false
    }
    traffic_config {
      routes {
        served_model_name  = "e5_small_v2"
        traffic_percentage = 100
      }
    }
  }
}
```

fails due OpenAPI spec problem where this field is marked as `omitempty` although it's required...

alexott:
Done

PR #3385, opened 2024-03-20
Set ForceSendFields for boolean/integer values explicitly set to false/0

mgyucht:
## Changes
There are a number of issues that arise related to users trying to set an attribute in Terraform to `false` or `0`, but that attribute is not reflected in the API itself. See #3383 for example. This happens because these attributes are optional, so they are only serialized when set to something other than the zero value. For APIs that require the zero value explicitly, this needs to be set in `ForceSendFields`.

This PR changes DataToStructPointer and DiffToStructPointer to set ForceSendFields automatically for structures when 1) a ForceSendFields field is present on the structure, and 2) one or more of the fields is a boolean or integer that is set to 0, using the deprecated `GetOkExists` API.

As this subsumes the `SetForceSendFields` API, I've removed that code in this PR.

This revealed another potential bug in using Terraform to set `auto_stop_mins` for `databricks_sql_warehouse` to `0`. I've updated that test.

## Tests
Added unit tests for setting ForceSendFields appropriately.

- [ ] `make test` run locally
- [ ] using Go SDK


tanmay-db (approved):
Went through the PR, looks Good.

mgyucht:
We actually can't merge this as we currently store all fields in state, even for optional fields whose value is equal to the zero value. :( We'll need to revisit this in the future.

andrewnester:
The logic was fixed in this PR https://github.com/databricks/terraform-provider-databricks/pull/3385/commits/4bf843bac42c691f6e9e6b9843db0b758b0d886c

Instead of storing in state all fields we only store the one that explicitly defined in force_send_fields (which means they were explicitly set in state prior).

Integration tests are green

andrewnester (approved):

common/reflect_resource.go
```diff
@@ -617,10 +621,16 @@ func collectionToMaps(v any, s *schema.Schema, aliases map[string]map[string]str
 			}
 			v = v.Elem()
 		}
-		err := iterFields(v, []string{}, r.Schema, aliases, func(fieldSchema *schema.Schema,
-			path []string, valueField *reflect.Value) error {
+		fv := v.FieldByName("ForceSendFields")
+		forceSendFields := []string{}
+		// force send fields might not exist in struct, so checking it against zero value first
+		if fv != (reflect.Value{}) {
```
mgyucht, line 627: This is probably the more canonical way to check this.
```suggestion
		if fv.IsValid() {
```

common/reflect_resource.go
```diff
@@ -629,8 +639,16 @@ func collectionToMaps(v any, s *schema.Schema, aliases map[string]map[string]str
 					return fmt.Errorf("%s: %v", path, err)
 				}
 				data[fieldName] = nv
+			case schema.TypeBool, schema.TypeInt:
+				// If the field is not in the force send fields, we should not store it in the state
+				// Only bool and int values are in force send fields now
+				if len(forceSendFields) > 0 && !slices.Contains(forceSendFields, valueField.sf.Name) {
```
mgyucht, line 645: Is the `len` check needed?

common/reflect_resource.go
```diff
@@ -617,10 +621,16 @@ func collectionToMaps(v any, s *schema.Schema, aliases map[string]map[string]str
 			}
 			v = v.Elem()
 		}
-		err := iterFields(v, []string{}, r.Schema, aliases, func(fieldSchema *schema.Schema,
-			path []string, valueField *reflect.Value) error {
+		fv := v.FieldByName("ForceSendFields")
+		forceSendFields := []string{}
```
mgyucht, line 625:
```suggestion
		forceSendFields := nil
```

common/reflect_resource.go
mgyucht, line 645: We do need to store it in state if it is not a zero value though. I figured we could handle this by adding a case to `isValueNilOrEmpty` since the treatment is the same as that case.

common/reflect_resource.go
```diff
@@ -702,6 +719,22 @@ func StructToData(result any, s map[string]*schema.Schema, d *schema.ResourceDat
...
+			if fv != (reflect.Value{}) {
+				forceSendFields = fv.Interface().([]string)
+			}
+
+			// If the field is not in the force send fields, we should not store it in the state
+			// Only bool and int values are in force send fields now
+			if len(forceSendFields) > 0 && !slices.Contains(forceSendFields, valueField.sf.Name) {
+				log.Printf("[TRACE] skipping field %s", valueField.sf.Name)
+				return nil
+			}
```
mgyucht, line 735: Any chance we can define a single function to remove duplication between this and collectionToMaps?

common/reflect_resource_test.go
```diff
@@ -502,6 +502,33 @@ func TestDataToStructPointerWithResourceProviderStruct(t *testing.T) {
 	DataToStructPointer(d, s, &dummyCopy)
 }
 
+type structWithForceSendFields struct {
+	BoolOpt         bool     `json:"bool_opt,omitempty"`
+	NonJson         bool     `json:"-,omitempty"`
+	ForceSendFields []string `json:"-"`
+}
+
+func TestDataToStructPointerWithImpplicitlyZeroFields(t *testing.T) {
```
mgyucht, line 511:
```suggestion
func TestDataToStructPointerWithImplicitlyZeroFields(t *testing.T) {
```

sql/resource_sql_endpoint_test.go
```diff
@@ -155,6 +155,7 @@ func TestResourceSQLEndpointCreateNoAutoTermination(t *testing.T) {
 				AutoStopMins:       0,
 				EnablePhoton:       true,
 				SpotInstancePolicy: "COST_OPTIMIZED",
+				ForceSendFields:    []string{"AutoStopMins"},
```
mgyucht, line 158: This is awesome!

sql/resource_sql_global_config_test.go
```diff
@@ -110,8 +110,9 @@ func TestResourceSQLGlobalConfigDelete(t *testing.T) {
 				Method:   "PUT",
 				Resource: "/api/2.0/sql/config/warehouses",
 				ExpectedRequest: map[string]any{
-					"data_access_config": []any{},
-					"security_policy":    "DATA_ACCESS_CONTROL",
+					"data_access_config":        []any{},
+					"security_policy":           "DATA_ACCESS_CONTROL",
+					"enable_serverless_compute": false,
```
mgyucht, line 115: This makes me just a bit nervous. Could we add one unit test to verify that if `enable_serverless_compute` is true in state that deleting it results in `"enable_serverless_compute": true` in the request?

mgyucht (commented):
Overall looking good, couple comments

common/reflect_resource.go
andrewnester: @mgyucht yes, if there are no forceSendFields set we don't want to execute this check because otherwise no fields  will be written to TF state at all

common/reflect_resource.go
andrewnester: > This is probably the more canonical way to check this.

These are not equivalent checks, we need to check that fv is "zero" value (empty reflect.Value struct) which is valid value.

common/reflect_resource.go
```diff
@@ -661,21 +670,53 @@ func isValueNilOrEmpty(valueField *reflect.Value, fieldPath string) bool {
...
+		return valueField.Int() == 0
+	}
+	return false
+}
+
+func isIntOrBoolFieldNeedToBeSent(root reflect.Value, valueField field) bool {
+	fv := root.FieldByName("ForceSendFields")
+	var forceSendFields []string
+	// force send fields might not exist in struct, so checking it against zero value first
+	if fv != (reflect.Value{}) {
```
mgyucht, line 687: if not valid, return true

common/reflect_resource.go
```diff
@@ -661,21 +670,53 @@ func isValueNilOrEmpty(valueField *reflect.Value, fieldPath string) bool {
...
+	switch valueField.Kind() {
+	case reflect.Bool:
+		return !valueField.Bool()
+	case reflect.Int, reflect.Int32, reflect.Int64:
+		return valueField.Int() == 0
+	}
+	return false
+}
+
+func isIntOrBoolFieldNeedToBeSent(root reflect.Value, valueField field) bool {
```
mgyucht, line 683: isIntOrBoolExplicitlySet

PR #3391, opened 2024-03-21
Bump go-sdk version to 0.36.0

edwardfeng-db:
## Changes

- Bump up go-sdk version to `0.36.0`

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


hectorcast-db (approved):

alexott (changes requested):
We already had PR to bump versions, but we need to make code changes to accommodate it: https://github.com/databricks/terraform-provider-databricks/pull/3389

catalog/data_catalogs.go
```diff
@@ -4,14 +4,16 @@ import (
...
+	"github.com/databricks/databricks-sdk-go/service/catalog"
 	"github.com/databricks/terraform-provider-databricks/common"
 )
 
 func DataSourceCatalogs() common.Resource {
 	return common.WorkspaceData(func(ctx context.Context, data *struct {
 		Ids []string `json:"ids,omitempty" tf:"computed,slice_set"`
 	}, w *databricks.WorkspaceClient) error {
-		catalogs, err := w.Catalogs.ListAll(ctx)
+		var req catalog.ListCatalogsRequest = catalog.ListCatalogsRequest{}
```
alexott, line 15: why we need to have variable here, why not simple `ListAll(ctx, catalog.ListCatalogsRequest{})` ?

catalog/data_catalogs_test.go
```diff
@@ -12,7 +12,7 @@ func TestCatalogsData(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "GET",
-				Resource: "/api/2.1/unity-catalog/catalogs",
+				Resource: "/api/2.1/unity-catalog/catalogs?",
```
alexott, line 15: Can we fix SDK to not produce `?` when there is no parameters?

catalog/resource_storage_credential.go
```diff
@@ -130,11 +130,6 @@ func ResourceStorageCredential() common.Resource {
 			common.DataToStructPointer(d, storageCredentialSchema, &update)
 			update.Name = d.Id()
 			update.Force = force
-			// We need to set them to empty since these fields are computed
-			if _, ok := d.GetOk("aws_iam_role"); ok {
-				update.AwsIamRole.UnityCatalogIamArn = ""
-				update.AwsIamRole.ExternalId = ""
-			}
```
alexott, line 137: Hmmm, @nkvuong - what do you think, can we remove this?

alexott (changes requested):
We still need to update docs - delete references to removed attributes

catalog/resource_storage_credential.go
edwardfeng-db: It seems like these two are already marked as `output only` in the current doc, do we still need to make any change?

alexott: we need to remove them from the docs.

edwardfeng-db: Removed, PTAL

alexott (approved):

PR #3395, opened 2024-03-22
[Feature] Allow to filter jobs by name in `databricks_jobs` data source

dsfrederic:
## Changes

I've added a filter to obtain only job ids if the name contains a certain string. Inspired by the logic applied at databricks_clusters

image

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


jobs/data_jobs.go
```diff
@@ -9,12 +9,13 @@ import (
 
 func DataSourceJobs() common.Resource {
 	type jobsData struct {
-		Ids map[string]string `json:"ids,omitempty" tf:"computed"`
+		Ids        map[string]string `json:"ids,omitempty" tf:"computed"`
+		NameFilter string            `json:"job_name_contains,omitempty" tf:"optional"`
```
alexott, line 13: you don't need `tf:"optional"` if you have `omitempty`

jobs/resource_job.go
```diff
@@ -472,8 +472,20 @@ func (a JobsAPI) ListByName(name string, expandTasks bool) ([]Job, error) {
...
+
+	allJobs, err := a.ListByName("", false)
+	if err != nil {
+		return l, err
+	}
+	for _, job := range allJobs {
+		if strings.Contains(job.Settings.Name, nameFilter) {
+			l = append(l, job)
+		}
+	}
```
alexott, line 488: I wouldn't put this into the API, but do the filtering in the data source itself.

alexott (changes requested):

alexott (approved):

PR #3397, opened 2024-03-22
[BUG] Remove set in job libraries

bouland:
Remove Set to revert to ordered List

## Changes
I have developed 2 libraries  (A and B) but B depend on A.
They are installed consecutively on my databricks cluster from a static S3 resource link, so order matter as it don't use registry to resolve dependencies in a single pass.

The TypeSet will ensure unicity but have drawback to introduce shifting in order of installation, so i have just fallback to default TypeList.

Actually it's custom python wheels.

Rgds

## Tests
I have tested to build my fork on my github action runner and successfully solved the deployment issue (order is preserved)

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -218,7 +218,7 @@ type JobTaskSettings struct {
 	NewCluster        *clusters.Cluster   `json:"new_cluster,omitempty" tf:"group:cluster_type"`
 	JobClusterKey     string              `json:"job_cluster_key,omitempty" tf:"group:cluster_type"`
 	ComputeKey        string              `json:"compute_key,omitempty" tf:"group:cluster_type"`
-	Libraries         []libraries.Library `json:"libraries,omitempty" tf:"slice_set,alias:library"`
+	Libraries         []libraries.Library `json:"libraries,omitempty" tf:"slice,alias:library"`
 
 	NotebookTask    *NotebookTask       `json:"notebook_task,omitempty" tf:"group:task_type"`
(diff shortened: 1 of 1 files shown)


jobs/resource_job.go
```diff
@@ -218,7 +218,7 @@ type JobTaskSettings struct {
 	NewCluster        *clusters.Cluster   `json:"new_cluster,omitempty" tf:"group:cluster_type"`
 	JobClusterKey     string              `json:"job_cluster_key,omitempty" tf:"group:cluster_type"`
 	ComputeKey        string              `json:"compute_key,omitempty" tf:"group:cluster_type"`
-	Libraries         []libraries.Library `json:"libraries,omitempty" tf:"slice_set,alias:library"`
+	Libraries         []libraries.Library `json:"libraries,omitempty" tf:"slice,alias:library"`
```
edwardfeng-db, line 221: I don't think you need the `slice` here.

Just `tf:"alias:library"` should be enough

edwardfeng-db:
Thanks @bouland I'm wondering can we add a test for this?

alexott (changes requested):
We also need to modify the documentation: https://github.com/databricks/terraform-provider-databricks/blob/main/docs/resources/job.md?plain=1#L93

jobs/resource_job.go
alexott: by default arrays are converted into slices... `slice_set` was need if you don't want to depend on the order of elements

edwardfeng-db:
Fixed this in https://github.com/databricks/terraform-provider-databricks/pull/3398

Issue #1585, opened 2022-09-08
Cluster Libraries are not ordered


davidglevy:
As per the title, there is no ordering applied to libraries and it is a "surprise" with the underlying object being a Set rather than a List. This causes unpredictable results. For some configurations this may be harmless but can also result in clusters failing to provision.

Expectation is that library ordering in Terraform cluster definition is mapped to the order within the cluster provisioned.

nfx:
@davidglevy  there are no guarantees of libraries being installed in a certain order on the api level. please use [databricks_library](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/library) with `depends_on` attributes to achieve what you're describing.

edwardfeng-db:
Fixed in https://github.com/databricks/terraform-provider-databricks/pull/3398

PR #3398, opened 2024-03-22
Make libraries in `resource_job` `slice` instead of set

edwardfeng-db:
## Changes

- FIxing issue: https://github.com/databricks/terraform-provider-databricks/issues/1585
- Right now the libraries are not ordered, this PR fixes the issue by removing `slice_set` tag

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/docs/resources/job.md
+++ b/docs/resources/job.md
@@ -90,7 +90,7 @@ The resource supports the following arguments:
   continuous { }
   ```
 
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm, but we'll need to add a note into the changelog about behaviour change

Cc: @mgyucht

tanmay-db (approved):

PR #3401, opened 2024-03-25
Add table update trigger in terraform API

sebastien-db:
## Changes

Add table update trigger in terraform API.
This is the [design doc](https://docs.google.com/document/d/1mC_Yr3ADDAPGKggdyluUoRtsNm2fDBTNvG91Z0DetrI/edit?pli=1#heading=h.dd2cc57oc5wk) for reference.

## Tests

- [X] `make test` run locally
- [ ] covered with integration tests in `internal/acceptance`: having a table update is complex, so we decided to not do integration tests for now
- [X] using Go SDK


andrewnester (approved):
LGTM, but still requires review from TF provider team

jobs/resource_job.go
```diff
@@ -267,8 +267,16 @@ type FileArrival struct {
...
+	TableNames                    []string `json:"table_names"`
+	Condition                     string   `json:"condition,omitempty"`
+	MinTimeBetweenTriggersSeconds int32    `json:"min_time_between_triggers_seconds,omitempty"`
+	WaitAfterLastChangeSeconds    int32    `json:"wait_after_last_change_seconds,omitempty"`
+}
+
 type Trigger struct {
-	FileArrival *FileArrival `json:"file_arrival"`
+	FileArrival *FileArrival `json:"file_arrival",omitempty`
+	TableUpdate *TableUpdate `json:"table_update",omitempty`
```
nkvuong, line 279: is it possible to define both triggers for the same job? if not, then we need to use  `CustomizeSchemaPath().SetExactlyOneOf`

internal/acceptance/job_test.go
```diff
@@ -293,6 +293,102 @@ func TestAccJobControlRunState(t *testing.T) {
...
+			time.Sleep(5 * time.Second)
+		}
+		return lastErr
+	}
+	waitForRunToStart := func(ctx context.Context, client *common.DatabricksClient, id string) error {
+		return retryFor(ctx, client, id, errors.New("timed out waiting for job run to start"), checkIfRunHasStarted)
+	}
+	workspaceLevel(t, step{
+		Template: getJobTemplate(`
+			table_names = ["catalog.schema.table1", "catalog.schema.table2"],
```
nkvuong, line 385: these tables do not exist in our integration environment, so the job will not trigger

nkvuong: we can create a table with the resource `databricks_sql_table`, but to trigger an update would require sending a sql statement

it does feel complicated, so for this PR, you can test locally and confirm that it works. We can add better integration tests in a follow-up PR

jobs/resource_job.go
```diff
@@ -728,6 +736,8 @@ var jobSchema = common.StructToSchema(JobSettings{},
 		s["continuous"].ConflictsWith = []string{"schedule", "trigger"}
 		s["trigger"].ConflictsWith = []string{"schedule", "continuous"}
 
+		s["trigger"].ExactlyOneOf = []string{"file_arrival", "table_update"}
```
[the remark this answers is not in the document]
nkvuong: because it is a nested field, you need to follow the below example
```		
// we need to have only one of user name vs service principal in the run_as block
run_as_eoo := []string{"run_as.0.user_name", "run_as.0.service_principal_name"}
common.MustSchemaPath(s, "run_as", "user_name").ExactlyOneOf = run_as_eoo
common.MustSchemaPath(s, "run_as", "service_principal_name").ExactlyOneOf = run_as_eoo
```

docs/resources/job.md
```diff
@@ -209,10 +209,15 @@ This block describes the queue settings of the job:
...
   * `url` - (Required) string with URL under the Unity Catalog external location that will be monitored for new files. Please note that have a trailing slash character (`/`).
   * `min_time_between_triggers_seconds` - (Optional) If set, the trigger starts a run only after the specified amount of time passed since the last time the trigger fired. The minimum allowed value is 60 seconds.
   * `wait_after_last_change_seconds` - (Optional) If set, the trigger starts a run only after no file activity has occurred for the specified amount of time. This makes it possible to wait for a batch of incoming files to arrive before triggering a run. The minimum allowed value is 60 seconds.
+* `table_update` - (Optional) configuration block to define a trigger for Table Update events consisting of following attributes:
+  * `table_names` - (Required) A list of Delta tables to monitor for changes. The table name must be in the format `catalog_name.schema_name.table_name`.
+  * `condition` - (Optional) The table(s) condition based on which to trigger a job run.
```
nkvuong, line 218:
```suggestion
  * `condition` - (Optional) The table(s) condition based on which to trigger a job run. Valid values are `ANY_UPDATED` or `ALL_UPDATED`.
```

PR #3415, opened 2024-03-30
Add account workspace setting type

nkvuong:
## Changes
Add two new types to `settings/generic_setting.go` - `accountWorkspaceSettingDefinition` and `accountWorkspaceSetting`, to allow settings that can be set at both workspace & account level
This may simplify PRs like #3325

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/settings/generic_setting.go
+++ b/settings/generic_setting.go
@@ -149,7 +149,7 @@ var _ workspaceSettingDefinition[struct{}] = workspaceSetting[struct{}]{}
 
 type accountSettingDefinition[T any] genericSettingDefinition[T, *databricks.AccountClient]
 
-// An account setting is a setting that is scoped to a workspace.
+// An account setting is a setting that is scoped to an account.
 type accountSetting[T any] struct {
 	// The struct corresponding to the setting. The schema of the Terraform resource will be generated from this struct.
 	// This struct must have an Etag field of type string.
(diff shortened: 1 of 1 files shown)


settings/generic_setting.go
```diff
@@ -199,6 +199,100 @@ func (w accountSetting[T]) GetId(t *T) string {
...
+
+	// Optional function to generate resource ID from the settings. If not provided, will use predefined value `global`
+	generateIdFunc func(setting *T) string
+}
+
+func (aw accountWorkspaceSetting[T]) SettingStruct() T {
+	return aw.settingStruct
+}
+func (aw accountWorkspaceSetting[T]) Read(ctx context.Context, c *common.DatabricksClient, etag string) (*T, error) {
+	if c.Config.IsAccountClient() {
```
mgyucht, line 235: We can use `c.AccountOrWorksapceRequest` for all of these.

settings/generic_setting.go
```diff
@@ -149,7 +149,7 @@ var _ workspaceSettingDefinition[struct{}] = workspaceSetting[struct{}]{}
 
 type accountSettingDefinition[T any] genericSettingDefinition[T, *databricks.AccountClient]
 
-// An account setting is a setting that is scoped to a workspace.
+// An account setting is a setting that is scoped to an account.
```
mgyucht, line 152: Oops :)

mgyucht (approved):
LGTM, but suggest to use `c.AccountOrWorkspaceRequest` to minimize duplicating logic.

settings/generic_setting.go
nkvuong: `AccountOrWorkspaceRequest` is a method on `DatabricksClient`, so won't accept a type parameter 😢

tanmay-db (approved):

PR #3419, opened 2024-04-01
Add lookup by name or region to `databricks_metastore` data source

alexott:
## Changes

Sometimes it would be easier to fetch metastore information by a workspace region or by a metastore name compared to fetching metastore information by ID.  This PR adds ability to specify either `name` or `region` to perform a metastore lookup.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/data_metastore.go
+++ b/catalog/data_metastore.go
@@ -2,6 +2,8 @@ package catalog
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/service/catalog"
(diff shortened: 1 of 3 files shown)


catalog/data_metastore.go
```diff
@@ -10,15 +12,54 @@ import (
...
+			data.Metastore = minfo.MetastoreInfo
+		} else {
+			metastores, err := acc.Metastores.ListAll(ctx)
+			if err != nil {
+				return err
+			}
+			minfos := []catalog.MetastoreInfo{}
+			if data.Name != "" {
+				for _, v := range metastores {
+					if v.Name == data.Name {
```
nkvuong, line 41: would it be better to match names without case sensitivity - the metastore name gets converted to all lowercase on the backend

nkvuong (approved):
one small comment

catalog/data_metastore.go
alexott: ah, yes - that's what I wasn't sure about. Let me fix it.

alexott: done

moritzmeister:
Hey @alexott, does this work for you? no matter what I do, look up by id or by name, the lookup seems to work, but then the data source only contains `metastore_id         = "_"` an underscore as id.

Using 1.40.0 of the provider.

alexott:
Hmmm, `metastore_id` should be populated. The `id` isn't populated yet, but I'll fix it

```
  + some_metastore = {
      + id             = "_"
      + metastore_id   = "53fa5c35-8001-407c-aa97-b95ee8c37f8d"
```

moritzmeister:
Ahh ok, I think I was too tired last night, you are right `metastore_id` exists. We should add that to the docs.

alexott:
I've opened PR to populate `id` as well

PR #3431, opened 2024-04-03
Update `databricks_metastore_data_access` implementation to match `databricks_storage_credential`

nkvuong:
## Changes
- Add update method for `databricks_metastore_data_access`
- Roll back when default storage credential assignment failed
- Add notes to documentation

## Tests

- [x] `make test` run locally
- [x] using Go SDK


docs/resources/metastore_data_access.md
```diff
@@ -5,7 +5,9 @@ subcategory: "Unity Catalog"
 
 -> **Note** This resource could be used with account or workspace-level provider.
 
-Optionally, each [databricks_metastore](metastore.md) can have a default [databricks_storage_credential](storage_credential.md) defined as `databricks_metastore_data_access`. This will be used by Unity Catalog to access data in the root storage location if defined.
+Databricks recommends that you create a separate managed storage location for each catalog in your metastore (and you can do the same for schemas). Instead of creating a `databricks_metastore_data_access`, you should create a [databricks_storage_credential](storage_credential.md) isntead.
```
mgyucht, line 8:
```suggestion
Databricks recommends that you create a separate managed storage location for each catalog in your metastore (and you can do the same for schemas). Instead of creating a `databricks_metastore_data_access`, you should create a [databricks_storage_credential](storage_credential.md) instead.
```

docs/resources/metastore_data_access.md
```diff
@@ -5,7 +5,9 @@ subcategory: "Unity Catalog"
...
+
+However, If you opt to create a managed location at the metastore level and use it as the default storage for multiple catalogs and schemas, a default [databricks_storage_credential](storage_credential.md) can be defined as `databricks_metastore_data_access`. This will be used by Unity Catalog to access data in the root storage location if defined. Creating this resource requires account admin & metastore admin privileges.
```
mgyucht, line 10:
```suggestion
However, if you opt to create a managed location at the metastore level and use it as the default storage for multiple catalogs and schemas, a default [databricks_storage_credential](storage_credential.md) can be defined as `databricks_metastore_data_access`. This will be used by Unity Catalog to access data in the root storage location if defined. Creating this resource requires account admin & metastore admin privileges.
```

mgyucht (commented):
Should we do the same for the create method as well?

nkvuong:
> Should we do the same for the create method as well?

Refactoring Create is a bit tricky since we need `AccountOrWorkspaceRequest` to return an object to set the Id appropriately

alexott (commented):
looks good, we just need to resolve conflicts

catalog/resource_metastore_data_access.go
```diff
@@ -115,6 +66,11 @@ func ResourceMetastoreDataAccess() common.Resource {
 						},
 					})
 					if err != nil {
+						// Rollback
+						rollback_err := acc.StorageCredentials.DeleteByMetastoreIdAndStorageCredentialName(ctx, metastoreId, dac.CredentialInfo.Name)
+						if rollback_err != nil {
+							return fmt.Errorf("failed to rollback: %v", rollback_err)
```
alexott, line 72: Here we're hiding original error - should we put it here as well?

catalog/resource_metastore_data_access.go
```diff
@@ -130,9 +86,14 @@ func ResourceMetastoreDataAccess() common.Resource {
...
 						StorageRootCredentialId: dac.Id,
 					})
-				}
-				if err != nil {
-					return err
+					if err != nil {
+						// Rollback
+						rollback_err := w.StorageCredentials.DeleteByName(ctx, dac.Name)
+						if rollback_err != nil {
+							return fmt.Errorf("failed to rollback: %v", rollback_err)
```
alexott, line 93: same

alexott (commented):
in general - lgtm, but we need to discuss how we report errors during rollback, without losing the original error that triggered rollback.

@mgyucht wdyt?

alexott:
@nkvuong do we still need it?

Issue #1840, opened 2022-12-13
[ISSUE] Issue with `databricks_group_member` resource


atul-ram:
### Configuration
```hcl

variable "workspace_resource_id" {
  type = string
  description = "resource id of the databricks workspace"
  default = "XXXX"
}
provider "databricks" {
  azure_workspace_resource_id = var.databricks_workspace_id
}

terraform {
  required_providers {
    databricks = {
      source  = "databricks/databricks"
    }
    azuread = {
      source = "hashicorp/azuread"
      version = "2.31.0"
    }
  }
}

provider "databricks" {
  alias      = "azure_account"
  host       = "https://accounts.azuredatabricks.net"
  account_id = "00000000-0000-0000-0000-000000000000"
  auth_type  = "azure-cli"
}

# VARIABLES
/* azuread variables*/
variable "service_principal_name" {
  type        = string
  description = "data access service principal"
  default     = "blabla-d-01-sp"
}

# DATA
data "databricks_group" "admins" {
  //provider = databricks.azure_account
  display_name = "admins"
}

data "azuread_service_principal" "data_access_spn" {
  display_name = var.service_principal_name
}

# RESOURCES

resource "databricks_service_principal" "data_access_spn" {
  provider       = databricks.azure_account
  application_id = data.azuread_service_principal.data_access_spn.application_id
}


resource "databricks_group_member" "sp_admins" {

  group_id  = data.databricks_group.admins.id
  member_id = databricks_service_principal.data_access_spn.id

}

```

### Expected Behavior

the service principal should be added to the local admin group at the workspace level.

### Actual Behavior

```
C:\>terraform apply --auto-approve
databricks_service_principal.data_access_spn: Refreshing state... [id=3872500855341595]
databricks_group_member.sp_admins: Refreshing state... [id=759957192102856|3872500855341595]

Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply":

  # databricks_group_member.sp_admins has been deleted
  - resource "databricks_group_member" "sp_admins" {
      - group_id  = "759957192102856" -> null
      - id        = "759957192102856|3872500855341595" -> null
      - member_id = "3872500855341595" -> null
    }


Unless you have made equivalent changes to your configuration, or ignored the relevant attributes using ignore_changes, the following plan may include actions to undo or
respond to these changes.

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

[...]

PR #3434, opened 2024-04-04
Fix `databricks_entitlement` Update & Delete

mgyucht:
## Changes
It seems that updating entitlements may not be working properly, see #1840. I think there might be some issue with the way that updates are being sent to Databricks. Here's an example from a test I've written in this PR:
```
024-04-04T18:39:26.412+0200 [DEBUG] databricks: PATCH /api/2.0/preview/scim/v2/Groups/234408399693172
> {
>   "Operations": [
>     {
>       "op": "remove",
>       "path": "entitlements",
>       "value": [
>         {
>           "value": "allow-cluster-create"
>         },
>         {
>           "value": "allow-instance-pool-create"
>         },
>         {
>           "value": "databricks-sql-access"
>         },
>         {
>           "value": "workspace-access"
>         }
>       ]
>     },
>     {
>       "op": "add",
>       "path": "entitlements",
>       "value": [
>         {
>           "value": "allow-cluster-create"
>         },
>         {
>           "value": "allow-instance-pool-create"
>         },
>         {
>           "value": "databricks-sql-access"
>         },
>         {
>           "value": "workspace-access"
>         }
>       ]
>     }
>   ],
>   "schemas": [
>     "urn:ietf:params:scim:api:messages:2.0:PatchOp"
>   ]
> }
< HTTP/2.0 400 Bad Request
< {
<   "detail": "No such attribute with the name : entitlements in the current resource",
<   "schemas": [
<     "urn:ietf:params:scim:api:messages:2.0:Error"
<   ],
<   "scimType": "invalidPath",
<   "status": "400"
< }
```

I've modified the update implementation to more closely resemble the create implementation, using SCIM's `replace` operation to reset entitlements to the desired values. The only difference between the create and update pathways is that the ID is set in the create pathway. With additional testing added in this PR, this seems to address the issues users have been facing with entitlements.

Other cleanup in this PR:
* Change `addEntitlementsToSchema()` to accept `map[string]*schema.Schema`. `map`s are mutable so you don't need a pointer to update them.
* Only call `d.SetId()` in `Create`. This is save as changes to the user/group/SP will cause the resource to be recreated, so this doesn't need to be called in the update pathway.
* If no entitlements are set, deleting `databricks_entitlement` is a no-op, so exit early. This is the cause of the `No such attribute with the name : entitlements in the current resource` message when deleting a `databricks_entitlement` resource with no entitlements set.

## Tests
Added an integration test fixture harness to easily test entitlements. In particular, `databricks_entitlement` needs to be tested without a corresponding `databricks_user`/`databricks_group`/`databricks_service_principal` resource to ensure that its create/update/delete methods work properly. The test is written in a way where the first step's `PreConfig` creates a group, and then all steps' `databricks_entitlement` resources refer to that group. The group is cleaned up at the end of the test. I test three combinations:
* Create an empty databricks_entitlement, then add all of the allowed entitlements.
* Create a full databricks_entitlement, then remove all of the entitlements.
* Create a partial databricks_entitlement with some entitlements set to false.

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/entitlements_test.go
+++ b/internal/acceptance/entitlements_test.go
@@ -1,9 +1,14 @@
 package acceptance
 
 import (
+	"context"
+	"fmt"
 	"testing"
 
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/iam"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestAccEntitlementResource(t *testing.T) {
@@ -93,5 +98,54 @@ func TestAccServicePrincipalEntitlementsResourceOnAws(t *testing.T) {
 			allow_cluster_create       = true
 			allow_instance_pool_create = true
 		}`,
+	})
+}
+
+func TestAccIssue1860(t *testing.T) {
+	groupName := RandomName("entitlements-")
+	workspaceLevel(t, step{
+		PreConfig: func() {
+			w := databricks.Must(databricks.NewWorkspaceClient())
+			ctx := context.Background()
+			_, err := w.Groups.Create(ctx, iam.Group{
+				DisplayName: groupName,
+			})
+			assert.NoError(t, err)
+		},
+		Template: fmt.Sprintf(`
+		  data "databricks_group" "example" {
+			display_name = "%s"
+		  }
+		  
+		  resource "databricks_entitlements" "entitlements_users" {
+			group_id              = data.databricks_group.example.id
+		  }
+		`, groupName),
+	}, step{
+		Template: fmt.Sprintf(`
+		  data "databricks_group" "example" {
+			display_name = "%s"
+		  }
+		  resource "databricks_entitlements" "entitlements_users" {
+			group_id                   = data.databricks_group.example.id
+			allow_cluster_create       = true
+			allow_instance_pool_create = true
+			workspace_access           = true
+			databricks_sql_access      = true
+		  }
+		  `, groupName),
+	}, step{
+		Template: fmt.Sprintf(`
+		  data "databricks_group" "example" {
+			display_name = "%s"
+		  }
+		  resource "databricks_entitlements" "entitlements_users" {
+			group_id                   = data.databricks_group.example.id
+			allow_cluster_create       = false
+			allow_instance_pool_create = false
+			workspace_access           = true
+			databricks_sql_access      = true
+		  }
+		  `, groupName),
 	})
 }
--- a/internal/acceptance/entitlements_test.go
+++ b/internal/acceptance/entitlements_test.go
@@ -3,11 +3,14 @@ package acceptance
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/databricks/databricks-sdk-go"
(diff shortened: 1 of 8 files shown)


internal/acceptance/entitlements_test.go
```diff
@@ -95,3 +103,122 @@ func TestAccServicePrincipalEntitlementsResourceOnAws(t *testing.T) {
...
+			resource "databricks_entitlements" "entitlements_users" {
+				group_id              = data.databricks_group.example.id
+				%s
+			}
+		`, groupName, entitlementsBuf.String()),
+			Check: func(s *terraform.State) error {
+				groupId := s.RootModule().Resources["data.databricks_group.example"].Primary.ID
+				var res iam.Group
+				ctx := context.Background()
+				err := (*c).Do(ctx, "GET", fmt.Sprintf("/api/2.0/preview/scim/v2/Groups/%s?attributes=entitlements", groupId),
```
nkvuong, line 137: we should expose attributes as a parameter in the SDK to clean this up later

internal/acceptance/entitlements_test.go
```diff
@@ -95,3 +103,122 @@ func TestAccServicePrincipalEntitlementsResourceOnAws(t *testing.T) {
...
+}
+
+func TestAccEntitlementsAddToEmpty(t *testing.T) {
+	steps := makeEntitlementsSteps(t, [][]entitlement{
+		{},
+		{
+			{"allow_cluster_create", true},
+			{"allow_instance_pool_create", true},
+			{"workspace_access", true},
+			{"databricks_sql_access", true},
```
nkvuong, line 195:
```suggestion
			{"allow_cluster_create", false},
			{"allow_instance_pool_create", false},
			{"workspace_access", false},
			{"databricks_sql_access", false},
```

internal/acceptance/entitlements_test.go
```diff
@@ -95,3 +103,122 @@ func TestAccServicePrincipalEntitlementsResourceOnAws(t *testing.T) {
...
+	for i, entitlements := range entitlementsSteps {
+		steps[i] = makeEntitlementsStep(entitlements)
+	}
+	steps[0].PreConfig = makePreconfig(t, &c, groupName)
+	return steps
+}
+
+func makePreconfig(t *testing.T, c **httpclient.ApiClient, groupName string) func() {
+	return func() {
+		w := databricks.Must(databricks.NewWorkspaceClient())
```
nkvuong, line 172: would it make sense to initialize a WorkspaceClient in our CRUD methods, to catch if user is misconfiguring`databricks_entitlement` with an account-level provider?

nkvuong (approved):
a couple of small comments

internal/acceptance/entitlements_test.go
mgyucht: Yes, absolutely.

internal/acceptance/entitlements_test.go
mgyucht: I'll make a separate test for this, as I did want to have one test that sets all entitlements to true on create.

internal/acceptance/entitlements_test.go
mgyucht: Yes, we just have to be careful about where we do this validation (as in, it needs to happen during the apply phase and not plan).

PR #3454, opened 2024-04-11
Only suppress diff for `run_if` when it is `ALL_SUCCESS`

shreyas-goenka:
## Changes
This PR introduces the `SetSuppressDiffWithDefault` which allows us to suppress diff in one of the most common cases. The first use of it is for `run_if`.

When a `run_if` is not defined for a task, the platform assigns it a default of `ALL_SUCCESS`. This PR fixes the suppress diff behaviour to only suppress the diff if the platform returns the default value, and not suppress it in all cases.

This PR treats one of the symptoms of job tasks being reordered during diff computation, and tasks that do not have a run_if defined ending up with one in the final job.

## Tests
Unit tests and manually. Task reordering during diff computation no longer incorrectly sets run_if.

### DIFF
--- a/common/customizable_schema.go
+++ b/common/customizable_schema.go
@@ -90,7 +90,7 @@ func (s *CustomizableSchema) SetSuppressDiff() *CustomizableSchema {
 // the old value (ie value from state / platform) is equal to the default value.
 //
 // Often Databricks HTTP APIs will return values for fields that were not set by
-// the author in their terraform configuration This function allows us to suppress
+// the author in their terraform configuration. This function allows us to suppress
 // the diff in these cases.
 func (s *CustomizableSchema) SetSuppressDiffWithDefault(dv any) *CustomizableSchema {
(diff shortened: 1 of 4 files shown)


alexott (approved):
lgtm

alexott:
Are integration tests passing?

shreyas-goenka:
Integration tests are OK. The tests that are failing are unrelated to the change are also failing on main.

common/customizable_schema.go
```diff
@@ -82,6 +85,36 @@ func (s *CustomizableSchema) SetSuppressDiff() *CustomizableSchema {
 	return s
 }
 
+// SetSuppressDiffWithDefault suppresses the diff if the
+// new value (ie value from HCL config) is not set and
+// the old value (ie value from state / platform) is equal to the default value.
+//
+// Often Databricks HTTP APIs will return values for fields that were not set by
+// the author in their terraform configuration This function allows us to suppress
```
mgyucht, line 93:
```suggestion
// the author in their terraform configuration. This function allows us to suppress
```

mgyucht (approved):
Amazing!

PR #3458, opened 2024-04-12
Bump Go SDK to v0.38.0

mgyucht:
## Changes
Update the Databricks Go SDK dependency to v0.38.0.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -17,7 +17,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
-	"github.com/databricks/databricks-sdk-go/service/compute"
 	"github.com/databricks/terraform-provider-databricks/clusters"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/databricks/terraform-provider-databricks/libraries"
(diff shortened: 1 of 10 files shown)


jobs/resource_job.go
```diff
@@ -17,7 +17,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
-	"github.com/databricks/databricks-sdk-go/service/compute"
```
nordp, line 20: We should remove `ComputeKey` / `compute_key` as well.

nordp (changes requested):

nordp (approved):
LGTM for `jobs`

scim/resource_service_principal.go
```diff
@@ -224,13 +224,13 @@ func ResourceServicePrincipal() common.Resource {
 				if isForceDeleteRepos {
 					err = workspace.NewNotebooksAPI(ctx, c).Delete(fmt.Sprintf("/Repos/%v", appId), true)
 					if err != nil {
-						return fmt.Errorf("force_delete_repos: %w", err)
+						return fmt.Errorf("force_delete_repos: %s", err.Error())
```
mgyucht, line 227: This is needed due to changes in the behavior of `apierr.IsMissing()`. These errors were previously not considered to be missing because the status code used in the test is 400. However, in 0.39, this function changed to check whether the error inherits from `NotFound`. Because the status code returned from the API is RESOURCE_DOES_NOT_EXIST, this is considered missing. Using the error string here causes the error to not be considered missing.

scim/resource_service_principal_test.go
```diff
@@ -468,7 +468,7 @@ func TestResourceServicePrincipalDelete_NonExistingRepo(t *testing.T) {
 					ErrorCode: "RESOURCE_DOES_NOT_EXIST",
 					Message:   "Path (/Repos/abc) doesn't exist.",
 				},
-				Status: 400,
+				Status: 404,
```
mgyucht, line 471: This is the true error code returned.

tanmay-db (approved):

PR #3462, opened 2024-04-12
Exporter: improve emitting of directories

alexott:
## Changes

This includes:

* Don't try to emit directories for SQL objects if `directories` service isn't enabled
* Don't split/join strings to extract directory path - just do search from the right

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


nkvuong (approved):
lgtm

PR #3467, opened 2024-04-15
Fix integration tests for `databricks_entitlements`

mgyucht:
## Changes
The integration tests for `databricks_entitlements` are now failing. I believe this is caused by a race condition between `databricks_group` and `databricks_entitlements` resources. In practice, these resources should not be used at the same time; instead, you should the former when you need to define the group in TF and the latter if the group is defined externally to TF.

First, I've improved the test coverage for entitlements to cover all resource types (user, group, service principal). As part of this, I've rewritten the tests to manage the resources externally from Terraform's perspective, so there are no more race conditions. Because of this, I'm able to remove the other integration tests, including the flaky one.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -343,3 +343,6 @@ scripts/tt
 .metals
 
 provider/completeness.md
+
+# VS Code debugging binaries
+__debug_bin*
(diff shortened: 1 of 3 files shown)


tanmay-db (approved):
Went through the changes, LGTM.

PR #3469, opened 2024-04-15
Fix `databricks_user` and `databricks_service_principal` integration tests

mgyucht:
## Changes
The recent Go SDK upgrade caused a small regression for these test cases. If the home folder was never provisioned for the user/service principal, deleting with force_delete_home_dir should not cause deletion failures. Here, we only return the failure if the path deletion fails due to the folder not existing. The same logic is applied to repos.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/service_principal_test.go
+++ b/internal/acceptance/service_principal_test.go
@@ -3,7 +3,6 @@ package acceptance
 import (
 	"context"
 	"fmt"
-	"os"
 	"testing"
 
 	"github.com/databricks/databricks-sdk-go"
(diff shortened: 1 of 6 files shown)


internal/acceptance/service_principal_test.go
```diff
@@ -22,36 +21,37 @@ const awsSpn = `resource "databricks_service_principal" "this" {
...
+	}
+	uuid := createUuid()
+	var spId string
 	workspaceLevel(t, step{
 		Template: `
 			resource "databricks_service_principal" "a" {
-				application_id = "{var.RANDOM_UUID}"
+				application_id = "` + uuid + `"
 				force_delete_home_dir = true
 			}`,
```
alexott, line 35: Move template into the variable to avoid code duplication?

alexott (approved):

PR #3475, opened 2024-04-16
Make CustomizeSchemaPath able to pass information to skip unfeasible customizations

edwardfeng-db:
## Changes

- As we introduce the schema registry, there are nested customizations we'd like to reuse. 
- However, customizations such as `ConflictsWith` require a `path` and when the path contains `TypeList` with `MaxItems>1`, it is unfeasible to apply such customizations any more.
- This PR adds a new concept called `SchemaPathContext` and it is being maintained as we go through `typeToSchema` calls. It is passed into `customizeSchema` function callback and is used as an input to `CustomizeSchemaPath` so that we can skip certain unfeasible customizations.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/common/customizable_schema.go
+++ b/common/customizable_schema.go
@@ -2,7 +2,9 @@ package common
 
 import (
 	"fmt"
+	"log"
 	"slices"
+	"strings"
 
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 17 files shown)


common/customizable_schema.go
```diff
@@ -13,9 +15,43 @@ type CustomizableSchema struct {
...
+}
+
+func getEmptySchemaPathContext() SchemaPathContext {
+	return SchemaPathContext{
+		[]string{},
+		[]*schema.Schema{},
+	}
+}
+
+func CustomizeSchemaPath(ctx SchemaPathContext, s map[string]*schema.Schema, path ...string) *CustomizableSchema {
```
mgyucht, line 54: I feel that this is a less usable interface and requires more boilerplate for people to use. What design spaces can we consider here?

Personally, I think it would make a lot of sense for SchemaPathContext to include the schema itself. That way you only have one parameter to work with.

Another variation would be to include this information in `CustomizableSchema` and construct this within `reflect_resource.go` and pass it to users. We would need some methods to index into this structure then. Then a user could do something like:

```go
func (r Resource) CustomizeSchema(s *CustomizableSchema) *CustomizableSchema {
  s.SchemaPath("field1", "attribute").SetRequired()
  s.SchemaPath("field2").SetConflictsWith("field3")
  // It is now unsafe to modify the schema directly, but we can expose if needed.
  s.Underlying.ConflictsWith = []string{"field4"}
}
```
Then this is all entirely hidden from the user's perspective.

mgyucht (commented):
Suggestion on the core interface. Curious to hear your thoughts.

common/customizable_schema.go
edwardfeng-db: Yeah good point, I do agree that exposing a new argument is not very easy to use. I created an updated version here https://github.com/databricks/terraform-provider-databricks/pull/3481, can you take another look? Thanks!

PR #3479, opened 2024-04-17
Improve integration tests for `databricks_model_serving`

mgyucht:
## Changes
Integration tests for model serving were failing due to a change in library support in DBR 15. Additionally, this test has had a history of flakiness and difficulty of maintenance due to the test training a model itself.

To address these points, I've made a model fixture called `experiment-fixture-model` that can be used for this model serving test. This should reduce test time by removing the need to start a cluster & train a model, as well as remove the dependency on the actual training process in this test.

## Tests
Ran locally in Azure, and this passed.

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/acceptance/model_serving_test.go
+++ b/internal/acceptance/model_serving_test.go
@@ -1,14 +1,10 @@
 package acceptance
 
 import (
-	"context"
 	"fmt"
 	"testing"
 
-	"github.com/databricks/databricks-sdk-go"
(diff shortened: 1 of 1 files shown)


tanmay-db (approved):

mgyucht:
@alexott Yes, it is. I'll send you a DM with a link to the PR where the model is defined.

alexott (approved):

PR #3481, opened 2024-04-17
Refactor `CustomizeSchema` for `ResourceProvider` to skip unfeasible customizations

edwardfeng-db:
## Changes

- As we introduce the schema registry, there are nested customizations we'd like to reuse.
- However, customizations such as ConflictsWith require a path and when the path contains TypeList with MaxItems>1, it is unfeasible to apply such customizations any more.
- This PR adds a new concept called `SchemaPathContext` and it is being maintained as we go through typeToSchema calls. 
- This PR refactors the interface of `CustomizeSchema` function for `ResourceProvider` types so that this new `SchemaPathContext` concept is hidden and passed into those customization functions such as `SetConflictsWith automatically.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


clusters/resource_cluster.go
```diff
@@ -123,62 +123,62 @@ type ClusterSpec struct {
...
+	s.SchemaPath("init_scripts").SetMaxItems(10)
+	s.SchemaPath("init_scripts", "dbfs").SetDeprecated(DbfsDeprecationWarning)
+	s.SchemaPath("init_scripts", "dbfs", "destination").SetRequired()
+	s.SchemaPath("init_scripts", "s3", "destination").SetRequired()
+	s.SchemaPath("init_scripts", "volumes", "destination").SetRequired()
+	s.SchemaPath("init_scripts", "workspace", "destination").SetRequired()
+	s.SchemaPath("workload_type", "clients").SetRequired()
+	s.SchemaPath("workload_type", "clients", "notebooks").SetDefault(true)
+	s.SchemaPath("workload_type", "clients", "jobs").SetDefault(true)
+	s.SchemaPath("library").Schema.Set = func(i any) int {
```
edwardfeng-db, line 143: New interface would hide the details about `SchemaPathContext`.

And I think it is not very hard to use, might even be easier since we don't have to type `common.CustomizeSchemaPath` on each line.

Let me know what you think @mgyucht

edwardfeng-db: Meanwhile, only the new resources using `ResourceProvider` are onboarded to this change. The legacy ones will have their `CustomizeSchema` callback function unchanged because 
  1. there will be a lot of files we need to refactor if we change them 
  2. the problem that `SchemaPathContext` is trying to solve only applies to those using `ResourceProvider`, so it's not necessary
  3.  the legacy `CustomizeSchema` callback functions are using a lot of direct modifications of the `*map[string]*schema.Schema` object, which requires extra step to refactor into this new interface

common/customizable_schema.go
```diff
@@ -30,6 +90,16 @@ func CustomizeSchemaPath(s map[string]*schema.Schema, path ...string) *Customiza
 	return &CustomizableSchema{Schema: sch, path: path}
 }
 
+func (s *CustomizableSchema) SchemaPath(path ...string) *CustomizableSchema {
+	sch := MustSchemaPath(s.GetSchemaMap(), path...)
+	return &CustomizableSchema{Schema: sch, path: path, context: s.context}
+}
+
+func (s *CustomizableSchema) GetSchemaMap(path ...string) map[string]*schema.Schema {
```
mgyucht, line 98: `path` parameter is not used. This probably should not be here.

common/customizable_schema.go
```diff
@@ -30,6 +90,16 @@ func CustomizeSchemaPath(s map[string]*schema.Schema, path ...string) *Customiza
...
+	schemaMap := s.Schema.Elem.(*schema.Resource).Schema
```
mgyucht, line 99: If `Elem` is nil or not a `*schema.Resource` we should panic but with a better error message.

common/customizable_schema.go
```diff
@@ -149,31 +219,47 @@ func (s *CustomizableSchema) SetConflictsWith(value []string) *CustomizableSchem
 	if len(value) == 0 {
 		panic("SetConflictsWith cannot take in an empty list")
 	}
-	s.Schema.ConflictsWith = value
+	if s.pathContainsMultipleItemsList() {
```
mgyucht, line 222: Fine to return a warning in this case. However, if `path` is `nil` (i.e. this is for a top-level resource), we should panic if this happens, because it will never work as expected.

This applies to all cases.

common/customizable_schema.go
```diff
@@ -13,6 +15,64 @@ type CustomizableSchema struct {
 	Schema         *schema.Schema
 	path           []string
 	isSuppressDiff bool
+	context        SchemaPathContext
+}
+
+type SchemaPathContext struct {
+	// Path is used for refernces from resourceProviderRegistry as prefix, follows the format of `[a,0,b]`
+	path       []string
+	schemaPath []*schema.Schema
```
mgyucht, line 24: This is risky because it is easy for developers to incorrectly set this state. I would rather introduce a new pair struct and have a list of pairs here.

common/customizable_schema.go
```diff
@@ -13,6 +15,64 @@ type CustomizableSchema struct {
...
+		prefix = ""
+	}
+	prefixedPaths := make([]string, len(value))
+	for i, item := range value {
+		prefixedPaths[i] = prefix + item
+	}
+	return prefixedPaths
 }
 
 func CustomizeSchemaPath(s map[string]*schema.Schema, path ...string) *CustomizableSchema {
```
mgyucht, line 78: Can we remove this function?

mgyucht (commented):
Great external interface. Let's make sure we get the details right.

common/customizable_schema.go
edwardfeng-db: Ahh yeah my bad, copy pasta

common/customizable_schema.go
edwardfeng-db: Not right now because there are already resources using them in the old `StructToSchema` callbacks

common/reflect_resource.go
```diff
@@ -100,19 +100,21 @@ type RecursiveResourceProvider interface {
...
+func resourceProviderStructToSchema(v ResourceProvider, scp SchemaPathContext) map[string]*schema.Schema {
 	rv := reflect.ValueOf(v)
 	var scm map[string]*schema.Schema
 	aliases := map[string]map[string]string{}
 	if rpwa, ok := v.(ResourceProviderWithAlias); ok {
 		aliases = rpwa.Aliases()
 	}
 	if rrp, ok := v.(RecursiveResourceProvider); ok {
-		scm = typeToSchema(rv, aliases, getRecursionTrackingContext(rrp))
+		scm = typeToSchema(rv, aliases, getRecursionTrackingContext(rrp), scp)
```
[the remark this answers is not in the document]
edwardfeng-db: Yeah good point, I renamed `recursionTrackingContext` to just `trackingContext` and made `schemaPathContext` a part of it. This way it'll be more generic

common/customizable_schema.go
```diff
@@ -30,6 +90,16 @@ func CustomizeSchemaPath(s map[string]*schema.Schema, path ...string) *Customiza
 	return &CustomizableSchema{Schema: sch, path: path}
 }
 
+func (s *CustomizableSchema) SchemaPath(path ...string) *CustomizableSchema {
+	sch := MustSchemaPath(s.GetSchemaMap(), path...)
+	return &CustomizableSchema{Schema: sch, path: path, context: s.context}
```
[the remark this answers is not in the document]
edwardfeng-db: > Don't we need to update the set of schemas in the context so that we inspect all intermediary schemas when verifying that there are no lists on this schema path?
Not really here, we need to do it for the path leading to the field that is set to be conflicted with the current field. This `SchemaPath` is not for that, this is help us to navigate to the field that we would like to customize.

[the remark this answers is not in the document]
mgyucht: Ok, I understand. My main idea was that this implementation doesn't need to distinguish between potential parent fields in terms of the context and the fields specified in the `path` parameter. I suppose the main consequence is that you would not be able to catch misconfiguration where someone sets `ConflictsWith` inappropriately in the base structure (i.e. part of the specified `path` is a non-unary list). The provider framework should catch this in validation though, so I think that's OK.

mgyucht (approved):
Excellent work, thanks for this!

PR #3487, opened 2024-04-18
Add test config for ignoring failed integration tests

mgyucht:
## Changes
This PR adds a test config to ignore test failures in `TestAccGroupsUpdateDisplayName`, which is currently flaky due to transient read-after-write inconsistencies.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/test-config.yaml
+++ b/test-config.yaml
@@ -1,0 +1,5 @@
+ignored_tests:
+  go:
+    - package: github.com/databricks/terraform-provider-databricks/internal/acceptance
+      test_name: TestAccGroupsUpdateDisplayName
+      comment: Failures due to read-after-write inconsistency. Tracked in https://databricks.atlassian.net/browse/ES-1100061


test-config.yaml
```diff
@@ -0,0 +1,5 @@
+ignored_tests:
+  go:
+    - package: github.com/databricks/terraform-provider-databricks/internal/acceptance
+      test_name: TestAccGroupsUpdateDisplayName
+      comment: Failures due to read-after-write inconsistency. Tracked in https://databricks.atlassian.net/browse/ES-1100061
```
tanmay-db, line 5: Not sure if we should mention internal links in public repository but we do need a way to track them. Since we already mention them in other repositories for example: https://github.com/databricks/cli/blob/main/libs/sync/sync.go#L79. I think it should be okay.

tanmay-db (approved):

test-config.yaml
mgyucht: Given the precedent in the CLI, I think we're safe to continue. Those links are protected anyways from external users.

PR #3503, opened 2024-04-24
Set `id` attribute of `databricks_metastore` data source

alexott:
## Changes

In addition to filling `metastore_id` we set `id` to the same value

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


nkvuong (approved):
lgtm

PR #3508, opened 2024-04-25
Introduce `common.NoClientData` helper function and use it in `databricks_aws_crossaccount_policy` data source

### LINKED ISSUE #3505: [ISSUE] Issue with using `host` and `token` since version 1.40.0 - `invalid Databricks Account configuration`
### Configuration
Since version 1.40.0, the following databricks provider block yields this error: `Error: cannot read aws crossaccount policy: cannot get client aws crossaccount policy: invalid Databricks Account configuration`
```hcl
provider "databricks" {
  alias = "alias"
  host  = var.databricks_host_new
  token = var.databricks_token_new
}
```

I'm not sure yet if this only occurs when working with AWS cross account policies or not, though the error would suggest that.

Going back to version 1.39.0 fixes the issue and we are able to plan / apply as expected.

### Expected Behavior
With the above provider configuration, terraform plan / apply works without the provided error.

### Actual Behavior
On version 1.40.0 and above, the error listed above occurs.

### Steps to Reproduce

Yes, this works on version 1.39.0 but the error occurs on 1.40.0 and 1.41.0.

### Debug Output
The debug log doesn't show anything useful beyond what I shared above.

### Important Factoids
The error happens on a data block call for `databricks_aws_crossaccount_policy` in case that makes a difference, since I'm not sure yet if this happens on non AWS cross account ops.

N/A

alexott:
## Changes

The `databricks_aws_crossaccount_policy` and some other data sources really don't need any client to work, so they could be used with both account and workspace-level providers. To simplify things, the `common.NoClientData` was introduced, and the implementation `databricks_aws_crossaccount_policy` has changed to use it.

This fixes #3505

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/aws/data_aws_crossaccount_policy.go
+++ b/aws/data_aws_crossaccount_policy.go
@@ -7,7 +7,6 @@ import (
 	"regexp"
 	"slices"
 
-	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/terraform-provider-databricks/common"
 )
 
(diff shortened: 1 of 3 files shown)


nkvuong (approved):
lgtm

PR #3520, opened 2024-04-29
Add DATABRICKS_USER_AGENT_EXTRA

kanterov:
## Changes
Support DATABRICKS_USER_AGENT_EXTRA environment variable.

It contains a list of products to add as extras to the user agent header. Product strings must follow RFC 9110, contain version numbers, and can't contain comments. E.g:

```
DATABRICKS_USER_AGENT_EXTRA="databricks-cli/0.1.2 my-project/0.0.1"
```

## Tests
I added unit tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
+	"unicode"
 
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm

alexott:
The only question - should we document it or not in the [corresponding section of the docs](https://registry.terraform.io/providers/databricks/databricks/latest/docs#environment-variables)?

kanterov:
@alexott Good question. I don't have a strong opinion about that. One of the reasons why we need that is to propagate additional user agents from DABs. Let's see if anyone has an opinion on that.

mgyucht (approved):

mgyucht:
This is mainly a feature for people using our TF provider as part of a separate product rather than TF directly. As such, we probably are alright not adding it to docs initially, since it won't be relevant to 99% of our users.

PR #3528, opened 2024-05-01
Added support for popular column types for `resource_sql_table`

tanmay-db:
## Changes

Added support for: 
1. Mapping of 
`````
        "integer": "int",
	"long":    "bigint",
	"real":    "float",
	"short":   "smallint",
	"byte":    "tinyint",
	"decimal": "decimal(10,0)",
	"dec":     "decimal(10,0)",
	"numeric": "decimal(10,0)",
`````
2. Case insensitive column types
3. Suppress diffs for column types
4. Make the column type as computed optional. 

Note: Since the PR also adds case sensitive matching, it includes functionality from: https://github.com/databricks/terraform-provider-databricks/pull/3501. I have followed up with author to rebase and their unit test can be added here: https://github.com/databricks/terraform-provider-databricks/pull/3501#issuecomment-2088379422

## Tests

Added integration test, before changes it fails during `assertNoColumnTypeDiff` because of using "integer" and passes with changes.
image

- [ ] `make test` run locally
- [ ] using Go SDK

### CODE REVIEW

**Comment by tanmay-db** [comment]
long to bigint mapping works E2E in integration test and also another user mentions it here: https://github.com/databricks/terraform-provider-databricks/issues/3490#issuecomment-2066330433 but couldn't find this information in API docs.

File: catalog/resource_sql_table.go [hunk 1]
```diff
@@ -486,10 +486,22 @@ func columnChangesCustomizeDiff(d *schema.ResourceDiff, newTable *SqlTableInfo)
 	return nil
 }
 
+func getColumnType(columnType string) string {
+	caseInsensitiveColumnType := strings.ToLower(columnType)
+	switch caseInsensitiveColumnType {
+	case "integer":
+		return "int"
+	case "long":
+		return "bigint"
```
mgyucht, line 495: Map is really the right datastructure here:
```go
var columnTypeAliases = map[string]string {
  "integer": "int",
  "long": "bigint",
}
  if alias, ok := columnTypeAlias[columnType]; ok {
    return alias
  }
  return columnType
}

```

catalog/resource_sql_table.go
```diff
@@ -538,6 +550,12 @@ func ResourceSqlTable() common.Resource {
 
 			s["partitions"].ConflictsWith = []string{"cluster_keys"}
 			s["cluster_keys"].ConflictsWith = []string{"partitions"}
+			common.MustSchemaPath(s, "column", "type").DiffSuppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+				if new == "" {
+					return true
+				}
```
mgyucht, line 556: Why this? If type is removed, there is no diff?

mgyucht (commented):
One question and nit, but otherwise this looks good to me.

docs/resources/sql_table.md
```diff
@@ -136,7 +136,7 @@ For table columns
 Currently, changing the column definitions for a table will require dropping and re-creating the table
 
 * `name` - User-visible name of column
-* `type` - Column type spec (with metadata) as SQL text. Not supported for `VIEW` table_type.
+* `type` - (Optional) Column type spec (with metadata) as SQL text. Not supported for `VIEW` table_type.
```
tanmay-db, line 139: This type is omitempty but we don't specify this in our docs.

catalog/resource_sql_table.go
```diff
@@ -488,10 +488,23 @@ func columnChangesCustomizeDiff(d *schema.ResourceDiff, newTable *SqlTableInfo)
 	return nil
 }
 
+var columnTypeAliases = map[string]string{
```
nkvuong, line 491: there are additional aliases that are less used
- float | real - https://docs.databricks.com/en/sql/language-manual/data-types/float-type.html
- decimal | dec | numeric - https://docs.databricks.com/en/sql/language-manual/data-types/decimal-type.html
- smallint | short - https://docs.databricks.com/en/sql/language-manual/data-types/smallint-type.html
- tinyint | byte - https://docs.databricks.com/en/sql/language-manual/data-types/tinyint-type.html

nkvuong (approved):
looks good to me, minor comments about other niche aliases

catalog/resource_sql_table.go
nkvuong: `decimal` technically must be specified with precision & scale, i.e. `decimal(p,s)`. `decimal` just defaults to `decimal(10,0)`

mgyucht (approved):

fernandodakko:
There are also a couple of extra cases as stated here. Especially complex/nested types
https://github.com/pulumi/pulumi-databricks/issues/652

PR #3529, opened 2024-05-01
Added support for addition of multiple objects in `resource_share` 

tanmay-db:
## Changes

Objects in share resource is a list which leads to issues when they are added / removed due to ordering. This PR makes them as set and sets the name as hash function. 

Also this allows proper updates and removals because before this if the list items were to be deleted from middle of list, that could lead to wrong data being populated in the structure. 

## Tests

Integration test fails on main branch reproducing the issue. Passes on PR changes.  
- [x] `make test` run locally
- [ ] using Go SDK


tanmay-db:
Unit tests fails because of ordering in request / response -- need to fix those.

docs/resources/share.md
```diff
@@ -87,7 +85,7 @@ The following arguments are required:
 
 ### object Configuration Block
 
-* `name` (Required) - Full name of the object, e.g. `catalog.schema.name` for a tables, volumes and models, or `catalog.schema` for schemas.
+* `name` (Required) - Full name of the object, e.g. `catalog.schema.name` for a tables, volumes and models, or `catalog.schema` for schemas. Change forces creation of a new resource.
```
tanmay-db, line 88: Note: Since we now use name for hashing, we should create new resource on name change.

sharing/resource_share.go
```diff
@@ -2,9 +2,9 @@ package sharing
 
 import (
 	"context"
+	"sort"
 
 	"reflect"
-	"sort"
```
nkvuong, line 8: unnecessary change

sharing/resource_share.go
```diff
@@ -58,6 +58,12 @@ type ShareUpdates struct {
 	Updates []ShareDataChange `json:"updates"`
 }
 
+func (su *ShareUpdates) sortSharesByName() {
```
nkvuong, line 61: does the API expect the list of updates to be sorted?
```suggestion
func (su *ShareUpdates) sortUpdatesByName() {
```

sharing/resource_share.go
```diff
@@ -29,13 +29,13 @@ const (
...
 	Owner     string             `json:"owner,omitempty" tf:"suppress_diff"`
-	Objects   []SharedDataObject `json:"objects,omitempty" tf:"alias:object"`
+	Objects   []SharedDataObject `json:"objects,omitempty" tf:"slice_set,alias:object"`
 	CreatedAt int64              `json:"created_at,omitempty" tf:"computed"`
 	CreatedBy string             `json:"created_by,omitempty" tf:"computed"`
 }
 
 type SharedDataObject struct {
-	Name                     string      `json:"name"`
+	Name                     string      `json:"name" tf:"force_new"`
```
mgyucht, line 38: As discussed, let's see if this works without force_new.

sharing/resource_share.go
```diff
@@ -176,6 +184,9 @@ func (beforeSi ShareInfo) Diff(afterSi ShareInfo) []ShareDataChange {
 func ResourceShare() common.Resource {
 	shareSchema := common.StructToSchema(ShareInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
 		m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
+		m["object"].Set = func(i any) int {
+			return schema.HashString(i.(map[string]any)["name"].(string))
```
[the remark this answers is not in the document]
alexott: yep, we may have model & table with the same name

sharing/resource_share.go
tanmay-db: This is for consistency, we sort them in get. Also, this is required for our unit tests (expected request is a list and we use set now).

sharing/resource_share.go
tanmay-db: Yeah, I removed sorting from all places but we need this for our unit test so added it again, this change happened due to that.

sharing/resource_share.go
tanmay-db: If we remove force_new, then we need to check for element with empty Name (when we remove an object). This happens during dataToStruct. 

`````
if afterSdo.Name == "" {
    continue
}
`````

Integration test passes, I also added more steps. The comment gets updated. I need to check if this gets shown in diff.

Comment (tanmay-db) [reply sharing/resource_share.go]: Diff doesn't get shown when we update the comment, although the changes are reflected in the request we make and get from server. 

Diff only works for add/delete/rename

Comment (tanmay-db) [reply sharing/resource_share.go]: Update after discussion: https://github.com/databricks/terraform-provider-databricks/pull/3540.

File: sharing/resource_share.go [hunk 6]
```diff
@@ -176,6 +186,11 @@ func (beforeSi ShareInfo) Diff(afterSi ShareInfo) []ShareDataChange {
 func ResourceShare() common.Resource {
 	shareSchema := common.StructToSchema(ShareInfo{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
 		m["name"].DiffSuppressFunc = common.EqualFoldDiffSuppress
+		m["object"].Set = func(i any) int {
+			objectStruct := i.(map[string]any)
+			hashString := objectStruct["name"].(string) + objectStruct["data_object_type"].(string)
```
alexott, line 191: Should we add any separator between these fields?

@nkvuong do you remember - can we share the same object with different `shared_as`? Or if we can share same table with different `history_data_sharing_status`?

alexott (commented):
lgtm in general, but we need @nkvuong's opinion regarding the content of the hashing function...

sharing/resource_share.go
ScottReyn-db: > Diff doesn't get shown when we update the comment, although the changes are reflected in the request we make and get from server.
> 
> Diff only works for add/delete/rename

I realized this code doesn't allow updating the Share objects comment. Added that in #3629

PR #3532, opened 2024-05-02
Jobs GoSDK schema migration

edwardfeng-db:
## Changes

- Make `typeToSchema` check and reuse information for registered resources in the `resourceProviderRegistry`
- Taking `Library` out of the registry because the only customization is not needed in other resources
- Migrating jobs resource schema to use the GoSDK schema
  - Constructed resource provider with `CustomizeSchema` function that takes in all of the existing customizations
  - Tried making the new schema as similar as possible to the original schema
  - Updated the schema used in the resource as well as in `DataToStructPointer` and `StructToData` function calls
  - Updated `resource_job_test`
- Adding `CustomizeSchemaResourceSpecific` function so that customizations don't always have to be shared for registry references

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -123,6 +123,34 @@ type ClusterSpec struct {
 	Libraries []compute.Library `json:"libraries,omitempty" tf:"slice_set,alias:library"`
 }
 
+func (ClusterSpec) CustomizeSchemaResourceSpecific(s *common.CustomizableSchema) *common.CustomizableSchema {
+	s.SchemaPath("autotermination_minutes").SetDefault(60)
+	s.AddNewField("default_tags", &schema.Schema{
+		Type:     schema.TypeMap,
+		Computed: true,
+	})
+	s.AddNewField("is_pinned", &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			if old == "" && new == "false" {
+				return true
+			}
+			return old == new
+		},
+	})
+	s.AddNewField("state", &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: true,
+	})
+	s.AddNewField("url", &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: true,
+	})
+	return s
+}
+
 func (ClusterSpec) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
 	s.SchemaPath("cluster_source").SetReadOnly()
 	s.SchemaPath("enable_elastic_disk").SetComputed()
@@ -158,7 +186,6 @@ func (ClusterSpec) CustomizeSchema(s *common.CustomizableSchema) *common.Customi
 	s.SchemaPath("aws_attributes", "zone_id").SetCustomSuppressDiff(ZoneDiffSuppress)
 	s.SchemaPath("azure_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "gcp_attributes"})
 	s.SchemaPath("gcp_attributes").SetSuppressDiff().SetConflictsWith([]string{"aws_attributes", "azure_attributes"})
-	s.SchemaPath("autotermination_minutes").SetDefault(60)
 	s.SchemaPath("autoscale", "max_workers").SetOptional()
 	s.SchemaPath("autoscale", "min_workers").SetOptional()
 	s.SchemaPath("cluster_log_conf", "dbfs", "destination").SetRequired()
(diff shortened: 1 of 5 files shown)


clusters/resource_cluster.go
```diff
@@ -498,5 +498,4 @@ func resourceClusterDelete(ctx context.Context, d *schema.ResourceData, c *commo
 
 func init() {
 	common.RegisterResourceProvider(compute.ClusterSpec{}, ClusterSpec{})
-	common.RegisterResourceProvider(compute.Library{}, LibraryResource{})
```
edwardfeng-db, line 501: Removing it from the registry because the only customization in `library` is the added required `cluster_id` field, it is only needed in `library` and not anywhere else.

jobs/resource_job_test.go
```diff
@@ -2466,10 +2478,11 @@ func TestResourceJobUpdate_InstancePoolToNodeType(t *testing.T) {
...
-							NodeTypeID:           "node-type-id",
-							DriverNodeTypeID:     "driver-node-type-id",
-							InstancePoolID:       "instance-pool-id-worker",
-							DriverInstancePoolID: "instance-pool-id-driver",
+							AutoterminationMinutes: 60,
+							NodeTypeID:             "node-type-id",
+							DriverNodeTypeID:       "driver-node-type-id",
+							InstancePoolID:         "instance-pool-id-worker",
+							DriverInstancePoolID:   "instance-pool-id-driver",
 						},
```
edwardfeng-db, line 2486: Only change to this file is adding `AutoterminationMinutes` to `60` as that is the default value being pulled in from the registry

jobs/resource_job.go
```diff
@@ -445,6 +445,184 @@ type JobsAPI struct {
...
+		},
+		"jobs.Task": {
+			"libraries": "library",
+		},
+	}
+	return aliases
+}
+
+func (JobSettingsResource) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
+	// ======= Replicate what was originally in CustomizeSchema call back func in the manually written schema. ======
```
mgyucht, line 493: This probably doesn't need to be here as a comment in source control, but it could be a comment on the PR :)

jobs/resource_job.go
```diff
@@ -445,6 +445,184 @@ type JobsAPI struct {
...
+	fixWebhookNotifications(s.GetSchemaMap())
+	fixWebhookNotifications(common.MustSchemaMap(s.GetSchemaMap(), "task"))
+
+	// Suppress diff if the platform returns ALL_SUCCESS for run_if in a task
+	s.SchemaPath("task", "run_if").SetSuppressDiffWithDefault(jobs.RunIfAllSuccess)
+	s.SchemaPath("task", "for_each_task", "task", "run_if").SetSuppressDiffWithDefault(jobs.RunIfAllSuccess)
+
+	s.SchemaPath("format").SetComputed()
+
+	// ========== To keep consisteny with the manually maintained schema, should be reverted once full migration is done. ======
```
mgyucht, line 567:
```suggestion
	// ========== To keep consistency with the manually maintained schema, should be reverted once full migration is done. ======
```

jobs/resource_job_test.go
mgyucht: Is this correct? What does auto-termination even mean for a job cluster? That default seems to make most sense for interactive clusters.

mgyucht (commented):
Is the autotermination change intentional? Seems like that may be needed/desired only for `databricks_cluster` resource.

jobs/resource_job_test.go
edwardfeng-db: Updated, it is no longer there

edwardfeng-db:
Triggered integration tests and confirmed that no failure is related to this change

clusters/resource_cluster.go
mgyucht: Add this back in after making cluster ID field resource-specific.

common/reflect_resource.go
```diff
@@ -101,7 +127,23 @@ type RecursiveResourceProvider interface {
 }
 
 // Takes in a ResourceProvider and converts that into a map from string to schema.
+// Used by typeToSchema when encountering types registered in the resourceProviderRegistry.
+func resourceProviderStructToSchemaNestedResource(v ResourceProvider, scp schemaPathContext) map[string]*schema.Schema {
```
mgyucht, line 131: Let's just inline this and avoid the (for now) unnecessary abstraction.

mgyucht (commented):
Great job! Almost there. We should remove cluster_id and cluster_source from wherever "new_cluster" appears in this schema, otherwise the new schema looks good.

mgyucht (approved):

PR #3537, opened 2024-05-02
Retry API calls until context timeout

mgyucht:
## Changes
The SDK currently sets a default retry timeout of 5 minutes for all API requests. This is convenient for users who call APIs with contexts with no timeout or that can't be cancelled, like context.Background(). However, in Terraform, the context timeout is determined by a combination of instance-level timeouts, resource-level timeouts, and a default timeout of 20m.

This PR changes the provider to use an indefinite retry timeout for all API requests. This allows API calls to be retried for the entire length of the resource creation operation, e.g. in case there is a low rate limit for an API and the resources in question cannot all be created within 5 minutes.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -279,6 +279,12 @@ func configureDatabricksClient(ctx context.Context, d *schema.ResourceData) (any
 			cfg.AuthType = newer
 		}
 	}
+	cfg.EnsureResolved()
+	// Unless set explicitly, the provider will retry indefinitely until context is cancelled
+	// by either a timeout or interrupt.
+	if cfg.RetryTimeoutSeconds == 0 {
+		cfg.RetryTimeoutSeconds = -1
+	}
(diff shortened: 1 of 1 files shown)


provider/provider.go
```diff
@@ -279,6 +279,12 @@ func configureDatabricksClient(ctx context.Context, d *schema.ResourceData) (any
 			cfg.AuthType = newer
 		}
 	}
+	cfg.EnsureResolved()
```
alexott, line 282: do we still have the lazy authentication here?

mgyucht: Yeah. `EnsureResolved` is actually called in `client.New()`. It makes sure that we've captured settings from the users .databrickscfg profile and/or environment variables, but it doesn't authenticate at all.

alexott (approved):

PR #3540, opened 2024-05-03
Added support for CustomizeDiff for resource_share

tanmay-db:
## Changes

We are trying to see whether using CustomizeDiff works since using force_new isn't for any object update isn't feasible. This is a followup to the PR: https://github.com/databricks/terraform-provider-databricks/pull/3529

Please see the highlighted comment below. 

This is a great use case for prioritising moving towards using terraform plugin framework (https://github.com/databricks/terraform-provider-databricks/pull/3510) as we can access config, state and plan easily: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/accessing-values whereas here we would have to use more hacks and it would lead to piling up of technical debt even if we get to have it working somehow. 

This situation could occur for other similar resources with lists. 

## Tests

Integration test
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/sharing/resource_share.go
+++ b/sharing/resource_share.go
@@ -2,6 +2,7 @@ package sharing
 
 import (
 	"context"
+	"strconv"
 
 	"reflect"
 	"sort"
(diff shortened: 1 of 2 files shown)


sharing/resource_share.go
```diff
@@ -173,13 +174,63 @@ func (beforeSi ShareInfo) Diff(afterSi ShareInfo) []ShareDataChange {
...
+			objectsListOld := getSharedDataObjectsList(objectsOld, objectSchemaMap)
+			objectsListNew := getSharedDataObjectsList(objectsNew, objectSchemaMap)
+			for i, newObject := range objectsListNew {
+				if i >= len(objectsListOld) {
+					break
+				}
+				if newObject["name"] == objectsListOld[i]["name"] && newObject["data_object_type"] == objectsListOld[i]["data_object_type"] {
+					for k := range objectSchemaMap {
+						oldFieldVal := objectsListOld[i][k]
+						// To Discuss:
```
tanmay-db, line 205: Highlight

PR #3544, opened 2024-05-03
Update data_aws_crossaccount_policy to format resource string with account and region

frosforever:
String with formatting wasn't passed to `fmt.Sprintf` resulting in the actual policy containing '%s' e.g. 'arn:aws:ec2:%s:%s:volume/*'. Passing the string to `fmt.Sprintf` results in the desired iam policy json.


alexott (approved):

alexott:
@frosforever can you please fix the tests?

frosforever:
@alexott Done! Didn't realize it was checking the exact length as a proxy for what to expect in the policy. Not obvious to me how useful a test that is besides verifying that the policy hasn't accidentally changed, but I'll leave that as a future exercise. 

Thanks for the review and for maintaining the provider!

PR #3550, opened 2024-05-06
Robust retries for workspace get-status

mgyucht:
## Changes
To reduce flakiness of Terraform when calling Workspace Get Status, I've added retries around `GetStatusByPath` for all usages of this API in the Terraform provider (aside from the exporter, which depends on a field that isn't yet exposed in the SDK). This should help increase stability of fetching the status of notebooks and directories in the provider.

This change introduces a utility to retry on timeout for this common use case. We may want to add this to the Go SDK if we feel it is generally useful as a retry option (e.g. `retries.OnRequestTimeout`).

Longer term, we should handle these retries for idempotent APIs directly in the SDK.

## Tests
Unit tests verify that this retries on timeouts but otherwise doesn't retry.

- [ ] `make test` run locally
- [ ] using Go SDK


alexott (approved):
code wise looks good, we just need to fix imports

hectorcast-db (approved):

PR #3551, opened 2024-05-07
Added state upgrader to remove `max_clusters_per_user` set to zero

andrewnester:
## Changes
Added state upgrader to remove `max_clusters_per_user` set to zero

## Tests
Manually (deploying policy with 1.41 version, upgrading to the fix and applying the changes. Changes succeed and schema updated).

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/policies/data_cluster_policy.go
+++ b/policies/data_cluster_policy.go
@@ -2,14 +2,16 @@ package policies
 
 import (
 	"context"
+	"log"
 
 	"github.com/databricks/databricks-sdk-go"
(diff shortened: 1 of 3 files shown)


hectorcast-db:
The change looks good. However, do we understand why this issue started happening now? I don't see recent changes on these resources which could cause this.

alexott:
@hectorcast-db #3385 could cause it with a combination with having an object with defaults in the state...

hectorcast-db (approved):

PR #3444, opened 2024-04-09
Added `databricks_automatic_cluster_update_workspace_setting` resource

nkvuong:
## Changes
- Added `databricks_automatic_cluster_update_workspace_setting` resource

## Tests

- [x] `make test` run locally
- [x] relevant change in `docs/` folder
- [x] using Go SDK
- [x] tested manually

PR #3563, opened 2024-05-10
Added `databricks_enhanced_security_monitoring_workspace_setting ` resource for ESC (Enhanced Compliance and Security) settings

harshshah-db:
## Changes
Added enhanced_security_monitoring settings. Will add compliance_security_profile in a follow up PR.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


settings/resource_enhanced_security_monitoring_setting.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/settings"
+)
+
+// Enhanced Security Monitoring setting
+var enhancedSecurityMonitoringFieldMask = strings.Join([]string{
+	"enhanced_security_monitoring_workspace.is_enabled",
+}, ",")
+var enhancedSecurityMonitoringSetting = workspaceSetting[settings.EnhancedSecurityMonitoringSetting]{
```
nkvuong, line 15: my understanding is that this setting can be set at both account & workspace level - can we use `accountWorkspaceSetting` instead

settings/all_settings.go
```diff
@@ -15,7 +15,9 @@ import (
 //  3. Add a new entry to the AllSettingsResources map below. The final resource name will be "databricks_<SETTING_NAME>_setting".
 func AllSettingsResources() map[string]common.Resource {
 	return map[string]common.Resource{
-		"default_namespace":         makeSettingResource[settings.DefaultNamespaceSetting, *databricks.WorkspaceClient](defaultNamespaceSetting),
-		"restrict_workspace_admins": makeSettingResource[settings.RestrictWorkspaceAdminsSetting, *databricks.WorkspaceClient](restrictWsAdminsSetting),
+		"default_namespace":                      makeSettingResource[settings.DefaultNamespaceSetting, *databricks.WorkspaceClient](defaultNamespaceSetting),
+		"restrict_workspace_admins":              makeSettingResource[settings.RestrictWorkspaceAdminsSetting, *databricks.WorkspaceClient](restrictWsAdminsSetting),
+		"enhanced_security_monitoring_workspace": makeSettingResource[settings.EnhancedSecurityMonitoringSetting, *databricks.WorkspaceClient](enhancedSecurityMonitoringSetting),
+		"automatic_cluster_update_workspace":     makeSettingResource[settings.AutomaticClusterUpdateSetting, *databricks.WorkspaceClient](automaticClusterUpdateSetting),
```
nkvuong, line 21: could you keep the changes for automaticClusterUpdate in a separate PR? I've actually started it in #3444 if you want to review

harshshah-db: thanks, approved that PR and removed automaticClusterUpdate from this.

settings/resource_enhanced_security_monitoring_setting.go
harshshah-db: those are 2 separate settings with different fields:

1. Account level - https://github.com/databricks/universe/blob/ea7ffc3ed4b3ef62407dc7ca74e1210ab926e84a/settings-policy/api/proto/external/service.proto#L63
2. Workspace level - https://github.com/databricks/universe/blob/ea7ffc3ed4b3ef62407dc7ca74e1210ab926e84a/settings-policy/api/proto/external/service.proto#L66

Should they still be combined?

nkvuong: this is interesting - so structure wise they are the same - <something>.<is_enabled>, but the top-level field name is different. From a usability perspective, it is easier for users to have a single resource, but let's see how much work it is to add aliasing in TF

harshshah-db: I don't think they are the same - the account level has the field`is_enforced` and workspace level is `is_enabled`. The account level is only used for setting defaults for new workspaces (new workspaces auto inherit the account level setting if set).

settings/all_settings.go
harshshah-db: @nkvuong would you please approve this PR if it looks good? Want to merge this if no additional comments.

nkvuong (approved):

harshshah-db:
> @harshshah-db there is no Jenkins here ;-)

yeah just figured that :) How do I add it to the merge queue?

PR #3569, opened 2024-05-13
Exporter: Don't list directories in the incremental mode

alexott:
## Changes

Because directories don't have timestamps, we're ending up emitting all of them anyway...
Instead, rely on the behavior of them emitted as dependencies for notebooks

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


nkvuong (approved):
looks good to me. my understanding is that in incremental mode, we won't emit new directories that are empty. if so, do we need to note it down?

PR #3571, opened 2024-05-13
add `databricks_table` data source

nkvuong:
## Changes
- Add `databricks_table` data source. This should simplify use cases where customers need to retrieve a specific table for permissions granting & other purposes

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/docs/data-sources/table.md
+++ b/docs/data-sources/table.md
@@ -19,9 +19,7 @@ data "databricks_table" "fct_transactions" {
 }
 
 resource "databricks_grants" "things" {
-  for_each = data.databricks_tables.things.ids
-
-  table = each.value
+  table = data.databricks_table.fct_transactions.name
 
   grant {
     principal  = "sensitive"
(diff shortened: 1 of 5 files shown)


docs/data-sources/table.md
```diff
@@ -0,0 +1,58 @@
...
+resource "databricks_grants" "things" {
+  for_each = data.databricks_tables.things.ids
+
+  table = each.value
+
+  grant {
+    principal  = "sensitive"
+    privileges = ["SELECT", "MODIFY"]
+  }
+}
```
alexott, line 30: This doesn't match the table ID, and it's not a list. We need to get something else here as an example.

alexott (changes requested):
we need to fix docs example...

alexott (approved):

tanmay-db (approved):

jdavidheiser:
thank you for pushing this over the finish line!

PR #3577, opened 2024-05-14
Jobs Methods GoSDK Migration 

edwardfeng-db:
## Changes

- Migrated `Read, Update, Create` methods to the go-sdk migration for the jobs resource
  - Updated cluster related methods to accept `compute.ClusterSpec` as a valid input in `resource_cluster.go`
  - Created `jobs_api_go_sdk.go` to replicate go-sdk versions of necessary methods that were part of `JobsApi`
  - Updated jobs manually written schema to be consistent with the go-sdk schema, mostly taking out some `omitempty`
  - Updated the `Read, Update, Create` methods to have two branches, if it's api2.0, make it run the exact same code as before, otherwise, use the go-sdk to perform the same actions
  - Updated unit tests to include some missing fields that are newly marked as required, such as `task_key`
  - Updated doc about schema changes
  - Removed `autotermination_minutes` from the cluster spec used by jobs because it is confirmed to be never useful for jobs

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -24,7 +24,7 @@ var clusterSchemaVersion = 4
 
 const (
 	numWorkerErr                     = "NumWorkers could be 0 only for SingleNode clusters. See https://docs.databricks.com/clusters/single-node.html for more details"
-	unsupportedExceptCreateOrEditErr = "unsupported type %T, must be either %scompute.CreateCluster or %scompute.EditCluster. Please report this issue to the GitHub repo"
+	unsupportedExceptCreateOrEditErr = "unsupported type %T, must be one of %scompute.CreateCluster, %scompute.ClusterSpec or %scompute.EditCluster. Please report this issue to the GitHub repo"
 )
 
 func ResourceCluster() common.Resource {
@@ -127,8 +127,15 @@ func Validate(cluster any) error {
 		profile = c.SparkConf["spark.databricks.cluster.profile"]
 		master = c.SparkConf["spark.master"]
 		resourceClass = c.CustomTags["ResourceClass"]
+	case compute.ClusterSpec:
+		if c.NumWorkers > 0 || c.Autoscale != nil {
+			return nil
+		}
+		profile = c.SparkConf["spark.databricks.cluster.profile"]
+		master = c.SparkConf["spark.master"]
+		resourceClass = c.CustomTags["ResourceClass"]
 	default:
-		return fmt.Errorf(unsupportedExceptCreateOrEditErr, cluster, "", "")
+		return fmt.Errorf(unsupportedExceptCreateOrEditErr, cluster, "", "", "")
 	}
 	if profile == "singleNode" && strings.HasPrefix(master, "local") && resourceClass == "SingleNode" {
 		return nil
@@ -140,6 +147,34 @@ func Validate(cluster any) error {
 // Long term, ModifyRequestOnInstancePool() in clusters_api.go will be removed once all the resources using clusters are migrated to Go SDK.
 func ModifyRequestOnInstancePool(cluster any) error {
 	switch c := cluster.(type) {
+	case *compute.ClusterSpec:
+		// Instance profile id does not exist or not set
+		if c.InstancePoolId == "" {
+			// Worker must use an instance pool if driver uses an instance pool,
+			// therefore empty the computed value for driver instance pool.
+			c.DriverInstancePoolId = ""
+			return nil
+		}
+		if c.AwsAttributes != nil {
+			// Reset AwsAttributes
+			awsAttributes := compute.AwsAttributes{
+				InstanceProfileArn: c.AwsAttributes.InstanceProfileArn,
+			}
+			c.AwsAttributes = &awsAttributes
+		}
+		if c.AzureAttributes != nil {
+			c.AzureAttributes = &compute.AzureAttributes{}
+		}
+		if c.GcpAttributes != nil {
+			gcpAttributes := compute.GcpAttributes{
+				GoogleServiceAccount: c.GcpAttributes.GoogleServiceAccount,
+			}
+			c.GcpAttributes = &gcpAttributes
+		}
+		c.EnableElasticDisk = false
+		c.NodeTypeId = ""
+		c.DriverNodeTypeId = ""
+		return nil
 	case *compute.CreateCluster:
 		// Instance profile id does not exist or not set
 		if c.InstancePoolId == "" {
@@ -197,20 +232,35 @@ func ModifyRequestOnInstancePool(cluster any) error {
 		c.DriverNodeTypeId = ""
 		return nil
 	default:
-		return fmt.Errorf(unsupportedExceptCreateOrEditErr, cluster, "*", "*")
+		return fmt.Errorf(unsupportedExceptCreateOrEditErr, cluster, "*", "*", "*")
 	}
 }
 
 // This method is a duplicate of FixInstancePoolChangeIfAny(d *schema.ResourceData) in clusters/clusters_api.go that uses Go SDK.
 // Long term, FixInstancePoolChangeIfAny(d *schema.ResourceData) in clusters_api.go will be removed once all the resources using clusters are migrated to Go SDK.
 // https://github.com/databricks/terraform-provider-databricks/issues/824
-func FixInstancePoolChangeIfAny(d *schema.ResourceData, cluster *compute.EditCluster) {
-	oldInstancePool, newInstancePool := d.GetChange("instance_pool_id")
-	oldDriverPool, newDriverPool := d.GetChange("driver_instance_pool_id")
-	if oldInstancePool != newInstancePool &&
-		oldDriverPool == oldInstancePool &&
-		oldDriverPool == newDriverPool {
-		cluster.DriverInstancePoolId = cluster.InstancePoolId
+func FixInstancePoolChangeIfAny(d *schema.ResourceData, cluster any) error {
+	switch c := cluster.(type) {
+	case *compute.ClusterSpec:
+		oldInstancePool, newInstancePool := d.GetChange("instance_pool_id")
+		oldDriverPool, newDriverPool := d.GetChange("driver_instance_pool_id")
+		if oldInstancePool != newInstancePool &&
+			oldDriverPool == oldInstancePool &&
+			oldDriverPool == newDriverPool {
+			c.DriverInstancePoolId = c.InstancePoolId
+		}
+		return nil
+	case *compute.EditCluster:
+		oldInstancePool, newInstancePool := d.GetChange("instance_pool_id")
+		oldDriverPool, newDriverPool := d.GetChange("driver_instance_pool_id")
+		if oldInstancePool != newInstancePool &&
+			oldDriverPool == oldInstancePool &&
+			oldDriverPool == newDriverPool {
+			c.DriverInstancePoolId = c.InstancePoolId
+		}
+		return nil
+	default:
+		return fmt.Errorf(unsupportedExceptCreateOrEditErr, cluster, "*", "*", "*")
 	}
 }
 
@@ -220,7 +270,6 @@ type ClusterSpec struct {
 }
 
 func (ClusterSpec) CustomizeSchemaResourceSpecific(s *common.CustomizableSchema) *common.CustomizableSchema {
-	s.SchemaPath("autotermination_minutes").SetDefault(60)
 	s.AddNewField("default_tags", &schema.Schema{
 		Type:     schema.TypeMap,
 		Computed: true,
@@ -243,6 +292,11 @@ func (ClusterSpec) CustomizeSchemaResourceSpecific(s *common.CustomizableSchema)
 	s.AddNewField("url", &schema.Schema{
 		Type:     schema.TypeString,
 		Computed: true,
+	})
+	s.AddNewField("autotermination_minutes", &schema.Schema{
+		Type:     schema.TypeInt,
+		Optional: true,
+		Default:  60,
 	})
 	return s
 }
@@ -300,6 +354,8 @@ func (ClusterSpec) CustomizeSchema(s *common.CustomizableSchema) *common.Customi
 			Schema: common.StructToSchema(MountInfo{}, nil),
 		},
 	})
+	// Adding it back in the resource specific customization function because it is not relevant for other resources.
+	s.RemoveField("autotermination_minutes")
 
 	return s
 }
@@ -453,7 +509,10 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
 		if err = ModifyRequestOnInstancePool(&cluster); err != nil {
 			return err
 		}
-		FixInstancePoolChangeIfAny(d, &cluster)
+		err = FixInstancePoolChangeIfAny(d, &cluster)
+		if err != nil {
+			return err
+		}
 
 		// We can only call the resize api if the cluster is in the running state
 		// and only the cluster size (ie num_workers OR autoscale) is being changed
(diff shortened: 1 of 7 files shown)


jobs/jobs_api_go_sdk.go
```diff
@@ -0,0 +1,289 @@
...
+func removeAutoterminationMinutesFromForceSendFields(clusterSpec *compute.ClusterSpec) error {
+	newForceSendFields := []string{}
+	for _, field := range clusterSpec.ForceSendFields {
+		if field != "AutoterminationMinutes" {
+			newForceSendFields = append(newForceSendFields, field)
+		}
+	}
+	clusterSpec.ForceSendFields = newForceSendFields
+	return nil
+}
```
edwardfeng-db, line 173: Another option would be to remove `autotermination_minutes` from the jobs schema, as it should never be set from the jobs side

edwardfeng-db:
Verified that nightly integration tests passed for this PR

jobs/jobs_api_go_sdk.go
```diff
@@ -0,0 +1,307 @@
...
+	sortTasksByKey(cj)
+	for _, task := range cj.Tasks {
+		sort.Slice(task.DependsOn, func(i, j int) bool {
+			return task.DependsOn[i].TaskKey < task.DependsOn[j].TaskKey
+		})
+		sortWebhookNotifications(task.WebhookNotifications)
+	}
+}
+
+func sortTasksByKey(cj *jobs.CreateJob) {
```
tanmay-db, line 47: Moving these methods to Go SDK might be better.

jobs/jobs_api_go_sdk.go
```diff
@@ -0,0 +1,307 @@
...
+		if errors.As(err, &apiErr) && apiErr.StatusCode != 404 {
+			return err
+		}
+	}
+
+	return StopActiveRun(jobID, c.d.Timeout(schema.TimeoutUpdate), w, ctx)
+}
+
+// Removing unnecesary fields out of ClusterSpec's ForceSendFields because the current terraform plugin sdk
+// has a limitation that it will always set unspecified values to the zero value.
```
tanmay-db, line 161: Some context on how the plugin framework could help in case of null vs zero: 

> SDKv2 also does not expose Terraform's null or unknown value concepts unambiguously, typically returning both null and unknown values as the same zero-value for the underlying Go type.

> The framework type system fully exposes null, unknown, and known value states. You can reliably query each value with the IsNull() or IsUnknown() methods.

https://developer.hashicorp.com/terraform/plugin/framework/migrating/benefits

jobs/jobs_api_go_sdk.go
tanmay-db: I think if all the clusters are automated for jobs resource, then we could remove autotermination_minutes from the schema.

clusters/resource_cluster.go
```diff
@@ -197,20 +232,44 @@ func ModifyRequestOnInstancePool(cluster any) error {
...
+	case *compute.ClusterSpec:
+		oldInstancePool, newInstancePool := d.GetChange("instance_pool_id")
+		oldDriverPool, newDriverPool := d.GetChange("driver_instance_pool_id")
+		if oldInstancePool != newInstancePool &&
+			oldDriverPool == oldInstancePool &&
+			oldDriverPool == newDriverPool {
+			c.DriverInstancePoolId = c.InstancePoolId
+		}
+		return nil
+	case *compute.CreateCluster:
```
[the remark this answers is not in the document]
edwardfeng-db: Yeah the go-sdk expects a `CreateCluster` for the create cluster call, but most of the fields are exactly the same as UpdateCluster and ClusterSpec

docs/resources/job.md
```diff
@@ -127,7 +127,7 @@ This block describes individual tasks:
   * `sql_task`
 * `library` - (Optional) (Set) An optional list of libraries to be installed on the cluster that will execute the job.
 * `depends_on` - (Optional) block specifying dependency(-ies) for a given task.
-* `job_cluster_key` - (Optional) Identifier of the Job cluster specified in the `job_cluster` block.
+* `job_cluster_key` - (Required) Identifier of the Job cluster specified in the `job_cluster` block.
```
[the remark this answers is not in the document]
edwardfeng-db: I think it was already required in the api, just that we weren't requiring it in the terraform level, so it should be fine. I confirmed with the jobs team

jobs/jobs_api_go_sdk.go
edwardfeng-db: That is true, can do that later

clusters/resource_cluster.go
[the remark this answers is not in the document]
edwardfeng-db: Yeah you're right, it is not really needed for create

docs/resources/job.md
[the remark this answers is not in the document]
tanmay-db: I see

jobs/resource_job.go
```diff
@@ -446,6 +446,45 @@ type JobsAPI struct {
...
+		"tasks":        "task",
+		"parameters":   "parameter",
+		"job_clusters": "job_cluster",
+		"environments": "environment",
+	},
+	"jobs.JobCreateStruct": {
+		"tasks":        "task",
+		"parameters":   "parameter",
+		"job_clusters": "job_cluster",
+		"environments": "environment",
```
tanmay-db, line 461: Since these are repeated, we could have them in a single map to reduce redundancy.

jobs/jobs_api_go_sdk.go
```diff
@@ -0,0 +1,307 @@
...
+			if !slices.Contains(unnecessaryFieldNamesForAwsAttributes, field) {
+				newAwsAttributesForceSendFields = append(newAwsAttributesForceSendFields, field)
+			}
+		}
+		clusterSpec.AwsAttributes.ForceSendFields = newAwsAttributesForceSendFields
+	}
+	return nil
+}
+
+func prepareJobSettingsForUpdateGoSdk(d *schema.ResourceData, js JobSettingsResource) {
```
tanmay-db, line 193: This should be `*JobSettingsResource`?

jobs/jobs_api_go_sdk.go
```diff
@@ -0,0 +1,307 @@
...
+		if task.NewCluster != nil {
+			clusters.ModifyRequestOnInstancePool(task.NewCluster)
+			clusters.FixInstancePoolChangeIfAny(d, &task.NewCluster)
+			removeUnnecessaryFieldsFromForceSendFields(task.NewCluster)
+		}
+	}
+	for i := range js.JobClusters {
+		clusters.ModifyRequestOnInstancePool(&js.JobClusters[i].NewCluster)
+		clusters.FixInstancePoolChangeIfAny(d, &js.JobClusters[i].NewCluster)
+		removeUnnecessaryFieldsFromForceSendFields(&js.JobClusters[i].NewCluster)
```
tanmay-db, line 209: As these 3 methods are repeated in order, we could have a wrapper function for them.

tanmay-db (approved):
Had some comments, LGTM

jobs/jobs_api_go_sdk.go
edwardfeng-db: Looks like we don't have to, but yeah changing it to be safe

clusters/resource_cluster.go
```diff
@@ -244,6 +293,11 @@ func (ClusterSpec) CustomizeSchemaResourceSpecific(s *common.CustomizableSchema)
 		Type:     schema.TypeString,
 		Computed: true,
 	})
+	s.AddNewField("autotermination_minutes", &schema.Schema{
+		Type:     schema.TypeInt,
+		Optional: true,
+		Default:  60,
+	})
```
edwardfeng-db, line 300: Only adding it as a resource specific field for cluster because job resource never needs it

jobs/resource_job.go
```diff
@@ -608,34 +624,16 @@ func (JobSettingsResource) CustomizeSchema(s *common.CustomizableSchema) *common
 	s.SchemaPath("task", "sql_task", "alert", "subscriptions").SetRequired()
 	s.SchemaPath("task", "for_each_task", "task", "sql_task", "alert", "subscriptions").SetRequired()
 
-	s.SchemaPath("health", "rules", "metric").SetOptional()
```
nkvuong, line 611: would this break existing user tf code since these fields are not optional anymore?

clusters/resource_cluster.go
```diff
@@ -140,6 +147,34 @@ func Validate(cluster any) error {
 // Long term, ModifyRequestOnInstancePool() in clusters_api.go will be removed once all the resources using clusters are migrated to Go SDK.
 func ModifyRequestOnInstancePool(cluster any) error {
 	switch c := cluster.(type) {
+	case *compute.ClusterSpec:
```
nkvuong, line 150: should these supporting functions be split into a clusters_api_go_sdk.go file? I don't like the code duplication here, but this is more of an issue with the underlying sdk structs

docs/resources/job.md
```diff
@@ -357,7 +357,7 @@ This block describes the queue settings of the job:
 This block is used to specify Git repository information & branch/tag/commit that will be used to pull source code from to execute a job. Supported options are:
 
 * `url` - (Required) URL of the Git repository to use.
-* `provider` - (Optional, if it's possible to detect Git provider by host name) case insensitive name of the Git provider.  Following values are supported right now (could be a subject for change, consult [Repos API documentation](https://docs.databricks.com/dev-tools/api/latest/repos.html)): `gitHub`, `gitHubEnterprise`, `bitbucketCloud`, `bitbucketServer`, `azureDevOpsServices`, `gitLab`, `gitLabEnterpriseEdition`.
+* `provider` - (Required, if it's possible to detect Git provider by host name) case insensitive name of the Git provider.  Following values are supported right now (could be a subject for change, consult [Repos API documentation](https://docs.databricks.com/dev-tools/api/latest/repos.html)): `gitHub`, `gitHubEnterprise`, `bitbucketCloud`, `bitbucketServer`, `azureDevOpsServices`, `gitLab`, `gitLabEnterpriseEdition`.
```
nkvuong, line 360: don't need this change - we try to detect `provider` based on host name

docs/resources/job.md
```diff
@@ -468,8 +468,8 @@ This syntax uses Jobs API 2.0 to create a job with a single task. Only a subset
 
 The job cluster is specified using either of the below argument:
 
-* `new_cluster` - (Optional) Same set of parameters as for [databricks_cluster](cluster.md) resource.
-* `existing_cluster_id` - (Optional) If existing_cluster_id, the ID of an existing [cluster](cluster.md) that will be used for all runs of this job. When running jobs on an existing cluster, you may need to manually restart the cluster if it stops responding. We strongly suggest to use `new_cluster` for greater reliability.
+* `new_cluster` - (Required) Same set of parameters as for [databricks_cluster](cluster.md) resource.
```
nkvuong, line 471: exactly one of these two arguments is required, (Required) is misleading

jobs/resource_job_test.go
```diff
@@ -1973,7 +1975,7 @@ func TestResourceJobCreateFromGitSourceWithoutProviderFail(t *testing.T) {
 			}
 		}
 	`,
-	}.ExpectError(t, "git source is not empty but Git Provider is not specified and cannot be guessed by url &{Url:https://custom.git.hosting.com/databricks/terraform-provider-databricks Provider: Branch: Tag:0.4.8 Commit: JobSource:<nil>}")
```
nkvuong, line 1976: this test should not be changed

jobs/resource_job_test.go
```diff
@@ -2436,21 +2442,14 @@ func TestResourceJobUpdate_NodeTypeToInstancePool(t *testing.T) {
...
 		InstanceState: map[string]string{
-			"new_cluster.0.node_type_id":                      "node-type-id-worker",
-			"new_cluster.0.driver_node_type_id":               "node-type-id-driver",
 			"task.0.new_cluster.0.node_type_id":               "node-type-id-worker-task",
 			"task.0.new_cluster.0.driver_node_type_id":        "node-type-id-driver-task",
 			"job_cluster.0.new_cluster.0.node_type_id":        "node-type-id-worker-job",
 			"job_cluster.0.new_cluster.0.driver_node_type_id": "node-type-id-driver-job",
 		},
 		HCL: `
-		new_cluster = {
```
nkvuong, line 2447: why is this block removed?

nkvuong (commented):
have left a few questions regarding to changes in our unit tests, and also potential breaking changes for git provider, and some other optional -> required fields

jobs/resource_job.go
edwardfeng-db: I think it was not optional on the API level anyway, as it is consistent with the openapi spec. So I don't think it will be a breaking change

clusters/resource_cluster.go
edwardfeng-db: Yes, that is a good point. I can move them over in a later refactor PR given that this one is already really big

docs/resources/job.md
edwardfeng-db: Ah right, let me revert this part

docs/resources/job.md
edwardfeng-db: Is it tho? I think they are both marked required on the openapi spec, as well as on the public facing [documentation](https://docs.databricks.com/api/workspace/jobs/create#job_clusters). Does that mean the openapi spec is wrong?

jobs/resource_job_test.go
edwardfeng-db: I see, yeah if we are inferring the value of provider then I agree this shouldn't be changed. Let me update it

jobs/resource_job_test.go
edwardfeng-db: Because this config was wrong, it is intending to use the jobs 2.1 API because it has `task` specified, but top level `new_cluster` is never consumed by the 2.1 API, so this `new_cluster` block is never necessary

common/reflect_resource.go
```diff
@@ -617,7 +617,7 @@ func iterFields(rv reflect.Value, path []string, s map[string]*schema.Schema, al
 			return fmt.Errorf("inconsistency: %s has omitempty, but is not optional", fieldName)
 		}
 		defaultEmpty := reflect.ValueOf(fieldSchema.Default).Kind() == reflect.Invalid
-		if fieldSchema.Optional && defaultEmpty && !omitEmpty {
+		if !isGoSDK && fieldSchema.Optional && defaultEmpty && !omitEmpty {
```
edwardfeng-db, line 620: Adding additional check because we are customizing `git_provider` to be optional while it doesn't have `omitEmpty` on the go-sdk struct.

docs/resources/job.md
nkvuong: actually, this part of the documentation is related to `Single-task syntax (deprecated)`, so we should not update the doc

docs/resources/job.md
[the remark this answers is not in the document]
nkvuong: iirc, it is exactlyoneof the 3 `job_cluster_key`, `existing_cluster_id`, `new_cluster`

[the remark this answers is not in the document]
edwardfeng-db: Right, good catch, let me update it

docs/resources/job.md
edwardfeng-db: Updated

jobs/resource_job_test.go
```diff
@@ -1884,6 +1885,7 @@ func TestResourceJobCreateFromGitSource(t *testing.T) {
 				import_from_git_branch = "main"
 				dirty_state = "NOT_SYNCED"
 			}
+			provider = "gitHub"
```
nkvuong, line 1888: don't need this

nkvuong (approved):
looks good to me, thanks for this

PR #3564, opened 2024-05-10
Add databricks_compliance_security_profile_workspace_setting resource

harshshah-db:
## Tests

- [x] `make test` run locally
- [x] relevant change in `docs/` folder
- [ ] covered with integration tests in `internal/acceptance`
- [ ] relevant acceptance tests are passing
- [x] using Go SDK

PR #3581, opened 2024-05-15
Fix bug in collectionToMaps

edwardfeng-db:
## Changes

- As mentioned in this [thread](https://github.com/databricks/terraform-provider-databricks/pull/3564#issuecomment-2104175779), there is a bug with `collectionToMaps`
- The reason is that in `collectionToMap`, in the callback function we pass into `iterFields`, we are not handling the case where a list schema's Elem` is a `*schema.Schema`.
- This PR basically takes the [block](https://github.com/databricks/terraform-provider-databricks/blob/ff8f028786eb146887cace71b6781c892c42a25e/common/reflect_resource.go#L773-L778) in the callback function in `StructToData` into `collectionToMaps` 

## Tests

- Tested by executing the failing unit test on the mentioned PR, verified it works
- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/common/reflect_resource.go
+++ b/common/reflect_resource.go
@@ -667,11 +667,16 @@ func collectionToMaps(v any, s *schema.Schema, aliases map[string]map[string]str
 			fieldPath := strings.Join(path, ".")
 			switch fieldSchema.Type {
 			case schema.TypeList, schema.TypeSet:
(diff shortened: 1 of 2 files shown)


mgyucht (commented):
Can you add a unit test for this? Otherwise LGTM

PR #3586, opened 2024-05-16
Handle a special use-case of user home directory permissions

favoretti:
## Changes
In a special case when one would want to add rights for extra users to either view or edit things in one's home directory - the user's own `CAN_MANAGE` permission is implicit. 

One can specify it explicitly in the resource, however, during deletion that `CAN_MANAGE` permission can not be removed, resulting in a failed run that can't be fixed without deleting resource from the state (which leaves other permissions hanging).

Hence here's an attempt to ignore or implicitly add it.

This works when permission is supplied with `directory_path` that matches the `user_name`, however, this also doesn't work when we specify `user_id`... 

Therefore as WIP for now..

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/hashicorp/hcl/v2 v2.23.0
 	github.com/hashicorp/terraform-json v0.24.0
 	github.com/hashicorp/terraform-plugin-framework v1.14.1
-	github.com/hashicorp/terraform-plugin-framework-validators v0.17.0
+	github.com/hashicorp/terraform-plugin-framework-validators v0.18.0
 	github.com/hashicorp/terraform-plugin-go v0.26.0
 	github.com/hashicorp/terraform-plugin-log v0.9.0
 	github.com/hashicorp/terraform-plugin-mux v0.18.0
--- a/go.sum
+++ b/go.sum
@@ -117,8 +117,8 @@ github.com/hashicorp/terraform-json v0.24.0 h1:rUiyF+x1kYawXeRth6fKFm/MdfBS6+lW4
 github.com/hashicorp/terraform-json v0.24.0/go.mod h1:Nfj5ubo9xbu9uiAoZVBsNOjvNKB66Oyrvtit74kC7ow=
 github.com/hashicorp/terraform-plugin-framework v1.14.1 h1:jaT1yvU/kEKEsxnbrn4ZHlgcxyIfjvZ41BLdlLk52fY=
 github.com/hashicorp/terraform-plugin-framework v1.14.1/go.mod h1:xNUKmvTs6ldbwTuId5euAtg37dTxuyj3LHS3uj7BHQ4=
-github.com/hashicorp/terraform-plugin-framework-validators v0.17.0 h1:0uYQcqqgW3BMyyve07WJgpKorXST3zkpzvrOnf3mpbg=
-github.com/hashicorp/terraform-plugin-framework-validators v0.17.0/go.mod h1:VwdfgE/5Zxm43flraNa0VjcvKQOGVrcO4X8peIri0T0=
+github.com/hashicorp/terraform-plugin-framework-validators v0.18.0 h1:OQnlOt98ua//rCw+QhBbSqfW3QbwtVrcdWeQN5gI3Hw=
+github.com/hashicorp/terraform-plugin-framework-validators v0.18.0/go.mod h1:lZvZvagw5hsJwuY7mAY6KUz45/U6fiDR0CzQAwWD0CA=
 github.com/hashicorp/terraform-plugin-go v0.26.0 h1:cuIzCv4qwigug3OS7iKhpGAbZTiypAfFQmw8aE65O2M=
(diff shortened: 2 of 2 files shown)


permissions/resource_permissions.go
```diff
@@ -215,6 +215,16 @@ func (a PermissionsAPI) Delete(objectID string) error {
 				accl.AccessControlList = append(accl.AccessControlList, change)
 			}
 		}
+
+		// handle special case when we add extra permission to a user home dir
+		if v, ok := d.GetOk("directory_path"); ok {
```
alexott, line 220: We need to think how we will handle this for the case when we specify object ID instead of directory path...

alexott (commented):
We also need to test how it will behave if we'll try to change permissions for user/SP from manage to lower, like, CAN_READ...

permissions/resource_permissions.go
favoretti: Yeah, that's a tricky one my brain refused to wrap itself around at 3am.. Directory ID won't match `user_name` of course, but fetching a user record based on a user id seemed like an overkill. Open for any hints here.

favoretti:
> We also need to test how it will behave if we'll try to change permissions for user/SP from manage to lower, like, CAN_READ...

This particular use-case is only intended for a single corner-case permission: user `X` to `/Users/X`, so just this particular user's home dir. That case of `CAN_MANAGE` is implicit, you can't lower it I believe.

alexott:
I'm clarifying with the team who is responsible for workspace-level permissions

permissions/resource_permissions.go
alexott: yes, it's possible to get the directory path by object iD

permissions/resource_permissions_test.go
alexott, line 1: I think that we'll need to expand tests to cover this stuff.

permissions/resource_permissions.go
alexott, line 220: We may need to disallow changes only for the current user - then we'll need to add a check like `if me == accessControl.UserName || me == accessControl.ServicePrincipalName {` like [below](https://github.com/databricks/terraform-provider-databricks/pull/3586/files#diff-e21ed6630273de6b0cb95643ddfc8cda8082f1ed5c5ce39b53210622ad90f55aR339). We also need to handle the case of home directory of the service principal, not only ordinary users.

alexott:
I got confirmation from the dev team:

> Currently, the `DirectoryPermissionsHandler` prevents users from changing the `CAN_MANAGE` permissions of a user on its own home folder.

favoretti:
@alexott So, to summarize what I need to do to get this merged:
1. Add SPN home directories to the use-case
2. Handle `object_id` next to currently implemented `directory_path`
3. Write tests

Correct?

alexott:
yes, it looks like. @mgyucht I think it makes sense to discuss in the next office hours

permissions/resource_permissions.go
```diff
@@ -215,6 +215,16 @@ func (a PermissionsAPI) Delete(objectID string) error {
...
+			if v.(string) == fmt.Sprintf("/Users/%s", acl.UserName) && acl.AllPermissions[0].PermissionLevel == "CAN_MANAGE" {
```
tanmay-db, line 221: Why do we only compare with the first index of the list? Just to confirm, is it necessar that CAN_MANAGE will always be at 0 index?

tanmay-db (commented):
The logic looks good to me overall, I had a concern on checking CAN_MANAGE with first index of the list.

permissions/resource_permissions.go
favoretti: @tanmay-db That is... a great question, considering that this PR has been hanging here for a while - I'm not sure... I'm unfortunately not that knee-deep in databricks provider code, would you mind suggesting an alternative?

tanmay-db: @alexott and me discussed this and we think going ahead with checking all the items in the list would be more safer.

tanmay-db: Just to check, after this extending this to SPN and adding tests would be left before we can merge this right?

favoretti: @tanmay-db Something like this?

favoretti: > Just to check, after this extending this to SPN and adding tests would be left before we can merge this right?

Expanding to SPN homedir you mean? Yeah, I'll need to work on that too, although I have no ideas what the format of SPN's home dir is... is it `/Users/<spn_id>`? If so, which ID is it, Azure Object ID, Azure app ID or Databricks internal ID? :)

As for tests... I'll do my best, but would really appreciate some help, I'm incredibly new to your codebase :)

alexott: it's SPN's Application ID (UUID)

tanmay-db: @favoretti, yes, to optimise a little, we can add break in the loop when we find the permission level to CAN_MANAGE.

Issue #3596, opened 2024-05-21
[ISSUE] Issue with case sensitive group names in `databricks_group` resource


Joe-Koch:
### Configuration

```hcl
data "databricks_group" "Admins_uppercase" {
  display_name = "Admins"
}

output "Admins_uppercase_group_id" {
  value = data.databricks_group.Admins_uppercase.id
}

data "databricks_group" "admins_lowercase" {
  display_name = "admins"
}

output "admins_lowercase_group_id" {
  value = data.databricks_group.admins_lowercase.id
}
```

### Expected Behavior
The Databricks group named "admins" (lowercase) and one named "Admins" (uppercase) should have distinct group ID's to accurately reflect their group ID's in Databricks.

### Actual Behavior
They both output the group ID of the lower case admins group.

The display names are correct, though. Below returns Admins and admins.

```
output "Admins_uppercase_name" {
  value = data.databricks_group.Admins_uppercase.display_name
}

output "admins_lowercase_group_name" {
  value = data.databricks_group.admins_lowercase.display_name
}
```

### Steps to Reproduce

1. Have a Databricks group named "Admins" (uppercase) and one named "admins" (lowercase).
2. `terraform plan`

### Terraform and provider versions
```
Terraform v1.7.5
on darwin_arm64
+ provider registry.terraform.io/databricks/databricks v1.44.0
```

### Is it a regression?
No

### Debug Output

Relevant portion:

```
[DEBUG] provider.terraform-provider-databricks: GET /api/2.0/preview/scim/v2/Groups?filter=displayName eq "Admins"
< HTTP/2.0 200 OK
< {
<   "Resources": [
<     {
<       "displayName": "admins",
<       "id": "00",
<     },
<     {
<       "displayName": "Admins",
<       "id": "01",
<     }
<   ],
<   "totalResults": 2
< }
```

### Important Factoids
Databricks enforces uniqueness of group names with case sensitivity, so groups like "Admins" and "admins" can exist in the same workspace. 
When you filter for Databricks groups matching a name with the SCIM API, the name match is not case-sensitive, and it might return multiple groups. The current code just grabs the first group from the results when multiple groups are returned.

### Would you like to implement a fix?
Yes! Fix submitted.

PR #3597, opened 2024-05-21
[Fix] Fix issue with case sensitive group names in `databricks_group` resource

Joe-Koch:
- Fixes Issue #3596, issue with case sensitive group names in databricks_group resource
- Updated ReadByDisplayName function to ensure exact case-sensitive matching of group display names.
- This addresses the issue where groups with similar names but different cases were not correctly distinguished.

## Changes

Databricks enforces uniqueness of group names with case sensitivity, so groups like "Admins" and "admins" can exist in the same workspace. When you filter for Databricks groups matching a name with the SCIM API, the name match is not case-sensitive, and it might return multiple groups. The current code just grabs the first group from the results when multiple groups are returned. With this code change, if the length of the returned group list is greater than 1, it loops through the returned groups and searches for a match on DisplayName. 

## Tests

Added a test in `groups_test.go` to check that when there's a lower-case "name" group and an upper-case "Name" group,  `ReadByDisplayName` will return the right one. 

- [X] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/groups.go
+++ b/scim/groups.go
@@ -59,7 +59,18 @@ func (a GroupsAPI) ReadByDisplayName(displayName, attributes string) (group Grou
 		err = fmt.Errorf("cannot find group: %s", displayName)
 		return
 	}
-	group = groupList.Resources[0]
+	if len(groupList.Resources) == 1 {
+		group = groupList.Resources[0]
+		return
+	}
+	for _, g := range groupList.Resources {
+		if g.DisplayName == displayName {
+			group = g
+			return
+		}
+	}
+
+	err = fmt.Errorf("cannot find group: %s", displayName)
 	return
 }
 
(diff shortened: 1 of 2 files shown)


Joe-Koch:
@tanmay-db  Is there anything else needed to get this approved? Thank you

tanmay-db:
Hi @Joe-Koch, thanks for the PR. I will take a look and reach back to you.

scim/groups.go
```diff
@@ -59,7 +59,18 @@ func (a GroupsAPI) ReadByDisplayName(displayName, attributes string) (group Grou
 		err = fmt.Errorf("cannot find group: %s", displayName)
 		return
 	}
-	group = groupList.Resources[0]
+	if len(groupList.Resources) == 1 {
+		group = groupList.Resources[0]
```
alexott, line 63: Should we enforce case-sensitive names here as well?

tanmay-db (commented):

PR #3611, opened 2024-05-23
Do not suppress diff if it is explicitly changed to zero

andrewnester:
## Changes
Do not suppress diff if it is explicitly changed to zero

## Tests
Added regression tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/reflect_resource.go
+++ b/common/reflect_resource.go
@@ -366,7 +366,9 @@ func chooseFieldName(typeField reflect.StructField) string {
 func diffSuppressor(fieldName string, v *schema.Schema) func(k, old, new string, d *schema.ResourceData) bool {
 	zero := fmt.Sprintf("%v", v.Type.Zero())
 	return func(k, old, new string, d *schema.ResourceData) bool {
-		if new == zero && old != zero {
+		// HasChange allows to check if the field is explicitly changed in the configuration.
+		// If the field is explicitly set in the configuration, we should not suppress the diff.
+		if new == zero && old != zero && !d.HasChange(k) {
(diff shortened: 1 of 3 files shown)


pietern (commented):
LGTM, but will have to defer to @alexott and @tanmay-db for a stamp.

Diff suppression is a beast and there could be side effects.

common/reflect_resource.go
```diff
@@ -366,7 +366,9 @@ func chooseFieldName(typeField reflect.StructField) string {
 func diffSuppressor(fieldName string, v *schema.Schema) func(k, old, new string, d *schema.ResourceData) bool {
 	zero := fmt.Sprintf("%v", v.Type.Zero())
 	return func(k, old, new string, d *schema.ResourceData) bool {
-		if new == zero && old != zero {
+		// HasChange allows to check if the field is explicitly changed in the configuration.
+		// If the field is explicitly set in the configuration, we should not suppress the diff.
+		if new == zero && old != zero && !d.HasChange(k) {
```
tanmay-db, line 371: I think we should also have this in `SetSuppressDiffWithDefault` as well: https://github.com/databricks/terraform-provider-databricks/blob/main/common/customizable_schema.go#L159. 

````
	// Suppress diff if the new value (ie value from HCL config) is not set and
	// the old value (ie value from state / platform) is equal to the default value.
	s.Schema.DiffSuppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
		if new == zero && old == sv {
			return true
		}
		return false
	}
````

Otherwise LGTM, stamping to unblock merge.

tanmay-db (approved):

alexott (approved):

PR #3615, opened 2024-05-24
Enhance DataAwsAssumeRolePolicy to allow the role to assume itself

unterstein:
## Changes
AWS roles used to be implicitly able to assume itself. As of somewhen last year this is no longer the case, therefore whenever a role wants to assume itself, it has to be configured explicitly in the assume role policy. We are using the `DataAwsAssumeRolePolicy` in our terraform stacks for a lot of roles but recently learned that this might cause issues. Therefore we would like to add the capabilities to this terraform resource, that the self assumption is also possible.

When creating roles with terraform, it is a little tricky to do this self assumption. Since ARNs are predictable, you could think, that you just can predict the ARN of your -to-be-created-role and add it to the list of `Allow-sts:AssumeRole` directly. But AWS will perform a check and will fail if the ARN is not already present. Since we are just about to create the role with that ARN, this is not possible. However, going in and adding the concrete check for the role as `Condition-ArnLike`, we will postpone the check if the ARN is actually existing to the runtime. At this time the ARN is present and the check will work as expected.

We do this kind of IAM approach for role self assumptions at a lot of other places ever since AWS announced their policy change last year and we didn't had problems with it so far.

## Tests
I ran `make test` and verified that this addition is not breaking existing behaviour/existing tests. I added another test and verified the output policy of my test:
```
        	            	{
        	            	  "Version": "2012-10-17",
        	            	  "Statement": [
        	            	    {
        	            	      "Effect": "Allow",
        	            	      "Action": "sts:AssumeRole",
        	            	      "Principal": {
        	            	        "AWS": "arn:aws:iam::414351767826:root"
        	            	      },
        	            	      "Condition": {
        	            	        "StringEquals": {
        	            	          "sts:ExternalId": "abc"
        	            	        }
        	            	      }
        	            	    },
        	            	    {
        	            	      "Sid": "ExplicitSelfRoleAssumption",
        	            	      "Effect": "Allow",
        	            	      "Action": "sts:AssumeRole",
        	            	      "Principal": {
        	            	        "AWS": "arn:aws:iam::aws-account:role/root"
        	            	      },
        	            	      "Condition": {
        	            	        "ArnLike": {
        	            	          "aws:PrincipalArn": "arn:aws:iam::aws-account:role/role-name"
        	            	        }
        	            	      }
        	            	    }
        	            	  ]
        	            	}
```

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/aws/data_aws_assume_role_policy.go
+++ b/aws/data_aws_assume_role_policy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -31,6 +32,7 @@ func DataAwsAssumeRolePolicy() common.Resource {
 	return common.Resource{
 		Read: func(ctx context.Context, d *schema.ResourceData, m *common.DatabricksClient) error {
 			externalID := d.Get("external_id").(string)
+			predictedRoleArn := d.Get("predicted_role_arn").(string)
 			policy := awsIamPolicy{
 				Version: "2012-10-17",
 				Statements: []*awsIamPolicyStatement{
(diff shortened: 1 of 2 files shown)


nkvuong:
@unterstein thank you for getting around to this - do you mind updating the doc for the data source as well?

unterstein:
@nkvuong thank you for your feedback. Is the place I added the documentation the right one and is it ok that way? Please feel free to adjust the docs in this PR to your liking.

thank you 🙏

nkvuong:
@unterstein I just checked out documentation, and the cross-account policy does not require support for self assumption. It is only the Unity Catalog IAM role that requires support for self assumption, but we currently do not have a data source for that

unterstein:
@nkvuong we were informed by our partner from databricks, that we need to make some of our roles self assuming. For all roles that were called out, we use this particular `DataAwsAssumeRolePolicy` to generate it's assume role policy. In order to fulfil the request to make those role able to assume themselves, we need to adjust this component.

nkvuong:
@unterstein confusingly, there are 2 different trust policies necessary, one for workspace creation & one for Unity Catalog. I've raised a separate PR #3622 for the second type of trust policies

unterstein:
@nkvuong we don't use databricks provider components for our universe role and we didn't got those roles flagged from databricks. Where as we got the workspaces called out that we need to fix them.

alexott:
@nkvuong should we close this one?

alexott:
Closing as it's not required.

PR #3631, opened 2024-05-30
Added `LocalSsdCount` in `GcpAttributes` to ForceSendFields for `databricks_cluster` resource

tanmay-db:
## Changes

We always force send the local ssd count if gcp attribute is specified. 

## Tests

Updated unit test
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -383,6 +383,9 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *commo
 	if createClusterRequest.Autoscale == nil {
 		createClusterRequest.ForceSendFields = []string{"NumWorkers"}
 	}
+	if createClusterRequest.GcpAttributes != nil {
+		createClusterRequest.GcpAttributes.ForceSendFields = []string{"LocalSsdCount"}
+	}
 	clusterWaiter, err := clusters.Create(ctx, createClusterRequest)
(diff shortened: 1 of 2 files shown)


edwardfeng-db (approved):
Great, thanks!

andrewnester:
Does it mean it will send `LocalSsdCount` always even if not explicitly specified in terraform file? Or we do some GetOkExists checks somewhere?

clusters/resource_cluster.go
```diff
@@ -383,6 +383,9 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *commo
 	if createClusterRequest.Autoscale == nil {
 		createClusterRequest.ForceSendFields = []string{"NumWorkers"}
 	}
+	if createClusterRequest.GcpAttributes != nil {
+		createClusterRequest.GcpAttributes.ForceSendFields = []string{"LocalSsdCount"}
```
andrewnester, line 387: We should call GetOkExists for local_ssd_count to check that it was explicitly set and only then send it. The reason is that if not specified in request, the default value for local_ssd_count on backend is not 0 (can be 1, 2, 4 depending on type of instance used).

andrewnester (changes requested):

andrewnester (approved):

PR #3633, opened 2024-05-31
Delta share rework set

ScottReyn-db:
## Changes
1. Delta share `Objects` are moved to `Set` and use default Set hashing function
2. Delta share `Partitions` are moved to a `Set` with default hashing function

This allows for adding and removing `Objects` and their associated `Partitions` in hcl *and* for them to show up in terraform diffs
 
## Tests
Added additional tests to share test suite for this updated functionality
Updated the share data test to test each field. This is because `schema.Set` does not implement `Equal` making it impossible to do a `Equal` assertation.

- [X] `make test` run locally
- [X] using Go SDK


sharing/resource_share.go
```diff
@@ -230,33 +197,155 @@ func ResourceShare() common.Resource {
...
 			err = NewSharesAPI(ctx, c).update(d.Id(), ShareUpdates{
-				Updates: changes,
+				Updates: apiChanges,
 			})
 			if err != nil {
-				if d.HasChange("owner") {
-					// Rollback
-					old, new := d.GetChange("owner")
+				// Rollback
+				if d.HasChanges("owner", "comment") {
```
ScottReyn-db, line 338:
```suggestion
				if d.HasChange("owner") {
```
Hold over from an older PR where I introduced the `Comment` field to the delta share object.

sharing/resource_share.go
```diff
@@ -230,33 +197,172 @@ func ResourceShare() common.Resource {
...
+				}
+			}
+
+			for _, raw := range objectsToAdd {
+				rawMap := raw.(map[string]interface{})
+
+				name, ok := rawMap["name"]
+				if !ok || name == "" {
+					continue
+				}
```
ScottReyn-db, line 309: This commit is off. 
```suggestion
				rawName, ok := rawMap["name"]
				name := rawName.(string)
				if !ok || name == "" {
					continue
				}
```

PR #3646, opened 2024-06-05
Run goreleaser action in snapshot mode from merge queue

pietern:
## Changes

Running this action before doing an actual release should prevent situations like #3645 where it doesn't work on release.

## Tests

See workflows.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,6 @@ on:
   push:
     tags:
       - '*'
-  pull_request:
-    types: [opened, synchronize]
   merge_group:
     types: [checks_requested]
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,7 +39,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v6
         with:
           version: ~> v2
-          args: release --clean ${{ !startsWith(github.ref, 'refs/tags/v') && '--snapshot' }}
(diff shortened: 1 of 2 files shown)


pietern:
This change still includes the pull request trigger to test it out, but must be removed before merging.

We can't run the goreleaser action on PRs because PRs from forks don't have access to secrets.

pietern:
It works: https://github.com/databricks/terraform-provider-databricks/actions/runs/9381922308/job/25832223352

.github/workflows/release.yml
```diff
@@ -19,21 +21,25 @@ jobs:
...
 
-      - name: Pull external libraries
-        run: make vendor
-
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
         with:
           version: ~> v2
-          args: release --clean
+          args: release --clean ${{ !startsWith(github.ref, 'refs/tags/v') && '--snapshot' || '' }}
```
pietern, line 42: Also see https://docs.github.com/en/actions/learn-github-actions/expressions#example

mgyucht: What is `github.ref`'s value when something is in the merge queue?

mgyucht (approved):
Does the merge fail if this fails? It's not a required check. Still good that we at least run it once.

.github/workflows/release.yml
pietern: Something like `gh-readonly-queue/main/...`, definitely not `refs/tags/...`.

PR #3662, opened 2024-06-10
Fix detection of local file changes

hectorcast-db:
## Changes
Fix detection of local file changes. Reuse logic by workspace files.
 
## Tests

- [X] Manual test
- [X] `make test` run locally
- N/A relevant change in `docs/` folder
- [X] using Go SDK

### DIFF
--- a/storage/resource_file.go
+++ b/storage/resource_file.go
@@ -3,9 +3,7 @@ package storage
 import (
 	"bytes"
 	"context"
-	"encoding/base64"
 	"io"
-	"os"
 
 	"github.com/databricks/databricks-sdk-go/service/files"
 	"github.com/databricks/terraform-provider-databricks/common"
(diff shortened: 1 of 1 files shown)


storage/resource_file.go
```diff
@@ -14,27 +12,11 @@ import (
...
-		reader = io.NopCloser(bytes.NewReader(decodedString))
-		if err != nil {
-			return nil, err
-		}
+	content, err := workspace.ReadContent(data)
+	if err != nil {
+		return nil, err
 	}
-	return reader, err
+	return io.NopCloser(bytes.NewReader(content)), nil
```
alexott, line 19: With files up to 5Gb supported, reading them completely into memory may not be a good idea.

alexott (commented):
We also need a test for the Update operation...

storage/resource_file.go
```diff
@@ -13,6 +16,32 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func calculateMD5(filePath string) (string, error) {
```
edwardfeng-db, line 19: Good to name it `calculateMD5Chunks`?

edwardfeng-db (approved):
LGTM

alexott (commented):
We can calculate MD5 together with reading it, instead of calculating it separately... Something like this: https://stackoverflow.com/questions/57850902/is-it-possible-to-calculate-md5-of-the-file-and-write-file-to-the-disk-in-the-sa

PR #3669, opened 2024-06-12
Update jobs library fields from set to list

edwardfeng-db:
## Changes

- Ordering is important, updating the library fields in jobs schema from set to list
- Previously it wasn't updated because during the go-sdk schema migration we wanted to minimize breaking schema changes

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/jobs/resource_job.go
+++ b/jobs/resource_job.go
@@ -219,7 +219,7 @@ type JobTaskSettings struct {
 	ExistingClusterID string            `json:"existing_cluster_id,omitempty" tf:"group:cluster_type"`
 	NewCluster        *clusters.Cluster `json:"new_cluster,omitempty" tf:"group:cluster_type"`
 	JobClusterKey     string            `json:"job_cluster_key,omitempty" tf:"group:cluster_type"`
-	Libraries         []compute.Library `json:"libraries,omitempty" tf:"slice_set,alias:library"`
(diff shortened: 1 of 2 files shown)


pietern (commented):
LGTM

Can this schema evolution happen automatically? I believe the underlying representation of a list or a set is the same, so suspect it should be fine, but could you double-check with an existing deployment with 2+ libraries, and then redeploying with this change?

hectorcast-db:
LGMT. What is the impact in customers? Will this cause an update on the TF resources?

edwardfeng-db:
Tested it out locally, I see that there'll be a diff for the ordering @hectorcast-db @pietern, but I'm not sure if this is considered a breaking thing for customers

>       ~ library {
>           ~ jar = "dbfs://ff/gg/hh.jar" -> "dbfs://aa/bb/cc.jar"
>         }
>       ~ library {
>           ~ jar = "dbfs://aa/bb/cc.jar" -> "dbfs://ff/gg/hh.jar"
>         }

hectorcast-db (approved):
We should add a note in the release notifying of possible diffs and updates.

alexott:
We need to be careful with slice set and complex blocks - we already had an issue with DLT pipelines: https://github.com/databricks/terraform-provider-databricks/issues/1401 - we need to test for `pypi` blocks, and similar things and make sure that it won't cause the configuration drift

Issue #3558, opened 2024-05-08
[ISSUE] Issue with `databricks_cluster` resource using exporter does not include cluster libraries 


alex-lopes-databricks:
Regular workspace with clusters and libraries installed on it 
When you use the exporter to export the cluster definitions, the libraries definition are not exported with the cluster
<br>
Images below
<br><br>
![image (7)](https://github.com/databricks/terraform-provider-databricks/assets/91564092/ed5dbd9a-d8c8-40bd-b4d4-503b1c391481)
![image (8)](https://github.com/databricks/terraform-provider-databricks/assets/91564092/d89ed966-6188-479b-b75c-b4ecb2e0775a)

PR #3674, opened 2024-06-14
Fixed: Issue with `databricks_cluster` resource using `exporter` does not include cluster libraries

tanmay-db:
## Changes

compute.InstallLibraries doesn't have Aliases and since it isn't local we can't add alias using Aliases(), to solve this we abstract out the compute.Library with alias and use it to while write to resource data

Fixes: https://github.com/databricks/terraform-provider-databricks/issues/3558

## Tests

Created a cluster with library and used exporter on main branch which shows no library block reproducing the issue. After changes, libraries are shown. 

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -645,7 +645,7 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
 		return err
 	}
 
-	var clusterLibraries ClusterSpec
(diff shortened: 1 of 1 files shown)


alexott:
Thank you very much - we may need to add a comment about non-local aliases to the `common`

hectorcast-db (approved):

alexott (approved):

PR #3682, opened 2024-06-19
Rename default_namespace_setting documentation to correct name

al-lac:
## Changes
Adapt documentation to use correct name for the resource. Was changed to a singular name in #2925. 

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


hectorcast-db (approved):
I think we should also change the title of doc page

al-lac:
> Thanks for the contribution. We should also change the title of doc page `databricks_namespace_setting` -> `databricks_default_namespace_setting`

Done 👍

Issue #3691, opened 2024-06-23
[ISSUE] Issue with `databricks_quality_monitor` resource


touchida:
When a `quality_monitor` resource with the `schedule` field is created, the `terraform plan` command always shows an unexpected change for the `pause_status` read-only field.

### Configuration

```hcl
data "databricks_sql_warehouse" "all" {
  name = "sandbox"
}

resource "databricks_sql_table" "myTestTable" {
  catalog_name       = "sandbox"
  schema_name        = "things"
  name               = "bar"
  table_type         = "MANAGED"
  data_source_format = "DELTA"
  warehouse_id       = data.databricks_sql_warehouse.all.id

  column {
    name = "timestamp"
    type = "int"
  }
}

resource "databricks_quality_monitor" "testTimeseriesMonitor" {
  table_name         = "sandbox.things.${databricks_sql_table.myTestTable.name}"
  assets_dir         = "/Shared/provider-test/databricks_quality_monitoring/${databricks_sql_table.myTestTable.name}"
  output_schema_name = "sandbox.things"
  time_series {
    granularities = ["1 hour"]
    timestamp_col = "timestamp"
  }
  schedule {
    quartz_cron_expression = "0 0 * ? * *"
    timezone_id            = "JST"
  }
  warehouse_id = data.databricks_sql_warehouse.all.id
}
```

### Expected Behavior

 When we don't modify the configuration after the `terraform apply` command, the `terraform plan` should not find any change.

### Actual Behavior

The `terraform plan` command shows a change for the `pause_status` field unexpectedly.
Note that it is a read-only field according to the Databricks REST API documentation, and so we cannot specify it in the configuration.
```
$ terraform plan
...
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the
following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # databricks_quality_monitor.testTimeseriesMonitor will be updated in-place
  ~ resource "databricks_quality_monitor" "testTimeseriesMonitor" {
        id                         = "sandbox.things.bar"
        # (9 unchanged attributes hidden)

      ~ schedule {
          - pause_status           = "UNPAUSED" -> null
            # (2 unchanged attributes hidden)
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
...
```

### Steps to Reproduce

1. `terraform apply`
2. `terraform plan`

### Terraform and provider versions

```
[...]

PR #3692, opened 2024-06-23
Make the schedule.pause_status field read-only

touchida:
## Changes

This PR will resolve https://github.com/databricks/terraform-provider-databricks/issues/3691.
- Make the `schedule.pause_status` field read-only.
- Remove the description of the field from the documentation.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

I locally confirmed that the `terraform plan` command no longer showed any change for the `schedule.pause_status` field.

### DIFF
--- a/catalog/resource_quality_monitor.go
+++ b/catalog/resource_quality_monitor.go
@@ -53,6 +53,7 @@ func ResourceQualityMonitor() common.Resource {
 			common.CustomizeSchemaPath(m, "profile_metrics_table_name").SetReadOnly()
 			common.CustomizeSchemaPath(m, "status").SetReadOnly()
 			common.CustomizeSchemaPath(m, "dashboard_id").SetReadOnly()
+			common.CustomizeSchemaPath(m, "schedule", "pause_status").SetReadOnly()
 			return m
 		},
 	)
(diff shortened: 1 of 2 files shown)


docs/resources/quality_monitor.md
```diff
@@ -112,7 +112,6 @@ table.
 * `schedule` - The schedule for automatically updating and refreshing metric tables.  This block consists of following fields:
     * `quartz_cron_expression` - string expression that determines when to run the monitor. See [Quartz documentation](https://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html) for examples.
     * `timezone_id` - string with timezone id (e.g., `PST`) in which to evaluate the Quartz expression.
-    * `pause_status` - optional string field that indicates whether a schedule is paused (`PAUSED`) or not (`UNPAUSED`).
```
vperiyasamy, line 115: Do we distinguish read/write fields here? Or is this only for passable arguments?

catalog/resource_quality_monitor.go
```diff
@@ -53,6 +53,7 @@ func ResourceQualityMonitor() common.Resource {
 			common.CustomizeSchemaPath(m, "profile_metrics_table_name").SetReadOnly()
 			common.CustomizeSchemaPath(m, "status").SetReadOnly()
 			common.CustomizeSchemaPath(m, "dashboard_id").SetReadOnly()
+			common.CustomizeSchemaPath(m, "schedule", "pause_status").SetReadOnly()
```
vperiyasamy, line 56: I don't see a related designation for the overall `schedule` resource - is it set as an optional field overall and this only marks the subfield as read only?

docs/resources/quality_monitor.md
alexott: it's for passable arguments...

catalog/resource_quality_monitor.go
alexott: yes, `schedule` is optional automatically if it's marked as `omitempty`. This one is marked as read-only because it comes from backend, and isn't set by user

alexott (approved):

vperiyasamy (approved):

touchida:
@alexott @vperiyasamy 
Thanks for your review!

Issue #3492, opened 2024-04-19
[FEATURE] Identity column specification on `databricks_sql_table`


Dedvall:
### Use-cases
Be able to define Identity column on a table the `databricks_sql_table` needs some means to define identity columns. After the release of provider version 1.40 it seems impossible to create a table with an Identity column (Issue #3490) 

The Identity column options documented [here](https://www.databricks.com/blog/2022/08/08/identity-columns-to-generate-surrogate-keys-are-now-available-in-a-lakehouse-near-you.html) needs to be possible in Terraform too.

### Attempted Solutions
Until provider version 1.40 this was possible 
```
resource "databricks_sql_table" "table_foo" {
  name               = "foo"
  catalog_name       = databricks_catalog.main.name
  schema_name        = databricks_schema.main.name
  table_type         = "MANAGED"
  data_source_format = "DELTA"

  column {
    name     = "bar"
    type     = "bigint GENERATED BY DEFAULT AS IDENTITY"
    nullable = "false"
  }
```
After that provider version above fails with errors on plan. As #3490 

### Proposal
Something similar to this could be possible
```
resource "databricks_sql_table" "table_foo" {
  name               = "foo"
  catalog_name       = databricks_catalog.main.name
  schema_name        = databricks_schema.main.name
  table_type         = "MANAGED"
  data_source_format = "DELTA"

  column {
    name     = "bar"
    type     = "bigint "
    nullable = "false"
    identity = "default" // possible values FALSE (default) & ALWAYS
  }
```

### References
#3490
[SQL Reference](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-table-using.html)
[Identity release blog](https://www.databricks.com/blog/2022/08/08/identity-columns-to-generate-surrogate-keys-are-now-available-in-a-lakehouse-near-you.html)

aru-trackunit:
I would also appreciate that feature

hshahconsulting:
Resolved in #4035

PR #3709, opened 2024-06-26
[Feature] Support for identity column in sql table

hshahconsulting:
## Changes

As Databricks supports identity columns, I have added support for identity columns in databricks_sql_table.
Closes [[FEATURE] Identity column specification on databricks_sql_table](https://github.com/databricks/terraform-provider-databricks/issues/3492)
## Tests
 added unit test that asserts a correct statement has been generated when an identity column type is specified

- [x] `make test` run locally
- [x] implement integration test


catalog/resource_sql_table_test.go
```diff
@@ -33,6 +33,36 @@ func TestResourceSqlTableCreateStatement_External(t *testing.T) {
...
+			},
+		},
+	}
+	stmt := ti.buildTableCreateStatement()
+	assert.Contains(t, stmt, "CREATE EXTERNAL TABLE `main`.`foo`.`bar`")
+	assert.Contains(t, stmt, "USING DELTA")
+	assert.Contains(t, stmt, "(`id` BIGINT GENERATED BY DEFAULT AS IDENTITY NOT NULL, `name`  NOT NULL COMMENT 'a comment')")
+	assert.Contains(t, stmt, "LOCATION 's3://ext-main/foo/bar1' WITH (CREDENTIAL `somecred`)")
+	assert.Contains(t, stmt, "COMMENT 'terraform managed'")
+}
```
hectorcast-db, line 64: Can you also add an integration test to this file?
https://github.com/databricks/terraform-provider-databricks/blob/main/internal/acceptance/sql_table_test.go#L13

hectorcast-db (changes requested):
Hi @hshahconsulting 
Thanks for this contribution. I looks quite good. I left a couple of comment. 

@edwardfeng-db Can you also take a look at this PR?

catalog/resource_sql_table_test.go
hshahconsulting: Done

edwardfeng-db (approved):

hshahconsulting:
@hectorcast-db I have made the changes according to the feedback. Anything else that needs to be done? Also, @tanmay-db said that other changes may be needed. ny follow up on that?

tanmay-db:
Hi @hshahconsulting, thanks for creating the PR. Can you please check if this works locally for you? You can create a local binary to test: https://github.com/databricks/terraform-provider-databricks/blob/main/CONTRIBUTING.md#developing-provider

So the steps would be to:
1. Create local binary (run `make` in root of repository over your PR changes)
2. Override the provider using ~/.terraformrc as mentioned in the link
3. Do terraform apply as you were doing before over your HCL. 

Note: Please use dev environment / test data for this (just in case there are some issues).

catalog/resource_sql_table.go
```diff
@@ -24,6 +24,7 @@ var MaxSqlExecWaitTimeout = 50
 type SqlColumnInfo struct {
 	Name     string `json:"name"`
 	Type     string `json:"type_text,omitempty" tf:"alias:type,computed"`
+	Identity string `json:"identity,omitempty" tf:"default:false"`
```
hectorcast-db, line 27: We can use the IdentifyColumn here 
```
Identity IdentityColumn `json....`
```
and then there is no need to convert types below.
In that case, you will need to create a 3rd const below for the default case.

catalog/resource_sql_table_test.go
hectorcast-db: Sorry for the missunderstanding: I was talking to add a test in the linked file so we can test this against the a real workspace.

hectorcast-db (commented):
I see some issue in the list of commits. It seems that this PR is changing preexisting commits. Can you take a look at it?

catalog/resource_sql_table_test.go
hshahconsulting: My bad, sorry. Will add.

hshahconsulting:
> I see some issue in the list of commits. It seems that this PR is changing preexisting commits. Can you take a look at it?

I made a mistake with a previous rebase. The branch has now been cleaned up.

catalog/resource_sql_table.go
hshahconsulting: Done

catalog/resource_sql_table_test.go
hshahconsulting: I have added an integration test. However, I do not know how to run it.

hshahconsulting:
> Hi @hshahconsulting, thanks for creating the PR. Can you please check if this works locally for you? You can create a local binary to test: https://github.com/databricks/terraform-provider-databricks/blob/main/CONTRIBUTING.md#developing-provider
> 
> So the steps would be to:
> 
> 1. Create local binary (run `make` in root of repository over your PR changes)
> 2. Override the provider using ~/.terraformrc as mentioned in the link
> 3. Do terraform apply as you were doing before over your HCL.
> 
> Note: Please use dev environment / test data for this (just in case there are some issues).

I am trying to run this locally but I cannot stop it from downloading the terraform provider from the internet. We have a version.tf with this code

`terraform {
  required_version = ">=1.7.1"
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~> 3.89"
    }
    databricks = {
      version = "~>1.50"
    }
  }
  backend "azurerm" {
    resource_group_name  = "unity-tf-backend-rg"
    storage_account_name = "infradevseunitytfbackend"
    container_name       = "tf-backend"
    key                  = "terraform.tfstate"
  }
}
`

catalog/resource_sql_table_test.go
hectorcast-db: I did run the tests and they succeed. Thanks!

hshahconsulting: Wohooo! Would you still like me to test locally or is this good enough? I have trouble of setting up what Tanmay suggested.

catalog/resource_sql_table.go
```diff
@@ -188,7 +194,25 @@ func (ti *SqlTableInfo) getOrCreateCluster(clusterName string, clustersAPI clust
...
+	case IdentityColumnAlways:
+		identityType = "BIGINT GENERATED ALWAYS AS IDENTITY"
+	case IdentityColumnDefault:
+		identityType = "BIGINT GENERATED BY DEFAULT AS IDENTITY"
+	}
+	return identityType
+}
+
 func (ti *SqlTableInfo) serializeColumnInfo(col SqlColumnInfo) string {
+	var colType string
```
hectorcast-db, line 209: While testing this, I saw some cases where Identify was empty. This happens when SqlColumnInfo is build by this package (and not directly by TF). 
In such cases, the `colType` ends empty and will cause issues.
We can replace the `serializeIdentityColumnInfo` by a `getColumnType`

```
func (ci *SqlColumnInfo) getColumnType() string {
	var colType string
	switch ci.Identity {
	case IdentityColumnAlways:
		colType = "BIGINT GENERATED ALWAYS AS IDENTITY"
	case IdentityColumnDefault:
		colType = "BIGINT GENERATED BY DEFAULT AS IDENTITY"
        default:
                colType = col.Type
	}
	return colType

```

and then remove the if/else from here:
```
var colType = getColumnType()
```

hshahconsulting: Done

catalog/resource_sql_table_test.go
hectorcast-db: No needed to.

hectorcast-db (commented):
Content looks good. 
I just saw that there are issues with the git history again. Can you fix that? Then we can run the automatic workflows and merge.

catalog/resource_sql_table_test.go
hectorcast-db: No need to since tests are passing. Thank!

hshahconsulting:
> Content looks good. I just saw that there are issues with the git history again. Can you fix that? Then we can run the automatic workflows and merge.

I am not sure what the issue is this time. I created this branch from main and then pasted in the code. Which commit is bad?

hshahconsulting:
> Content looks good. I just saw that there are issues with the git history again. Can you fix that? Then we can run the automatic workflows and merge.

Please check again @hectorcast-db

hectorcast-db:
Hi @hshahconsulting 

I did run all integration tests and this changes causes some issues related to the usage of the `false` default:

* This test fails:
https://github.com/databricks/terraform-provider-databricks/blob/main/internal/acceptance/sql_table_test.go#L397
The given reason is that the `diff` after apply is not empty. 
```
 error=
  | After applying this test step, the plan was not empty.
  | stdout:
  | 
  | 
  | Terraform used the selected providers to generate the following execution
  | plan. Resource actions are indicated with the following symbols:
  |   ~ update in-place
  |  <= read (data resources)
  | 
  | Terraform will perform the following actions:
  | 
  |   # data.databricks_table.this will be read during apply
  |   # (depends on a resource or a module with changes pending)
  |  <= data "databricks_table" "this" {
  |       + id   = (known after apply)
  |       + name = "sandbox########.bar"
  |     }
  | 
  |   # databricks_sql_table.mytable will be updated in-place
  |   ~ resource "databricks_sql_table" "mytable" {
  |         id                 = "sandbox########.bar"
  |         name               = "bar"
  |         # (7 unchanged attributes hidden)
  | 
  |       ~ column {
  |           + identity = "false"
  |             name     = "id"
  |             # (3 unchanged attributes hidden)
  |         }
  |     }
  | 
  | Plan: 0 to add, 1 to change, 0 to destroy.
```
* This change causes a diff on all existing resources, which forces customers to update all their tables. Not a great experience, and we should avoid it if possible.

Making the default an empty string should solve both issues.

hshahconsulting:
> Hi @hshahconsulting
> 
> I did run all integration tests and this changes causes some issues related to the usage of the `false` default:
> 
> * This test fails:
>   https://github.com/databricks/terraform-provider-databricks/blob/main/internal/acceptance/sql_table_test.go#L397
>   The given reason is that the `diff` after apply is not empty.
> 
> ```
>  error=
>   | After applying this test step, the plan was not empty.
>   | stdout:
>   | 
>   | 
>   | Terraform used the selected providers to generate the following execution
>   | plan. Resource actions are indicated with the following symbols:
>   |   ~ update in-place
>   |  <= read (data resources)
>   | 
>   | Terraform will perform the following actions:
>   | 
>   |   # data.databricks_table.this will be read during apply
>   |   # (depends on a resource or a module with changes pending)
>   |  <= data "databricks_table" "this" {
>   |       + id   = (known after apply)
>   |       + name = "sandbox########.bar"
>   |     }
>   | 
>   |   # databricks_sql_table.mytable will be updated in-place
>   |   ~ resource "databricks_sql_table" "mytable" {
>   |         id                 = "sandbox########.bar"
>   |         name               = "bar"
>   |         # (7 unchanged attributes hidden)
>   | 
>   |       ~ column {
>   |           + identity = "false"
>   |             name     = "id"
>   |             # (3 unchanged attributes hidden)
>   |         }
>   |     }
>   | 
>   | Plan: 0 to add, 1 to change, 0 to destroy.
> ```
> 
> * This change causes a diff on all existing resources, which forces customers to update all their tables. Not a great experience, and we should avoid it if possible.
> 
> Making the default an empty string should solve both issues.

Is this correct?

	Identity IdentityColumn `json:"identity,omitempty"`
	
I have never coded in Go Lang before so please excuse my beginner level questions hehe :D

hshahconsulting:
Please check. I have set the default to empty string.

hshahconsulting:
@hectorcast-db Sorry for bumping but anything else needs fixing?

hectorcast-db:
Tests just finished, and there is still one issue. However, I don't know the API well enough to recommend a solution. I have contacted the service team and I am waiting for a response. I will describe the issue in case you know how to fix it:

The new test you added is failing. Terraform always reads the server data after an apply, and then validates that there is no diff between the server status and the local status. 

However, the server does not return the Identity, so there is always a diff.

We need to be able to reconstruct the "Identity" value inside the READ method based on the server response.

hshahconsulting:
@hectorcast-db Hello Hector. Do you have any news regarding the Service Team planning to take a look at this?

hshahconsulting:
@hectorcast-db thanks. Could you paste the logs from the latest run of the integration tests where this fails, after I set the default value for the identity to an empty string? Perhaps I could figure something out by looking at it.

hshahconsulting:
Duplicate of #4035

PR #3713, opened 2024-06-27
Add codegen template for tfsdk struct

parthban-db:
## Changes

Added a codegen template for for generating structs similar to go-sdk. But has the following changes.

- The type of each primitive type is tf types.
- ForceSendFields are not present.
- The tfsdk tags has the aliased names

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


service/billing/model.go
```diff
@@ -0,0 +1,532 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// Budget configuration to be created.
+type Budget struct {
+	Alerts []BudgetAlert `tfsdk:"alerts"`
```
edwardfeng-db, line 14: I'm wondering if list can be done with `types.List` from the tfsdk?

edwardfeng-db: Looks like it has the same problem with map?

parthban-db: Yep, It has same problem like map.

.codegen/model.go.tmpl
```diff
@@ -0,0 +1,77 @@
...
+	{{- else if .IsExternal }}{{.Package.Name}}.{{.PascalName}}
+	{{- else if .IsAny}}any
+	{{- else if .IsEmpty}}{{.PascalName}}
+	{{- else if .IsString}}types.String
+	{{- else if .IsBool}}types.Bool
+	{{- else if .IsInt64}}types.Int64
+	{{- else if .IsFloat64}}types.Float64
+	{{- else if .IsInt}}types.Int64
+	{{- else if .IsByteStream}}io.ReadCloser
+	{{- else if .ArrayValue }}[]{{template "type" .ArrayValue}}
```
edwardfeng-db, line 71: Can we add a comment here just to document why we decided to use the go native slice type?

.codegen/model.go.tmpl
```diff
@@ -0,0 +1,77 @@
...
+	{{- else if .MapValue }}map[string]{{template "type" .MapValue}}
```
edwardfeng-db, line 72: Same here, can we add a comment about the go native decision. Also, let's document why the key has to be a string instead of `types.String`

edwardfeng-db (approved):
LGTM, let's get @hectorcast-db take another look

hectorcast-db (approved):
LGMT, just a question

.codegen.json
```diff
@@ -1,14 +1,14 @@
...
+    "packages": {
+        ".codegen/model.go.tmpl": "service/{{.Name}}_tf/model.go"
+    },
     "version": {
         "common/version.go": "version = \"$VERSION\""
     },
     "toolchain": {
         "required": [
             "go"
-        ],
```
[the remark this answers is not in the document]
parthban-db: added again

PR #3729, opened 2024-07-02
[Feature] Lakeview dashboard resource created

Divyansh-db:
## Changes
Created Lakeview dashboard resource and make unit and integration tests for it. Also, started writing its documentation

## Tests
unit tests in dashboards/resource_lakeview_dashboard_test.go
integration/acceptance tests in internal/acceptance/lakeview_dashboard_test.go


dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+			s["dashboard_id"].Computed = true
+			s["display_name"].Optional = false
+			s["display_name"].Required = true
+			s["etag"].Computed = true
+			s["lifecycle_state"].Computed = true
+			s["parent_path"].Optional = false
+			s["parent_path"].Required = true
+			s["parent_path"].ForceNew = true
+			s["path"].Computed = true
+			s["serialized_dashboard"].Optional = true
```
hectorcast-db, line 66: This is already optional in the Go SDK. You only need to customize things which are not expressed in the Go SDK.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+			s["serialized_dashboard"].ConflictsWith = []string{"file_path"}
+			s["serialized_dashboard"].DiffSuppressFunc = func(k, old, new string, d *schema.ResourceData) bool {
+				_, ok := d.GetOk("file_path")
```
hectorcast-db, line 69: What is the logic here? Shouldn't you read the content from the file and check the md5?
See how files does it
https://github.com/databricks/terraform-provider-databricks/blob/main/workspace/file_resource.go#L93

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+			dashboard := dashboards.GetDashboardRequest{
+				DashboardId: d.Id(),
+			}
+			resp, err := w.Lakeview.Get(ctx, dashboard)
+			if err != nil {
+				return err
+			}
+			// fmt.Println(d.Get("md5").(string))
+			d.Set("dashboard_change_detected", (resp.Etag != d.Get("etag").(string)) || func() bool {
+				if v, ok := d.GetOk("file_path"); ok {
```
hectorcast-db, line 162: THis will work, but I think we only need to detect remote changes here with the Etag. The DiffSuppress function above should take care of the local changes.

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,50 @@
+---
+subcategory: "Workspace"
+---
+# lakeview_dashboard Resource
```
hectorcast-db, line 4: databricks_lakeview_dashboard

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Example Usage
+
+Dashboard using `serialized_dashboard` attribute:
+
+```hcl
+resource "databricks_lakeview_dashboard" "dashboard" {
+			display_name			= 	"New Dashboard"
+			warehouse_id			=	"{env.TEST_DEFAULT_WAREHOUSE_ID}"
+			serialized_dashboard	=	"{\"pages\":[{\"name\":\"new_name\",\"displayName\":\"New Page\"}]}"
```
hectorcast-db, line 16: nit, formatting

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+Dashboard using `file_path` attribute:
+
+```hcl
+resource "databricks_lakeview_dashboard" "dashboard" {
+            display_name			= 	"New Dashboard"
+            warehouse_id			=	"{env.TEST_DEFAULT_WAREHOUSE_ID}"
+            file_path				=	"${path.module}/dashboard.json"
+            embed_credentials		=	false // Optional
```
hectorcast-db, line 29: formatting

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,412 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAccLakeviewDashboard(t *testing.T) {
+	workspaceLevel(t, step{
+		Template: `
+		resource "databricks_lakeview_dashboard" "d1" {
+			display_name			= 	"Monthly Traffic Report"
+			warehouse_id			=	"{env.TEST_DEFAULT_WAREHOUSE_ID}"
+			file_path				=	"/Users/divyansh.vijayvergia/terraform-provider-databricks/Experimental-Divyansh/json_file"
```
hectorcast-db, line 22: Don't use absolute paths, nor your username

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,412 @@
+package acceptance
+
+import (
```
hectorcast-db, line 3: In general, you should check that the resource has been correctly created in the workspace. You can use the Go SDK client to access the dashboard.

internal/acceptance/lakeview_dashboard_test.go
Divyansh-db: Yes. I forgot to change this here.

dashboards/resource_lakeview_dashboard.go
Divyansh-db: I tried creating custom diff function for file as well. But somehow, checking md5 hash of file in read function works better. (now there is no custom diff for `file_path` field

dashboards/resource_lakeview_dashboard.go
Divyansh-db: So the logic was:
If you have a file path mentioned by the user, do not calculate the diff using `serialized_dashboard` field. The md5 diff will be calculated in the read function.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+func readFileContent(source string) ([]byte, error) {
+	log.Printf("[INFO] Reading %s", source)
+	f, err := os.Open(source)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	reader := bufio.NewReader(f)
+	return io.ReadAll(reader)
+}
```
edwardfeng-db, line 29: I think we have sharable helper function for this already? cc @hectorcast-db

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Calculates MD5 hash of the given content
+func calculateMd5Hash(content []byte) string {
+	return fmt.Sprintf("%x", md5.Sum(content))
+}
```
edwardfeng-db, line 34: Same for this, I think we did this for other resources as well, can we share those?

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+		if err != nil {
+			return "", "", err
+		}
+	} else {
+		log.Printf("[INFO] Reading `serialized_json` of %d bytes", len(jsonStr))
+		content = []byte(jsonStr)
+	}
+	md5Hash = calculateMd5Hash(content)
+	return string(content), md5Hash, nil
+}
```
edwardfeng-db, line 50: Same for this one

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+* `file_path` - (Optional) The path to the dashboard JSON file. Conflicts with `serialized_dashboard`.
+* `embed_credentials` - (Optional) Whether to embed credentials in the dashboard. Default is `true`.
+* `parent_path` - (Required) The workspace path of the folder containing the dashboard. Includes leading slash and no trailing slash.
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The unique ID of the Lakeview dashboard.
+
```
edwardfeng-db, line 50: Can we document our decision on publishing directly instead of drafting?

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+				return err
+			}
+			d.Set("md5", md5Hash)
+			new_dashboard.SerializedDashboard = content
+			created_dashboard, err := w.Lakeview.Create(ctx, new_dashboard)
+			if err != nil {
+				return err
+			}
+
+			time.Sleep(2 * time.Second)
```
edwardfeng-db, line 123: Why do we need to sleep for 2 seconds here?

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+			d.Set("etag", resp.Etag)
+
+			// Publish the dashboard
+			// fmt.Println("Publishing the dashboard", d.Get("embed_credentials").(bool))
+			_, err = w.Lakeview.Publish(ctx, dashboards.PublishRequest{
+				DashboardId:      created_dashboard.DashboardId,
+				WarehouseId:      d.Get("warehouse_id").(string),
+				EmbedCredentials: d.Get("embed_credentials").(bool),
+				ForceSendFields:  []string{"EmbedCredentials"},
```
edwardfeng-db, line 139: Could be good to add a comment explaining why we need to add `EmbedCredentials` to `ForceSendFields`

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,412 @@
...
+				if !ok {
+					return fmt.Errorf("dashboard resource not found")
+				}
+				fmt.Println(dashboard.Primary.Attributes["serialized_dashboard"])
+				return nil
+			},
+		),
+	}, step{
+		PreConfig: func() {
+			tmpDir := "/tmp/LakeviewDashboardTest"
```
edwardfeng-db, line 290: I suggest randomize these tmpDirs, for example https://github.com/databricks/terraform-provider-databricks/blob/75236a645b7d86819e43281e6c2623bbc5527528/internal/acceptance/cluster_test.go#L13

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,412 @@
...
+			},
+		),
+	}, step{
+		Template: `
+		resource "databricks_lakeview_dashboard" "d1" {
+			display_name			= 	"Monthly Traffic Report"
+			warehouse_id			=	"{env.TEST_DEFAULT_WAREHOUSE_ID}"
+			file_path				=	"/tmp/LakeviewDashboardTest/Dashboard.json"
+			parent_path				= 	"/Shared/Teams"
+		}
```
edwardfeng-db, line 398: These templates can probably be constructed with a helper function which takes in parameters to update the corresponding fields, that will make the test look way more concise.

dashboards/resource_lakeview_dashboard.go
hectorcast-db: https://github.com/databricks/terraform-provider-databricks/blob/main/workspace/file_resource.go#L22
I think we can make it public

dashboards/resource_lakeview_dashboard.go
Divyansh-db: I removed it now. It was a hack for when API was not working correctly

dashboards/resource_lakeview_dashboard.go
Divyansh-db: sure

dashboards/resource_lakeview_dashboard.go
Divyansh-db: I took inspiration from that function only. But, is it not considered good to have a new function in the code. (The function I used is bit different and some of the things in old function are depricated)

dashboards/resource_lakeview_dashboard.go
hectorcast-db: Read should be reserved for server side reading. Read can be called in other occasions. We should do the file reading during the diff supress instead.

dashboards/resource_lakeview_dashboard.go
hectorcast-db: We can update the deprecated code. The deprecation message tells you to just replace the line with your line. These 2 functions do the same. Reusing the code keeps it easier in the future (less code, less maintenance).

internal/acceptance/lakeview_dashboard_test.go
hectorcast-db: Helper to create random things
https://github.com/databricks/terraform-provider-databricks/blob/main/qa/testing.go#L50

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,187 @@
...
+					if err != nil {
+						return false
+					}
+					return (new_file_hash == d.Get("md5").(string))
+				}
+				_, new_json_hash, err := readSerializedJsonContent(new, "")
+				if err != nil {
+					return false
+				}
+				return d.Get("md5").(string) == new_json_hash && !d.Get("dashboard_change_detected").(bool)
```
hectorcast-db, line 63: Can't we simplify this?
```
if _, new_hash, err := readSerializedJsonContent(new, d.GetOk("file_path").(string)); err != nil {
					return false
				}
				log.Printf("[INFO] Suppressing %s diff: %v", d.Id(), old == d.Get("md5"))
return old == new_hash && !d.Get("dashboard_change_detected")
```

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+			d.Set("md5", md5Hash)
+			dashboard.SerializedDashboard = content
+			resp, err := w.Lakeview.Update(ctx, dashboard)
+			if err != nil {
+				return err
+			}
+			d.Set("etag", resp.Etag)
+
+			// Publish the dashboard
```
hectorcast-db, line 160: nit, add the same comment as before

dashboards/resource_lakeview_dashboard_test.go
```diff
@@ -0,0 +1,181 @@
...
+		State: map[string]any{
+			"display_name":         "Dashboard name",
+			"warehouse_id":         "abc",
+			"parent_path":          "/path", // Ensure this is consistent
+			"serialized_dashboard": "serialized_dashboard_updated",
+		},
+		InstanceState: map[string]string{
+			"display_name":         "Dashboard name",
+			"warehouse_id":         "abc",
+			"parent_path":          "/path", // Ensure this is consistent
```
hectorcast-db, line 154: What does this comment mean?

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,54 @@
+---
+subcategory: "Workspace"
+---
+# databricks_lakeview_dashboard Resource
+
+This resource allows you to manage [Databricks Lakeview Dashboards](https://docs.databricks.com/api/workspace/lakeview). To manage [Lakeview Dashboards](https://docs.databricks.com/api/workspace/lakeview) you must have a warehouse access on your databricks workspace.
```
hectorcast-db, line 6: Nit, but from the user reading the final doc, it feels like [Databricks Lakeview Dashboards] and [Lakeview Dashboards] are two different things/links.

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,481 @@
...
+type TemplateStruct struct {
+	DisplayName         string
+	WarehouseId         string
+	ParentPath          string
+	SerializedDashboard string
+	FilePath            string
+	EmbedCredentials    string
+}
+
+func MakeTemplate(template TemplateStruct) string {
```
hectorcast-db, line 26: This does not need to be public.

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,481 @@
...
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/dashboards"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/databricks/terraform-provider-databricks/qa"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type TemplateStruct struct {
```
hectorcast-db, line 17: This does not need to be public

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,481 @@
...
+				return err
+			}
+			dashboard, err := w.Lakeview.Get(ctx, dashboards.GetDashboardRequest{
+				DashboardId: id,
+			})
+			if err != nil {
+				return err
+			}
+			assert.Equal(t, "Monthly Traffic Report", dashboard.DisplayName)
+			fmt.Println(dashboard.SerializedDashboard)
```
hectorcast-db, line 99: Let's not print the dashboard. In case of test failure, the http calls are printed, so we will see the dashboard.

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,481 @@
...
```
hectorcast-db, line 128: Same here and below

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,481 @@
...
+			require.NoError(t, err)
+			return nil
+		}),
+	}, step{
+		Template: MakeTemplate(template.SetAttributes(map[string]string{
```
hectorcast-db, line 133: The fact that we keep updating the template, but reuse some attributes seems a bit off to me. I would prefer to see here all attributes for the current step. @edwardfeng-db what do you think?

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,481 @@
...
+			dashboard, err := w.Lakeview.Get(ctx, dashboards.GetDashboardRequest{
+				DashboardId: id,
+			})
+			if err != nil {
+				return err
+			}
+			assert.Equal(t, "Monthly Traffic Report", dashboard.DisplayName)
+			fmt.Println(dashboard.SerializedDashboard)
+			require.NoError(t, err)
+			require.NotEqual(t, dashboard.UpdateTime, dashboard.CreateTime)
```
hectorcast-db, line 329: You also should check that it has been updated by tf. This `UpdateTime` could be the one by your manual modification.

internal/acceptance/lakeview_dashboard_test.go
Divyansh-db: So where should I put it? Should I also put it in util.go in common?

internal/acceptance/lakeview_dashboard_test.go
Divyansh-db: So what can we check in the 'check' function that would be relevant?

internal/acceptance/lakeview_dashboard_test.go
hectorcast-db: Just make it lowercase:
```
func makeTemplate(...
```

internal/acceptance/lakeview_dashboard_test.go
hectorcast-db: Not sure. First I would think about validating the Dashabord. But service team does not promise a static format. Maybe check that the length of the serialized dashabord is > 10 (to ensure it's not empty). And add a note

docs/resources/lakeview_dashboard.md
Divyansh-db: Moved Databricks out of the link

internal/acceptance/lakeview_dashboard_test.go
Divyansh-db: So basically also add the check to compare update time from GET and from the local terraform state?

internal/acceptance/lakeview_dashboard_test.go
Divyansh-db: Would not that be same as writing the full template. I thing the reasoning that Edward gave was that in tests with many steps if we change only one attribute then why to write the full template. Also, the change would be more visible if we just mention that.

hectorcast-db: It's a bit less than the whole template (4 lines vs ~10). And mappings have better syntax highlighting. But yes, we don't want to make it too big.

internal/acceptance/lakeview_dashboard_test.go
hectorcast-db: Either time, or the etag.

internal/acceptance/lakeview_dashboard_test.go
edwardfeng-db: I'd say I would still advocate for having a helper methods, but if you have strong opinion against it, don't have to block on me.

I'll leave the decision to @Divyansh-db

hectorcast-db: I like having helper methods. I just found it a bit hard to read. But it could be because the git view is not as great as the vscode one. Let's go with this. We can change it later.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+			s["warehouse_id"].Required = true
+
+			s["file_path"] = &schema.Schema{
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"serialized_dashboard"},
+			}
+			s["md5"] = &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
```
hectorcast-db, line 68: I am not sure this needs to be computed. Looking at the `files` resource, it should not. Marking it as computed also supresses diffs, which can have unintended consequences.

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The unique ID of the Lakeview dashboard.
+
+## Notes
+* Currently, only one of `serialized_dashboard` or `file_path` can be used throughout the lifecycle of the dashboard. If you want to switch from one to the other, you must first destroy the dashboard resource and then recreate it with the new attribute.
+* Currently, all the dashboards managed by Terraform will be published by default. There is no option to keep the dashboards in draft state. If required in future, functionality to manage the publish status of the dashboard will be added.
```
hectorcast-db, line 54: Let's refrase this. We don't want to overpromise. Let's just say
`Dashboards managed by Terraform are published automatically.`
Remove `default` and the `future option`, since that can be misunderstood.

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,489 @@
...
+			warehouse_id = dashboard.WarehouseId
+			etag = dashboard.Etag
+			// As the format of the serialized dashboard is not fixed, we can only check if it is not empty
+			assert.NotEqual(t, "", dashboard.SerializedDashboard)
+			return nil
+		}),
+	}, step{
+		PreConfig: func() {
+			w, err := databricks.NewWorkspaceClient(&databricks.Config{})
+			if err != nil {
```
hectorcast-db, line 410: Test should fail if there is an error. Use this instead of logging (it will also log automatically)
```
require.NoError(t, err)
```

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,489 @@
...
+				return err
+			}
+			assert.Equal(t, "Monthly Traffic Report", dashboard.DisplayName)
+			require.NoError(t, err)
+			resource.TestCheckResourceAttr("databricks_lakeview_dashboard.d1", "etag", dashboard.Etag)
+			// As the format of the serialized dashboard is not fixed, we can only check if it is not empty
+			assert.NotEqual(t, "", dashboard.SerializedDashboard)
+			os.Remove("/tmp/LakeviewDashboardTest/Dashboard.json")
+			os.Remove("/tmp/LakeviewDashboardTest")
```
hectorcast-db, line 485: Tests may run in parallel:
* Do not remove the folder.
* Either the file or the folder name should contain random characters.
Note that the OS does clean the /tmp folder automatically, so removing the file is ok, but not needed.

Divyansh-db: Got it.

docs/resources/lakeview_dashboard.md
Divyansh-db: got it

dashboards/resource_lakeview_dashboard.go
Divyansh-db: 2 points:
1) In files, it is marked as optional. I know it is not exposed to the user but it is better that user cannot change this field (even unintentionally)
2) We do not have a Suppress on this field. We use this field indirectly in Diffsuppress for other field which fetches the 'md5' field from state so might not cause issue
But I am open to your opinion

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"github.com/databricks/databricks-sdk-go/service/dashboards"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Calculates MD5 hash of the given content
+func calculateMd5Hash(content []byte) string {
+	return fmt.Sprintf("%x", md5.Sum(content))
+}
```
tanmay-db, line 17: We could move it to common/util.go similar to what we did for ReadFileContents since it's more general in nature.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+		if err != nil {
+			return "", "", err
+		}
+	} else {
+		log.Printf("[INFO] Reading `serialized_json` of %d bytes", len(jsonStr))
+		content = []byte(jsonStr)
+	}
+	md5Hash = calculateMd5Hash(content)
+	return string(content), md5Hash, nil
+}
```
tanmay-db, line 33: Same for this, it's more common in nature. Maybe right now only lakeview resource will be using this method but it would be nice it we refactor it now to avoid situations like moving `ReadFileContent` to common/util.go that we are doing in this PR in future.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+
+// ResourceLakeviewDashboard manages lakeview dashboards
+func ResourceLakeviewDashboard() common.Resource {
+	s := common.StructToSchema(dashboards.Dashboard{},
```
tanmay-db, line 37: This is quite big, let's split this into smaller parts so it's easier on code readability + maintenance. 

eg: Have a method (resourceLakeviewDashboardSchema) to return the schema

Also let's use `common.CustomizeSchemaPath` to modify the schema.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+			return s
+		})
+	return common.Resource{
+		Schema: s,
+		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var new_dashboard dashboards.CreateDashboardRequest
```
tanmay-db, line 88: Let's use camelCase

> Names in Go should in general not contain underscores.
Ref for more details: https://google.github.io/styleguide/go/decisions.html#naming

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+			d.Set("dashboard_change_detected", (resp.Etag != d.Get("etag").(string)))
+			return common.StructToData(resp, s, d)
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var dashboard dashboards.UpdateDashboardRequest
```
tanmay-db, line 137: nit: using updateDashboardRequest for variable name would be more readable.

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) The display name of the Lakeview dashboard.
+* `warehouse_id` - (Required) The warehouse ID used to run the dashboard.
+* `serialized_dashboard` - (Optional) The contents of the dashboard in serialized string form. Conflicts with `file_path`.
+* `file_path` - (Optional) The path to the dashboard JSON file. Conflicts with `serialized_dashboard`.
+* `embed_credentials` - (Optional) Whether to embed credentials in the dashboard. Default is `true`.
```
tanmay-db, line 43: Should we go ahead with default embed credentials?

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The unique ID of the Lakeview dashboard.
+
+## Notes
+* Currently, only one of `serialized_dashboard` or `file_path` can be used throughout the lifecycle of the dashboard. If you want to switch from one to the other, you must first destroy the dashboard resource and then recreate it with the new attribute.
+* Currently, dashboards managed by Terraform will be published automatically.
```
tanmay-db, line 54: Might be good to not use `currently`, it implied we will change this later (not sure if we plan to).

internal/acceptance/lakeview_dashboard_test.go
```diff
@@ -0,0 +1,477 @@
...
+	`, template.DisplayName, template.WarehouseId, template.ParentPath)
+	if template.SerializedDashboard != "" {
+		templateString += fmt.Sprintf(`	serialized_dashboard	=	"%s"
+	`, template.SerializedDashboard)
+	}
+	if template.FilePath != "" {
+		templateString += fmt.Sprintf(`	file_path				=	"%s"
+	`, template.FilePath)
+	}
+	if template.EmbedCredentials != "" {
```
tanmay-db, line 42: Not sure if spacing is accidental indentation issue.

tanmay-db (commented):
Thanks for the PR, left a few comments.

dashboards/resource_lakeview_dashboard.go
edwardfeng-db: +1

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,179 @@
...
+			s["create_time"].Computed = true
+			s["dashboard_id"].Computed = true
+			s["display_name"].Optional = false
+			s["display_name"].Required = true
+			s["etag"].Computed = true
+			s["lifecycle_state"].Computed = true
+			s["parent_path"].Optional = false
+			s["parent_path"].Required = true
+			s["parent_path"].ForceNew = true
+			s["path"].Computed = true
```
edwardfeng-db, line 48: We can use the common methods like `setComputed` etc. See example [here](https://github.com/databricks/terraform-provider-databricks/blob/ac06fdbfff9b41a84817c4251ad0c2bc8a645016/jobs/resource_job.go#L517)

edwardfeng-db: For adding new fields and setting fields properties, we can use `s.SchemaPath().<method>`

dashboards/resource_lakeview_dashboard.go
edwardfeng-db, line 37: I think it'd be better to create a type and implement the methods like this one [here](https://github.com/databricks/terraform-provider-databricks/blob/6371eeab97f5b6cea348ee8e30c74c85b7d50c9d/jobs/resource_job.go#L485-L511). This is our latest convention for doing things.

docs/resources/lakeview_dashboard.md
Divyansh-db: It is written in the API specs that default is true
https://docs.databricks.com/api/workspace/lakeview/publish

internal/acceptance/lakeview_dashboard_test.go
Divyansh-db: It is intentional (just to make the request template in debug log well formatted)

dashboards/resource_lakeview_dashboard.go
Divyansh-db: Schema custom function cannot be applied if the Schema struct is of non-local type

Divyansh-db: Screenshot 2024-07-08 at 07 27 45

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Example Usage
+
+Dashboard using `serialized_dashboard` attribute:
+
+```hcl
+resource "databricks_lakeview_dashboard" "dashboard" {
+			display_name			= 	"New Dashboard"
+			warehouse_id			=	"{env.TEST_DEFAULT_WAREHOUSE_ID}"
+			serialized_dashboard	        =	"{\"pages\":[{\"name\":\"new_name\",\"displayName\":\"New Page\"}]}"
```
nkvuong, line 16: should just look up the sql warehouse from a data source


```suggestion
data "databricks_sql_warehouse" "starter" {
  name = "Starter Warehouse"
}

resource "databricks_lakeview_dashboard" "dashboard" {
  display_name			= 	"New Dashboard"
  warehouse_id			=	data.databricks_sql_warehouse.starter.id
  serialized_dashboard	        =	"{\"pages\":[{\"name\":\"new_name\",\"displayName\":\"New Page\"}]}"
```

dashboards/resource_lakeview_dashboard_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/databricks/databricks-sdk-go/service/dashboards"
+	"github.com/databricks/terraform-provider-databricks/qa"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLakeviewDashboardCreate(t *testing.T) {
+	d, err := qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
```
nkvuong, line 14: could we avoid using `Fixtures`, and use `MockWorkspaceClientFunc` instead for testing?

dashboards/resource_lakeview_dashboard_test.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+		Resource: ResourceLakeviewDashboard(),
+		Create:   true,
+		State: map[string]any{
+			"display_name":         "Dashboard name",
+			"warehouse_id":         "abc",
+			"parent_path":          "/path",
+			"serialized_dashboard": "serialized_json",
+		},
+	}.Apply(t)
```
nkvuong, line 65: you could rewrite this with `ApplyAndExpectData`

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,155 @@
...
+			if err != nil {
+				return err
+			}
+
+			d.Set("etag", createdDashboard.Etag)
+
+			// We need to 'Force send' the EmbedCredentials field because it is 'omitempty' and if it is not set, it will be ignored. This is a workaround to force the field to be sent if the user wants to set 'embed_credentials' to false.
+			_, err = w.Lakeview.Publish(ctx, dashboards.PublishRequest{
+				DashboardId:      createdDashboard.DashboardId,
+				WarehouseId:      d.Get("warehouse_id").(string),
```
nkvuong, line 82:
```suggestion
				WarehouseId:      createdDashboard.WarehouseId,
```

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,155 @@
...
+				EmbedCredentials: d.Get("embed_credentials").(bool),
+				ForceSendFields:  []string{"EmbedCredentials"},
+			})
+			if err != nil {
```
nkvuong, line 86: if we could not publish, should we roll back & delete the dashboard created?

docs/resources/lakeview_dashboard.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The unique ID of the Lakeview dashboard.
+
+## Notes
+* Only one of `serialized_dashboard` or `file_path` can be used throughout the lifecycle of the dashboard. If you want to switch from one to the other, you must first destroy the dashboard resource and then recreate it with the new attribute.
+* Dashboards managed by Terraform will be published automatically.
```
nkvuong, line 54: would it be better to have an additional Terraform field as `is_published`, and let Terraform publish/unpublish as required?

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,155 @@
...
+			return common.StructToData(resp, s, d)
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var updateDashboardRequest dashboards.UpdateDashboardRequest
+			common.DataToStructPointer(d, s, &updateDashboardRequest)
+			updateDashboardRequest.DashboardId = d.Get("dashboard_id").(string)
```
nkvuong, line 115:
```suggestion
			updateDashboardRequest.DashboardId = d.Id()
```

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,155 @@
...
+			if err != nil {
+				return err
+			}
+			err = w.Lakeview.Trash(ctx, dashboards.TrashDashboardRequest{
+				DashboardId: d.Id(),
+			})
+			if err != nil {
+				return err
+			}
+			return nil
```
nkvuong, line 152:
```suggestion
			return w.Lakeview.Trash(ctx, dashboards.TrashDashboardRequest{
				DashboardId: d.Id(),
			})
```

dashboards/resource_lakeview_dashboard.go
Divyansh-db: I think its a good point. We should delete the dashboard as well.

dashboards/resource_lakeview_dashboard.go
```diff
@@ -0,0 +1,151 @@
...
+func ResourceLakeviewDashboard() common.Resource {
+	s := common.StructToSchema(dashboards.Dashboard{},
+		func(s map[string]*schema.Schema) map[string]*schema.Schema {
+			s["create_time"].Computed = true
+			s["dashboard_id"].Computed = true
+			s["display_name"].Optional = false
+			s["display_name"].Required = true
+			s["etag"].Computed = true
+			s["lifecycle_state"].Computed = true
+			s["parent_path"].Optional = false
```
alexott, line 21: It's more to the corresponding team, but I'm wondering why we still don't allow to change parent path - this already leads to a lot of problems with redash dashboards.

@mgyucht should we ask lakeview team about plans for it?

docs/resources/lakeview_dashboard.md
nkvuong: this page will be part of Terraform registry documentation, so we want a relatively self contained example. The approach of environment variable is specific to our acceptance tests, so won't be of use for other customers.

The Starter Warehouse comes by default with all workspaces, but I think we deleted that from our internal deco workspaces

dashboards/resource_dashboard.go
```diff
@@ -0,0 +1,170 @@
...
+	// Default values
+	s.SchemaPath("embed_credentials").SetDefault(true)
+
+	// DiffSuppressFunc
+	s.SchemaPath("serialized_dashboard").SetCustomSuppressDiff(customDiffSerializedDashboard)
+
+	return s
+}
+
+var dashboardsSchema = common.StructToSchema(Dashboard{}, nil)
```
nkvuong, line 58: nit
```suggestion
var dashboardSchema = common.StructToSchema(Dashboard{}, nil)
```

dashboards/resource_dashboard_test.go
```diff
@@ -0,0 +1,171 @@
...
+				DashboardId:         "xyz",
+				DisplayName:         "Dashboard name",
+				SerializedDashboard: "serialized_json_2",
+				WarehouseId:         "abc",
+				UpdateTime:          "2125678",
+			}, nil)
+		},
+		Resource: ResourceDashboard(),
+		Create:   true,
+		State: map[string]any{
```
nkvuong, line 52: nit: we normally use `HCL` instead of `State`, as it better reflects the target configuration

dashboards/resource_dashboard_test.go
```diff
@@ -0,0 +1,171 @@
...
+				DisplayName:         "Dashboard name",
+				SerializedDashboard: "serialized_dashboard_updated_2",
+				WarehouseId:         "abc",
+				UpdateTime:          "2125679",
+			}, nil)
+		},
+		Resource:    ResourceDashboard(),
+		Update:      true,
+		ID:          "xyz",
+		RequiresNew: true,
```
nkvuong, line 133: we only need this is the update forces a new resource
```suggestion
```

docs/resources/dashboard.md
```diff
@@ -0,0 +1,54 @@
...
+
+This resource allows you to manage Databricks [Dashboards](https://docs.databricks.com/en/dashboards/index.html). To manage [Dashboards](https://docs.databricks.com/en/dashboards/index.html) you must have a warehouse access on your databricks workspace.
+
+## Example Usage
+
+Dashboard using `serialized_dashboard` attribute:
+
+```hcl
+resource "databricks_dashboard" "dashboard" {
+			display_name			= 	"New Dashboard"
```
nkvuong, line 14: we seem to have too many tabs here, could you run `make fmt-docs` as well

nkvuong (approved):
a few nits, otherwise looks good to me

docs/resources/dashboard.md
Divyansh-db: I am unable to use terrafmt. but I manually formatted the docs

alexott: you just need to install it as per: https://github.com/katbyte/terrafmt?tab=readme-ov-file#local-install

alexott:
What about permissions on Dashboards? Do we have a follow-up item for it?

edwardfeng-db (approved):

PR #3733, opened 2024-07-03
[Feature] Add `active` attribute to `databricks_user` data source

### LINKED ISSUE #3731: [FEATURE] New feature request - add attribute 'active' to databricks_user Data Source
### Use-cases
Our users are provisioned from AAD using SCIM. When the user is deleted in AAD he is set to 'inactive' in Databricks accounts.
There is currently no way to get that property using the data block `databricks_user` (while it is supported for the resource `databricks_user`).

### Attempted Solutions
We are now also (trying to) get the user using azure_ad provider, and comparing properties, but this is not a nice solution.

### Proposal
Add attribute `active` to data block `databricks_user`, as it is included in the [rest api](https://docs.databricks.com/api/azure/account/accountusers/get).

### References

- https://docs.databricks.com/api/azure/account/accountusers/get

840:
## Changes
Add `active` attribute to `databricks_user` data source.

Resolves #3731 

## Tests
Acceptance test and verified using the following code:
```
terraform {
  required_providers {
    databricks = {
    }
  }
}

provider "databricks" {
  profile = "<REDACTED>"
}

data "databricks_user" "this" {
  user_name = "<REDACTED>"
}

output "user" {
  value = data.databricks_user.this.active
}
```

Output:
```
❯ terraform apply
data.databricks_user.this: Reading...
data.databricks_user.this: Read complete after 1s [id=<REDACTED>]

Changes to Outputs:
  + user = true
```

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/data_user.go
+++ b/scim/data_user.go
@@ -67,6 +67,10 @@ func DataSourceUser() common.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"active": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
 		},
 		Read: func(ctx context.Context, d *schema.ResourceData, m *common.DatabricksClient) error {
 			usersAPI := NewUsersAPI(ctx, m)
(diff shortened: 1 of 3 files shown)


alexott (approved):
It looks good - should we extend it to service principals as well?

scim/data_user.go
```diff
@@ -67,6 +67,10 @@ func DataSourceUser() common.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"active": {
```
nkvuong, line 70: can we switch to GoSDK & `common.WorkspaceDataWithParams` for this data source?

840: I think the issue is that this data source is being used for both the workspace- and account-level. Resources that are being used at both levels cannot use `common.WorkspaceDataWithParams`, right?

PR #3737, opened 2024-07-04
[Fix] Tolerate `databricks_workspace_conf` deletion failures

### LINKED ISSUE #3316: [ISSUE] Issue destroying databricks_workspace_conf after setting maxTokenLifetimeDays field
### Configuration

```hcl
resource "databricks_workspace_conf" "this" {
custom_config = {
"enableIpAccessLists" : true,
"maxTokenLifetimeDays": 90
}
}
```
### Expected Behavior

It should be possible to destroy the resource databricks_workspace_conf or to remove the "maxTokenLifetimeDays" field after setting it to some value

### Actual Behavior

Currently destroying databricks_workspace_conf is not allowed after setting maxTokenLifetimeDays to a value. It throws the error │ Error: cannot delete workspace conf: Some values are not allowed: {"maxTokenLifetimeDays":""}

### Steps to Reproduce

1. create a databricks_workspace_conf resource 

2) Remove or comment the field maxTokenLifetimeDays and run apply or try destroying the entire resource. Terraform will return 

Error: cannot update workspace conf: Some values are not allowed: {"maxTokenLifetimeDays":""}

3) Try adding the field "maxTokenLifetimeDays": 90 back and terraform will pass

4) Try destroying the resource after adding maxTokenLifetimeDays back, Terraform will fail with the same error even after adding it back

It looks like once the value maxTokenLifetimeDays is set once, the resource gets into a state where it cannot be destroyed in Terraform

provider registry.terraform.io/databricks/databricks v1.37.1

### Debug Output

2024-02-29T10:07:57.769-0600 [DEBUG] provider.terraform-provider-databricks_v1.37.1: PATCH /api/2.0/workspace-conf
> {
>   "enableIpAccessLists": "false",
>   "maxTokenLifetimeDays": ""
> }
< HTTP/2.0 400 Bad Request
< {
<   "error_code": "BAD_REQUEST",
<   "message": "Some values are not allowed: {\"maxTokenLifetimeDays\":\"\"}"
< }: @caller=/home/runner/work/terraform-provider-databricks/terraform-provider-databricks/logger/logger.go:33 tf_provider_addr=registry.terraform.io/databricks/databricks tf_req_id=3cac0fb4-cf9b-da69-2908-8c39c9484e2b @module=databricks tf_resource_type=databricks_workspace_conf tf_rpc=ApplyResourceChange timestamp=2024-02-29T10:07:57.768-0600
2024-02-29T10:07:57.769-0600 [ERROR] provider.terraform-provider-databricks_v1.37.1: Response contains error diagnostic: tf_resource_type=databricks_workspace_conf diagnostic_summary="cannot delete workspace conf: Some values are not allowed: {\"maxTokenLifetimeDays\":\"\"}" tf_provider_addr=registry.terraform.io/databricks/databricks tf_req_id=3cac0fb4-cf9b-da69-2908-8c39c9484e2b @caller=/home/runner/work/terraform-provider-databricks/terraform-provider-databricks/vendor/github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/diag/diagnostics.go:62 diagnostic_detail="" diagnostic_severity=ERROR tf_proto_version=5.4 tf_rpc=ApplyResourceChange @module=sdk.proto timestamp=2024-02-29T10:07:57.768-0600
2024-02-29T10:07:57.778-0600 [DEBUG] State storage *statemgr.Filesystem declined to persist a state snapshot
2024-02-29T10:07:57.778-0600 [ERROR] vertex "databricks_workspace_conf.this (destroy)" error: cannot delete workspace conf: Some values are not allowed: {"maxTokenLifetimeDays":""}

Error: cannot delete workspace conf: Some values are not allowed: {"maxTokenLifetimeDays":""}

mgyucht:
## Changes
Currently, deleting `databricks_workspace_conf` can fail in some cases. The provider currently tries to do a best-effort to reset config values to an "original" state by setting all boolean fields to `false` and string fields to `""`. However, the provider isn't aware of three key bits of information: the set of allowed keys in the workspace configuration, the types of values accepted for those keys, and the default values for new workspaces (the last of which can vary from workspace to workspace).

As a result, it does a best-effort approach to interpret the value for each key as a boolean, setting it to `false` on delete if it can be interpreted as a boolean, otherwise setting the value for the key to `""`. This can fail, however, if a user has specified a value like `TRue`, this may not be interpreted as a boolean by the provider. The provider then tries to reset its value to `""`, which fails.

This PR adds two layers of protection around deletion: first, it attempts to lower-case the value before checking if it is a boolean to handle possible issues with case-sensitivity with the values in workspace_conf. Secondly, if the configuration update fail due to invalid key types, the provider will print a warning message guiding users to our docs page to help them review their workspace config, but it will consider the resource to be successfully deleted.

As part of this, I added a small internal module to provide links to docs pages when the provider needs to print them in log messages.

Resolves #3316, #1802.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


workspace/resource_workspace_conf.go
```diff
@@ -138,7 +142,17 @@ func ResourceWorkspaceConf() common.Resource {
...
+			// Tolerate errors like the following on deletion:
+			// cannot delete workspace conf: Some values are not allowed: {"enableGp3":"","enableIpAccessLists":""}
+			// The API for workspace conf is quite limited and doesn't support a generic "reset to original state"
+			// operation. So if our attempted reset fails, don't fail resource deletion.
+			var apiErr *apierr.APIError
+			if errors.As(err, &apiErr) && strings.HasPrefix(apiErr.Message, "Some values are not allowed") {
+				tflog.Warn(ctx, fmt.Sprintf("Failed to delete databricks_workspace_conf: %s. The workspace configuration may not be fully restored to its original state. For more information, see %s", apiErr.Message, docs.ResourceDocumentationUrl("workspace_conf")))
+				return nil
+			}
+			return err
```
pietern, line 155: Do you know if this could lead to partial update, or if the conf API rejects everything if the payload contains at least 1 invalid setting?

alexott: I have the same question - if everything is rejected, should we set each setting separately?

mgyucht: IIUC, you're asking if partial updates succeed for most settings but fail for a few, we could restore most settings to their original state and minimize manual work for customers. If the API partially updates settings but fails for others, then I think we will get this desired behavior. Otherwise, the error message will include the settings that the customer needs to manually check. I will double check with the backing team to understand what might happen with this.

alexott (approved):

PR #3751, opened 2024-07-09
[Feature] Introduced Terraform Plugin Framework

tanmay-db:
## Changes

This PR introduces plugin framework to the Databricks Terraform Provider. 
- Introduce plugin framework
- Add core support for resource and data source. 
- Add provider schema
- Add support to configure databricks client
- Add support for provider fixture
- Add support to get raw config to be used in provider tests and further in unit testing framework. 
- Add Quality Monitor resource using plugin framework
- Add UC Volumes data source using plugin framework
- Fix bugs related in configuring client
- Fix bugs where provider schema is used instead of resource schema
- Fix miscellaneous bugs 

## Tests

- Run existing provider tests also for plugin framework. All passes. 
- Manually tested the above 2 resources E2E (apply works for data source, for resource -- we need customise schema support for `SetReadOnly`)

Note: We don't have unit and integration test support as of now, so going ahead with commiting the .tf files for these resources so we can test the changes in the meantime. Once we have proper support, these can be converted to example or removed. 

image


pluginframework/resource_quality_monitor.go
```diff
@@ -0,0 +1,196 @@
...
+func (r *QualityMonitorResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Terraform schema for Databricks Lakehouse Monitor. MonitorInfo struct is used to create the schema",
+		Attributes: common.PluginFrameworkResourceStructToSchemaMap(catalog_tf.MonitorInfo{}, func(c common.CustomizableSchemaPluginFramework) common.CustomizableSchemaPluginFramework {
+			c.AddNewField("skip_builtin_dashboard", schema.BoolAttribute{Optional: true, Required: false})
+			c.AddNewField("warehouse_id", schema.StringAttribute{Optional: true, Required: false})
+			c.SetRequired("assets_dir")
+			c.SetRequired("output_schema_name")
+			c.SetRequired("table_name")
+			// TODO: Uncomment this once SetReadOnly is supported in the plugin framework
```
edwardfeng-db, line 61: I don't know how to actually achieve this. As I saw in the provider schema types, all of the `isReadOnly` functions are hard coded to be returning false

edwardfeng-db (approved):
LGTM, thanks!

PR #3778, opened 2024-07-16
[Feature] Automatically create `parent_path` folder if it doesn't exist

alexott:
## Changes

This makes behavior consistent with other workspace objects - notebooks, workspace files, etc.

I have the skeleton of the test for it, but it doesn't work because I can't mock the same API call two times with different return results

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/dashboards/resource_dashboard.go
+++ b/dashboards/resource_dashboard.go
@@ -2,6 +2,8 @@ package dashboards
 
 import (
 	"context"
+	"log"
+	"strings"
 
 	"github.com/databricks/databricks-sdk-go/service/dashboards"
 	"github.com/databricks/terraform-provider-databricks/common"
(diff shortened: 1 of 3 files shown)


dashboards/resource_dashboard.go
```diff
@@ -76,7 +78,17 @@ func ResourceDashboard() common.Resource {
 			newDashboardRequest.SerializedDashboard = content
 			createdDashboard, err := w.Lakeview.Create(ctx, newDashboardRequest)
 			if err != nil {
-				return err
+				if isParentDoesntExistError(err) {
```
mgyucht, line 81: To remove a layer of nesting, we can do:
```go
if err != nil && isParentDoesntExistError(err)
```
since the else case of the outer condition is empty.

mgyucht (approved):
LGTM, one nit.

dashboards/resource_dashboard.go
alexott: thanks - fixed!

PR #3779, opened 2024-07-16
[Exporter] Add support for exporting of Lakeview dashboards

alexott:
## Changes

This adds support for exporting of `databricks_dashboard` resource and dependencies.

Current limitations:

- No support for incremental mode
- List operation fails on big lists (not clear if it's a problem with API or Go SDK)

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -3129,6 +3129,7 @@ var resourcesMap map[string]importable = map[string]importable{
 		Import: func(ic *importContext, r *resource) error {
 			path := r.Data.Get("path").(string)
 			if strings.HasPrefix(path, "/Repos") {
+				ic.emitRepoByPath(path)
 				return nil
 			}
 			parts := strings.Split(path, "/")
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -16,6 +16,7 @@ import (
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/service/catalog"
 	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/databricks-sdk-go/service/dashboards"
 	"github.com/databricks/databricks-sdk-go/service/iam"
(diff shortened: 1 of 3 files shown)


exporter/importables.go
```diff
@@ -3086,4 +3088,95 @@ var resourcesMap map[string]importable = map[string]importable{
...
+				s = s + dname
+			}
+			s = s + "_" + d.Id()
+			return nameNormalizationRegex.ReplaceAllString(s, "_")
+		},
+		Import: func(ic *importContext, r *resource) error {
+			path := r.Data.Get("path").(string)
+			if strings.HasPrefix(path, "/Repos") {
+				return nil
+			}
```
tanmay-db, line 3133: curious -- repos path isn't supported?

exporter/importables.go
```diff
@@ -3086,4 +3088,95 @@ var resourcesMap map[string]importable = map[string]importable{
...
+			fileName, err := ic.saveFileIn("dashboards", name, []byte(r.Data.Get("serialized_dashboard").(string)))
+			if err != nil {
+				return err
+			}
+			r.Data.Set("file_path", fileName)
+			r.Data.Set("serialized_dashboard", "")
+
+			ic.emitPermissionsIfNotIgnored(r, "/dashboards/"+r.ID,
+				"dashboard_"+ic.Importables["databricks_dashboard"].Name(ic, r.Data))
+			parentPath := r.Data.Get("parent_path").(string)
```
tanmay-db, line 3151: we should also return nil if parent path contains repos? We do this in `Ignore` below

tanmay-db (approved):

exporter/importables.go
alexott: It won't get into Import anyway, because it will be skipped in List operation

exporter/importables.go
alexott: It's supported, but we don't export it because it's a part of the repo that is cloned separately

alexott: although you're correct - we need to emit repo as dependency

PR #3788, opened 2024-07-17
Implement basic functionality for StructToSchema

edwardfeng-db:
## Changes

- Baisc functionality for `StructToSchema`, everything is `Optional` now
- Need to do:
  - Implement equivalent for `CustomizableSchema`
  - Make tfsdk structs contain information about optional and computed
  - Handle recursive data types

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/common/reflect_resource_plugin_framework.go
+++ b/common/reflect_resource_plugin_framework.go
@@ -292,6 +292,7 @@ func checkTheStringInForceSendFields(fieldName string, forceSendFields []string)
 }
 
 func pluginFrameworkTypeToSchema(v reflect.Value) map[string]schema.Attribute {
+	scm := map[string]schema.Attribute{}
 	rk := v.Kind()
 	if rk == reflect.Ptr {
 		v = v.Elem()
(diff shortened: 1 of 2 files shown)


common/reflect_resource_plugin_framework.go
```diff
@@ -289,3 +290,105 @@ func checkTheStringInForceSendFields(fieldName string, forceSendFields []string)
...
+				panic("types.Map should never be used in tfsdk structs")
+			default:
+				// If it is a real stuct instead of a tfsdk type, recursively resolve it.
+				elem := typeField.Type
+				sv := reflect.New(elem)
+				nestedScm := pluginFrameworkTypeToSchema(sv)
+				scm[fieldName] = schema.SingleNestedAttribute{Attributes: nestedScm, Optional: true}
+			}
+		} else {
+			panic(fmt.Errorf("unknown type for field: %s", typeField.Name))
```
tanmay-db, line 384: should we also print the type as well here?

common/reflect_resource_plugin_framework.go
```diff
@@ -289,3 +290,105 @@ func checkTheStringInForceSendFields(fieldName string, forceSendFields []string)
 	}
 	return false
 }
+
+func pluginFrameworkTypeToSchema(v reflect.Value) map[string]schema.Attribute {
```
tanmay-db, line 294: This would also need to include aliases and tracking context that we currently have to typeToSchema for sdkv2? 

Example -- we use tracking context for depth:
````
		if tc.depthExceeded(typeField) {
			// Skip the field if recursion depth is over the limit.
			log.Printf("[TRACE] over recursion limit, skipping field: %s, max depth: %d", getNameForType(typeField.Type), tc.getMaxDepthForTypeField(typeField))
			continue
		}
````

and aliases while getting field name:

`````
fieldName := chooseFieldNameWithAliases(typeField, v.Type(), aliases)
`````

tanmay-db (approved):
Had some comments but core logic LGTM.

common/reflect_resource_plugin_framework.go
edwardfeng-db: For tracking context yes, alias no.

Tracking context is coming up as the next step for recursion depth limit.

In the future all of alias information will go directly into the proto layer with annotations, so we don't have to worry about them in tf

PR #3792, opened 2024-07-18
[Feature] Rewriting DLT pipelines using SDK

Divyansh-db:
## Changes
Rewrite DLT pipelines resource using SDK.

## Tests
All existing unit tests and integration tests are passing

- [ ] `make test` run locally
- [ ] using Go SDK


Divyansh-db:
Currently all changes are in a separate directory named `dlt`.  After changes are finalised, they will be shifted to pipelines directory

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+		// The solution here is to look for the Spark configuration to determine
+		// if the user only wants a single node cluster (only master, no workers).
+		if cluster.SparkConf["spark.databricks.cluster.profile"] == "singleNode" {
+			cluster.ForceSendFields = append(cluster.ForceSendFields, "NumWorkers")
+		}
+	}
+}
+
+func (w workspaceClient) Create(ctx context.Context, d *schema.ResourceData, timeout time.Duration) error {
+	var createPipelineRequest createPipelineRequestStruct
```
hectorcast-db, line 79: Don't create the createPipelineRequestStruct. Use the Go `CreatePipeline` directly. Add any necessary field and/or alias to the `Pipeline` you define above.

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+	// ValidateFunc fields
+	s.SchemaPath("channel").SetValidateFunc(validation.StringInSlice([]string{"current", "preview"}, true))
+	s.SchemaPath("edition").SetValidateFunc(validation.StringInSlice([]string{"pro", "core", "advanced"}, true))
+
+	return s
+}
+
+var pipelineSchema = common.StructToSchema(Pipeline{}, nil)
+
+type workspaceClient struct {
```
hectorcast-db, line 279: You want to get rid of this.

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+				}
+				return retry.NonRetryableError(err)
+			}
+			message := fmt.Sprintf("Pipeline %s is in state %s, not yet deleted", id, i.State)
+			log.Printf("[DEBUG] %s", message)
+			return retry.RetryableError(fmt.Errorf(message))
+		})
+}
+
+func (w workspaceClient) waitForState(ctx context.Context, id string, timeout time.Duration, desiredState pipelines.PipelineState) error {
```
hectorcast-db, line 145: The SDK has a `WaitGetPipelineRunning` method. You don't need this function.

dlt/resource_dlt_pipeline.go
Divyansh-db: I think when we do DataToStructPointer we also need Aliases set for the struct that we want to fill. That's why I created a separate struct for create response and created aliases for it

dlt/resource_dlt_pipeline.go
Divyansh-db: I will do that but I am curious what is the shortcoming in defining member function for Workspace client.

hectorcast-db: I am not sure of a shortcoming per se. But we want to keep the code "uniform" so it's easier/faster to understand for new people.

hectorcast-db: Ok, i thought about one. It's not recommended to use external types in an anonymous field. This external library is own by us, so it may be ok.

dlt/resource_dlt_pipeline.go
hectorcast-db: You can do it wihout creating a struct:
https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_catalog.go#L48
Not sure if we want one struct or not. If we keep the `createPipelineRequestStruct`, let's put all methods together so it's more clear why this struct exists.

@edwardfeng-db What do you prefer?

edwardfeng-db: I think I prefer what Divyansh has here, because it is the same pattern with what I did with Clusters and Jobs.

I do agree that there are sharable contents in those `Aliases` functions tho, which we can extract and make more succinct

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+type updatePipelineRequestStruct struct {
+	pipelines.EditPipeline
+}
+
+// Name not keeping twice
+type Pipeline struct {
+	pipelines.PipelineSpec
+	PipelineId          string                              `json:"pipeline_id,omitempty"`
```
hectorcast-db, line 37: Some of these things are computed, correct? I know it's not very well documented, but we can try to mark them as such.

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+		`^dbfs:/pipelines/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
+	res := defaultStorageRegex.MatchString(old)
+	if new == "" && res {
+		log.Printf("[DEBUG] Suppressing diff for %v: platform=%#v config=%#v", k, old, new)
+		return true
+	}
+	return false
+}
+
+func (Pipeline) Aliases() map[string]map[string]string {
```
hectorcast-db, line 180: Let's move this next to the Pipeline definition

dlt/resource_dlt_pipeline.go
hectorcast-db: Also, in particular fot the `pipeline_id`, this is an alias for `id`, correct?

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+	return s
+}
+
+func (updatePipelineRequestStruct) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
+	return s
+}
+
+func (Pipeline) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
+	// Required fields
+	// s.SchemaPath("name").SetRequired()
```
hectorcast-db, line 221: Are those required? Remove any commented code which we don't need.

dlt/resource_dlt_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+			if err != nil {
+				return err
+			}
+			wsClient := workspaceClient{w}
+			err = wsClient.Create(ctx, d, d.Timeout(schema.TimeoutCreate))
+			if err != nil {
+				return err
+			}
+
+			// why is it required
```
hectorcast-db, line 297: I assume because it's useful for customers to have the url of a job (much more than the url of a users). Let's add a note.

dlt/resource_dlt_pipeline.go
Divyansh-db: So basically the get response returns a `pipeline_id` as well as a `spec` object which contains `id`. But they are essentially the same thing

hectorcast-db: Then I think it's better to use the alias option

internal/acceptance/pipeline_test.go
```diff
@@ -94,14 +96,15 @@ func TestAccPipelineResource_CreatePipeline(t *testing.T) {
 }
 
 func TestAccAwsPipelineResource_CreatePipeline(t *testing.T) {
+	pipelineName := fmt.Sprintf("pipeline-acceptance-aws-%s", qa.RandomName())
 	workspaceLevel(t, step{
 		Template: `
 		locals {
-			name = "pipeline-acceptance-aws-{var.RANDOM}"
+			name = "` + pipelineName + `"
```
alexott, line 103: we have sticky random variant as well: `{var.STICKY_RANDOM}`

exporter/importables.go
```diff
@@ -1996,26 +2008,27 @@ var resourcesMap map[string]importable = map[string]importable{
...
 						Resource: "databricks_cluster_policy",
-						ID:       cluster.PolicyID,
+						ID:       cluster.PolicyId,
 					})
 				}
-				ic.emitInitScriptsLegacy(cluster.InitScripts)
-				ic.emitSecretsFromSecretsPathMap(cluster.SparkConf)
+				// ic.emitInitScriptsLegacy(cluster.InitScripts)
+				ic.emitInitScripts(cluster.InitScripts)
+				ic.emitSecretsFromSecretsPathMao(cluster.SparkConf)
```
alexott, line 2031: it should be `emitSecretsFromSecretsPathMap`

pipelines/data_pipelines_test.go
alexott, line 1: Why we comment out tests?

provider/provider.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/databricks/terraform-provider-databricks/mws"
 	"github.com/databricks/terraform-provider-databricks/permissions"
 	"github.com/databricks/terraform-provider-databricks/pipelines"
+
```
alexott, line 35: ? we don't need it

pipelines/data_pipelines_test.go
Divyansh-db: I was working in a seperate folder with same files which were conflicted with this. Corrected it now

exporter/importables.go
```diff
@@ -1943,8 +1945,16 @@ var resourcesMap map[string]importable = map[string]importable{
...
-			pipelinesList, err := api.List(50, "")
+			w, err := ic.Client.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			pipelinesList, err := w.Pipelines.ListPipelinesAll(ic.Context, pipelines.ListPipelinesRequest{
+				MaxResults: 50,
+			})
+
+			// api := pipelines.NewPipelinesAPI(ic.Context, ic.Client)
```
hectorcast-db, line 1956: Remove this

exporter/importables.go
```diff
@@ -1996,25 +2008,26 @@ var resourcesMap map[string]importable = map[string]importable{
...
-				if cluster.PolicyID != "" {
+				if cluster.PolicyId != "" {
 					ic.Emit(&resource{
 						Resource: "databricks_cluster_policy",
-						ID:       cluster.PolicyID,
+						ID:       cluster.PolicyId,
 					})
 				}
-				ic.emitInitScriptsLegacy(cluster.InitScripts)
+				// ic.emitInitScriptsLegacy(cluster.InitScripts)
```
hectorcast-db, line 2029: Remove commented code

pipelines/resource_pipeline.go
```diff
@@ -7,319 +7,188 @@ import (
...
-	err := a.client.Post(a.ctx, "/pipelines", s, &resp)
+	createdPipeline, err := w.Pipelines.Create(ctx, createPipelineRequest.CreatePipeline)
 	if err != nil {
-		return "", err
+		return err
 	}
-	id := resp.PipelineID
-	err = a.waitForState(id, timeout, StateRunning)
+	var id string
+	if d.Get("dry_run").(bool) {
```
hectorcast-db, line 47: Can you add an explanation and link to the docs on why this is like this?

pipelines/resource_pipeline.go
```diff
@@ -7,319 +7,188 @@ import (
...
+func Update(w *databricks.WorkspaceClient, ctx context.Context, d *schema.ResourceData, timeout time.Duration) error {
+	var updatePipelineRequest updatePipelineRequestStruct
+	common.DataToStructPointer(d, pipelineSchema, &updatePipelineRequest)
+	updatePipelineRequest.EditPipeline.PipelineId = d.Id()
+	adjustForceSendFields(&updatePipelineRequest.Clusters)
 
-func (a PipelinesAPI) Update(id string, s PipelineSpec, timeout time.Duration) error {
-	adjustForceSendFields(&s)
-	err := a.client.Put(a.ctx, "/pipelines/"+id, s)
+	err := w.Pipelines.Update(ctx, updatePipelineRequest.EditPipeline)
```
hectorcast-db, line 81: Does `dry_run` logic applies here? What if the user changes the `dry_run` option?

Divyansh-db: The `dry_run` doesn't affect the update call

alexott: but really - do we need to support `dry_run` at all?

hectorcast-db: I just learned what it is (I was asking above). And I don't think we need to support it.

pipelines/resource_pipeline.go
hectorcast-db: Based on the documentation of `dry_run`, I don't think we need to support it.

edwardfeng-db (approved):
LGTM, can you document the newly added fields and features also in the PR description? I think we need to mention those in the release note

PR #3813, opened 2024-07-23
Make plugin framework infra functions work for Enum types

edwardfeng-db:
## Changes

- Make the two converter functions + StructToSchema work for Enum types
- We do not need to use `types.String` for Enum types as they do not have a force send zero value problem
- Adding a unit test test case

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott:
Do we have only string enum types in SDK? Or we may have int-based enums as well?

edwardfeng-db:
@alexott Based on the codegen template I only see strings

alexott (approved):
lgtm

PR #3820, opened 2024-07-25
[Feature] Notification Destination resource

Divyansh-db:
## Changes
Made a new Notification Destination resource

- [x] using Go SDK


settings/resource_notification_destination.go
```diff
@@ -0,0 +1,117 @@
...
+			return nil
+		},
+		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			readND, err := w.NotificationDestinations.Get(ctx, settings.GetNotificationDestinationRequest{
+				Id: d.Id(),
+			})
```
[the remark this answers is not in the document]
Divyansh-db: I understand your points but I think both cases are a little different.
-> in the case you mentioned, sensitive fields are supposed to be returned by the `get` call
-> In out case, Only the boolean confirmations of whether the fields are set or not are sent (no sensitive info)

On a side note, while I was looking at the mlflow_webhook resource, I could not find where did they specify the `access_tokens` as sensitive? It's not done in the SDK and neither in provider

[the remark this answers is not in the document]
Divyansh-db: It's a bit more complicated because now if we think of it, even if user changes anything in configuration, it will not be detected (as no useful information is returned in `get`). So, we will need to use `md5` as we used in dashboard resource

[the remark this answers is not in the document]
alexott: regarding mlfow webhook - most probably it should be fixed.

regarding the Read operation - the Get won't bring the data back, and as result, StructToData will fill the parameters with empty values, so when you run `terraform plan` after apply, it should show you the diff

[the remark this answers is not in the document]
alexott: This is what I'm talking about - this is after I did `apply`, but `plan` still shows changes:

```
  # databricks_notification_destination.slack will be updated in-place
  ~ resource "databricks_notification_destination" "slack" {
        id               = "acf3f051-d075-42f6-99b0-e87768045030"
        # (2 unchanged attributes hidden)

      ~ config {
          ~ slack {
              + url     = "https://hooks.slack.com/services/..."
                # (1 unchanged attribute hidden)
            }
        }
    }

```

alexott (changes requested):
API doesn't return sensitive values, so we need to handle it similarly to other resources (like, MLflow webhook or connections)

settings/resource_notification_destination.go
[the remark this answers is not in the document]
Divyansh-db: Got it. Also, can you tell me how can I test it against any web hooks so that I can check for the things locally

alexott:
For testing, you just need to provide a base URL; I was just using `"https://hooks.slack.com/services/..."` and it worked

docs/resources/notification_destination.md
```diff
@@ -0,0 +1,99 @@
...
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `display_name` - (Required) The display name of the Notification Destination.
+* `config` - (Required) The configuration of the Notification Destination. It must be one of the following:
```
nkvuong, line 79:
```suggestion
* `config` - (Required) The configuration of the Notification Destination. It must contain exactly one of the following blocks:
```

internal/acceptance/notification_destination_test.go
```diff
@@ -0,0 +1,44 @@
...
+	display_name := "Email Notification Destination - " + qa.RandomName()
+	workspaceLevel(t, step{
+		Template: `
+		resource "databricks_notification_destination" "this" {
+			display_name = "` + display_name + `"
+			config {
+				email {
+					addresses = ["` + qa.RandomEmail() + `"]
+				}
+			}
```
alexott, line 24: It would be really nice to add test cases for all supported types - to catch configuration drift, etc.

Divyansh-db:
I could not figure out on how to set the values of nested fields (fields inside the structs) directly through d.Set(...). I could not find similar integration tests which test setting nested fields (There is no integration test for mlflow_webhook as well). Used a way around to get things done

settings/resource_notification_destination.go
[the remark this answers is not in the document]
Divyansh-db: I tried doing the same thing but in step 3, I would require to set (say, `slack_url`) a nested field. For that, I tried doing
d.Set("config.0.slack.0.url", <name_of_temp_struct_in_step_1>.Config.SlackConfig.Url)
but it cannot be set like this, it throws the following error
[ERROR] setting state: config.0.slack.0.url: can only set full list
I also tried doing:
d.Set("config.0.slack", <name_of_temp_struct_in_step_1>.Config.SlackConfig), it also throws the same error.
Even I cannot set "config" directly.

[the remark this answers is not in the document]
alexott: You have a structure that you already read via API - just fill missing fields in it, and that's all - you don't need to use `d.Set`

internal/acceptance/notification_destination_test.go
```diff
@@ -0,0 +1,304 @@
...
+				Id: id,
+			})
+			if err != nil {
+				return err
+			}
+			assert.Equal(t, settings.DestinationType("EMAIL"), ndResource.DestinationType)
+			assert.Equal(t, display_name, ndResource.DisplayName)
+			require.NoError(t, err)
+			return nil
+		}),
```
alexott, line 53: We can move this check into the reusable function, reducing the amount of the code

settings/resource_notification_destination.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/settings"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func setStruct(s *settings.NotificationDestination, readND *settings.NotificationDestination) {
+	switch readND.DestinationType {
+	case settings.DestinationTypeSlack:
+		readND.Config.Slack.Url = s.Config.Slack.Url
```
alexott, line 15: we need to add a check that `s.Config.Slack` (and other blocks) aren't null - when we importing the resource, source data doesn't exist, so it will crash with null pointer

settings/resource_notification_destination.go
```diff
@@ -0,0 +1,177 @@
...
+
+func Create(ctx context.Context, d *schema.ResourceData, w *databricks.WorkspaceClient) error {
+	var newNDrequest settings.CreateNotificationDestinationRequest
+	common.DataToStructPointer(d, ndSchema, &newNDrequest)
+	createdND, err := w.NotificationDestinations.Create(ctx, newNDrequest)
+	if err != nil {
+		return err
+	}
+	d.SetId(createdND.Id)
+	d.Set("destination_type", createdND.DestinationType)
```
alexott, line 41: Technically, it's done by the Read method

settings/resource_notification_destination.go
```diff
@@ -0,0 +1,177 @@
...
+		Id: d.Id(),
+	})
+	if err != nil {
+		return err
+	}
+	setStruct(&tempND, readND)
+	return common.StructToData(readND, ndSchema, d)
+}
+
+func detectConfigTypeChange(d *schema.ResourceData) bool {
```
alexott, line 59: I'm not sure that we need this function. Also, it doesn't cover changes in `display_name`

settings/resource_notification_destination.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func Update(ctx context.Context, d *schema.ResourceData, w *databricks.WorkspaceClient) error {
+	if detectConfigTypeChange(d) {
+		err := Delete(ctx, d, w)
+		if err != nil {
+			return err
+		}
+		return Create(ctx, d, w)
+	}
```
alexott, line 84: According to API it should be possible to update it inline: https://docs.databricks.com/api/workspace/notificationdestinations/update - the main problem is that if we do delete/create, then ID of destination will change and this may break other integrations

alexott (changes requested):

settings/resource_notification_destination.go
Divyansh-db: That is true, but the update API does not allow to change the type of notification destination. I created an integration test `TestAccConfigTypeChange` for this exact puspose. If we do not delete and create a new resouce and try to update  it instead, the update API response is `bad request` and states that you can't change config types using PATCH call

settings/resource_notification_destination.go
Divyansh-db: This function only covers the fields which would require this resource to be deleted and recreated. The `display_name` can be change using the update call

settings/resource_notification_destination.go
alexott: We need to raise an internal ticket to update the OpenAPI spec to reflect this - right now, it's not documented.  But it's better first to confirm that this is intended behavior. 

Instead of implicitly recreating the resource, we may need to mark these blocks as `force_new`, so Terraform will delete/create automatically. We need to do this because Terraform needs to know that `id` is changing, and this change is propagated to the dependent resources. Otherwise, they could be broken after such an update. 

We also need to document that behavior.

Divyansh-db: 1) which team should I contact for this.
2) I will try to make fields force new rather than explicitly doing it.

Divyansh-db: Also, is raising the ticket a blocking change for this PR?

internal/acceptance/notification_destination_test.go
```diff
@@ -0,0 +1,217 @@
...
+		Id: id,
+	})
+	if err != nil {
+		return err
+	}
+	assert.Equal(t, config_type, ndResource.DestinationType)
+	assert.Equal(t, display_name, ndResource.DisplayName)
+	require.NoError(t, err)
+	return nil
+}
```
alexott, line 29: It's not critical, but yon can further simplify the code by returning a function, so instead of

```
		Check: resourceCheck("databricks_notification_destination.this", func(ctx context.Context, client *common.DatabricksClient, id string) error {
			return check(t, ctx, client, id, display_name, settings.DestinationTypeEmail)
		}),
```

you can write something like:

```
Check: makeCheckFunction(dsetinationType, displayName)
```

docs/resources/notification_destination.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - The unique ID of the Notification Destination.
+
+
+## Note
+* If the type of notification destination is changed, the existing notification destination will be deleted and a new notification destination will be created with the new type.
```
alexott, line 102: It's better to move this note close to the arguments reference, and use this syntax to render it as a block:

```
-> **Note** If the type of notification destination is changed ...
```

alexott (approved):
Code looks good, also tested it myself. Just few small comments

PR #3821, opened 2024-07-25
[Exporter] Improve exporting of `databricks_model_serving`

alexott:
## Changes

Changes include:

- emitting references to UC models
- emitting references to UC catalog and schema for auto-capture
- emitting secrets from environment variables and external models configuration
- generation of some required fields
- expanding test coverage

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/importables.go
+++ b/exporter/importables.go
@@ -20,6 +20,7 @@ import (
 	"github.com/databricks/databricks-sdk-go/service/iam"
 	sdk_jobs "github.com/databricks/databricks-sdk-go/service/jobs"
 	"github.com/databricks/databricks-sdk-go/service/ml"
+	"github.com/databricks/databricks-sdk-go/service/serving"
 	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/databricks-sdk-go/service/sharing"
 	"github.com/databricks/databricks-sdk-go/service/sql"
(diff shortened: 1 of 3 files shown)


exporter/importables.go
```diff
@@ -2160,16 +2162,96 @@ var resourcesMap map[string]importable = map[string]importable{
...
-				(strings.HasPrefix(pathString, "config.0.served_models.") &&
-					strings.HasSuffix(pathString, ".scale_to_zero_enabled")) {
+			if pathString == "config.0.traffic_config" || pathString == "config.0.auto_capture_config.0.enabled" ||
+				(pathString == "config.0.auto_capture_config.0.table_name_prefix" && d.Get(pathString).(string) != "") {
 				return false
 			}
+			if res := servedEntityFieldExtractionRegex.FindStringSubmatch(pathString); res != nil {
+				field := res[2]
+				log.Printf("[DEBUG] ShouldOmitField: extracted field from %s: '%s'", pathString, field)
+				switch field {
```
tanmay-db, line 2232: should we have a default case here?

exporter/importables.go
```diff
@@ -2160,16 +2162,96 @@ var resourcesMap map[string]importable = map[string]importable{
...
+				case "scale_to_zero_enabled", "name":
+					return false
+				case "workload_size", "workload_type":
+					return d.Get(pathString).(string) == ""
+				}
+			}
 			return defaultShouldOmitFieldFunc(ic, pathString, as, d)
 		},
+		ShouldGenerateField: func(ic *importContext, pathString string, as *schema.Schema, d *schema.ResourceData) bool {
+			// We need to generate some fields even if they have zero value...
```
tanmay-db, line 2242: For my understanding, why do we want to generate these fields? Also, are there other fields for which we would like to do this?

tanmay-db (approved):

exporter/importables.go
alexott: Default case could be also complex, so it's better to pass through and allow the last line of the function (call to the `defaultShouldOmitFieldFunc`) handle it

PR #3831, opened 2024-07-29
[Internal] refactored `databricks_mws_permission_assignment` to Go SDK

nkvuong:
## Changes
- refactored `databricks_mws_permission_assignment` to Go SDK
- added `TestUcAccAssignGroupToWorkspace` to flaky tests

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/mws/resource_mws_permission_assignment.go
+++ b/mws/resource_mws_permission_assignment.go
@@ -2,92 +2,38 @@ package mws
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/iam"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func NewPermissionAssignmentAPI(ctx context.Context, m any) PermissionAssignmentAPI {
-	return PermissionAssignmentAPI{m.(*common.DatabricksClient), ctx}
-}
-
-type PermissionAssignmentAPI struct {
-	client  *common.DatabricksClient
-	context context.Context
-}
-
-type Permissions struct {
-	Permissions []string `json:"permissions"`
-}
-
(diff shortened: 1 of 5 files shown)


mws/resource_mws_permission_assignment.go
```diff
@@ -2,131 +2,88 @@ package mws
...
 func ResourceMwsPermissionAssignment() common.Resource {
-	type entity struct {
-		WorkspaceId int64    `json:"workspace_id"`
-		PrincipalId int64    `json:"principal_id"`
-		Permissions []string `json:"permissions" tf:"slice_as_set"`
-	}
-	s := common.StructToSchema(entity{},
-		common.NoCustomize)
+	s := common.StructToSchema(iam.UpdateWorkspaceAssignments{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
```
alexott, line 25: Should we modify type to be a Set?

mws/resource_mws_permission_assignment.go
```diff
@@ -2,131 +2,88 @@ package mws
...
+			list, err := acc.WorkspaceAssignment.ListByWorkspaceId(ctx, common.MustInt64(workspaceId))
 			if err != nil {
 				return err
 			}
-			permissions, err := list.ForPrincipal(common.MustInt64(principalId))
+			permissions, err := getPermissionsByPrincipal(*list, common.MustInt64(principalId))
 			if err != nil {
 				return err
 			}
 			return common.StructToData(permissions, s, d)
```
alexott, line 75: We need to set workspace ID and principal ID explicitly - otherwise, import won't work correctly

alexott (approved):

alexott:
Hmmm, it looks like integration test is failing... Not sure if it's because of the eventual consistency or not

nkvuong:
@alexott it is to do with the creation of a new group, which has eventual inconsistency itself. Maybe changing the integration test to a service principal would help?

alexott:
Yes, I don't remember that we had problems with users/SPNs

mgyucht (approved):

PR #3852, opened 2024-08-04
Make tfsdk struct enums into types.String

edwardfeng-db:
## Changes

- These enums have to be strings instead of the indirect types that they used to be, otherwise `.get` and `.set` does not work properly
- First part of making the acceptance test work as specified here https://github.com/databricks/terraform-provider-databricks/pull/3841

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/.codegen/model.go.tmpl
+++ b/.codegen/model.go.tmpl
@@ -79,7 +79,7 @@ func (f *{{.PascalName}}) Type() string {
 	{{- else if .ArrayValue }}[]{{template "type" .ArrayValue}}
 	{{- else if .MapValue }}map[string]{{template "type" .MapValue}}
 	{{- else if .IsObject }}{{.PascalName}}
-	{{- else if .Enum }}{{.PascalName}}
+	{{- else if .Enum }}types.String
 	{{- else}}any /* MISSING TYPE */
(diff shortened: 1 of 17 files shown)


.codegen/model.go.tmpl
```diff
@@ -79,7 +79,7 @@ func (f *{{.PascalName}}) Type() string {
 	{{- else if .ArrayValue }}[]{{template "type" .ArrayValue}}
 	{{- else if .MapValue }}map[string]{{template "type" .MapValue}}
 	{{- else if .IsObject }}{{.PascalName}}
-	{{- else if .Enum }}{{.PascalName}}
+	{{- else if .Enum }}types.String
```
tanmay-db, line 82: Just to confirm -- are all enums always strings? or can some be int as well? I took a look at few of them and those seem to be string but just want to confirm if we have a constraint somewhere that enums are strings only.

tanmay-db (approved):

.codegen/model.go.tmpl
edwardfeng-db: from looking at the codegen template for the gosdk structs, yes, that seems to be the case

PR #3853, opened 2024-08-04
Update resource quality monitor in plugin framework

edwardfeng-db:
## Changes

- Part of making acceptance test pass: https://github.com/databricks/terraform-provider-databricks/pull/3841
- Update `ResourceQualityMonitor` to make it pass acceptance test
  - Commented out adding two new fields temporarily because we need to have those fields in the tfsdk struct, otherwise `.get()` and `.set()` don't work
  - Updated the target struct for `.get` calls to be `MonitorInfo` because terraform requires that to be consistent with the schema
  - Added logic at the end of `Create` and `Update` to read the latest `monitorInfo` and set the state

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/pluginframework/resource_quality_monitor.go
+++ b/pluginframework/resource_quality_monitor.go
@@ -3,7 +3,6 @@ package pluginframework
 import (
 	"context"
 	"fmt"
-	"go/types"
 	"time"
 
 	"github.com/databricks/databricks-sdk-go"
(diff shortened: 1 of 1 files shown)


pluginframework/resource_quality_monitor.go
```diff
@@ -150,15 +168,21 @@ func (r *QualityMonitorResource) Update(ctx context.Context, req resource.Update
...
-	var updateMonitorTfSDK catalog_tf.UpdateMonitor
-	diags := req.Plan.Get(ctx, &updateMonitorTfSDK)
+	var monitorInfoTfSDK catalog_tf.MonitorInfo
+	diags := req.Plan.Get(ctx, &monitorInfoTfSDK)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// Plan is not adding `dashboard_id`, but it is in the state.
+	diags = req.State.GetAttribute(ctx, path.Root("dashboard_id"), &monitorInfoTfSDK.DashboardId)
```
edwardfeng-db, line 178: Not sure why this is the case, it seems like we need to manually add it as it is only in the state

tanmay-db (approved):

PR #3854, opened 2024-08-04
Update reflect_resource_plugin_framework to make acceptance test pass

edwardfeng-db:
## Changes

- Subpart of https://github.com/databricks/terraform-provider-databricks/pull/3841
- Updated reflect_resource_plugin_framework to handle cases
  - When a field in the given source struct is nil, the converter function has to be able to handle it
  - Handle enums in tfsdk structs as types.String
  - When a field does not exist on the destination struct, ignore the field
- Added unit test for the cases mentioned above

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/reflect_resource_plugin_framework.go
+++ b/common/reflect_resource_plugin_framework.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"log"
 	"reflect"
 	"strings"
 
@@ -24,7 +25,7 @@ func TfSdkToGoSdkStruct(tfsdk interface{}, gosdk interface{}, ctx context.Contex
 	}
 	destVal = destVal.Elem()
 
-	if srcVal.Kind() != reflect.Struct || destVal.Kind() != reflect.Struct {
+	if srcVal.Kind() != reflect.Struct {
 		panic("input should be structs")
 	}
 
@@ -51,7 +52,8 @@ func TfSdkToGoSdkStruct(tfsdk interface{}, gosdk interface{}, ctx context.Contex
 func tfSdkToGoSdkSingleField(srcField reflect.Value, destField reflect.Value, srcFieldName string, forceSendFieldsField *reflect.Value, ctx context.Context) error {
 
 	if !destField.IsValid() {
-		panic(fmt.Errorf("destination field is not valid: %s", destField.Type().Name()))
+		// Skip field that destination struct does not have.
+		return nil
 	}
 
 	if !destField.CanSet() {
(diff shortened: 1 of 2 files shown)


common/reflect_resource_plugin_framework.go
```diff
@@ -24,20 +25,20 @@ func TfSdkToGoSdkStruct(tfsdk interface{}, gosdk interface{}, ctx context.Contex
 	}
 	destVal = destVal.Elem()
 
-	if srcVal.Kind() != reflect.Struct || destVal.Kind() != reflect.Struct {
+	if srcVal.Kind() != reflect.Struct {
```
tanmay-db, line 28: This is for the invalid type right? Do we know where that is coming from?

common/reflect_resource_plugin_framework.go
```diff
@@ -62,6 +64,10 @@ func tfSdkToGoSdkSingleField(srcField reflect.Value, destField reflect.Value, sr
 	if srcFieldValue == nil {
 		return nil
 	} else if srcField.Kind() == reflect.Ptr {
+		if srcField.IsNil() {
+			// Skip nils
```
tanmay-db, line 68: Should we have destination field as nil as well or is skipping leads to the same result?

common/reflect_resource_plugin_framework.go
```diff
@@ -87,21 +93,52 @@ func tfSdkToGoSdkSingleField(srcField reflect.Value, destField reflect.Value, sr
...
+					addToForceSendFields(srcFieldName, forceSendFieldsField)
+				}
 			}
 		case types.List:
 			panic("types.List should never be used, use go native slices instead")
 		case types.Map:
 			panic("types.Map should never be used, use go native maps instead")
 		default:
+			if srcField.IsZero() {
+				// Skip nils
```
tanmay-db, line 129: Skip Zeros probably would be better as comment as we are checking for zero values.

tanmay-db (approved):

common/reflect_resource_plugin_framework.go
edwardfeng-db: Yeah, sometimes destVal doesn't exist, but maybe I can make this check a bit more robust

common/reflect_resource_plugin_framework.go
edwardfeng-db: Skipping leads to the same result

common/reflect_resource_plugin_framework.go
edwardfeng-db: Yeah, good call, let me update it

PR #3867, opened 2024-08-08
Update init test and migrate imports for plugin framework

edwardfeng-db:
## Changes

- Part of https://github.com/databricks/terraform-provider-databricks/pull/3841
- Migrated imports following the documentation [here](https://developer.hashicorp.com/terraform/plugin/testing/migrating)
  - 
image

- Made `run` in `init_test` able to take both plugin framework resources as well as sdkv2 resources

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/acceptance/init_test.go
+++ b/internal/acceptance/init_test.go
@@ -24,9 +24,7 @@ import (
 	dbproviderlogger "github.com/databricks/terraform-provider-databricks/logger"
 	"github.com/databricks/terraform-provider-databricks/provider"
 	"github.com/databricks/terraform-provider-databricks/qa"
-	"github.com/hashicorp/terraform-plugin-framework/providerserver"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
-	"github.com/hashicorp/terraform-plugin-mux/tf5to6server"
 	"github.com/hashicorp/terraform-plugin-mux/tf6muxserver"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	terraform_sdk_v2 "github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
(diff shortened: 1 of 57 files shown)


internal/acceptance/init_test.go
```diff
@@ -138,7 +143,38 @@ func run(t *testing.T, steps []step) {
...
+				},
+			}
+
+			muxServer, err := tf6muxserver.NewMuxServer(ctx, providers...)
+
+			if err != nil {
+				return nil, err
+			}
+
+			return muxServer.ProviderServer(), nil
```
mgyucht, line 174: There should be one definition of this shared by both the main provider and integration tests. Can we combine these together?

internal/acceptance/init_test.go
```diff
@@ -207,10 +242,16 @@ func run(t *testing.T, steps []step) {
...
+
+					// Convert InstanceState from terraform-plugin-framework to terraform-plugin-sdk v2
+					sdkInstanceState := convertToSDKInstanceState(is.Primary)
+
+					if r, ok := sdkPluginProvider.ResourcesMap[p[0]]; ok {
+						// Only checking for sdkv2 resources.
+						dia := r.ReadContext(ctx, r.Data(sdkInstanceState), client)
+						if dia != nil {
+							return fmt.Errorf("%v", dia)
+						}
```
mgyucht, line 254: I'm wondering if we really need this at all, even for SDK-based resources. If the test framework refreshes resources to read them again, then I think it is actually OK to remove this entirely. What do you think?

edwardfeng-db: I think maybe for now let's keep it just to minimize the amount of changes we introduce?

internal/acceptance/init_test.go
edwardfeng-db: Yeah good point, I can update it

edwardfeng-db: Updated, PTAL

mgyucht (approved):
LGTM!

PR #3874, opened 2024-08-09
[Feature] Add `databricks_mlflow_models` data source

### LINKED ISSUE #3791: [FEATURE] Add data source support for databricks_mlflow_models
### Use-cases
Currently there is only a data source for a singular databricks_mlflow_model which requires you to have a name already, but this means that we cannot filter our local variables by only existing model names due to "resource attributes that cannot be determined until apply" when filtering based on `try(data.databricks_mlflow_model.dataproduct_mlflow_models[key].id, null) != null`. We want to validate that mlflow registered models exist at plan time so that it does not throw an error during the apply an is instead caught in the PR plan generation stage, and will not throw an error during apply.

### Attempted Solutions
We have the option to pull via the mastercard/rest from [databricks mlflow list models](https://docs.databricks.com/api/workspace/modelregistry/listmodels), or to manually validate that nobody is adding missing models at PR-time, but native support would be better. 

### Proposal
Create a data source `databricks_mlflow_models` which returns a list of valid model names, as is the case for the `databricks_external_locations` data source.

### References
No existing reference

840:
## Changes
Add `databricks_mlflow_models` data source

## Tests
Add integration and acceptance test

- [x] `make test` run locally
- [x] using Go SDK

Resolves #3791

### DIFF
--- a/docs/data-sources/mlflow_models.md
+++ b/docs/data-sources/mlflow_models.md
@@ -17,6 +17,24 @@ output "model" {
 }
 ```
 
+```hcl
+data "databricks_mlflow_models" "this" {}
+
+check "model_list_not_empty" {
+  assert {
+    condition = length(data.databricks_mlflow_models.this.names) != 0
+    error_message = "Model list is empty."
+  }
+}
+
+check "model_list_contains_model" {
+  assert {
+    condition = contains(data.databricks_mlflow_models.this.names, "model_1")
+    error_message = "model_1 is missing in model list."
+  }
+}
+```
+
 ## Attribute Reference
 
 This data source exports the following attributes:
--- a/docs/data-sources/mlflow_models.md
+++ b/docs/data-sources/mlflow_models.md
@@ -1,0 +1,24 @@
+---
+subcategory: "MLflow"
+---
+# databricks_mlflow_models Data Source
+
+-> **Note** This data source could be only used with workspace-level provider!
+
+Retrieves a list of [databricks_mlflow_model](../resources/mlflow_model.md) objects, that were created by Terraform or manually, so that special handling could be applied.
+
+## Example Usage
+
+```hcl
+data "databricks_mlflow_models" "this" {}
+
+output "model" {
+  value = data.databricks_mlflow_models.this
+}
+```
+
+## Attribute Reference
+
+This data source exports the following attributes:
+
+* `names` - List of names of [databricks_mlflow_model](./mlflow_model.md)
--- a/internal/acceptance/data_mlflow_models_test.go
+++ b/internal/acceptance/data_mlflow_models_test.go
@@ -8,8 +8,8 @@ import (
 )
 
 func TestAccDataMlflowModels(t *testing.T) {
-	workspaceLevel(t,
-		step{
+	WorkspaceLevel(t,
+		Step{
 			Template: `
 			resource "databricks_mlflow_model" "this" {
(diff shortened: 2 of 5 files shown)


docs/data-sources/mlflow_models.md
```diff
@@ -0,0 +1,24 @@
...
+
+Retrieves a list of [databricks_mlflow_model](../resources/mlflow_model.md) objects, that were created by Terraform or manually, so that special handling could be applied.
+
+## Example Usage
+
+```hcl
+data "databricks_mlflow_models" "this" {}
+
+output "model" {
+  value = data.databricks_mlflow_models.this
```
nkvuong, line 16: should we show a more realistic example, like checking if a model exists using `check` block?

e.g. https://developer.hashicorp.com/terraform/language/checks

internal/acceptance/data_mlflow_models_test.go
```diff
@@ -0,0 +1,26 @@
...
+		step{
+			Template: `data "databricks_mlflow_models" "this" {}`,
+			Check: func(s *terraform.State) error {
+				r, ok := s.RootModule().Resources["data.databricks_mlflow_models.this"]
+				if !ok {
+					return fmt.Errorf("data not found in state")
+				}
+				names := r.Primary.Attributes["names.#"]
+				if names == "" {
+					return fmt.Errorf("names are empty: %v", r.Primary.Attributes)
```
nkvuong, line 21: this assumes we have at least a single model in our integration test environment, which may not be correct

mlflow/data_mlflow_models_test.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/service/ml"
+	"github.com/databricks/terraform-provider-databricks/qa"
+)
+
+func TestDataSourceModels(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
```
nkvuong, line 12: can we use Mocks instead of Fixtures? there should be quite a few existing examples

docs/data-sources/mlflow_models.md
840: Done

mlflow/data_mlflow_models_test.go
840: Done

internal/acceptance/data_mlflow_models_test.go
840: Done

mlflow/data_mlflow_models_test.go
840: Changed to Mocks

internal/acceptance/data_mlflow_models_test.go
840: The test will now create a model to guarantee a model exists

internal/acceptance/data_mlflow_models_test.go
```diff
@@ -0,0 +1,61 @@
...
+			  tags {
+				key   = "key1"
+				value = "value1"
+			  }
+			  tags {
+				key   = "key2"
+				value = "value2"
+			  }
+			}`,
+		},
```
alexott, line 28: Do we really need this step? Why can't we do it one step?

alexott (approved):

PR #3880, opened 2024-08-09
Make reflect_resource and customizable_schema work for data source

edwardfeng-db:
## Changes

- Currently these util functions only work for provider resources not data sources, this one adds support for both

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


common/reflect_resource_plugin_framework.go
```diff
@@ -503,3 +610,14 @@ func PluginFrameworkResourceStructToSchemaMap(v any, customizeSchema func(Custom
...
+func PluginFrameworkDataSourceStructToSchemaMap(v any, customizeSchema func(CustomizableSchemaPluginFrameworkDataSource) CustomizableSchemaPluginFrameworkDataSource) map[string]dataschema.Attribute {
+	attributes := pluginFrameworkDataSourceTypeToSchema(reflect.ValueOf(v))
+
+	if customizeSchema != nil {
+		cs := customizeSchema(*ConstructCustomizableSchemaDataSource(attributes))
+		return cs.ToAttributeMap()
+	} else {
+		return attributes
+	}
+}
```
edwardfeng-db, line 623: I am not sure if duplicating everything is a good idea. But the other option is to create if else branches in a single function, might also be pretty ugly.

common/databricks_schema_attribute.go
```diff
@@ -0,0 +1,571 @@
...
+	a.Required = false
+	return a
+}
+
+func (a StringAttribute) SetDeprecated(msg string) Attribute {
+	a.DeprecationMessage = msg
+	return a
+}
+
+func (a StringAttribute) AddValidators(v any) Attribute {
```
mgyucht, line 75: Can we use expected type of the validator here instead of `any`?

common/databricks_schema_attribute.go
```diff
@@ -0,0 +1,571 @@
...
+type Attribute interface {
+	ToDataSourceAttribute() dataschema.Attribute
+	ToResourceAttribute() schema.Attribute
+	SetOptional() Attribute
+	SetRequired() Attribute
+	SetSensitive() Attribute
+	SetComputed() Attribute
+	SetReadOnly() Attribute
+	SetDeprecated(string) Attribute
+	AddValidators(any) Attribute
```
mgyucht, line 21: Remove this since the types will be disparate

common/databricks_schema_attribute.go
```diff
@@ -0,0 +1,571 @@
...
+}
+
+type StringAttribute struct {
```
mgyucht, line 24: Each attribute definition should be in its own file.

common/databricks_schema_attribute.go
```diff
@@ -0,0 +1,571 @@
...
+
+func (a StringAttribute) SetDeprecated(msg string) Attribute {
+	a.DeprecationMessage = msg
+	return a
+}
+
+func (a StringAttribute) AddValidators(v any) Attribute {
+	a.Validators = append(a.Validators, v.(validator.String))
+	return a
+}
```
mgyucht, line 78: For types that implement an interface:
```
var _ InterfaceType = ConcreteType{}
e.g.
var _ Attribute = StringAttribute{}
```

common/databricks_schema_attribute.go
```diff
@@ -0,0 +1,571 @@
...
+	a.DeprecationMessage = msg
+	return a
+}
+
+func (a BoolAttribute) AddValidators(v any) Attribute {
+	a.Validators = append(a.Validators, v.(validator.Bool))
+	return a
+}
+
+type MapAttribute struct {
```
mgyucht, line 248: For this type and the ones below, doccomments

mgyucht:
Replaced by #3922 and #3927, closing.

PR #3901, opened 2024-08-14
[Exporter] Better support for notebooks with /Workspace path

### LINKED ISSUE #3195: Exporter: Investigate effect of using `/Workspace` path everywhere

alexott:
## Changes

Jobs, DLT and other services are switching to use paths starting /Workspace for all objects (notebooks, ...), not only for workspace files.  This PR improves handling of such objects in exporter.

Resolves #3195

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/exporter_test.go
+++ b/exporter/exporter_test.go
@@ -1160,7 +1160,7 @@ func TestImportingJobs_JobList(t *testing.T) {
 							},
 						},
 						NotebookTask: &jobs.NotebookTask{
-							NotebookPath: "/Test",
+							NotebookPath: "/Workspace/Test",
 						},
 						PipelineTask: &jobs.PipelineTask{
 							PipelineID: "123",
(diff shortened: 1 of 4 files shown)


mgyucht (approved):
Approval for exporter change

PR #3909, opened 2024-08-15
[Fix] Handle cluster deletion in `databricks_library` read

### LINKED ISSUE #3679: [ISSUE] Issue with `databricks_library` resource when cluster is automatically deleted
### Configuration
```hcl
  provider      = databricks.workspace
  cluster_name  = var.cluster_name
  # ...
}

  provider   = databricks.workspace

  cluster_id = databricks_cluster.this.cluster_id
  # ...
}

resource "databricks_library" "cluster-library" {
  provider  = databricks.workspace
  for_each = var.cluster_libraries

  cluster_id = databricks_cluster.this.cluster_id

  # ...
}
```

### Expected Behavior
When a cluster is deleted automatically after being terminated for more than 30 days (databricks [default](https://docs.databricks.com/en/compute/clusters-manage.html#cluster-terminate) behaviour) the next run of a terraform plan should recreate the cluster and all of the libraries attached

### Actual Behavior
The plan does detect that the cluster and the permissions were deleted and need to be recreated but fails with the `databricks_library` resource because the old cluster no longer exists:
 
```
Error: cannot read library: Cluster xxxx does not exist
```

### Steps to Reproduce

1. Create a cluster  with the `databricks_cluster` resource and add libraries to it using the `databricks_library` resource
2. Permanently delete the cluster
3. Run a tf plan again

```
Terraform v1.7.4
+ provider registry.terraform.io/databricks/databricks v1.47.0
+ provider registry.terraform.io/hashicorp/aws v4.33.0
```

### Is it a regression?
Yes, from `<= v1.39` this worked correctly, i.e. all the resources (cluster, permissions and libraries) were recreated after they were permanently deleted

### Debug Output
NA

### Important Factoids
NA

No

alexott:
## Changes

We now don't return an error when reading a library that belongs to the cluster deleted outside of the Terraform.

Resolves #3679

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"crypto/md5"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -11,6 +12,8 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+
+	"github.com/databricks/databricks-sdk-go/apierr"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -85,4 +88,19 @@ func ReadSerializedJsonContent(jsonStr, filePath string) (serJSON string, md5Has
 	}
 	md5Hash = CalculateMd5Hash(content)
 	return string(content), md5Hash, nil
+}
+
+// Suppress the error if it is 404
+func IgnoreNotFoundError(err error) error {
+	var apiErr *apierr.APIError
+	if !errors.As(err, &apiErr) {
+		return err
+	}
+	if apiErr.StatusCode == 404 {
+		return nil
+	}
+	if strings.Contains(apiErr.Message, "does not exist") {
+		return nil
+	}
+	return err
 }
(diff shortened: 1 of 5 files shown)


common/util.go
```diff
@@ -86,3 +89,18 @@ func ReadSerializedJsonContent(jsonStr, filePath string) (serJSON string, md5Has
...
+	var apiErr *apierr.APIError
+	if !errors.As(err, &apiErr) {
+		return err
+	}
+	if apiErr.StatusCode == 404 {
+		return nil
+	}
+	if strings.Contains(apiErr.Message, "does not exist") {
+		return nil
+	}
```
mgyucht, line 104: Does this work? This will allow us to handle mapping of API responses to error codes consistently in the SDK, so we don't need to maintain custom logic in the TF provider.
```suggestion
	if errors.Is(err, databricks.ErrResourceDoesNotExist) {
		return nil
	}
```

alexott: `strings.Contains(apiErr.Message, "does not exist")` is used for cases when it's not 404...

mgyucht (approved):
Sounds good to leave this as is.

PR #3893, opened 2024-08-13
[Feature] Introduced Terraform Plugin Framework

tanmay-db:
## Changes

Created PR from `terraform-plugin-framework` branch over https://github.com/databricks/terraform-provider-databricks/pull/3892 (as it's not merged yet). 

## Tests

Unit tests. Nightly tests are running over the PR. 
- [ ] `make test` run locally
- [ ] relevant change in `docs/` folder
- [ ] covered with integration tests in `internal/acceptance`
- [ ] relevant acceptance tests are passing
- [ ] using Go SDK

PR #3914, opened 2024-08-19
[Internal] Add converter functions and tests for plugin framework

edwardfeng-db:
## Changes

- Addressed comments from https://github.com/databricks/terraform-provider-databricks/pull/3893
  - Created new package for plugin framework and added converter functions between tfsdk and gosdk structs
  - Added unit tests
  - Moved `ListAllFields` from `common` into `internal/reflect_utils`

## Tests

- [x] `make test` run locally
- [x] using Go SDK


pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,221 @@
...
+		stringResult := stringMethod.Call(nil)
+		if len(stringResult) == 1 {
+			return stringResult[0].Interface().(string)
+		} else {
+			panic("num get string has more than one result")
+		}
+	} else {
+		panic("enum does not have valid .String() method")
+	}
+}
```
edwardfeng-db, line 212: See comment [here](https://github.com/databricks/terraform-provider-databricks/pull/3893/files#r1721764318) for why type casting doesn't necessarily make the code simpler

internal/reflect_utils/reflect_utils.go
```diff
@@ -0,0 +1,25 @@
+package reflect_utils
+
+import "reflect"
+
+type Field struct {
+	Sf reflect.StructField
```
mgyucht, line 6: Can we just name this `StructField`? I always forget what `Sf` means

internal/reflect_utils/reflect_utils.go
```diff
@@ -0,0 +1,25 @@
...
+	V  reflect.Value
+}
+
+func ListAllFields(v reflect.Value) []Field {
```
mgyucht, line 10: Can you add a doccomment here?

pluginframework/common/converters_test.go
```diff
@@ -0,0 +1,256 @@
+package pluginframework
```
mgyucht, line 1: Can we move these files into `internal/pluginframework/converters`? Keep functions internal that don't need to be exposed publicly, plus converters is a better package name than common or pluginframework.

internal/reflect_utils/reflect_utils.go
```diff
@@ -0,0 +1,25 @@
+package reflect_utils
```
mgyucht, line 1: I hate the name `utils` in packages... see https://go.dev/wiki/CodeReviewComments#package-names. If we don't want to clash with `reflect` from Go's standard library, let's call this `tfreflect` instead.

pluginframework/common/converters_test.go
```diff
@@ -0,0 +1,256 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/stretchr/testify/assert"
+)
+
+type DummyTfSdk struct {
+	Enabled           types.Bool                  `tfsdk:"enabled" tf:"optional"`
+	Workers           types.Int64                 `tfsdk:"workers" tf:""` // Test required field
```
mgyucht, line 15: Is the `tf` option present on every field?

pluginframework/common/converters_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+type DummyNestedGoSdk struct {
+	Name            string   `json:"name"`
+	Enabled         bool     `json:"enabled"`
+	ForceSendFields []string `json:"-"`
+}
+
+// Function to construct individual test case with a pair of matching tfSdkStruct and gosdkStruct.
+// Verifies that the conversion both ways are working as expected.
+func ConverterTestCase(t *testing.T, description string, tfSdkStruct DummyTfSdk, goSdkStruct DummyGoSdk) {
```
mgyucht, line 93: This is great, thank you for splitting this into different test cases. However, these will all be reported as the same test case.

Instead, let's follow the pattern of table driven tests: https://go.dev/wiki/TableDrivenTests. The general structure will look like this:
```go
var tests = []struct{
  name string
  tfSdkStruct DummyTfSdk
  goSdkStruct DummyGoSdk
}{
  {"case 1", DummyTfSdk{}, DummyGoSdk{}},
  ... more cases
}

func TestConverter(t *testing.T) {
  for _, tc := range tests {
    t.Run(tc.name, func() { ConverterTestCase(t, tc.name, tc.tfSdkStruct, tc.goSdkStruct) })
  }
}
```

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+		boolVal := srcFieldValue.(bool)
+		// check if the value is non-zero or if the field is in the forceSendFields list
+		if boolVal || forceSendField {
+			destField.Set(reflect.ValueOf(types.BoolValue(boolVal)))
+		} else {
+			destField.Set(reflect.ValueOf(types.BoolNull()))
+		}
+	case reflect.Int64:
+		// convert any kind of integer to int64
+		intVal := srcField.Convert(reflect.TypeOf(int64(0))).Interface().(int64)
```
mgyucht, line 115: Does this work?
```suggestion
		intVal := srcField.Convert(reflect.TypeOf(int64(0))).Int()
```

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+		// check if the value is non-zero or if the field is in the forceSendFields list
+		if intVal != 0 || forceSendField {
+			destField.Set(reflect.ValueOf(types.Int64Value(int64(intVal))))
+		} else {
+			destField.Set(reflect.ValueOf(types.Int64Null()))
+		}
+	case reflect.Float64:
+		// convert any kind of float to float64
+		float64Val := srcField.Convert(reflect.TypeOf(float64(0))).Interface().(float64)
```
mgyucht, line 124: Does this work?
```suggestion
		float64Val := srcField.Convert(reflect.TypeOf(float64(0))).Float()
```

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+			destField.Set(reflect.ValueOf(types.BoolValue(boolVal)))
+		} else {
+			destField.Set(reflect.ValueOf(types.BoolNull()))
+		}
+	case reflect.Int64:
+		// convert any kind of integer to int64
+		intVal := srcField.Convert(reflect.TypeOf(int64(0))).Interface().(int64)
+		// check if the value is non-zero or if the field is in the forceSendFields list
+		if intVal != 0 || forceSendField {
+			destField.Set(reflect.ValueOf(types.Int64Value(int64(intVal))))
```
mgyucht, line 118: Unnecessary cast?
```suggestion
			destField.Set(reflect.ValueOf(types.Int64Value(intVal)))
```

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+		} else {
+			destField.Set(reflect.ValueOf(types.StringNull()))
+		}
+	case reflect.Struct:
+		if srcField.IsZero() {
+			// Skip zeros
+			return nil
+		}
+		// resolve the nested struct by recursively calling the function
+		if GoSdkToTfSdkStruct(srcFieldValue, destField.Addr().Interface(), ctx).ErrorsCount() > 0 {
```
mgyucht, line 151:
```suggestion
		if GoSdkToTfSdkStruct(srcFieldValue, destField.Addr().Interface(), ctx).HasErrors() {
```

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+			strVal = srcFieldValue.(string)
+		}
+		// check if the value is non-zero or if the field is in the forceSendFields list
+		if strVal != "" || forceSendField {
+			destField.Set(reflect.ValueOf(types.StringValue(strVal)))
+		} else {
+			destField.Set(reflect.ValueOf(types.StringNull()))
+		}
+	case reflect.Struct:
+		if srcField.IsZero() {
```
mgyucht, line 146: Do we need to worry about ForceSendFields here?

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+			// Skip zeros
+			return nil
+		}
+		// resolve the nested struct by recursively calling the function
+		if GoSdkToTfSdkStruct(srcFieldValue, destField.Addr().Interface(), ctx).ErrorsCount() > 0 {
+			panic("Error converting gosdk to tfsdk struct")
+		}
+	case reflect.Slice:
+		if srcField.IsNil() {
```
mgyucht, line 155: Here we don't need to worry about FSF because our API can't distinguish between no list and empty list in requests.

pluginframework/common/go_to_tf.go
mgyucht: Did we ever figure out why this is?

pluginframework/common/go_to_tf.go
```diff
@@ -0,0 +1,222 @@
...
+			return diag.Diagnostics{diag.NewErrorDiagnostic(err.Error(), "gosdk to tfsdk field conversion failure")}
+		}
+	}
+	return nil
+}
+
+func goSdkToTfSdkSingleField(srcField reflect.Value, destField reflect.Value, forceSendField bool, ctx context.Context) error {
+
+	if !destField.CanSet() {
+		panic(fmt.Errorf("destination field can not be set: %s", destField.Type().Name()))
```
mgyucht, line 84: Anywhere that we panic, we should print an error that says something like "This is a bug. Please report this to the maintainers at github.com/databricks/terraform-provider-databricks."

pluginframework/common/tf_to_go.go
```diff
@@ -0,0 +1,220 @@
...
+
+		destField := destVal.FieldByName(srcFieldName)
+
+		err := tfSdkToGoSdkSingleField(srcField, destField, srcFieldName, &forceSendFieldsField, ctx)
+		if err != nil {
+			return diag.Diagnostics{diag.NewErrorDiagnostic(err.Error(), "tfsdk to gosdk field conversion failure")}
+		}
+	}
+
+	return diag.Diagnostics{}
```
mgyucht, line 67: return nil?

pluginframework/common/tf_to_go.go
```diff
@@ -0,0 +1,220 @@
...
+			if srcField.IsZero() {
+				return
+			}
+
+			// Find the Set method
+			destVal := reflect.New(destField.Type())
+			setMethod := destVal.MethodByName("Set")
+			if !setMethod.IsValid() {
+				panic(fmt.Sprintf("set method not found on enum type: %s", destField.Type().Name()))
+			}
```
mgyucht, line 170: We don't have the same case with enum pointers as above?

mgyucht (commented):
Looking great, few comments but this is much nicer!

pluginframework/common/converters_test.go
edwardfeng-db: Yes

pluginframework/common/tf_to_go.go
```diff
@@ -0,0 +1,220 @@
...
+//	types.Float64 -> float64
+//	types.String -> string
+//
+// NOTE:
+//
+//	ForceSendFields are populated for string, bool, int64, float64 on non null values
+//	types.list and types.map are not supported
+//	map keys should always be a string
+//	tfsdk structs use types.String for all enum values
+func TfSdkToGoSdkStruct(tfsdk interface{}, gosdk interface{}, ctx context.Context) diag.Diagnostics {
```
mgyucht, line 30: ctx should be first argument

pluginframework/common/tf_to_go.go
```diff
@@ -0,0 +1,220 @@
...
+		// If it is a real stuct instead of a tfsdk type, recursively resolve it.
+		if TfSdkToGoSdkStruct(srcFieldValue, destField.Addr().Interface(), ctx).ErrorsCount() > 0 {
+			panic("Error converting tfsdk to gosdk struct")
+		}
+	}
+}
+
+func addToForceSendFields(fieldName string, forceSendFieldsField *reflect.Value) {
+	if forceSendFieldsField == nil || !forceSendFieldsField.IsValid() || !forceSendFieldsField.CanSet() {
+		log.Printf("[Debug] forceSendFieldsField is nil, invalid or not settable. %s", fieldName)
```
mgyucht, line 208: Use `tflog.Debug`

pluginframework/common/tf_to_go.go
mgyucht: Not needed because we always instantiate the memory for this field in the same way two lines above.

pluginframework/common/go_to_tf.go
edwardfeng-db: Not really - discussed offline, this is actuallly not supported by the gosdk, so we don't need FSF

pluginframework/common/go_to_tf.go
edwardfeng-db: Discussed offline and added comment, this case where `CanAddr()` returns false happens in unit tests due to the value being a constant

internal/tfreflect/reflect_utils.go
```diff
@@ -0,0 +1,27 @@
+package tfreflect
+
+import "reflect"
+
+type Field struct {
+	StructField reflect.StructField
+	Value       reflect.Value
+}
+
+// Given a reflect.Value of a struct, list all of the fields for both struct field and
```
mgyucht, line 10: Small nit:
> Comments should begin with the name of the thing being described and end in a period.

https://go.dev/wiki/CodeReviewComments#comment-sentences

pluginframework/converters/converters_test.go
```diff
@@ -0,0 +1,228 @@
...
+	name        string
+	tfSdkStruct DummyTfSdk
+	goSdkStruct DummyGoSdk
+}{
+	{
+		"string conversion",
+		DummyTfSdk{Description: types.StringValue("abc")},
+		DummyGoSdk{Description: "abc", ForceSendFields: []string{"Description"}},
+	},
+	{
```
mgyucht, line 113: Can we add one test case for DummyTfSdk value set to zero values for each of the primitive types and verifying that ForceSendFields is set properly?

pluginframework/converters/go_to_tf.go
```diff
@@ -0,0 +1,224 @@
...
+			strVal = getStringFromEnum(srcField)
+		} else {
+			strVal = srcFieldValue.(string)
+		}
+		// check if the value is non-zero or if the field is in the forceSendFields list
+		if strVal != "" || forceSendField {
+			destField.Set(reflect.ValueOf(types.StringValue(strVal)))
+		} else {
+			destField.Set(reflect.ValueOf(types.StringNull()))
+		}
```
mgyucht, line 145: Is the else case needed? The default value should be empty string IIUC.

pluginframework/converters/go_to_tf.go
```diff
@@ -0,0 +1,224 @@
...
+
+	"github.com/databricks/databricks-sdk-go/logger"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/databricks/terraform-provider-databricks/internal/tfreflect"
+)
+
+// Converts a gosdk struct into a tfsdk struct, with the folowing rules.
```
mgyucht, line 16: Ditto here, just put the function name first.

pluginframework/converters/tf_to_go.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/databricks/databricks-sdk-go/logger"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/databricks/terraform-provider-databricks/internal/tfreflect"
+)
+
+// Converts a tfsdk struct into a gosdk struct, with the folowing rules.
```
mgyucht, line 17: Ditto here, just put the function name first.

mgyucht (approved):
LGTM, with a few nits.

pluginframework/converters/converters_test.go
edwardfeng-db: Sounds good, added

pluginframework/converters/go_to_tf.go
edwardfeng-db: Setting `StringNull` just means that we are explicitly saying this is the case that is "unset" thus when we convert it back to gosdk struct, we know whether to add it into FSF or not

PR #3916, opened 2024-08-19
[Internal] Migrate Share resource to Go SDK

rauchy:
## Changes

This PR is my first attempt at migrating the Share resource from using hand-crafted API calls to the Databricks Go SDK. I’ve updated the `Read` function to use the SDK’s `Get` method. Next, I’ll work on migrating the remaining CRUD operations. I’ve run the existing tests to ensure everything is still working as expected.

Update: also included Create & Update here, which makes the Share resource go through the Go SDK for all CRUD operations.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/sharing/data_share.go
+++ b/sharing/data_share.go
@@ -3,6 +3,7 @@ package sharing
 import (
 	"context"
 
+	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/service/sharing"
 	"github.com/databricks/terraform-provider-databricks/common"
 )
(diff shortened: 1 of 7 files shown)


sharing/data_share.go
```diff
@@ -3,20 +3,28 @@ package sharing
...
 )
 
 func DataSourceShare() common.Resource {
 	type ShareDetail struct {
-		Name      string             `json:"name,omitempty" tf:"computed"`
-		Objects   []SharedDataObject `json:"objects,omitempty" tf:"computed,slice_set,alias:object"`
-		CreatedAt int64              `json:"created_at,omitempty" tf:"computed"`
-		CreatedBy string             `json:"created_by,omitempty" tf:"computed"`
+		Name      string                     `json:"name,omitempty" tf:"computed"`
+		Objects   []sharing.SharedDataObject `json:"objects,omitempty" tf:"computed,slice_set,alias:object"`
```
tanmay-db, line 13: Some changes would be required before we can do this, for example: 
`StartVersion` in `SharedDataObject` is annotated `tf:"suppress_diff" ` whereas it is not in Go SDK struct.

Also, DataObjectType is string in `SharedDataObject` but SharedDataObjectDataObjectType in Go SDK which I think is okay but needs to be verified if this doesn't cause any other issue (eg: if its used in test, then that would need to be upgraded, if it's not in test, might be good to add a test).

sharing/resource_share.go
```diff
@@ -208,17 +206,28 @@ func ResourceShare() common.Resource {
 			return nil
 		},
 		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
-			si, err := NewSharesAPI(ctx, c).get(d.Id())
```
tanmay-db, line 211: si.sortSharesByName()
	si.suppressCDFEnabledDiff()

would also be needed here as we were doing it before + would be required before we do StructToData below.

tanmay-db (commented):
Thanks for the PR, looks great, just had a few comments. I have triggered nightly test run.

sharing/data_share.go
rauchy: 👍 addressed in 20b08cbbb0ba0cbbff333aa72deba511ec81c637

tanmay-db (approved):

sharing/resource_share.go
rauchy: Addressed in [910decf](https://github.com/databricks/terraform-provider-databricks/pull/3916/commits/910decfab1c52c5f522d3c5e69714aaf1d714bed)

PR #3917, opened 2024-08-19
[Fix] Fixed logging for underlying Go SDK

tanmay-db:
## Changes

Go SDK doesn't contain valid context for terraform logger, due to this we get a nil logger from context: https://github.com/hashicorp/terraform-plugin-log/blob/main/tflog/provider.go#L40. 

We always use the context from terraform and ignore the one from Go SDK. 

## Tests

`TF_LOG=TRACE terraform apply` won't print: https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_default.go#L54 on main, prints it after change. Also tested manually for some debug statements.  

Ran integration tests and see debug logs.
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -22,44 +22,55 @@ func (tfLogger *TfLogger) Enabled(_ context.Context, _ logger.Level) bool {
 	return true
 }
 
-func (tfLogger *TfLogger) Tracef(_ context.Context, format string, v ...any) {
+func (tfLogger *TfLogger) Tracef(ctx context.Context, format string, v ...any) {
 	if tfLogger == nil {
-		tflog.Trace(loggerContext, fmt.Sprintf(format, v...), nil)
+		tflog.Trace(getLoggerValidContext(ctx), fmt.Sprintf(format, v...), nil)
 	} else {
-		tflog.SubsystemTrace(loggerContext, tfLogger.Name, fmt.Sprintf(format, v...), nil)
+		tflog.SubsystemTrace(getLoggerValidContext(ctx), tfLogger.Name, fmt.Sprintf(format, v...), nil)
 	}
 }
 
-func (tfLogger *TfLogger) Debugf(_ context.Context, format string, v ...any) {
+func (tfLogger *TfLogger) Debugf(ctx context.Context, format string, v ...any) {
 	if tfLogger == nil {
-		tflog.Debug(loggerContext, fmt.Sprintf(format, v...), nil)
+		tflog.Debug(getLoggerValidContext(ctx), fmt.Sprintf(format, v...), nil)
 	} else {
-		tflog.SubsystemDebug(loggerContext, tfLogger.Name, fmt.Sprintf(format, v...), nil)
+		tflog.SubsystemDebug(getLoggerValidContext(ctx), tfLogger.Name, fmt.Sprintf(format, v...), nil)
 	}
 }
 
-func (tfLogger *TfLogger) Infof(_ context.Context, format string, v ...any) {
+func (tfLogger *TfLogger) Infof(ctx context.Context, format string, v ...any) {
 	if tfLogger == nil {
-		tflog.Info(loggerContext, fmt.Sprintf(format, v...), nil)
+		tflog.Info(getLoggerValidContext(ctx), fmt.Sprintf(format, v...), nil)
 	} else {
-		tflog.SubsystemInfo(loggerContext, tfLogger.Name, fmt.Sprintf(format, v...), nil)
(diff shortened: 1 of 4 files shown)


logger/logger.go
```diff
@@ -22,45 +24,57 @@ func (tfLogger *TfLogger) Enabled(_ context.Context, _ logger.Level) bool {
...
+func getLoggerValidContext(ctx context.Context) context.Context {
+	logger := ctx.Value("provider")
+	if logger == nil {
+		return loggerContext
 	}
+	return ctx
 }
 
-func SetLogger() {
+func SetLogger(ctx context.Context) {
```
renaudhartert-db, line 76: I'm a little puzzled by this function (especially the previous version) as it is resetting something. If I understand correctly, it could be rewritten as follows:

```go
func SetLogger(ctx context.Context) {
	logger.DefaultLogger = nil
	loggerContext = ctx
}
```

Is that the intended behavior? If so, I think it would benefit from a comment to explain why we are resetting the defaultLogger.

tanmay-db: `DefaultLogger` is defined in Go SDK. We set it only at this place inside terraform repo (we also set this to SimpleLogger in GoSDK). We can't re-write this to nil as we need to specify the type to `TfLogger`. 

So when this is run inside Go SDK (as part of terraform provider), we get the logger based on DefaultLogger: https://github.com/databricks/databricks-sdk-go/blob/main/logger/logger.go#L47

tanmay-db: Both SimpleLogger (defined in Go SDK) and TfLogger (defined in TF repo) implements the Logger (defined in Go SDK) interface

renaudhartert-db:
I'm not sure I understand what the problem really is. Is it correct that we want the Go SDK to log using the Terraform Logger? 

If so, I'm not sure I understand why the Go SDK contexts are problematic. Aren't these given by the terraform provider?

logger/logger.go
renaudhartert-db: OK, I see that a `nil` TfLogger is actually a valid logger.

renaudhartert-db:
> Go SDK contexts are problematic as these aren't given by terraform. We define context separately there, for example while Configuring we use refreshCtx that we set earlier as context.Background(): https://github.com/databricks/databricks-sdk-go/blob/main/config/config.go#L365

Thanks! That is the piece I was missing. We're essentially trying to hack around a bug in the Go SDK. 

I'd like to consider an alternative solution that better better encapsulate the hack so that the code is a little more future proof and less coupled. 

My understanding is that the context that is being passed to `SetLogger` contains all the pieces we need to use. If that is correct, we could leverage that to build a new `TfLogger` that explicitly encapsulates these pieces. Something like that:

```go
type TfLogger struct {
  name string 
  ctx context.Context
  useSubsystem bool
} 

func (tfl *TfLogger) Errorf(_ context.Context, format string, v ...any) {
  if useSubsystem {
    tflog.SubsystemError(tfl.ctx, tfl.name, format, v)
  } else {
    tflog.Error(tfl.ctx, tfl.name, format, v)
  }
}

func SetLogger(tfl *TfLogger) {
  logger.DefaultLogger = tfl
}
```

What I like about this solution is that it doesn't rely on any global variable on the Terraform provider side. It also fully commits to a single interface to interact with Go SDK logging (i.e. the `DefaultLogger`) and literally ignores the context given to the logging methods — which I believe is a great thing as I'm having a hard time seeing the current logger interface remain in SDK V1.

logger/logger.go
```diff
@@ -9,7 +9,15 @@ import (
...
 type TfLogger struct {
-	Name string
+	ctx context.Context
+}
+
+func NewTfLogger(ctx context.Context) *TfLogger {
+	return &TfLogger{ctx: ctx}
+}
+
+func SetTfLogger(tfLogger *TfLogger) {
```
tanmay-db, line 19: We could have another method such as SetTfLoggerFromContext which will get a new logger and set it but I think it will lead to a loss of readability of code.

SetTfLogger(NewTfLogger(ctx)) clearly say we get a new logger first and set it.

logger/logger.go
```diff
@@ -9,7 +9,15 @@ import (
 )
 
 type TfLogger struct {
-	Name string
+	ctx context.Context
+}
+
+func NewTfLogger(ctx context.Context) *TfLogger {
```
renaudhartert-db, line 15: [nit] I would just add a small comment here to explain that this function is expecting a certain type of context that has contains a logger and not any context.

renaudhartert-db (approved):
LGTM — thanks for navigating me through the PR Tanmay.

logger/logger.go
tanmay-db: Added

tanmay-db:
Thanks for the review @renaudhartert-db

PR #3925, opened 2024-08-21
[Fix] Fix `databricks_sql_table`

### LINKED ISSUE #3894: [ISSUE] Issue with `databricks_sql_table` resource properties
### Configuration

See [sourcefile](https://github.com/databricks/universe-dev/blob/731d3516b385cab8da17111b51d1658ea41a2907/terraform/common/db-sentinel/teams/cleanroom/cleanroom_team_stack.jsonnet) 
```hcl
  {
    resource: {
      databricks_catalog: {
        t_clroom_catalog: {
          name: "t_clroom_sentinel-catalog-dcw",
          owner: "owner",
        },
      },
      databricks_schema: {
        t_clroom_schema: {
          name: "t_clroom_schema",
          owner: "owner",
          catalog_name: "t_clroom_sentinel-catalog-dcw",
        },
      },
      databricks_sql_table: {
        t_clroom_sql_table: {
          name: "t_clroom_sql_table",
          owner: "owner",
          catalog_name: "t_clroom_sentinel-catalog-dcw",
          schema_name: "t_clroom_schema",
          table_type: "MANAGED",
          data_source_format: "DELTA",
          properties: {
            "delta.enableDeletionVectors": "false",
          },
          column: [
            {
              name: "id",
              type: "int",
            }
          ],
        },
      }
    }
  }
```

### Expected Behavior

The table should get created

### Actual Behavior

```
Error: cannot create sql table: cannot execute CREATE TABLE `t_clroom_sentinel-catalog-dcw`.`t_clroom_schema`.`t_clroom_sql_table` (`id` int, `name` string) USING DELTA COMMENT 'Default sql table asset for clean room probes. Managed by TF' TBLPROPERTIES ();: [PARSE_SYNTAX_ERROR] Syntax error at or near ')'. SQLSTATE: 42601
```
The error comes from the optional **properties**. From [slack discussion](https://databricks.slack.com/archives/C0279U6643G/p1723266795743099?thread_ts=1723225952.186029&cid=C0279U6643G), seems like the SQL statement constructed by TF has empty `TBLPROPERTIES` - what likely happened is that the property specified in HCL is automatically ignored by Terraform, due to https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_sql_table.go#L84

mgyucht:
## Changes
`databricks_sql_table` has been difficult to maintain due to the fact that Databricks adds table properties and options that users haven't directly configured. Historically, to handle this, we have tried to construct properties carefully by tracking which properties are added by Databricks automatically and removing them. This has two challenges:
1. We don't know ahead of time which properties are set by Databricks, so we don't know which properties we need to remove to compute an accurate diff.
2. Users are not able to specify different values for those default properties: they are hidden from the `serializeProperties()` and `serializeOptions()` functions in the resource implementation.

This PR takes a different approach for handling these fields by storing the properties/options returned by the API in `effective_properties` and `effective_options` fields, respectively. These fields are computed and non-user-settable, so users cannot override them. When creating/updating the table, we only include the properties specified by the user. However, when determining if there is a diff, we compare the effective properties fetched by the `terraform plan` triggering an update of the table state with the user-specified properties. If the user specified any properties that are not part of the effective properties/options, or if the user's value clashes with the value provided, this will trigger a diff. However, if a non-user-specified property has a difference, then that does not trigger a diff.

As a side-effect, this PR also fixes the treatment of `TBLPROPERTIES` and `OPTIONS` as reported in #3894. 

Fixes #3894.
## Tests
Unit tests to verify the diff behavior with this change.
Unit test to ensure that TBLPROPERTIES is correctly set.

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -42,7 +42,9 @@ type SqlTableInfo struct {
 	ViewDefinition        string            `json:"view_definition,omitempty"`
 	Comment               string            `json:"comment,omitempty"`
 	Properties            map[string]string `json:"properties,omitempty" tf:"computed"`
+	EffectiveProperties   map[string]string `json:"effective_properties,omitempty" tf:"computed"`
 	Options               map[string]string `json:"options,omitempty" tf:"force_new"`
+	EffectiveOptions      map[string]string `json:"effective_options,omitempty" tf:"computed"`
 	ClusterID             string            `json:"cluster_id,omitempty" tf:"computed"`
 	WarehouseID           string            `json:"warehouse_id,omitempty"`
 	Owner                 string            `json:"owner,omitempty" tf:"computed"`
@@ -51,7 +53,7 @@ type SqlTableInfo struct {
 	sqlExec sql.StatementExecutionInterface
 }
 
-func (ti *SqlTableInfo) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
+func (ti SqlTableInfo) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
 
 	caseInsensitiveFields := []string{"name", "catalog_name", "schema_name"}
 	for _, field := range caseInsensitiveFields {
@@ -74,6 +76,9 @@ func (ti *SqlTableInfo) CustomizeSchema(s *common.CustomizableSchema) *common.Cu
 	s.SchemaPath("column", "type").SetCustomSuppressDiff(func(k, old, new string, d *schema.ResourceData) bool {
 		return getColumnType(old) == getColumnType(new)
 	})
+	for _, field := range []string{"effective_properties", "effective_options"} {
+		s.SchemaPath(field).SetReadOnly()
+	}
 	return s
 }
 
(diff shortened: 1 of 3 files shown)


catalog/resource_sql_table.go
```diff
@@ -51,6 +53,35 @@ type SqlTableInfo struct {
...
+	s.SchemaPath("warehouse_id").SetConflictsWith([]string{"cluster_id"})
+
+	s.SchemaPath("partitions").SetConflictsWith([]string{"cluster_keys"})
+	s.SchemaPath("cluster_keys").SetConflictsWith([]string{"partitions"})
+	s.SchemaPath("column", "type").SetCustomSuppressDiff(func(k, old, new string, d *schema.ResourceData) bool {
+		return getColumnType(old) == getColumnType(new)
+	})
+	for _, field := range []string{"effective_properties", "effective_options"} {
+		s.SchemaPath(field).SetReadOnly()
+	}
```
alexott, line 81: if we set them to `tf:computed` they should be set as readonly automatically

catalog/resource_sql_table_test.go
```diff
@@ -1332,6 +1334,196 @@ func TestResourceSqlTableCreateTable_ExistingSQLWarehouse(t *testing.T) {
...
+				Response: SqlTableInfo{
+					Name:                  "bar",
+					CatalogName:           "main",
+					SchemaName:            "foo",
+					TableType:             "EXTERNAL",
+					DataSourceFormat:      "DELTA",
+					StorageLocation:       "s3://ext-main/foo/bar1",
+					StorageCredentialName: "somecred",
+					Comment:               "terraform managed",
+					Properties: map[string]string{
```
nkvuong, line 1390: `delta.enableDeletionVectors` should appear in the response properties as well?

nkvuong (approved):
this looks good to me, just a nit on the new unit test

apamildner:
Thank you very much for this! Can we do a new release so this commit is available to us? It has been long awaited 👍

PR #3928, opened 2024-08-21
[Internal] Add StructToSchema for Plugin Framework

edwardfeng-db:
## Changes

- Addressed comments in https://github.com/databricks/terraform-provider-databricks/pull/3893
- Adding `StructToSchema` and corresponding unit tests, as well as tests for `CustomizableSchema`

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/pluginframework/tfschema/struct_to_schema.go
+++ b/internal/pluginframework/tfschema/struct_to_schema.go
@@ -1,0 +1,145 @@
+package tfschema
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/databricks/terraform-provider-databricks/internal/tfreflect"
+	dataschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func typeToSchema(v reflect.Value) map[string]AttributeBuilder {
+	scm := map[string]AttributeBuilder{}
+	rk := v.Kind()
+	if rk == reflect.Ptr {
+		v = v.Elem()
+		rk = v.Kind()
+	}
+	if rk != reflect.Struct {
+		panic(fmt.Errorf("schema value of Struct is expected, but got %s: %#v", rk.String(), v))
+	}
+	fields := tfreflect.ListAllFields(v)
+	for _, field := range fields {
+		typeField := field.StructField
+		fieldName := typeField.Tag.Get("tfsdk")
+		if fieldName == "-" {
+			continue
+		}
+		isOptional := fieldIsOptional(typeField)
+		kind := typeField.Type.Kind()
+		if kind == reflect.Ptr {
+			elem := typeField.Type.Elem()
+			sv := reflect.New(elem).Elem()
+			nestedScm := typeToSchema(sv)
+			scm[fieldName] = SingleNestedAttributeBuilder{Attributes: nestedScm, Optional: isOptional, Required: !isOptional}
+		} else if kind == reflect.Slice {
+			elemType := typeField.Type.Elem()
+			if elemType.Kind() == reflect.Ptr {
+				elemType = elemType.Elem()
+			}
+			if elemType.Kind() != reflect.Struct {
+				panic(fmt.Errorf("unsupported slice value for %s: %s", fieldName, elemType.Kind().String()))
+			}
+			switch elemType {
+			case reflect.TypeOf(types.Bool{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.BoolType, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.Int64{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.Int64Type, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.Float64{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.Float64Type, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.String{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.StringType, Optional: isOptional, Required: !isOptional}
+			default:
+				// Nested struct
+				nestedScm := typeToSchema(reflect.New(elemType).Elem())
+				scm[fieldName] = ListNestedAttributeBuilder{NestedObject: NestedAttributeObject{Attributes: nestedScm}, Optional: isOptional, Required: !isOptional}
+			}
+		} else if kind == reflect.Map {
+			elemType := typeField.Type.Elem()
+			if elemType.Kind() == reflect.Ptr {
+				elemType = elemType.Elem()
+			}
+			if elemType.Kind() != reflect.Struct {
+				panic(fmt.Errorf("unsupported map value for %s: %s", fieldName, elemType.Kind().String()))
+			}
+			switch elemType {
+			case reflect.TypeOf(types.Bool{}):
+				scm[fieldName] = MapAttributeBuilder{ElementType: types.BoolType, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.Int64{}):
+				scm[fieldName] = MapAttributeBuilder{ElementType: types.Int64Type, Optional: isOptional, Required: !isOptional}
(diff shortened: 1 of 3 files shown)


internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return scm
+}
+
+func fieldIsOptional(field reflect.StructField) bool {
+	tagValue := field.Tag.Get("tf")
+	return strings.Contains(tagValue, "optional")
+}
+
+func ResourceStructToSchema(v any, customizeSchema func(CustomizableSchema) CustomizableSchema) schema.Schema {
```
mgyucht, line 115: Can we add a doccomment for each public function?

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+
+func typeToSchema(v reflect.Value) map[string]AttributeBuilder {
+	scm := map[string]AttributeBuilder{}
+	rk := v.Kind()
+	if rk == reflect.Ptr {
+		v = v.Elem()
+		rk = v.Kind()
+	}
+	if rk != reflect.Struct {
+		panic(fmt.Errorf("schema value of Struct is expected, but got %s: %#v", rk.String(), v))
```
mgyucht, line 22: Can we link to the common panic error message here?

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+			sv := reflect.New(elem).Elem()
+			nestedScm := typeToSchema(sv)
+			scm[fieldName] = SingleNestedAttributeBuilder{Attributes: nestedScm, Optional: isOptional, Required: !isOptional}
+		} else if kind == reflect.Slice {
+			elemType := typeField.Type.Elem()
+			if elemType.Kind() == reflect.Ptr {
+				elemType = elemType.Elem()
+			}
+			if elemType.Kind() != reflect.Struct {
+				panic(fmt.Errorf("unsupported slice value for %s: %s", fieldName, elemType.Kind().String()))
```
mgyucht, line 44: ditto on the error message

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+				nestedScm := typeToSchema(reflect.New(elemType).Elem())
+				scm[fieldName] = ListNestedAttributeBuilder{NestedObject: NestedAttributeObject{Attributes: nestedScm}, Optional: isOptional, Required: !isOptional}
+			}
+		} else if kind == reflect.Map {
+			elemType := typeField.Type.Elem()
+			if elemType.Kind() == reflect.Ptr {
+				elemType = elemType.Elem()
+			}
+			if elemType.Kind() != reflect.Struct {
+				panic(fmt.Errorf("unsupported map value for %s: %s", fieldName, elemType.Kind().String()))
```
mgyucht, line 66: Ditto on the error message

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+			case types.Bool:
+				scm[fieldName] = BoolAttributeBuilder{Optional: isOptional, Required: !isOptional}
+			case types.Int64:
+				scm[fieldName] = Int64AttributeBuilder{Optional: isOptional, Required: !isOptional}
+			case types.Float64:
+				scm[fieldName] = Float64AttributeBuilder{Optional: isOptional, Required: !isOptional}
+			case types.String:
+				scm[fieldName] = StringAttributeBuilder{Optional: isOptional, Required: !isOptional}
+			case types.List:
+				panic("types.List should never be used in tfsdk structs")
```
mgyucht, line 93: Ditto

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+			case types.Map:
+				panic("types.Map should never be used in tfsdk structs")
```
mgyucht, line 95: Ditto

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+			continue
+		}
+		isOptional := fieldIsOptional(typeField)
+		kind := typeField.Type.Kind()
+		if kind == reflect.Ptr {
+			elem := typeField.Type.Elem()
+			sv := reflect.New(elem).Elem()
+			nestedScm := typeToSchema(sv)
+			scm[fieldName] = SingleNestedAttributeBuilder{Attributes: nestedScm, Optional: isOptional, Required: !isOptional}
+		} else if kind == reflect.Slice {
```
mgyucht, line 38:
```suggestion
		value := field.Value
		if kind == reflect.Ptr {
			kind = typeField.Type.Elem().Kind()
			value := field.Elem()
		}
		if kind == reflect.Slice {
```
Then on line 83 we use this `value` instead of `field.Value`. This means that pointers always receive consistent treatment as if they were not pointers from the perspective of schema calculation. It's basically the equivalent to lines 17-20.

internal/pluginframework/tfschema/struct_to_schema.go
```diff
@@ -0,0 +1,145 @@
...
+			switch elemType {
+			case reflect.TypeOf(types.Bool{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.BoolType, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.Int64{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.Int64Type, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.Float64{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.Float64Type, Optional: isOptional, Required: !isOptional}
+			case reflect.TypeOf(types.String{}):
+				scm[fieldName] = ListAttributeBuilder{ElementType: types.StringType, Optional: isOptional, Required: !isOptional}
+			default:
```
mgyucht, line 55: For these `default` cases, these are always types defined in the Go SDK right?

internal/pluginframework/tfschema/struct_to_schema_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+type TestNestedMapTfSdk struct {
+	NestedMap map[string]DummyNested `tfsdk:"nested_map" tf:"optional"`
+}
+
+type TestPointerTfSdk struct {
+	Nested *DummyNested `tfsdk:"nested" tf:"optional"`
+}
+
+var tests = []struct {
```
mgyucht, line 54: Can we add a couple more test cases:
* Input is a pointer to a struct
* Input has a field that is a pointer to a struct

internal/pluginframework/tfschema/struct_to_schema_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	data_scm := DataSourceStructToSchema(testStruct, nil)
+	data_state := tfsdk.State{
+		Schema: data_scm,
+	}
+	// Assert we can properly set the state, this means the schema and the struct are consistent.
+	assert.True(t, !data_state.Set(context.Background(), testStruct).HasError(), fmt.Sprintf("DataSourceStructToSchema - %s", description))
+}
+
+func TestStructToSchemaConversion(t *testing.T) {
```
mgyucht, line 136: Awesome. It would also be great to have some negative test cases: where is this expected to fail? E.g.

```go
func testStructToSchemaPanics(t *testing.T, struct any, expectedError string) {
  defer func() {
    err := recover()
    if err == nil {
      t.Fail(fmt.Sprintf("did not fail with error %q", expectedError))
    }
    if !strings.Contains(err.Error(), expectedError) {
      t.Fail(fmt.Sprintf("error %w did not include expected error message %q", err, expectedError))
    }
  }
  ResourceStructToSchema(struct)
}
```
We can add test cases for unsupported paths.

mgyucht (commented):
Few comments on this one, but mostly looking great!

internal/pluginframework/tfschema/struct_to_schema_test.go
edwardfeng-db: Sounds good, I think the case for "input is a pointer to a struct" already exists

mgyucht (approved):

PR #3933, opened 2024-08-22
[Internal] Bump Go SDK to v0.45.0

tanmay-db:
## Changes

Bumps the Go SDK to latest version and fix some tests related to changes in:
- https://github.com/databricks/databricks-sdk-go/pull/1014
- https://github.com/databricks/databricks-sdk-go/pull/1021

## Tests

Unit tests, nightly will run over release PR.
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/client_test.go
+++ b/common/client_test.go
@@ -314,7 +314,7 @@ func TestGetJWTProperty_Authenticate_Fail(t *testing.T) {
 	_, err := client.GetAzureJwtProperty("tid")
 	require.Error(t, err)
 	assert.True(t, strings.HasPrefix(err.Error(),
-		"default auth: azure-cli: cannot get access token: This is just a failing script"))
+		"default auth: azure-cli: cannot get account info"))
 }
 
 type mockInternalUserService struct {
--- a/go.mod
+++ b/go.mod
@@ -3,11 +3,12 @@ module github.com/databricks/terraform-provider-databricks
 go 1.22
 
 require (
-	github.com/databricks/databricks-sdk-go v0.44.0
+	github.com/databricks/databricks-sdk-go v0.45.0
 	github.com/golang-jwt/jwt/v4 v4.5.0
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hashicorp/hcl/v2 v2.21.0
+	github.com/hashicorp/terraform-plugin-framework v1.11.0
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -49,7 +50,6 @@ require (
 	github.com/hashicorp/logutils v1.0.0 // indirect
 	github.com/hashicorp/terraform-exec v0.21.0 // indirect
 	github.com/hashicorp/terraform-json v0.22.1 // indirect
-	github.com/hashicorp/terraform-plugin-framework v1.11.0 // indirect
```
tanmay-db, line 52: This automatically happened when I upgraded the Go SDK. Leaving it for now as it's a no-op change.

hectorcast-db (approved):

renaudhartert-db (approved):

common/client_test.go
```diff
@@ -314,7 +314,7 @@ func TestGetJWTProperty_Authenticate_Fail(t *testing.T) {
 	_, err := client.GetAzureJwtProperty("tid")
 	require.Error(t, err)
 	assert.True(t, strings.HasPrefix(err.Error(),
-		"default auth: azure-cli: cannot get access token: This is just a failing script"))
+		"default auth: azure-cli: cannot get account info"))
```
tanmay-db, line 317: Note: Earlier we used to do: 

`````
	if ee, ok := err.(*exec.ExitError); ok {
		return nil, fmt.Errorf("cannot get access token: %s", string(ee.Stderr))
	}
`````
and so we get the error from stderr which is "This is just a failing script" that we defined in test data.

We would need similar to for when we run: https://github.com/databricks/databricks-sdk-go/pull/1021/files#diff-eda1575b7fd7a235cdaeb9aefc3b7f84f8c4f5596726674ba187b39fab75ce13R166 but this would require another go sdk bump + release (or we can add the fix in go sdk and use pseudo version for bump)

I think, it should be okay to go with this change since we only need to check for the failure prefix and we can track to do this in next Go SDK and TF release.

Comment (tanmay-db) [reply common/client_test.go:317]: Test data ref for writing to stderr

```
if [ "yes" == "$FAIL" ]; then 
    >&2 /bin/echo "This is just a failing script."
    exit 1
fi
```

tanmay-db: + @mgyucht for visibility on this as it was done in PR: https://github.com/databricks/databricks-sdk-go/pull/1021.

renaudhartert-db (approved):

PR #3940, opened 2024-08-22
[Internal] Refactor provider and related packages

tanmay-db:
## Package organization for Providers
We are migrating the resource from SDKv2 to Plugin Framework provider and hence both of them exist in the codebase. For uniform code convention, readability and development, they are organized in the `internal/providers` directory under root as follows: 
- `providers`: Contains the changes that `depends` on both internal/providers/sdkv2 and internal/providers/pluginfw packages, eg: `GetProviderServer`.
- `common`: Contains the changes `used by` both internal/providers/sdkv2 and internal/providers/pluginfw packages, eg: `ProviderName`.
- `pluginfw`: Contains the changes specific to Plugin Framework. This package shouldn't depend on sdkv2 or common.
- `sdkv2`: Contains the changes specific to SDKv2. This package shouldn't depend on pluginfw or common.

This has been updated in CONTRIBUTING.md

Package comments have been added, also some tests have been refactored to reduce redundancy example: we define test data path at multiple places. 

## Tests

Modified existing tests
- [ ] `make test` run locally
- [ ] using Go SDK


CONTRIBUTING.md
```diff
@@ -118,6 +118,17 @@ provider_installation {
...
 After installing the necessary software for building provider from sources, you should be able to run `make coverage` to run the tests and see the coverage.
 
+## Package organization for Providers
+We are migrating the resource from SDKv2 to Plugin Framework provider and hence both of them exist in the codebase. For uniform code convention, readability and development, they are defined under internal directory under root as: 
+- `common/provider`: This package contains the changes common to both the providers example provider factory and tests
+  - imports are declared as: 
+  	- `pluginframeworkprovider` "github.com/databricks/terraform-provider-databricks/internal/pluginframework/provider"
+	  - `sdkv2provider` "github.com/databricks/terraform-provider-databricks/internal/sdkv2/provider"
+- `sdkv2/provider`: This package contains the changes related to only sdkv2 provider and shouldn't depend on pluginframework provider
+- `pluginframework/provider`: This package contains the changes related to only plugin framework provider and shouldn't depend on sdkv2
```
renaudhartert-db, line 128: I think it would be good to start building the habit of documenting our packages, even internal ones. Having these instructions next to the code is located will increase the likelihood of people reading them.

See "Packages" section: https://tip.golang.org/doc/comment#packages

CONTRIBUTING.md
```diff
@@ -118,6 +118,17 @@ provider_installation {
...
+
+Note: To avoid circular dependency and make code easy to read, any common changes for both provider will go into interna/common/provider package.
```
renaudhartert-db, line 130: Phrasing nit + fix a couple of typos.

```suggestion
Note: Common dependencies for both providers will go into the internal/common/provider package.
```

CONTRIBUTING.md
renaudhartert-db, line 128: Have we consider a different structure where all these will be subpackages of a `providers` package?

```
- providers/
  - common/
  - sdkv2/
  - pluginfw/
``` 

This is a little more idiomatic, makes it clear that all of this is provider related, and solve the need for renaming packages.

renaudhartert-db (changes requested):

CONTRIBUTING.md
```diff
@@ -118,6 +118,17 @@ provider_installation {
 
 After installing the necessary software for building provider from sources, you should be able to run `make coverage` to run the tests and see the coverage.
 
+## Package organization for Providers
+We are migrating the resource from SDKv2 to Plugin Framework provider and hence both of them exist in the codebase. For uniform code convention, readability and development, they are defined under internal directory under root as: 
+- `common/provider`: This package contains the changes common to both the providers example provider factory and tests
```
mgyucht, line 123: There is no `internal/common/provider` package in this PR or on main. Are you planning to introduce this, or is this a stale comment?

CONTRIBUTING.md
```diff
@@ -118,6 +118,17 @@ provider_installation {
...
+  - imports are declared as: 
+  	- `pluginframeworkprovider` "github.com/databricks/terraform-provider-databricks/internal/pluginframework/provider"
+	  - `sdkv2provider` "github.com/databricks/terraform-provider-databricks/internal/sdkv2/provider"
```
mgyucht, line 126: Small suggestion just to shorten the length of the renamed import slightly
```suggestion
  	- `fwprovider` "github.com/databricks/terraform-provider-databricks/internal/pluginframework/provider"
	  - `sdkv2provider` "github.com/databricks/terraform-provider-databricks/internal/sdkv2/provider"
```

mgyucht (commented):
+1 to Renaud's comments: let's organize all provider logic in subdirectories under `internal/provider` and put this documentation in `internal/provider/pkg.go`.

CONTRIBUTING.md
tanmay-db: There is internal/common/provider in the PR:
image

Maybe I am getting this wrong, do you mean imports for common provider?
We use that in main.go

`"github.com/databricks/terraform-provider-databricks/internal/common/provider"`

CONTRIBUTING.md
tanmay-db: Agree, that's better for organisation. I will update the structure.

internal/providers/common/provider.go
```diff
@@ -1,10 +1,12 @@
-package provider
+// Package common contains the changes that depends on both internal/providers/sdkv2 and internal/providers/pluginfw packages
+package common
```
renaudhartert-db, line 2: I think the name "common" is more appropriate for shared dependencies (e.g. what is currently in `internal/provider.go`). What about putting this in `providers/tf6/tf6.go` or  even `providers/providers.go`?

I like `providers/providers.go` as it reinforces that this is wrapper of more specialized implementations.

internal/providers/internal/provider.go
```diff
@@ -0,0 +1,10 @@
+// Package internal contains the changes used by both internal/providers/sdkv2 and internal/providers/pluginfw packages.
+//
+// Note: This is different from internal/providers/common which contains the changes that *depends* on both:
+// internal/providers/sdkv2 and internal/providers/pluginfw packages. Whereas, internal/providers/internal package contains
+// the changes *used* by both internal/providers/sdkv2 and internal/providers/pluginfw packages.
+package internal
```
renaudhartert-db, line 6: I actually think that `common` would be a good name for this (see other comment). Wondering: any reason we don't merge this into the other `common` package (i.e. the one with the version)?

internal/providers/internal/provider.go
```diff
@@ -0,0 +1,10 @@
...
+
+func GetProviderName() string {
+	return "databricks-tf-provider"
+}
```
renaudhartert-db, line 10: Should this be a constant?
```suggestion
const ProviderName = "databricks-tf-provider"
```

internal/providers/sdkv2/provider.go
```diff
@@ -1,4 +1,7 @@
-package provider
+// Package sdkv2 contains the changes specific to the SDKv2
```
renaudhartert-db, line 1: [Nit] try to keep the same name as the package: `sdkv2/provider.go` -> `sdkv2/skv2.go`.

internal/providers/sdkv2/tests/coverage_test.go
```diff
@@ -1,4 +1,4 @@
-package provider
+package tests
```
renaudhartert-db, line 1: I haven't checked much what these tests are actually testing. Though, if they are testing `sdkv2/provider.go` then move them in `sdkv2` next to the code they are testing.

internal/providers/internal/provider.go
tanmay-db: I agree with having providers.go but I think we shouldn't have `common` package name for shared dependencies going forward since it will lead to loss of readability and it can grow up to contain more code changes. Eg: currently in root common we have reflection, retries, test utils, customise schema etc... all of which should ideally be under their separate package. 

Also from Naming convention as mentioned in go.dev: https://go.dev/blog/package-names
> "Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused."

In the current context of plugin framework and sdkv2 -- they share the same tests that are defined in providers/common/tests in the PR so I thought of moving them to common but we can also move them to top level where we define providers.go, providers_test.go, providers_utils.go. 

Happy to go ahead with moving `internal/providers/internal` to `internal/providers/common` if we decide to move forward with that. + @mgyucht for discussion.

Regarding merging this into the other common package: 
- I think version and provider name should be part of providers package and not the root common package since providers package is more suitable to contain those changes (as they are related to it). 
- Right now version.go is updated automatically when we do the release and I think we should move version.go to providers package as well but this can be done later (+ need to test release automation isn't broken)

internal/providers/internal/provider.go
tanmay-db: Yes, better to have this as a constant. Good catch.

internal/providers/sdkv2/tests/coverage_test.go
tanmay-db: These don't test the provider.go. They check for whether the right stub is generated for unit tests and check for coverage of resources by creating completeness.md. 

One of the tests is skipped and I think these are probably dead code so need to confirm if they are needed or not. Since they are not testing provider.go (now sdkv2.go), so not moving them there.

internal/providers/common/provider.go
tanmay-db: Moved them to providers.go to top level

internal/providers/internal/provider.go
tanmay-db: update: Rest changes have been addressed except if we want to use the `common` package name, will make the change if we decide to move forward with it.

tanmay-db: updated to common

renaudhartert-db (approved):
Thanks Tanmay!

tanmay-db:
Thanks for the review @renaudhartert-db

PR #3943, opened 2024-08-23
[internal] Migrate imports for terraform plugin framework

edwardfeng-db:
## Changes

- Migrated imports following the documentation [here](https://developer.hashicorp.com/terraform/plugin/testing/migrating)
  - 
image

- Made `run` in `init_test` able to take both plugin framework resources as well as sdkv2 resources

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


edwardfeng-db:
Integration tests don't have failures related to this infra import change.

internal/providers/providers.go
```diff
@@ -9,12 +9,37 @@ import (
...
+	sdkV2Provider *schema.Provider
+}
+
+func (o *sdkV2ProviderOption) Apply(options *serverOptions) {
+	options.sdkV2Provider = o.sdkV2Provider
+}
+
+// WithSdkV2Provider takes in a *schema.Provider from sdk v2 and returns a ServerOption.
+// This is typically used in acceptance test for a test step to have a custom provider override.
+func WithSdkV2Provider(sdkV2Provider *schema.Provider) ServerOption {
```
tanmay-db, line 39: This seems to be only used in internal/acceptance/mws_workspaces_test.go? If yes, then probably best to move it to providers_test_utils.go

tanmay-db (approved):
Had one comment, rest LGTM

go.sum
```diff
@@ -109,17 +109,25 @@ github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320 h1:1/D3zfFHttUK
 github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320/go.mod h1:EiZBMaudVLy8fmjf9Npq1dq9RalhveqZG5w/yz3mHWs=
 github.com/hashicorp/go-hclog v1.5.0 h1:bI2ocEMgcVlz55Oj1xZNBsVi900c7II+fWDyV9o+13c=
 github.com/hashicorp/go-hclog v1.5.0/go.mod h1:W4Qnvbt70Wk/zYJryRzDRU/4r0kIg0PVHBcfoyhpF5M=
+github.com/hashicorp/go-hclog v1.6.3 h1:Qr2kF+eVWjTiYmU7Y31tYlP1h0q/X3Nl3tPGdaB11/k=
```
mgyucht, line 112: Can you run `go mod tidy`?

internal/providers/providers.go
```diff
@@ -9,12 +9,37 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-mux/tf6muxserver"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+type serverOptions struct {
+	sdkV2Provider           *schema.Provider
+	pluginFrameworkProvider provider.Provider
+}
+
+type ServerOption interface {
```
mgyucht, line 25: nit: add doccomment for public type

internal/providers/providers.go
```diff
@@ -9,12 +9,37 @@ import (
...
+
+type sdkV2ProviderOption struct {
+	sdkV2Provider *schema.Provider
+}
+
+func (o *sdkV2ProviderOption) Apply(options *serverOptions) {
+	options.sdkV2Provider = o.sdkV2Provider
+}
+
+// WithSdkV2Provider takes in a *schema.Provider from sdk v2 and returns a ServerOption.
```
mgyucht, line 37:
```suggestion
// WithSdkV2Provider allows overriding the SDKv2 provider used when creating a Terraform provider with muxing. 
```

edwardfeng-db:
Update - holding off the merge. There's actually some other acceptance test failure I noticed that's related to this. Investigating

internal/providers/providers.go
```diff
@@ -23,8 +48,22 @@ import (
 // implementations are then combined using a multiplexing server, which allows multiple
 // Protocol v6 providers to be served together. The function returns the multiplexed
 // ProviderServer, or an error if any part of the process fails.
-func GetProviderServer(ctx context.Context) (tfprotov6.ProviderServer, error) {
-	sdkPluginProvider := sdkv2.DatabricksProvider()
+//
+// The function takes in a list of ServerOption objects for acceptance test purposes where we can override
+// and create custom providers.
```
mgyucht, line 53:
```suggestion
// GetProviderServer constructs the Databricks Terraform provider server. By default, it combines the default SDKv2-based provider and the default plugin framework-based provider using muxing. The providers used by the muxed server can be overridden using ServerOptions. 
```

internal/providers/providers.go
mgyucht: I think this needs to stay here since it is part of the (internal) public interface for GetProviderServer. Alternatively, we could have two copies of `GetProviderServer()`, one with only the context argument and a customizable one, but that seems redundant to me.

mgyucht (commented):
Let's make sure all integration tests pass before merging this.

PR #3956, opened 2024-08-27
[Fix] Refactor `databricks_permissions` and allow the current user to set their own permissions

### LINKED ISSUE #2407: [ISSUE] Issue with `databricks_permissions` resource: cannot change permissions when I am owner
### Configuration
```hcl

  job_id = databricks_job.this.id
  access_control {
    user_name        = "my_email@mydomain.com"
    permission_level = "IS_OWNER"
  }

}
```

### Expected Behavior
`terraform apply` works fine and the permissions for the job are properly applied

### Actual Behavior
I get the following error: `Error: cannot create permissions: it is not possible to decrease administrative permissions for the current user`

### Steps to Reproduce
1.  Create a job with Terraform with the permissions block I linked above

1.19.0

### Additional information

- In the UI I see am the creator and owner of the job for which I am trying to set permissions
- I have other jobs where I am owner (they run as my user) but not creator and those work fine when I apply terraform with the same permissions block for that job
- I am a workspace admin
- Also I found a related broken link in the docs: the `removes` link [here](https://github.com/databricks/terraform-provider-databricks/blob/master/docs/resources/permissions.md?plain=1#L11) gives a 404

mgyucht:
## Changes
In https://github.com/databricks/terraform-provider-databricks/commit/c441517af5ed3f2c6d793d64d5cf5d4e1ca0dc68, a check was added to prevent users from assigning any permissions for themselves in `databricks_permissions`. This unfortunately makes it impossible for users to legitimately assign themselves as the owner of certain resources, such as jobs, if they are currently owned by a different principal. This PR removes this unnecessary restriction. If the user requests to set permissions for an object in a way that is incompatible with the object, such as removing themselves as owner, the failure will be propagated from the backend to the user instead. This does not make any changes to the way the ownership ACLs are set up (e.g. for resources that require an owner, like jobs, if the Terraform user did not specify an owner, the provider will still set the current user as the owner).

This PR also refactors the permissions resource substantially. The logic for implementing each resource type's permissions, including the field name, object type and resource-specific modifications, are all colocated with the resource's own definition. The type encapsulating this is called`resourcePermissions`. As a result, the control flow is easier to follow:
* Read reads from the API, customizes the response in a resource-specific way, maps the response to the TF state representation, and stores, or marks as deleted if there are no permissions.
* Create and Update read the desired permissions from ResourceData, perform some validation, perform resource-specific, then puts the update with an owner if not specified.
* Delete resets the ACLs to only admins + resource-specific customizations.

Customizations are defined in the permissions/read and permissions/update packages. For update, a mini expression language is defined to support conditional application of customizations.

Lastly, this PR also migrates the resource to the Databricks SDK.

Fixes #2407.

## Tests
This PR adds integration test coverage for the `databricks_permissions` resource for nearly all supported resource types. I wasn't able to run the integration test for `authorization = "passwords"` because password-based auth is deprecated, nor for serving endpoints because of a likely race condition.

Integration tests cover all permission levels, and all principal types. Included is special edge case testing for root directory and all registered models.

- [ ] `make test` run locally
- [ ] using Go SDK


permissions/resource_permissions_test.go
```diff
@@ -455,37 +455,21 @@ func TestResourcePermissionsCustomizeDiff_ErrorOnCreate(t *testing.T) {
 		access_control {
 			permission_level = "WHATEVER"
 		}`,
-	}.ExpectError(t, "permission_level WHATEVER is not supported with cluster_id objects")
-}
-
-func TestResourcePermissionsCustomizeDiff_ErrorOnPermissionsDecreate(t *testing.T) {
```
mgyucht, line 461: This behavior is changed in this PR: users are allowed to change their permissions as long as they maintain management-level permissions.

permissions/resource_permissions_test.go
```diff
@@ -921,19 +905,19 @@ func TestResourcePermissionsCreate_SQLA_Endpoint_WithOwnerError(t *testing.T) {
...
+				ExpectedRequest: AccessControlChangeListApiRequest{
+					AccessControlList: []AccessControlChangeApiRequest{
 						{
 							UserName:        TestingUser,
 							PermissionLevel: "CAN_USE",
 						},
 						{
 							UserName:        TestingAdminUser,
-							PermissionLevel: "IS_OWNER",
+							PermissionLevel: "CAN_MANAGE",
```
mgyucht, line 916: Order of permissions won't affect the behavior in the API.

permissions/resource_permissions.go
```diff
@@ -108,315 +28,137 @@ type PermissionsAPI struct {
...
+	err = mapping.validate(entity, currentUser)
+	if err != nil {
+		return err
 	}
-	log.Printf("[DEBUG] PUT %s %v", objectID, objectACL)
-	return a.client.Put(a.context, urlPathForObjectID(objectID), objectACL)
-}
-
-// safePutWithOwner is a workaround for the limitation where warehouse without owners cannot have IS_OWNER set
-func (a PermissionsAPI) safePutWithOwner(objectID string, objectACL AccessControlChangeList, originalAcl []AccessControlChange) error {
```
tanmay-db, line 207: Is this something we should also have in Go SDK and hence might be better to move it there?

permissions/acl_customizers.go
```diff
@@ -0,0 +1,103 @@
...
+
+func removeAdminPermissionsCustomizer(ctx aclReadCustomizerContext, acl *iam.ObjectPermissions) (*iam.ObjectPermissions, error) {
+	// Remove all permissions for the 'admins' group.
+	filteredAcl := make([]iam.AccessControlResponse, 0, len(acl.AccessControlList))
+	for _, a := range acl.AccessControlList {
+		if a.GroupName != "admins" {
+			filteredAcl = append(filteredAcl, a)
+		}
+	}
+	acl.AccessControlList = filteredAcl
```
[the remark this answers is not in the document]
tanmay-db: Also since we are always returning nil as error in current implementation -- should we remove returning it / add error pathway?

tanmay-db (approved):
LGTM, had some comments.

permissions/resource_permissions.go
mgyucht: TBH I don't think it really belongs in the Go SDK. I would actually consider this more along the lines of an API issue for the permissions API for warehouses. This may not even be needed anymore now that warehouses and other SQL objects are using the global permissions API endpoints, but I need to test it out.

permissions/acl_customizers.go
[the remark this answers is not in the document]
mgyucht: Good suggestion. I can change this method to return a new ObjectPermissions instance and not modify the original one.

permissions/permission_definitions.go
```diff
@@ -0,0 +1,699 @@
...
+		},
+		{
+			field:             "sql_dashboard_id",
+			objectType:        "dashboard",
+			requestObjectType: "dbsql-dashboards",
+			allowedPermissionLevels: map[string]permissionLevelOptions{
+				"CAN_EDIT":   {isManagementPermission: false},
+				"CAN_RUN":    {isManagementPermission: false},
+				"CAN_MANAGE": {isManagementPermission: true},
+				// This was originally called CAN_VIEW in the preview API, but was renamed to CAN_READ in the GA API.
```
nkvuong, line 560: we are blocking `CAN_READ`, which is a valid configuration here?

permissions/permission_definitions.go
```diff
@@ -0,0 +1,699 @@
...
+				"CAN_VIEW": {isManagementPermission: false},
+			},
+			idMatcher: func(id string) bool {
+				return strings.HasPrefix(id, "/dbsql-dashboards/") || strings.HasPrefix(id, "/sql/dashboards/")
```
nkvuong, line 564: aren't we normalizing the ids to `dashboard` in the read function?

permissions/resource_permissions.go
```diff
@@ -108,315 +28,137 @@ type PermissionsAPI struct {
...
-	}
-	return isDbsqlPermissionsWorkaroundNecessary(objectID)
-}
-
-func isOwnershipWorkaroundNecessary(objectID string) bool {
-	return strings.HasPrefix(objectID, "/jobs") || strings.HasPrefix(objectID, "/pipelines") || strings.HasPrefix(objectID, "/sql/warehouses")
-}
-
-func (a PermissionsAPI) getObjectCreator(objectID string) (string, error) {
+// safePutWithOwner is a workaround for the limitation where warehouse without owners cannot have IS_OWNER set
```
nkvuong, line 31: I wonder if this bug is tracked by the sql warehouse team, as this results in 2 additional functions just for this edge case

nkvuong (commented):
just a few minor comments, I will run a few test locally as well to confirm

permissions/permission_definitions.go
mgyucht: At first, I wasn't sure we would be able to support it, because on the read pathway we would need to convert CAN_READ to CAN_VIEW when users used CAN_VIEW. I've made some tweaks and think it should be possible to support both now since the read pathway has access to the current state.

permissions/permission_definitions.go
mgyucht: Yes, but I think this is needed for `databricks_permissions` resources deployed on older versions of the provider.

nkvuong (approved):

NicholasFiorentini:
I suspect this change has introduced a regression. See https://github.com/databricks/terraform-provider-databricks/issues/4143.

PR #3958, opened 2024-08-27
[Internal] Added `databricks_quality_monitor` resource and `databricks_volumes` data source to plugin framework

tanmay-db:
## Changes

Add support for Quality Monitor resource and Volumes data source in plugin framework. Also add utility method to get workspace and account client with diagnostics.

Subpart of: https://github.com/databricks/terraform-provider-databricks/pull/3751 + Addressed comments from: https://github.com/databricks/terraform-provider-databricks/pull/3893/files#diff-8edec0f6289f63bfb61c7c6fb76fc1a61eec69c33fc2ba3a4ee6e77a8103f85b

Note: The changes for flipping `databricks_quality_monitor_pluginframework` -> `databricks_quality_monitor` in pluginframework along with making this legacy for sdkv2 i.e. `databricks_quality_monitor` -> `databricks_quality_monitor_legacy` will be done in separate PR. Same for `databricks_volumes`

## Tests

Existing unit tests.

Integration and Unit tests for this resource will be merged after this PR is merged.

- [ ] `make test` run locally
- [ ] relevant change in `docs/` folder
- [ ] covered with integration tests in `internal/acceptance`
- [ ] relevant acceptance tests are passing
- [ ] using Go SDK

PR #3975, opened 2024-08-31
[Internal] Add integration tests for volumes and quality monitor plugin framework

edwardfeng-db:
## Changes

- Add integration tests for resource and data source introduced [here](https://github.com/databricks/terraform-provider-databricks/pull/3958)
- Small updates to `ConfigureDataSource` function to handle nil case
- In `data_volumes`, we don't support `computed` tf tag yet, so we need to do that in customizable schema

## Tests

- [x] `make test` run locally
- [x] using Go SDK


tanmay-db (approved):
LGTM. Synced offline about comments on nil Provider Data and moving files to their respective packages.

PR #3979, opened 2024-09-02
[Feature] Library plugin framework migration

edwardfeng-db:
## Changes

- Add library resource to plugin framework, not officially migrating because it has suffix `_pluginframework`
- Added integration test for creation and importing

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/resources/library/resource_library.go
+++ b/internal/providers/pluginfw/resources/library/resource_library.go
@@ -73,14 +73,22 @@ func (r *LibraryResource) Schema(ctx context.Context, req resource.SchemaRequest
 	resp.Schema = schema.Schema{
 		Description: "Terraform schema for Databricks Library",
 		Attributes: tfschema.ResourceStructToSchemaMap(LibraryExtended{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
-			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "cluster_id")
-			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "cran")
-			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "egg")
-			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "jar")
-			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "maven")
-			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "pypi")
-			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "requirements")
-			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "whl")
+			// c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "cluster_id")
+			// c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "cran")
+			// c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "egg")
+			// c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "jar")
+			// c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "maven")
+			// c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "pypi")
+			// c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "requirements")
+			// c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "whl")
+			for field, attribute := range c.ToAttributeMap() {
+				switch attribute.(type) {
+				case tfschema.StringAttributeBuilder:
+					c.AddPlanModifier(stringplanmodifier.RequiresReplace(), field)
+				case tfschema.SingleNestedAttributeBuilder:
+					c.AddPlanModifier(objectplanmodifier.RequiresReplace(), field)
+				}
+			}
 			return c
 		}),
 	}
--- a/internal/providers/pluginfw/resources/library/resource_library.go
+++ b/internal/providers/pluginfw/resources/library/resource_library.go
@@ -1,0 +1,227 @@
+package library
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/terraform-provider-databricks/clusters"
+	"github.com/databricks/terraform-provider-databricks/common"
+	pluginfwcommon "github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/common"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/converters"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/tfschema"
+	"github.com/databricks/terraform-provider-databricks/internal/service/compute_tf"
+	"github.com/databricks/terraform-provider-databricks/libraries"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/objectplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	"github.com/databricks/databricks-sdk-go"
+)
+
+const libraryDefaultInstallationTimeout = 15 * time.Minute
+
+var _ resource.ResourceWithConfigure = &LibraryResource{}
+
+func ResourceLibrary() resource.Resource {
+	return &LibraryResource{}
+}
+
+func parseId(id string) (string, string, error) {
+	split := strings.SplitN(id, "/", 2)
+	if len(split) != 2 {
+		return "", "", fmt.Errorf("invalid library id format, %s", common.TerraformBugErrorMessage)
+	}
+	return split[0], split[1], nil
+}
+
+func readLibrary(ctx context.Context, w *databricks.WorkspaceClient, waitParams compute.Wait, libraryRep string, libraryExtended *LibraryExtended) diag.Diagnostics {
+	res, err := libraries.WaitForLibrariesInstalledSdk(ctx, w, waitParams, libraryDefaultInstallationTimeout)
+	if err != nil {
+		return diag.Diagnostics{diag.NewErrorDiagnostic("failed to wait for library installation", err.Error())}
+	}
+
+	for _, v := range res.LibraryStatuses {
+		thisRep := v.Library.String()
+		if thisRep == libraryRep {
+			// This is not entirely necessary as we can directly write the fields in the config into the state, because there's no computed field.
+			diags := converters.GoSdkToTfSdkStruct(ctx, v.Library, libraryExtended)
+
+			if diags.HasError() {
+				return diags
+			}
+
+			libraryExtended.ClusterId = types.StringValue(waitParams.ClusterID)
+
+			return nil
+		}
+	}
+	return diag.Diagnostics{diag.NewErrorDiagnostic("failed to find the installed library", fmt.Sprintf("failed to find %s on %s", libraryRep, waitParams.ClusterID))}
+}
+
+type LibraryExtended struct {
+	compute_tf.Library
+	ClusterId types.String `tfsdk:"cluster_id"`
+}
+
+type LibraryResource struct {
+	Client *common.DatabricksClient
+}
+
+func (r *LibraryResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = "databricks_library_pluginframework"
+}
+
+func (r *LibraryResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Terraform schema for Databricks Library",
+		Attributes: tfschema.ResourceStructToSchemaMap(LibraryExtended{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "cluster_id")
+			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "cran")
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "egg")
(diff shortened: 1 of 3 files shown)


edwardfeng-db:
Integration tests all passed

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
+package clusters
```
mgyucht, line 1: Why not put this in `package library`?

Also, we should probably use singular nouns for the package name (e.g. library, not libraries)

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+
+func ResourceLibrary() resource.Resource {
+	return &LibraryResource{}
+}
+
+func parseId(id string) (string, string) {
+	split := strings.SplitN(id, "/", 2)
+	if len(split) != 2 {
+		return "unknown", "unknown"
+	}
```
mgyucht, line 38: I know you're porting this directly from the old implementation, but this is just gross...

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+
+			return nil
+		}
+	}
+	return diag.Diagnostics{diag.NewErrorDiagnostic("failed to find the installed library", fmt.Sprintf("failed to find %s on %s", libraryRep, waitParams.ClusterID))}
+}
+
+type LibraryExtended struct {
+	compute_tf.Library
+	ClusterId types.String `tfsdk:"cluster_id" tf:""`
```
mgyucht, line 68: Is `tf:""` necessary? can we omit it if it is not needed?

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+type LibraryResource struct {
+	Client *common.DatabricksClient
+}
+
+func (r *LibraryResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = "databricks_library_pluginframework"
```
mgyucht, line 76: Suggestion: we should use `req.ProviderName` here. Not that it matters too much, it just allows the provider name to be defined in one place.

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (r *LibraryResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Terraform schema for Databricks Library",
+		Attributes: tfschema.ResourceStructToSchemaMap(LibraryExtended{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
+			return c
+		}),
```
mgyucht, line 84: Can we do this?
```suggestion
		Attributes: tfschema.ResourceStructToSchemaMap(LibraryExtended{}, nil),
```

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	lib := LibraryExtended{}
+	resp.Diagnostics.Append(readLibrary(ctx, w, waitParams, libraryRep, &lib)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, lib)...)
+
```
mgyucht, line 113: Do we need to call the `readLibrary` method in `ImportState`? My understanding is that we only need to set whatever attributes the `Read()` method depends on. Basically, parse the ID, and set the clusterId and library state in the resource. Can you try this out?

See https://developer.hashicorp.com/terraform/plugin/framework/resources/import:

> Resources can implement the ImportState method, which must either specify enough Terraform state for the Read method to refresh [resource.Resource](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#Resource) or return an error.

This simplifies this method's implementation, removes redundant API requests, and minimizes need for refactoring of your resource implementation (the read logic can live in the `Read()` method).

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+	resp.Diagnostics.Append(readLibrary(ctx, w, waitParams, libraryRep, &installedLib)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, installedLib)...)
+}
+
+func (r *LibraryResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
mgyucht, line 191: I guess we need this implementation, but should this just fail?

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (r *LibraryResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = "databricks_library_pluginframework"
+}
+
+func (r *LibraryResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Terraform schema for Databricks Library",
+		Attributes: tfschema.ResourceStructToSchemaMap(LibraryExtended{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
```
mgyucht, line 82: Do we need to modify the plan to set all library attributes to the equivalent of ForceNew? Today, this is what is done by `common.Resource` for resources without update defined:
```
		// set ForceNew to all attributes with CRD
		queue := []*schema.Resource{
			{Schema: r.Schema},
		}
		for {
			head := queue[0]
			queue = queue[1:]
			for _, v := range head.Schema {
				if v.Computed {
					continue
				}
				if nested, ok := v.Elem.(*schema.Resource); ok {
					queue = append(queue, nested)
				}
				v.ForceNew = true
			}
			if len(queue) == 0 {
				break
			}
		}
```
I believe all that is needed is to set `<type>planmodifier.RequiresReplace()` plan modifier for all top-level fields.

Let's validate this behavior with an integration test as well.

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+		err := w.Libraries.Uninstall(ctx, compute.UninstallLibraries{
+			ClusterId: clusterID,
+			Libraries: []compute.Library{*v.Library},
+		})
+		if err != nil {
+			resp.Diagnostics.AddError("failed to uninstall library", err.Error())
+		}
+		return
+	}
+	resp.Diagnostics.AddError("failed to uninstall library", fmt.Sprintf("failed to find %s on %s", libraryRep, clusterID))
```
mgyucht, line 235: I see that you're porting the existing implementation, but this doesn't seem like an error to me. E.g. if the library was manually uninstalled. Can you leave a comment here about this?

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+	resp.Diagnostics.Append(converters.TfSdkToGoSdkStruct(ctx, libraryTfSDK, &libGoSDK)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	libraryRep := libGoSDK.String()
+	_, err := clusters.StartClusterAndGetInfo(ctx, w, clusterID)
+	if err != nil {
+		resp.Diagnostics.AddError("failed to start and get cluster", err.Error())
+		return
+	}
```
mgyucht, line 216: Wild that you need to start the cluster to list its libraries... again, seems like this is part of the original implementation.

internal/providers/pluginfw/resources/clusters/resource_library.go
```diff
@@ -0,0 +1,237 @@
...
+		thisRep := v.Library.String()
+		if thisRep == libraryRep {
+
+			diags := converters.GoSdkToTfSdkStruct(ctx, v.Library, libraryExtended)
+
+			if diags.HasError() {
+				return diags
+			}
+
+			libraryExtended.ClusterId = types.StringValue(waitParams.ClusterID)
```
mgyucht, line 58: Noting that we're porting this resource, so this preserves the original behavior, but a question: do we actually need to do this? There are no computed fields in this resource. In theory, we don't even need to update the state at the end of the `Create()` or `Read()` methods.

internal/providers/pluginfw/resources/clusters/resource_library_test.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+func TestAccLibraryCreationPluginFramework(t *testing.T) {
```
mgyucht, line 12: Let's also test the update flow here.

internal/providers/pluginfw/resources/clusters/resource_library_test.go
```diff
@@ -0,0 +1,93 @@
...
+			instance_pool_id = "{env.TEST_INSTANCE_POOL_ID}"
+			autotermination_minutes = 10
+			num_workers = 1
+			spark_conf = {
+				"spark.databricks.repl.allowedLanguages" = "sql,python,r"
+				"spark.databricks.cluster.profile" = "serverless"
+			}
+			custom_tags = {
+				"ResourceClass" = "Serverless"
+			}
```
mgyucht, line 58: What is this configuration exactly?

internal/providers/pluginfw/resources/clusters/resource_library_test.go
```diff
@@ -0,0 +1,93 @@
...
+	acceptance.WorkspaceLevel(t,
+		acceptance.Step{
+			Template: `data "databricks_spark_version" "latest" {
+		}
+		resource "databricks_cluster" "this" {
+			cluster_name = "libs-{var.RANDOM}"
+			spark_version = data.databricks_spark_version.latest.id
+			instance_pool_id = "{env.TEST_INSTANCE_POOL_ID}"
+			autotermination_minutes = 10
+			num_workers = 1
```
mgyucht, line 51: Do we need a worker to test library installation? It will be half as expensive to run this test this way.

mgyucht (changes requested):
Thanks for this! I think we're largely on the right track here, but a few comments to address. I think there is quite a bit of legacy weirdness about this resource, commented on a few places below, so it might be worth leaving a comment at those spots to indicate that e.g. we think this is safe to remove or change but it is this way to preserve the original behavior for now.

internal/providers/pluginfw/resources/clusters/resource_library.go
edwardfeng-db: Yeah we should just make it fail

internal/providers/pluginfw/resources/clusters/resource_library.go
edwardfeng-db: Updated it into a panic, not sure if that sounds good

internal/providers/pluginfw/resources/clusters/resource_library.go
edwardfeng-db: Yes, that works.

internal/providers/pluginfw/resources/clusters/resource_library.go
edwardfeng-db: Yes we can

internal/providers/pluginfw/resources/clusters/resource_library.go
edwardfeng-db: Discussed offline, created ticket for this and not going to address it in this PR

internal/providers/pluginfw/resources/clusters/resource_library_test.go
edwardfeng-db: we can update it to 0 and use a single node cluster

internal/providers/pluginfw/resources/clusters/resource_library.go
edwardfeng-db: Yes it does seem like this is suboptimal in terms of efficiency but I tried doing this and I don't think we can always reverse this serialisation.

For example for `pypi` and `mvn`, we cannot deterministically parse them since the sub fields are just blindly concatenated together without any separator.

edwardfeng-db: As discussed offline, we are not going to support import for now for migration purposes. Will considering modifying the .String() function before introducing the importer

internal/providers/pluginfw/resources/clusters/resource_library_test.go
edwardfeng-db: Added, PTAL

internal/providers/pluginfw/resources/library/resource_library.go
```diff
@@ -0,0 +1,218 @@
...
+		Description: "Terraform schema for Databricks Library",
+		Attributes: tfschema.ResourceStructToSchemaMap(LibraryExtended{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "cluster_id")
+			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "cran")
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "egg")
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "jar")
+			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "maven")
+			c.AddPlanModifier(objectplanmodifier.RequiresReplace(), "pypi")
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "requirements")
+			c.AddPlanModifier(stringplanmodifier.RequiresReplace(), "whl")
```
edwardfeng-db, line 83: Is there a smarter way to do this?

mgyucht:
```go
for field, attribute := range c.ToAttributeMap() {
        switch attribute.(type) {
        case tfschema.StringAttributeBuilder:
                attribute.AddPlanModifier(stringplanmodifier.RequiresReplace)
        case tfschema.NestedAttributeBuilder:
                attribute.AddPlanModifier(objectplanmodifier.RequiresReplace)
        }
}
```

mgyucht (approved):
LGTM, one small suggestion on the schema customization!

PR #3987, opened 2024-09-03
[Fix] Modify error code when library cluster status is not found during read

### LINKED ISSUE #3984: [ISSUE] Issue with `databricks_library` resource
Given a non-pinned cluster that is configured with libraries managed by `databricks_library` resources, when the cluster configuration is deleted from the workspace, library resources are unable to be refreshed and the project enters a broken state.

### Configuration
```hcl
  local_disk = true
}

  long_term_support = true
}

resource "databricks_cluster" "cluster" {
  cluster_name            = "test"
  autotermination_minutes = 20
  num_workers             = 1
}

resource "databricks_library" "library" {
  cluster_id = databricks_cluster.cluster.id

  pypi {
    package = "databricks-cli"
  }
}
```

### Expected Behavior
Project can be redeployed without manual intervention. When the read function encounters a cluster not found error [here](https://github.com/databricks/terraform-provider-databricks/blob/a4b02256e80cc6b8a33160d63aca9363d20ed4ea/clusters/resource_library.go#L77), the library resource should be removed from state and return a nil error.

### Actual Behavior
Errors are propagated and the Terraform project requires manual intervention to be fixed

### Steps to Reproduce
1. Create `databricks_cluster` and `databricks_library` resources
2. Manually terminate the cluster and delete its configuration from the workspace
3. All subsequent terraform commands (refresh, plan, apply etc) for this project will fail

Terraform 1.5.6
Provider 1.51.0

### Is it a regression?
No

### Debug Output
N/A

### Important Factoids
N/A

Sure

bgrams:
## Changes
Set the API error code within `libraries.WaitForLibrariesInstalledSdk` to `NOT_FOUND` when the cluster status is not found.

The error code takes precedence over the status code in the [databricks-go-sdk](https://github.com/databricks/databricks-sdk-go/blob/590d59704619647b958eec09f8faf9fa952ddd86/apierr/unwrap.go#L35-L42) when error checking is performed to determine whether a resource should be [automatically removed from state](https://github.com/databricks/terraform-provider-databricks/blob/a4b02256e80cc6b8a33160d63aca9363d20ed4ea/common/resource.go#L133-L137) during a failed read. Only overriding the status code for a 400 Bad Request response from the `libraries/cluster-status` API otherwise has no effect due to this detail.

Fixes #3984 

## Tests
Unit test added

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_library_test.go
+++ b/clusters/resource_library_test.go
@@ -59,6 +59,26 @@ func TestLibraryCreate(t *testing.T) {
 	}.ApplyNoError(t)
 }
 
+func TestLibraryRead_ClusterDoesNotExist(t *testing.T) {
+	qa.ResourceFixture{
+		Resource: ResourceLibrary(),
+		Fixtures: []qa.HTTPFixture{
+			{
+				Method:   "GET",
+				Resource: "/api/2.0/libraries/cluster-status?cluster_id=abc",
+				Response: map[string]string{
+					"error_code": "INVALID_PARAMETER_VALUE",
+					"message":    "Cluster abc does not exist",
+				},
+				Status: 400,
+			},
+		},
+		Read:    true,
+		Removed: true,
+		ID:      "abc/whl:foo.whl",
+	}.ApplyNoError(t)
+}
+
 func TestLibraryDelete(t *testing.T) {
 	qa.ResourceFixture{
 		Resource: ResourceLibrary(),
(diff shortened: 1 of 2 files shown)


bgrams:
Hi @tanmay-db - I just noticed this addresses the same issue that is resolved by https://github.com/databricks/terraform-provider-databricks/pull/3909, which was merged a few hours after I opened this one. 

The changes are different but the net result is the same, so I'm not sure whether we still want these. Please feel free to close the PR.

Thank you @alexott for the fix 👍

tanmay-db:
Hi @bgrams, thanks for creating the PR. I think we can close this one since as you mentioned the linked PR is merged. Let me double check this.

PR #3988, opened 2024-09-03
Migrate databricks cluster resource and data source to plugin framework

tanmay-db:
##### Changes

Migrate the following to plugin framework
- databricks cluster resource
- databricks cluster data source
- databricks clusers data source

Note: These resources will be suffixed with `_pluginframework` and will be made default in another PR. 
## Tests

- Integration tests
- Unit tests
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/resources/cluster/data_cluster.go
+++ b/internal/providers/pluginfw/resources/cluster/data_cluster.go
@@ -1,0 +1,87 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/terraform-provider-databricks/common"
+	pluginfwcommon "github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/common"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/tfschema"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func DataSourceCluster() datasource.DataSource {
+	return &ClusterDataSource{}
+}
+
+var _ datasource.DataSourceWithConfigure = &ClusterDataSource{}
+
+type ClusterDataSource struct {
+	Client *common.DatabricksClient
+}
+
+type ClusterInfo struct {
+	Id          types.String            `json:"id,omitempty" tf:"computed"`
+	ClusterId   types.String            `json:"cluster_id,omitempty" tf:"computed"`
+	Name        types.String            `json:"cluster_name,omitempty" tf:"computed"`
+	ClusterInfo *compute.ClusterDetails `json:"cluster_info,omitempty" tf:"computed"`
+}
+
+func (d *ClusterDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = "databricks_cluster_pluginframework"
+}
+
+func (d *ClusterDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: tfschema.DataSourceStructToSchemaMap(ClusterInfo{}, nil),
+	}
+}
+
+func (d *ClusterDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if d.Client == nil {
+		d.Client = pluginfwcommon.ConfigureDataSource(req, resp)
+	}
+}
+
+func (d *ClusterDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	w, diags := d.Client.GetWorkspaceClient()
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var clusterInfo ClusterInfo
+	diags = req.Config.Get(ctx, &clusterInfo)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if clusterInfo.Name.ValueString() != "" {
+		clusters, err := w.Clusters.ListAll(ctx, compute.ListClustersRequest{})
+		if err != nil {
+			resp.Diagnostics.AddError("failed to list clusters", err.Error())
+			return
+		}
+		namedClusters := []compute.ClusterDetails{}
+		for _, clst := range clusters {
+			cluster := clst
+			if cluster.ClusterName == clusterInfo.Name.ValueString() {
+				namedClusters = append(namedClusters, cluster)
+			}
+		}
+		if len(namedClusters) == 0 {
+			resp.Diagnostics.AddError(fmt.Sprintf("there is no cluster with name '%s'", clusterInfo.Name.ValueString()), "")
+			return
+		}
+		if len(namedClusters) > 1 {
+			resp.Diagnostics.AddError(fmt.Sprintf("there is more than one cluster with name '%s'", clusterInfo.Name.ValueString()), "")
+			return
+		}
+		clusterInfo.ClusterInfo = &namedClusters[0]
+		resp.State.Set(ctx, clusterInfo)
+	}
+}
--- a/internal/providers/pluginfw/resources/cluster/data_clusters.go
+++ b/internal/providers/pluginfw/resources/cluster/data_clusters.go
@@ -1,0 +1,79 @@
+package cluster
+
+import (
+	"context"
+	"strings"
+
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/terraform-provider-databricks/common"
+	pluginfwcommon "github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/common"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/tfschema"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func DataSourceClusters() datasource.DataSource {
+	return &ClustersDataSource{}
+}
+
+var _ datasource.DataSourceWithConfigure = &ClustersDataSource{}
+
+type ClustersDataSource struct {
+	Client *common.DatabricksClient
+}
+
+type ClustersInfo struct {
+	Id                  string   `json:"id,omitempty" tf:"computed"`
+	Ids                 []string `json:"ids,omitempty" tf:"computed,slice_set"`
+	ClusterNameContains string   `json:"cluster_name_contains,omitempty"`
+}
+
+func (d *ClustersDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = "databricks_clusters_pluginframework"
+}
+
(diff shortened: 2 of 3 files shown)


internal/providers/pluginfw/resources/qualitymonitor/tests/resource_quality_monitor_test.go
```diff
@@ -48,7 +48,7 @@ resource "databricks_sql_table" "myInferenceTable" {
 
 `
 
-func TestUcAccQualityMonitorPluginFramework(t *testing.T) {
+func TestUcAccQualityMonitor(t *testing.T) {
```
tanmay-db, line 51: Since these are under pluginfw package, we can remove PluginFramework from the test name

internal/providers/pluginfw/resources/volume/data_volumes.go
```diff
@@ -37,10 +37,7 @@ func (d *VolumesDataSource) Metadata(ctx context.Context, req datasource.Metadat
 
 func (d *VolumesDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
 	resp.Schema = schema.Schema{
-		Attributes: tfschema.DataSourceStructToSchemaMap(VolumesList{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
-			c.SetComputed("ids")
-			return c
-		}),
+		Attributes: tfschema.DataSourceStructToSchemaMap(VolumesList{}, nil),
```
tanmay-db, line 40: We can set this to nil and move computed to struct tags.

internal/providers/pluginfw/resources/volume/data_volumes.go
```diff
@@ -69,7 +66,6 @@ func (d *VolumesDataSource) Read(ctx context.Context, req datasource.ReadRequest
 	if err != nil {
 		if apierr.IsMissing(err) {
 			resp.State.RemoveResource(ctx)
-			return
```
tanmay-db, line 72: Data source shouldn't return if the resource is missing. It should error. This is the same behaviour in SDKv2

internal/providers/pluginfw/pluginfw.go
```diff
@@ -48,6 +49,7 @@ func (p *DatabricksProviderPluginFramework) Resources(ctx context.Context) []fun
 func (p *DatabricksProviderPluginFramework) DataSources(ctx context.Context) []func() datasource.DataSource {
 	return []func() datasource.DataSource{
 		volume.DataSourceVolumes,
+		cluster.DataSourceCluster,
```
mgyucht, line 52: Can we sort these alphabetically?

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+type ClusterInfo struct {
+	ClusterId   types.String               `tfsdk:"cluster_id" tf:"optional,computed"`
+	Name        types.String               `tfsdk:"cluster_name" tf:"optional,computed"`
+	ClusterInfo *compute_tf.ClusterDetails `tfsdk:"cluster_info" tf:"optional,computed"`
+}
+
+func (d *ClusterDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = "databricks_cluster_pluginframework"
```
mgyucht, line 36: One thing I realized: the prefix `databricks` should be supplied in the `req` field, so you can do
```go
resp.TypeName = req.ProviderTypeName + "_cluster_pluginframework"
```

I realize that the implementation of `Metadata()` for the provider is probably incorrect. It is currently `databricks-tf-provider`. However, the docs for the TypeName are:
```
	// TypeName should be the provider type. For example, examplecloud, if
	// the intended resource or data source types are examplecloud_thing, etc.
```
This probably needs to be changed to `databricks`.

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+	resp.Diagnostics.Append(req.Config.Get(ctx, &clusterInfo)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if clusterInfo.Name.ValueString() != "" {
+		clusters, err := w.Clusters.ListAll(ctx, compute.ListClustersRequest{})
+		if err != nil {
+			if apierr.IsMissing(err) {
+				resp.State.RemoveResource(ctx)
+			}
```
mgyucht, line 68: When does `list` return missing? It would just return an empty list?

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+			if cluster.ClusterName == clusterInfo.Name {
+				namedClusters = append(namedClusters, cluster)
+			}
+		}
+		if len(namedClusters) == 0 {
+			resp.Diagnostics.AddError(fmt.Sprintf("there is no cluster with name '%s'", clusterInfo.Name.ValueString()), "")
+			return
+		}
+		if len(namedClusters) > 1 {
+			resp.Diagnostics.AddError(fmt.Sprintf("there is more than one cluster with name '%s'", clusterInfo.Name.ValueString()), "")
```
mgyucht, line 86: Suggestion: let's include the IDs of the clusters with this name in the `detail` parameter, as well as advice for the user (e.g. "When specifying a cluster name, the name must be unique. Alternatively, specify the cluster by ID using the `cluster_id` attribute.")

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+				resp.State.RemoveResource(ctx)
+			}
+			resp.Diagnostics.AddError("failed to list clusters", err.Error())
+			return
+		}
+		var clustersTfSDK []compute_tf.ClusterDetails
+		converters.GoSdkToTfSdkStruct(ctx, clusters, clustersTfSDK)
+		namedClusters := []compute_tf.ClusterDetails{}
+		for _, clst := range clustersTfSDK {
+			cluster := clst
```
mgyucht, line 76: Why this?

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+		clusters, err := w.Clusters.ListAll(ctx, compute.ListClustersRequest{})
+		if err != nil {
+			if apierr.IsMissing(err) {
+				resp.State.RemoveResource(ctx)
+			}
+			resp.Diagnostics.AddError("failed to list clusters", err.Error())
+			return
+		}
+		var clustersTfSDK []compute_tf.ClusterDetails
+		converters.GoSdkToTfSdkStruct(ctx, clusters, clustersTfSDK)
```
mgyucht, line 73: Do we need to check if this fails?

internal/providers/pluginfw/resources/cluster/tests/data_cluster_test.go
```diff
@@ -0,0 +1,20 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+)
+
+func TestAccDataSourceCluster(t *testing.T) {
+	acceptance.WorkspaceLevel(t, acceptance.Step{
```
mgyucht, line 10: Can we add test cases for the other pathways as well? E.g. cluster by name, multiple clusters with the same name, no such cluster with the given name/ID?

internal/providers/pluginfw/resources/cluster/tests/data_cluster_test.go
```diff
@@ -0,0 +1,20 @@
...
+		Template: `
+		data "databricks_cluster_pluginframework" "this" {
+			cluster_id = "{env.TEST_DEFAULT_CLUSTER_ID}"
+		}
+		
+		output "cluster_info" {
+			value = data.databricks_cluster_pluginframework.this.cluster_info
```
mgyucht, line 17: Is this output needed?

internal/providers/pluginfw/tfschema/struct_to_schema.go
```diff
@@ -110,6 +186,11 @@ func typeToSchema(v reflect.Value) map[string]AttributeBuilder {
 	return scm
 }
 
+func fieldIsComputed(field reflect.StructField) bool {
```
mgyucht, line 189: Can we add a test case for this?

mgyucht (commented):
Awesome work, some initial comments.

internal/providers/pluginfw/resources/cluster/data_cluster.go
tanmay-db: Thanks for the review, sounds good, since this needs to be done for quality monitor and volumes as well, I will do this in a separate PR to isolate changes.

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (d *ClusterDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if d.Client == nil {
+		d.Client = pluginfwcommon.ConfigureDataSource(req, resp)
+	}
+}
+
+func ValidateClustersList(ctx context.Context, clusters []compute_tf.ClusterDetails, clusterName string) diag.Diagnostics {
```
mgyucht, line 53: Why does this need to be exported?

internal/providers/pluginfw/resources/cluster/tests/acceptance/data_cluster_test.go
```diff
@@ -0,0 +1,45 @@
...
+func TestAccDataSourceClusterByName(t *testing.T) {
+	clusterName := acceptance.RandomName("cluster-test-name-")
+	acceptance.WorkspaceLevel(t, acceptance.Step{
+		Template: sparkVersionDataSourceTemplate + clusterResourceTemplate("this", clusterName),
+	}, acceptance.Step{
+		Template: sparkVersionDataSourceTemplate + clusterResourceTemplate("this", clusterName) + fmt.Sprintf(`
+		data "databricks_cluster_pluginframework" "this" {
+			cluster_name = "%s"
+		}`, clusterName),
+	})
```
mgyucht, line 44: Suggestion: instead of creating a new cluster (which is kind of slow), why don't we just use the cluster name of the test_default_cluster_id that we just fetched?

mgyucht (commented):
One small suggestion about the integration test. Otherwise, I don't think that we need separate testing packages. In golang, typically, the source code of tests resides in the same directory as the resources under test. I think that shouldn't be a problem to continue here. If we need to distinguish acceptance tests and unit tests in separate files, we can name acceptance tests `_acc_test.go`.

internal/providers/pluginfw/resources/cluster/tests/acceptance/data_cluster_test.go
tanmay-db: We don’t currently export them in env variables in our test infra so in cases the name changes (which is highly unlikely). I think going ahead with relying on default cluster’s name to be "DEFAULT Test Cluster"  should be good. It would be good to have an environment variable for this as well. I will create a tracking ticket for that.

internal/providers/pluginfw/resources/cluster/data_cluster.go
tanmay-db: Regarding having different directory for unit test — I was thinking that there would be minimum 9 files at top level (2 data sources, 1 resource and their unit and integration tests) so navigating might become difficult. But naming them _acc_test.go  to differentiate also works.

We can export ValidateClustersList (Unit tests) by keeping it in same <package> but acceptance tests have to be in <package>_test due to cyclic dependency.

mgyucht (approved):

Issue #3432, opened 2024-04-03
[FEATURE] Request "databricks_groups"  data resource to list all the groups in the databricks account.


db-wenxin:
Hi Team,

Right now the terraform provider only provides [databricks_group](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/group) data resource to query one databricks group at once and it currently does not have "skip_existing" or "skip_change" option.

It would be super helpful to have a **_databricks_groups_** data resource enabling users to query all existing groups using one data resource. This addition would facilitate the implementation of more complex logic directly using Terraform code.

Thank you very much!

simonvanderveldt:
Have some use-cases with SCIM provisioned groups we want to create some resources for that currently require hardcoding the groups because we have no way of looking them up. Lookup should ideally be possible both on workspace and account level (which seems to be impossible with the CLI?).

[edit] Some example use-cases we've run into thus far:
- Creating a serverless budget policy per team/group we onboard
- Give the teams/groups we onboard access to system tables
- Automatically give access/permissions to catalogs

PR #3992, opened 2024-09-04
[Feature] Added "databricks_groups" data resource to list all the groups in the databricks account.

sergiumpopa:
## Changes
Added a `databricks_groups` data resource to list all groups in the account - it addresses #3432.
The resource returns only the `display_name` field from the groups, which can then be used with the `databricks_group` data resource to retrieve more information.

It accepts `filter` as an argument, with the same functionality as the [API parameter](https://docs.databricks.com/api/workspace/groups/list#filter).

## Tests

- [x] `make test` run locally
- [x] using Go SDK


docs/data-sources/groups.md
```diff
@@ -0,0 +1,58 @@
...
+  value = data.databricks_pipelines.this.display_names
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `filter` - (Optional) Query by which the results have to be filtered. Supported operators are equals(`eq`), contains(`co`), starts with(`sw`) and not equals(`ne`).
```
alexott, line 38: We need to link the documentation about filter syntax. Like this: https://docs.databricks.com/api/workspace/groups/list

internal/acceptance/data_groups_test.go
```diff
@@ -0,0 +1,56 @@
...
+resource "databricks_group" "group2" {
+	display_name = "tf-groupfoo-{var.RANDOM}"
+}
+resource "databricks_group" "group3" {
+	display_name = "tf-groupbar-{var.RANDOM}"
+}
+
+#data "databricks_groups" "all" {}
+
+data "databricks_groups" "this" {
```
alexott, line 24: You need to add `depends_on` on the group resources, otherwise they could be created after data source is executed

scim/data_groups.go
```diff
@@ -0,0 +1,34 @@
...
+
+// DataSourceGroups searches from groups based on filter
+func DataSourceGroups() common.Resource {
+	type groupData struct {
+		DisplayNameFilter string   `json:"filter,omitempty"`
+		DisplayNames      []string `json:"display_names,omitempty" tf:"computed,slice_set"`
+	}
+
+	return common.WorkspaceData(func(ctx context.Context, data *groupData, w *databricks.WorkspaceClient) error {
+		groupSearch := iam.ListGroupsRequest{}
```
alexott, line 19: You need to specify that you want to return only display names, otherwise it will fetch all attributes and will fail on big workspaces.

Also, you need to configure page size.

scim/data_groups.go
```diff
@@ -0,0 +1,34 @@
...
+		if data.DisplayNameFilter != "" {
+			groupSearch = iam.ListGroupsRequest{Filter: data.DisplayNameFilter, Count: 100}
```
alexott, line 21: just do it as:

```
groupSearch.Filter = data.DisplayNameFilter
```

And `Count` defines maximum returned results, not page size

scim/data_groups_test.go
```diff
@@ -0,0 +1,97 @@
...
+		Resource:    DataSourceGroups(),
+		Read:        true,
+		NonWritable: true,
+		ID:          "_",
+	}.ExpectError(t, "i'm a teapot")
+}
+
+func TestDataSourceGroups(t *testing.T) {
+	qa.ResourceFixture{
+		Fixtures: []qa.HTTPFixture{
```
alexott, line 21: It's better to use Mock framework: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/data_external_location_test.go#L14

scim/scim.go
```diff
@@ -120,7 +120,7 @@ type GroupList struct {
 	StartIndex   int32   `json:"startIndex,omitempty"`
 	ItemsPerPage int32   `json:"itemsPerPage,omitempty"`
 	Schemas      []URN   `json:"schemas,omitempty"`
-	Resources    []Group `json:"resources,omitempty"`
+	Resources    []Group `json:"Resources,omitempty"`
```
alexott, line 123: Hmmm, we need to understand when this has changed...

alexott (changes requested):
We also need to discuss if it should work on account level as well

scim/data_groups.go
sergiumpopa: Are you sure `Count` is not used for page size in this particular api? This is what I saw in the [api model](https://github.com/databricks/databricks-sdk-go/blob/37cb031019b9cbe7e7388e6a14bcc6e54bbc13b5/service/iam/model.go#L486)

internal/acceptance/data_groups_test.go
sergiumpopa: Done

docs/data-sources/groups.md
```diff
@@ -0,0 +1,58 @@
...
+output "all_groups" {
+  value = data.databricksdatabricks_groups.all.display_names
+}
+```
+
+Get groups whose displayName contains `foo` or displayName contains `bar`.:
+
+```hcl
+data "databricks_groups" "this" {
+  filter = "displayName co "foo" or displayName co "bar""
```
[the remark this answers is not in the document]
sergiumpopa: You are correct, this was an oversight. Addressed now

docs/data-sources/groups.md
sergiumpopa: Done

scim/data_groups.go
sergiumpopa: Done

scim/data_groups_test.go
sergiumpopa: Done

scim/data_groups.go
alexott: Ah, you're right - SCIM API uses different paging method :-(

simonvanderveldt:
:wave: Has there been any progress/conclusions on the open questions for this?
Running into some use-cases where I need to get all the groups in the account.

simonvanderveldt:
Definitely still eagerly waiting for this functionality :)

PR #3993, opened 2024-09-04
[Internal] Make test utils public and move integration test for quality monitor

edwardfeng-db:
## Changes

- Make `Step`, `WorkspaceLevel`, `AccountLevel`, `UnityWorkspaceLevel`, `UnityAccountLevel` public 
- Rename `init_test.go` to `init.go` because variables exported from files ending with `_test` cannot be imported
- Moved integration test for quality monitor next to the resource definition

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
+++ b/internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
@@ -1,8 +1,10 @@
-package acceptance
+package qualitymonitor_test
 
 import (
 	"os"
 	"testing"
+
+	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
 )
 
 var commonPartQualityMonitoring = `resource "databricks_catalog" "sandbox" {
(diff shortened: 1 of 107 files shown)


internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
```diff
@@ -0,0 +1,153 @@
+package qualitymonitor_test
```
tanmay-db, line 1: I think this should be qualitymonitor only? 

Also, might be good to move volumes data source test to under volume package as well.

tanmay-db (approved):

internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
edwardfeng-db: This is an interesting point. There's some circular dependency issue with it just being `qualitymonitor` because the `init.go` file imports some provider factory, and provider factory indirectly imports from `qualitymonitor`

edwardfeng-db: So I think the easiest workaround is just to give the test file a different package name

PR #3994, opened 2024-09-04
[Internal] Support import in acceptance test + adding import state for quality monitor

edwardfeng-db:
## Changes

- Add `ImportState` method for `quality monitor` resource
- Make acceptance test infrastructure able to test importing behavior
- Add integration test for import state for quality monitor

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/acceptance/init.go
+++ b/internal/acceptance/init.go
@@ -56,6 +56,27 @@ func UnityAccountLevel(t *testing.T, steps ...Step) {
 	run(t, steps)
 }
 
+// BuildImportStateIdFunc constructs a function that returns the id attribute of a target resouce from the terraform state.
+// This is a helper function for conveniently constructing the ImportStateIdFunc field for a test step.
+func BuildImportStateIdFunc(resourceId, attr string) func(*terraform.State) (string, error) {
+	return func(s *terraform.State) (string, error) {
+		// Find the resource in the Terraform state.
+		rs, ok := s.RootModule().Resources[resourceId]
+		if !ok {
+			return "", fmt.Errorf("resource not found in state: %s", resourceId)
+		}
+
+		// Access the attribute directly from the state.
+		targetAttr := rs.Primary.Attributes[attr]
+		if targetAttr == "" {
+			return "", fmt.Errorf("attribute '%s' not found or empty in the resource", attr)
+		}
+
+		// Return the found attribute or the ID needed for the import.
+		return targetAttr, nil
+	}
+}
+
 // A step in a terraform acceptance test
 type Step struct {
 	// Terraform HCL for resources to materialize in this test step.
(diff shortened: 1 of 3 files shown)


internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
```diff
@@ -151,3 +153,43 @@ func TestUcAccUpdateQualityMonitorPluginFramework(t *testing.T) {
...
+			ImportState:  true,
+			ResourceName: "databricks_quality_monitor_pluginframework.testMonitorInference",
+			ImportStateIdFunc: func(s *terraform.State) (string, error) {
+				// Fetch the output value of 'table_name' from the state
+				rs, ok := s.RootModule().Outputs["table_name"]
+				if !ok || rs.Value == "" {
+					return "", fmt.Errorf("table_name output not found in the first step")
+				}
+				return rs.Value.(string), nil
+			},
```
mgyucht, line 192: Can we set `ImportStateVerify` to `true` here? I believe this will check that import resulted in the same state as the resource we created above. If we do, we may need to set `ImportStateVerifyIdentifierAttribute` to `table_name` as that is the "id" of the resource.

mgyucht, line 192: We should probably make a helper function for this, as this will be a very common case for testing import:
```go
func RetrieveAttributeFromResourceFunc(resourceId, attr string) func(*terraform.State) (string, error) {
  return func(s *terraform.State) (string, error) {
    rs, ok := s.RootModule().Resources[resourceId]
    if !ok {
      return "", fmt.Errorf("resource %s does not exist", resourceId)
    }
    id, ok := rs.Primary.Attributes[attr]
    if !ok || id.Value == "" {
      return "", fmt.Errorf("attribute %s not defined on resource %s", attr, resourceId)
    }
    return id, nil
  }
}
```

internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
```diff
@@ -151,3 +153,43 @@ func TestUcAccUpdateQualityMonitorPluginFramework(t *testing.T) {
...
+				  timestamp_col = "timestamp"
+				  prediction_col = "prediction"
+				  model_id_col = "model_id"
+				  problem_type = "PROBLEM_TYPE_REGRESSION"
+				} 
+			}
+
+			output "table_name" {
+				value = databricks_quality_monitor_pluginframework.testMonitorInference.table_name
+			}
```
mgyucht, line 179: Do we need to include an output for this test?

internal/acceptance/init.go
```diff
@@ -74,9 +74,16 @@ type Step struct {
...
-	ImportStateVerify         bool
-	ProtoV6ProviderFactories  map[string]func() (tfprotov6.ProviderServer, error)
+
+	// If true, will test the functionality of ImportState by importing the resource with ResourceName (must be set) and the ID of that resource.
+	// ID can be supplied with either ImportStateId or ImportStateIdFunc.
+	ImportState              bool
+	ImportStateId            string
+	ImportStateIdFunc        func(*terraform.State) (string, error)
+	ImportStateVerify        bool
+	ProtoV6ProviderFactories map[string]func() (tfprotov6.ProviderServer, error)
```
mgyucht, line 84: Could you move this field lower down, so all import-related fields are next to one another?

mgyucht (commented):
Few questions/suggestions. Thanks for adding this test!

internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
edwardfeng-db: Right, we don't, let me take it out

internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor_test.go
edwardfeng-db: Great idea, updated

mgyucht (approved):
Looks great!

PR #3998, opened 2024-09-05
[Internal] Migrate `databricks_clusters` data source to plugin framework

tanmay-db:
# WIP
- To be merged after: https://github.com/databricks/terraform-provider-databricks/pull/3988
- Integration tests are passing but we need to have set instead of list  so that's wip. 

## Changes

Migrate `databricks_clusters` data source to plugin framework 

## Tests

Integration tests. 
- [ ] `make test` run locally
- [ ] using Go SDK


internal/providers/pluginfw/resources/cluster/data_clusters.go
```diff
@@ -0,0 +1,81 @@
...
+	Client *common.DatabricksClient
+}
+
+type ClustersInfo struct {
+	Ids                 []types.String `tfsdk:"ids" tf:"optional,computed"`
+	ClusterNameContains types.String   `tfsdk:"cluster_name_contains" tf:"optional"`
+}
+
+func (d *ClustersDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = "databricks_clusters_pluginframework"
```
mgyucht, line 33: Ditto here

internal/providers/pluginfw/resources/cluster/data_clusters.go
```diff
@@ -0,0 +1,81 @@
...
+	resp.Diagnostics.Append(req.Config.Get(ctx, &clustersInfo)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	clusters, err := w.Clusters.ListAll(ctx, compute.ListClustersRequest{})
+	if err != nil {
+		if apierr.IsMissing(err) {
+			resp.State.RemoveResource(ctx)
+		}
```
mgyucht, line 65: Would this happen? Is it an error if there are no clusters? (Probably not.)

internal/providers/pluginfw/resources/cluster/data_clusters.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	ids := make([]types.String, 0, len(clusters))
+	name_contains := strings.ToLower(clustersInfo.ClusterNameContains.ValueString())
+	for _, v := range clusters {
+		match_name := strings.Contains(strings.ToLower(v.ClusterName), name_contains)
+		if name_contains != "" && !match_name {
+			continue
+		}
+		ids = append(ids, types.StringValue(v.ClusterId))
```
mgyucht, line 77: strings.Contains(any string, "") should always be true.
```suggestion
		if match_name {
			ids = append(ids, types.StringValue(v.ClusterId))
		}
```

internal/providers/pluginfw/resources/cluster/data_clusters.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		if apierr.IsMissing(err) {
+			resp.State.RemoveResource(ctx)
+		}
+		resp.Diagnostics.AddError("failed to list clusters", err.Error())
+		return
+	}
+
+	ids := make([]types.String, 0, len(clusters))
+	name_contains := strings.ToLower(clustersInfo.ClusterNameContains.ValueString())
```
mgyucht, line 71: can we use camelCase?

mgyucht (commented):
Initial comments

edwardfeng-db (commented):
Maybe the work for setComputed could be in a separate pr?

internal/providers/pluginfw/resources/cluster/data_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+var _ datasource.DataSourceWithConfigure = &ClusterDataSource{}
+
+type ClusterDataSource struct {
+	Client *common.DatabricksClient
+}
+
+type ClusterInfo struct {
+	ClusterId   types.String               `tfsdk:"cluster_id" tf:"optional,computed"`
+	Name        types.String               `tfsdk:"cluster_name" tf:"optional,computed"`
+	ClusterInfo *compute_tf.ClusterDetails `tfsdk:"cluster_info" tf:"optional,computed"`
```
edwardfeng-db, line 32: I think we should also try adding a proto level annotation for making a field computed in the tf configuration, so the generated tfsdk structs will have these

Issue #3866, opened 2024-08-08
[ISSUE] Tables with struct subcolumns which have dot in their name are redeployed on  `databricks_sql_table` resource


p-venclovas:
### Configuration

```hcl
resource "databricks_sql_table" "test_view" {
  provider                = databricks.workspace
  cluster_id              = data.databricks_cluster.test_cluster.id
  catalog_name            = "test"
  schema_name             = "test"
  name                    = "table_test"
  table_type              = "EXTERNAL"
  storage_credential_name = "test"
  data_source_format = "DELTA"
  storage_location = "test"
  view_definition = "select 1,\n\n\n\2,\t3"

  column {
    name     = "col1"
    type     = "struct<`x.y`:double>"
    nullable = true
   }

```

### Expected Behavior

After initial terraform apply, resource would be deployed. Running terraform plan afterwards, no changes should be detected in the resource.

### Actual Behavior

After initial terraform apply, resource was deployed. Running terraform plan afterwards, changes were detected in the struct column type.

Column type that was read and returned by the provider looks like: `struct<x.y:double>` (backticks removed) thus changes in the column type are detected.

### Steps to Reproduce

1. terraform apply
2. After successfully creating resource, run terraform plan

### Terraform and provider versions

Terraform v1.5.7. Databricks latest release.

### Would you like to implement a fix?

Yes, I would be happy to share the fix with the community.

p-venclovas:
Raised PR https://github.com/databricks/terraform-provider-databricks/pull/4001

PR #4001, opened 2024-09-06
[Fix] Tables with struct subcolumns which have dot in their name are redeployed on databricks_sql_table resource #3866

p-venclovas:
## Changes

Backticks are removed from the column type in order for struct subfield names to be compatible with the subfield names deployed. These subfields names are part of the column type. Example struct<\`x.z\`:string> == struct\<x.z:string>

## Tests

- [x] `make test` run locally
- [x] using Go SDK

tested using internal integration tests where tables were created and terraform plans were reviewed to have no changes.

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -489,7 +489,8 @@ var columnTypeAliases = map[string]string{
 }
 
 func getColumnType(columnType string) string {
-	caseInsensitiveColumnType := strings.ToLower(columnType)
+	columnTypeNoBackticks := strings.ReplaceAll(columnType, "`", "")
+	caseInsensitiveColumnType := strings.ToLower(columnTypeNoBackticks)
 	if alias, ok := columnTypeAliases[caseInsensitiveColumnType]; ok {
 		return alias
 	}


mgyucht (changes requested):
Thank you for raising this PR. The code changes look fine, but we should definitely add a test to ensure this works as expected. You mentioned running internal integration tests. Can you add one in `internal/acceptance/sql_table_test.go`? The integration tests are not very straightforward to run for external contributors now, but if you can at least add one, that would be terrific, and I can help make sure it works as expected.

tanmay-db:
Hi @p-venclovas, thanks for reaching out. This was recently raised -- regarding earlier comments, could you help add integration tests as mentioned here: https://github.com/databricks/terraform-provider-databricks/pull/4001#pullrequestreview-2345164000. It should be similar to other integration tests. Then I can help run these tests on our infrastructure / add changes if required.

nkvuong:
@tanmay-db this one is difficult to create an integration test, as it requires an external table with struct column - I tested this locally and it works

catalog/resource_sql_table.go
```diff
@@ -561,7 +561,8 @@ var columnTypeAliases = map[string]string{
 }
 
 func getColumnType(columnType string) string {
-	caseInsensitiveColumnType := strings.ToLower(columnType)
+	columnTypeNoBackticks := strings.ReplaceAll(columnType, "`", "")
+	caseInsensitiveColumnType := strings.ToLower(columnTypeNoBackticks)
```
renaudhartert-db, line 568: Could you add a comment to help future readers understand why we are doing this transformation? My understanding is that this client side normalization is necessary to match normalization that is happening on the backend side. Without that, TF will consider that the table are different while they are actually the same server side. 

Side question: Is that the only type of normalization that happens on the server side?

renaudhartert-db (changes requested):

catalog/resource_sql_table.go
alexott: Here is another example: https://github.com/databricks/terraform-provider-databricks/pull/4422

p-venclovas: @renaudhartert-db Yes, you got it right. I have added the comment to explain the reasoning behind it.

alexott (approved):

renaudhartert-db (approved):

Issue #3865, opened 2024-08-08
[ISSUE] Empty lines or tabs inside SQL query of view definition will trigger redeployment of  `databricks_sql_table` resource


p-venclovas:
### Configuration

```hcl
resource "databricks_sql_table" "test_view" {
  provider                = databricks.workspace
  cluster_id              = data.databricks_cluster.tabledeploy_cluster.id
  catalog_name            = "test"
  schema_name             = "test"
  name                    = "view_test"
  table_type              = "VIEW"
  view_definition = "select 1,\n\n\n\2,\t3"
}
```

### Expected Behavior
After initial terraform apply, resource would be deployed. Running terraform plan afterwards, no changes should be detected in the resource.

### Actual Behavior
After initial terraform apply, resource was deployed. Running terraform plan afterwards, changes were detected in the view definition.

View definition read and returned by the provider looks like: "select 1,\n2,    3" thus changes in the view_definition are detected.

### Steps to Reproduce

1. `terraform apply`
2.  After successfully creating resource, run `terraform plan`

### Terraform and provider versions

Terraform v1.5.7. Databricks latest release.

### Debug Output

After debugging databricks terraform provider,it became clear that when databricks terraform provider is reading the state of view, view definition is returned with the following formatting applied:

* Empty new lines are removed. E.g. "1\n\n\n2" -> "1\n2"
* Tab symbol is replaced by 4 spaces. E.g. "1\t2" -> "1    2"

### Would you like to implement a fix?

Yes, I would be happy to share the fix with the community.

There was an [attempt](https://github.com/databricks/terraform-provider-databricks/pull/3498) to resolve this for empty new lines specifically, but it was never merged.

p-venclovas:
Raised PR https://github.com/databricks/terraform-provider-databricks/pull/4001

PR #4003, opened 2024-09-06
[FIX] Empty lines or tabs inside SQL query of view definition will trigger redeployment of databricks_sql_table resource #3865

p-venclovas:
## Changes

Added formatting for new lines and tabs. This formatting is already applied on databricks side and it shows up when terraform read the state of view definition.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


tanmay-db:
Hi @p-venclovas, same here (https://github.com/databricks/terraform-provider-databricks/pull/4001#issuecomment-3046624183), could you add an integration tests.

nkvuong:
@tanmay-db the existing integration test was modified to account for this change, so this should be good to go

PR #4009, opened 2024-09-11
[Internal] Add `AddPlanModifer` method for AttributeBuilder

edwardfeng-db:
## Changes

- Add `AddPlanModifier` methods for all AttributeBuilders so that we can do things equivalent to set supress diff or force new 

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/tfschema/bool_attribute.go
+++ b/internal/providers/pluginfw/tfschema/bool_attribute.go
@@ -3,6 +3,7 @@ package tfschema
 import (
 	dataschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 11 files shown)


internal/providers/pluginfw/resources/library/resource_library.go
```diff
@@ -0,0 +1,244 @@
+package library
```
mgyucht, line 1: I think this is not expected?

internal/providers/pluginfw/resources/library/resource_library_test.go
```diff
@@ -0,0 +1,143 @@
+package library_test
```
mgyucht, line 1: Ditto, this is probably carried over from another PR?

mgyucht (approved):
LGTM, but don't include the library changes in this PR>

PR #4010, opened 2024-09-12
[Internal] Fix irregularities in plugin framework converter function errors

tanmay-db:
## Changes

I noticed few irregularities in converter functions while working on clusters plugin framework migration, creating a PR to address those: 
1. Typo in error message -- in `GoSdkToTfSdkStruct` we say: `tfsdk to gosdk struct conversion failure`. It should be `gosdk to tfsdk struct conversion failure`
2.  In `GoSdkToTfSdkStruct ` we don't specify destination value type but we do it in `TfSdkToGoSdkStruct`. 
3. Abstract the error message out to reduce redundancy 
4. Standardise similar types of error messages to be same so it's easier to maintain. 

## Tests

Added unit tests. 

- [ ] `make test` run locally
- [ ] using Go SDK


edwardfeng-db (approved):
Thanks for fixing it!

PR #4014, opened 2024-09-12
[Feature] Add support for filters in `databricks_clusters` data source

mgyucht:
## Changes
Version 1.50 of the Terraform Provider featured an upgrade to the Go SDK affecting cluster listing. The new List Clusters API returns all terminated clusters in the last 30 days without a limit. This results in the list operation taking considerably longer for some workspaces, especially workspaces with many jobs where clusters are frequently created. This impacts the `databricks_clusters` data source, which can be slow.

This PR partially addresses this by adding support for `filter_by` to the `databricks_clusters` API. Filters expressed here are pushed to the server and result in fewer clusters being returned by the API. Users of this data source can specify a particular cluster state, cluster source, pinned status, or cluster policy ID to limit the number of clusters returned by the API, drastically speeding up performance.

## Tests
Integration tests for `databricks_cluster` data source test setting the `filter_by` parameter's attributes.

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/data-sources/clusters.md
+++ b/docs/data-sources/clusters.md
@@ -27,6 +27,16 @@ data "databricks_clusters" "all_shared" {
 ## Argument Reference
 
 * `cluster_name_contains` - (Optional) Only return [databricks_cluster](../resources/cluster.md#cluster_id) ids that match the given name string.
+* `filter_by` - (Optional) Filters to apply to the listed clusters. See [filter_by Configuration Block](#filter_by-configuration-block) below for details.
+
+### filter_by Configuration Block
+
+The `filter_by` block controls the filtering of the listed clusters. It supports the following arguments:
+
+* `cluster_sources` - (Optional) List of cluster sources to filter by. Possible values are `API`, `JOB`, `MODELS`, `PIPELINE`, `PIPELINE_MAINTENANCE`, `SQL`, and `UI`.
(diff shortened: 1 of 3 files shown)


docs/data-sources/clusters.md
```diff
@@ -27,6 +27,16 @@ data "databricks_clusters" "all_shared" {
 ## Argument Reference
 
 * `cluster_name_contains` - (Optional) Only return [databricks_cluster](../resources/cluster.md#cluster_id) ids that match the given name string.
+* `filter_by` - (Optional) Filters to apply to the listed clusters. See [filter_by Configuration Block](#filter_by-configuration-block) below for details.
+
+### filter_by Configuration Block
+
+The `filter_by` block controls the filtering of the listed clusters. It supports the following arguments:
+
+* `cluster_sources` - (Optional) List of cluster sources to filter by. Possible values are `API`, `JOB`, `MODELS`, `PIPELINE`, `PIPELINE_MAINTENANCE`, `SQL`, and `UI`.
```
alexott, line 36: Can we add an artificial entry like, `INTERACTIVE` that will be expanded into `UI`, and `API` ? It's more understandable for people IMHO... 

Exporter does this: https://github.com/databricks/terraform-provider-databricks/blob/main/exporter/importables.go#L332

alexott (approved):
In general, code looks good - but should we have a default values for the filter, so it mimics the previous behavior?

PR #4021, opened 2024-09-16
[Fix] Permanently delete `ERROR` and `TERMINATED` state clusters if their creation fails

tanmay-db:
## Changes

If we get error or terminated cluster (after getting WaitGetClusterRunning in Create) then we permanently delete them. 

## Tests

Unit tests
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -472,6 +472,8 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *commo
 
 	clusterInfo, err := clusterWaiter.GetWithTimeout(timeout)
 	if err != nil {
+		// In case of error or terminated state, WaitGetClusterRunning return an error and we cleanup the cluster before returning
+		resourceClusterDelete(ctx, d, c)
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


clusters/resource_cluster.go
```diff
@@ -472,6 +472,8 @@ func resourceClusterCreate(ctx context.Context, d *schema.ResourceData, c *commo
 
 	clusterInfo, err := clusterWaiter.GetWithTimeout(timeout)
 	if err != nil {
+		// In case of "ERROR" or "TERMINATED" state, WaitGetClusterRunning returns an error and we should delete the cluster before returning
+		resourceClusterDelete(ctx, d, c)
```
alexott, line 476: we need to show the error if we can't delete the resource, etc.

alexott: Like this: https://github.com/databricks/terraform-provider-databricks/blob/main/vectorsearch/resource_vector_search_endpoint.go#L53

alexott (changes requested):
We should expose nested error...

clusters/resource_cluster.go
tanmay-db: Yes, I was just testing a few things with directly calling deletion method before sending it in review, will update it.

alexott (approved):

Issue #3468, opened 2024-04-15
[FEATURE] New data block `databricks_users`


w0ut0:
### Use-cases

I want to be able to get all users provisioned to my Databricks account (or, depending on provider configuration, workspace). My use case is adding them to a certain 'default' catalog (or group), and create a schema for every user that they can use as 'personal space'.

### Attempted Solutions

- I can use AzureAD provider to query the SCIM app, and go from there, but this is not clean (nor cloud-agnostic).
- I can also use data block [databricks_group](https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/group) to fetch the built-in group `account users` (or `users` in a workspace context).

### Proposal
A data block that uses the [account-level List users endpoint](https://docs.databricks.com/api/azure/account/accountusers/list). Should be have similar as existing [databricks_service_principals](https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/service_principals).

```hcl
data "databricks_users" "users" {
  display_name_contains = "@domain.org"
}
```

- display_name_contains - (Optional) Only return display name that match the given name string

I see it returning 2 possible outputs:
- either a map of user_id => display_name mapping, that can then be used in `databricks_user` data block to fetch further details (like how it is done for [databricks_service_principals](https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/service_principals), or
- a list of `users`, with some basic attributes like `id`, `external_id`, `user_name`, `display_name`,.. (similar to [azuread_users](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/data-sources/users))

### References

- [databricks_service_principals](https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/service_principals)
- [databricks_user](https://registry.terraform.io/providers/databricks/databricks/latest/docs/data-sources/user)
- [azuread_users](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/data-sources/users)
- [accounts rest api](https://docs.databricks.com/api/azure/account/accountusers/list)
- [workspace rest api](https://docs.databricks.com/api/azure/workspace/users/list)

Serpinex3:
In addition to @w0ut0 request, we would also be happy having a `databricks_groups` data source similar to `databricks_service_principals` with an optional attribute `display_name_contains`.

Something like this:
```hcl
data "databricks_groups" "groups" {
  display_name_contains = "xyz"
}
```

[...]

PR #4028, opened 2024-09-17
Add datasource `databricks_users`

dgomez04:
## Changes

Added `databricks_users` data source as requested in #3468 to retrieve multiple users from the account-level API with optional filtering by `display_name_contains` or `user_name_contains`, enabling bulk user management operations. 

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/docs/data-sources/users.md
+++ b/docs/data-sources/users.md
@@ -1,0 +1,63 @@
+---
+subcategory: "Security"
+---
+
+# databricks_users Data Source
+
+->**Note** This is an account-level data source.
+
+-> **Note** If you have a fully automated setup with workspaces created by [databricks_mws_workspaces](../resources/mws_workspaces.md) or [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace), please make sure to add [depends_on attribute](../guides/troubleshooting.md#data-resources-and-authentication-is-not-configured-errors) in order to prevent _default auth: cannot configure default credentials_ errors.
+
+Retrieves information about multiple [databricks_user](../resources/user.md) resources.
+
+## Example Usage
+
+Adding a subset of users to a group
+
+```hcl
+data "databricks_users" "company_users" {
+  user_name_contains = "@domain.org"
+}
+
+resource "databricks_group" "data_users_group" {
+  display_name = "Data Users"
+}
+
+resource "databricks_group_member" "add_users_to_group" {
+  for_each = { for user in data.databricks_users.company_users.users : user.id => user }
+  group_id  = databricks_group.data_users_group.id
+  member_id = each.value.id
+}
+```
+
+## Argument Reference
+
+This data source allows you to filter the list of users using the following optional arguments: 
+
+- `display_name_contains` - (Optional) A substring to filter users by their display name. Only users whose display names contain this substring will be included in the results.
+- `user_name_contains` - (Optional) A substring to filter users by their username. Only users whose usernames contain this substring will be included in the results.
+
+->**Note** You can specify **exactly one** of `display_name_contains` or `user_name_contains`. If neither is specified, all users will be returned.
+
+## Attribute Reference
+
+This data source exposes the following attributes:
+
+- `users` - A list of users matching the specified criteria. Each user has the following attributes:
+    - `id` - The ID of the user.
+    - `user_name` - The username of the user.
+    - `display_name` - The display name of the user. 
+
+## Related Resources
+
+The following resources are used in the same context:
+
+- [**databricks_user**](../resources/user.md): Resource to manage individual users in Databricks.
+
+- [**databricks_group**](../resources/group.md): Resource to manage groups in Databricks.
+
+- [**databricks_group_member**](../resources/group_member.md): Resource to manage group memberships by adding users to groups.
+
+- [**databricks_permissions**](../resources/permissions.md): Resource to manage access control in the Databricks workspace.
+
+- [**databricks_current_user**](current_user.md): Data source to retrieve information about the user or service principal that is calling the Databricks REST API.
--- a/internal/acceptance/data_users_test.go
+++ b/internal/acceptance/data_users_test.go
@@ -1,0 +1,67 @@
+package acceptance
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+func TestAccDataSourceDataUsers_DisplayNameContains(t *testing.T) {
+	accountLevel(t, step{
+		Template: `
+		data "databricks_data_users" "this" {
+			display_name_contains = "testuser"
+		}`,
+		Check: func(s *terraform.State) error {
+			r, ok := s.RootModule().Resources["data.databricks_data_users.this"]
+			if !ok {
+				return fmt.Errorf("data not found in state")
+			}
+			ids := r.Primary.Attributes["users.#"]
+			if ids == "" {
+				return fmt.Errorf("users is empty: %v", r.Primary.Attributes)
+			}
+			return nil
+		},
+	})
+}
+
+func TestAccDataSourceDataUsers_UserNameContains(t *testing.T) {
+	accountLevel(t, step{
+		Template: `
+		data "databricks_data_users" "this" {
+			user_name_contains = "example.com"
+		}`,
+		Check: func(s *terraform.State) error {
+			r, ok := s.RootModule().Resources["data.databricks_data_users.this"]
+			if !ok {
+				return fmt.Errorf("data not found in state")
+			}
+			usersCount := r.Primary.Attributes["users.#"]
+			if usersCount == "" || usersCount == "0" {
+				return fmt.Errorf("users list is empty: %v", r.Primary.Attributes)
+			}
+			return nil
+		},
+	})
+}
+
+func TestAccDataSourceDataUsers_NoFilters(t *testing.T) {
+	accountLevel(t, step{
+		Template: `
+		data "databricks_data_users" "this" {
+		}`,
+		Check: func(s *terraform.State) error {
+			r, ok := s.RootModule().Resources["data.databricks_data_users.this"]
+			if !ok {
+				return fmt.Errorf("data not found in state")
+			}
+			usersCount := r.Primary.Attributes["users.#"]
+			if usersCount == "" || usersCount == "0" {
+				return fmt.Errorf("users list is empty: %v", r.Primary.Attributes)
+			}
+			return nil
+		},
+	})
+}
--- a/scim/data_users.go
+++ b/scim/data_users.go
@@ -1,0 +1,52 @@
+package scim
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/iam"
+	"github.com/databricks/terraform-provider-databricks/common"
+)
+
+func DataSourceDataUsers() common.Resource {
+
+	type DataUsers struct {
+		Id                  string                   `json:"id,omitempty" tf:"computed"`
+		DisplayNameContains string                   `json:"display_name_contains,omitempty" tf:"computed"`
+		Users               []map[string]interface{} `json:"users,omitempty" tf:"computed"`
+	}
+
+	return common.AccountData(func(ctx context.Context, data *DataUsers, acc *databricks.AccountClient) error {
+		listRequest := iam.ListAccountUsersRequest{}
+
+		if data.DisplayNameContains != "" {
+			listRequest.Filter = fmt.Sprintf("displayName co \"%s\"", data.DisplayNameContains)
+		}
+
+		userList, err := acc.Users.ListAll(ctx, listRequest)
+
+		if err != nil {
+			return err
+		}
+
+		if len(userList) == 0 {
+			return fmt.Errorf("cannot find users with display name containing %s", data.DisplayNameContains)
+		}
+
+		var users []map[string]interface{}
+
+		for _, u := range userList {
+			user := map[string]interface{}{
+				"id":           u.Id,
+				"user_name":    u.UserName,
+				"display_name": u.DisplayName,
+			}
+			users = append(users, user)
+		}
+
+		data.Users = users
+
+		return nil
+	})
+}
--- a/scim/data_users.go
+++ b/scim/data_users.go
@@ -11,14 +11,21 @@ import (
 
 func DataSourceDataUsers() common.Resource {
 
+	type UserInfo struct {
+		Id          string `json:"id,omitempty" tf:"computed"`
+		UserName    string `json:"user_name,omitempty" tf:"computed"`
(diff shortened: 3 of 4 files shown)


alexott:
@dgomez04 to fix integration test - make this change: https://github.com/databricks/terraform-provider-databricks/pull/3874/commits/f38cdbc84efb90e84117a2de0417a267857f60d3

docs/data-sources/users.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Argument Reference
+
+This data source allows you to filter the list of users using the following optional arguments: 
+
+- `display_name_contains` - (Optional) A substring to filter users by their display name. Only users whose display names contain this substring will be included in the results.
+- `user_name_contains` - (Optional) A substring to filter users by their username. Only users whose usernames contain this substring will be included in the results.
+
+->**Note** You can specify **exactly one** of `display_name_contains` or `user_name_contains`. If neither is specified, all users will be returned.
```
mgyucht, line 38:
```suggestion
-> You can specify **at most one** of `display_name_contains` or `user_name_contains`. If neither is specified, all users will be returned.
```
TF docs adds the **Note** automatically when you use the `->` syntax.

internal/providers/pluginfw/pluginfw.go
```diff
@@ -62,6 +63,7 @@ func (p *DatabricksProviderPluginFramework) DataSources(ctx context.Context) []f
 		sharing.DataSourceShare,
 		sharing.DataSourceShares,
 		catalog.DataSourceFunctions,
+		user.DataSourceUsers,
```
mgyucht, line 66: You'll need to merge `main` to merge this PR. Make sure you add this in the list of not-migrated datasources.

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,113 @@
...
+
+var _ datasource.DataSourceWithConfigure = &UsersDataSource{}
+
+type UsersDataSource struct {
+	Client *common.DatabricksClient
+}
+
+type UsersInfo struct {
+	DisplayNameContains string     `json:"display_name_contains,omitempty" tf:"computed"`
+	UserNameContains    string     `json:"user_name_contains,omitempty" tf:"computed"`
```
mgyucht, line 28: This is a fairly prescriptive interface: it basically says you can search for users based on whether their display name or user name contains a certain string. However, the REST API permits much more flexibility, allowing users to specify a custom filter where they could check if the username contains multiple strings, for example. I don't think we need to introduce a new abstraction & validation specifically for Terraform. What do you think about exposing a Filter field:
```go
type usersInfo struct {
        Filter types.String `json:"filter,omitempty"`
        ...
}
```
This one should not be computed, and we can pass it directly to the API request.

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,113 @@
...
+	Users               []iam.User `json:"users,omitempty" tf:"computed"`
```
mgyucht, line 29: You probably need to use `iam_tf.User` here instead of `iam.User`, since this structure needs to use the types defined in the `types` module of the plugin framework in order to work with the `Get()` and `Set()` methods of the Plan.

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,113 @@
...
+		if AppendDiagAndCheckErrors(resp, diags) {
+			return
+		}
+		users, err = w.Users.ListAll(ctx, iam.ListUsersRequest{Filter: filter})
+		if err != nil && AppendDiagAndCheckErrors(resp, DiagsFromError("Error listing workspace users", err)) {
+			return
+		}
+	}
+
+	userInfo.Users = users
```
mgyucht, line 111: You will need to use the `GoSdkToTfSdkStruct` converter before saving this in the state.

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,113 @@
...
+	} else if userInfo.UserNameContains != "" {
+		filter = fmt.Sprintf("userName co \"%s\"", userInfo.UserNameContains)
+	}
+
+	var users []iam.User
+	var err error
+
+	if d.Client.Config.IsAccountClient() {
+		a, diags := d.Client.GetAccountClient()
+		if AppendDiagAndCheckErrors(resp, diags) {
```
mgyucht, line 93: Let's follow the code convention we've established here for error handling: https://github.com/databricks/terraform-provider-databricks/blob/main/CONTRIBUTING.md#code-conventions. I think we don't need AppendDiagAndCheckErrors or DiagsFromError, as we cover how to handle methods that return both `diag.Diagnostics` and `error` there.

internal/providers/pluginfw/resources/user/data_users_acc_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const dataSourceTemplate = `
+	resource "databricks_user" "user1" {
+		user_name = "testuser1@databricks.com"
```
mgyucht, line 14:
```suggestion
		user_name = "user-datasource-{var.STICKY_RANDOM}-1@databricks.com"
```
Tests run in a single testing environment. Tests use isolated state files for resources created during the tests, but the resources are created in a single environment. Two tests running in parallel may create and delete the same user. Use `{var.STICKY_RANDOM}` when creating resources that need unique names or other attributes to ensure they don't clash with one another.

internal/providers/pluginfw/resources/user/data_users_acc_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	resource "databricks_user" "user2" {
+		user_name = "testuser2@databricks.com"
+	}
+
+	data "databricks_users" "this" {
```
mgyucht, line 21: This needs a `depends_on` field to wait for the users to be created before listing.

mgyucht (changes requested):
Thanks for this! Good first cut. We'll need to make some changes before merging.

internal/providers/pluginfw/resources/user/data_users.go
dgomez04: Done 👍

internal/providers/pluginfw/resources/user/data_users_acc_test.go
dgomez04: Gotcha, thanks! 👍

internal/providers/pluginfw/resources/user/data_users_acc_test.go
dgomez04: Done 👍

internal/providers/pluginfw/resources/user/data_users.go
dgomez04: Done! 👍

internal/providers/pluginfw/resources/user/data_users.go
dgomez04: Makes sense, done! 👍

internal/providers/pluginfw/pluginfw.go
dgomez04: Done 👍

internal/acceptance/data_users_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const dataSourceTemplate = `
+	resource "databricks_user" "user1" {
+		user_name = "testuser1@databricks.com"
+	}
+
+	resource "databricks_user" "user2" {
+		user_name = "testuser2@databricks.com"
```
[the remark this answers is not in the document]
dgomez04: Done 👍

docs/data-sources/users.md
```diff
@@ -0,0 +1,80 @@
+---
+subcategory: "Security"
+---
+
+# databricks_users Data Source
+
+-> If you have a fully automated setup with workspaces created by [databricks_mws_workspaces](../resources/mws_workspaces.md) or [azurerm_databricks_workspace](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace), please make sure to add [depends_on attribute](../guides/troubleshooting.md#data-resources-and-authentication-is-not-configured-errors) in order to prevent _default auth: cannot configure default credentials_ errors.
```
alexott, line 7: Can we add a note that it works with both workspace & account-level providers?

docs/data-sources/users.md
```diff
@@ -0,0 +1,80 @@
...
+  group_id  = databricks_group.data_users_group.id
+  member_id = each.value.id
+}
+```
+
+## Argument Reference
+
+This data source allows you to filter the list of users using the following optional arguments: 
+
+- `filter` - (Optional) Query by which the results have to be filtered. If not specified, all users will be returned. Supported operators are equals (`eq`), contains (`co`), starts with (`sw`), and not equals (`ne`). Additionally, simple expressions can be formed using logical operators `and` and `or`.
```
alexott, line 35: Maybe we should add a note (or pointer to the doc) that will clarify if the search is case-sensitive or not?

docs/data-sources/users.md
```diff
@@ -0,0 +1,80 @@
...
+
+- `users` - A list of users matching the specified criteria. Each user has the following attributes:
+    - `id` - The ID of the user.
+    - `userName` - The username of the user.
+    - `emails` - All the emails associated with the Databricks user.
+    - `name`
+      - `givenName` - Given name of the Databricks user.
+      - `familyName` - Family name of the Databricks user.
+    - `displayName` - The display name of the user. 
+    - `roles` - Indicates if the user has the admin role.
```
alexott, line 59: it's not only for admin roles

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+var _ datasource.DataSourceWithConfigure = &UsersDataSource{}
+
+type UsersDataSource struct {
+	Client *common.DatabricksClient
+}
+
+type UsersInfo struct {
+	Filter types.String  `json:"filter,omitempty"`
```
alexott, line 30: What I would recommend as well - by default fetch only limited number of attributes, i.e., don't include `roles` - this could be done to decrease the load on the SCIM backend as fetching roles is quite heavyweight operation. 

We can add an additional parameter, like, `extra_attributes` that will receive the list of additional attributes (comma-separated).

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,97 @@
...
+	var err error
+
+	if d.Client.Config.IsAccountClient() {
+		a, diags := d.Client.GetAccountClient()
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
+		users, err = a.Users.ListAll(ctx, iam.ListAccountUsersRequest{Filter: usersInfo.Filter.ValueString()})
```
alexott, line 70: explicitly configure attributes, like this: https://github.com/databricks/terraform-provider-databricks/blob/main/exporter/util_scim.go#L289, but without `roles` and `groups` (and maybe `entitlements`) by default.

docs/data-sources/users.md
```diff
@@ -0,0 +1,80 @@
...
+      - `$ref`
+      - `value`
+      - `display`
+      - `primary`
+      - `type`
+    - `externalId` - reserved for future use. 
+    - `active` - Boolean that represents if this user is active.
```
alexott, line 66: We have more information in the `iam.User` struct - `groups`, `entitlements`.

alexott (changes requested):
looks good, but please consider my comments about additional attributes - this will decrease the load on SCIM backend

docs/data-sources/users.md
dgomez04: Done!

docs/data-sources/users.md
dgomez04: Done!

docs/data-sources/users.md
dgomez04: Added remaining information. Thanks for pointing it out!

internal/providers/pluginfw/resources/user/data_users.go
dgomez04: I've added a default attribute list and an additional `extra_attributes` parameter to the data source, should align with your ask. Thank you for the insight. 👍

internal/providers/pluginfw/resources/user/data_users.go
dgomez04: Done! 👍

dgomez04:
@alexott implemented the changes you requested. Appreciate the insights. Please give it a look and let me know how we want to proceed.

docs/data-sources/users.md
```diff
@@ -0,0 +1,99 @@
...
+    - User whose `displayName` equals "john": 
+    ```hcl
+      filter = "displayName eq \"john\""
+    ```
+    - User whose `displayName` contains "john" or `userName` contains "@domain.org": 
+    ```hcl
+      filter = "displayName co \"john\" or userName co \"@domain.org\""
+    ```
+
+- `extra_attributes` - (Optional) A list of additional user attributes to include in the results. By default, the data source returns the following attributes: `id`, `userName`, `displayName`, and `externalId`. Use this argument to request additional attributes as needed. The list of all available attributes can be found in the [API reference](https://docs.databricks.com/api/workspace/users/list).
```
alexott, line 51: Add that it's a comma-separated list

internal/providers/pluginfw/resources/user/data_users.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (d *UsersDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if d.Client == nil {
+		d.Client = pluginfwcommon.ConfigureDataSource(req, resp)
+	}
+}
+
+func (d *UsersDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var usersInfo UsersInfo
+	var attributes string = "id,userName,displayName,externalId,"
```
alexott, line 55: we can just do:

```
attributes := "id,userName,displayName,externalId"
```

and maybe omit the trailing `,`, and add it in the corresponding block

alexott (commented):
mostly good, small comments

internal/providers/pluginfw/resources/user/data_users.go
dgomez04: Done 👍

dgomez04:
@alexott I've implemented the last batch of requested changes. LMK how it looks.

alexott:
Superced by #4537

PR #4040, opened 2024-09-23
[Internal] Update plugin framework schema to use ListNestedBlocks

edwardfeng-db:
## Changes

- To make sure we can release the resources with plugin framework that's compatible with our sdkv2 resources, we need to use `ListNestedBlocks` for struct types so that the `x.0.y` access pattern can be preserved, also in the HCL we don't need to add extra `=` for blocks.
- To achieve this, we need corresponding changes in AttributeBuilder and CustomizableSchema
- Updated tfsdk struct generations to generate list types for previous struct types, so that the set and get functions will still work
- Updated the converter functions to cover cases that are structs in `gosdk` but slices in `tfsdk`
- Updated acceptance tests to remove `=` for block types

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/.codegen/model.go.tmpl
+++ b/.codegen/model.go.tmpl
@@ -24,7 +24,7 @@ import (
 type {{.PascalName}} struct {
     {{- range .Fields}}
     {{.Comment "    // " 80}}
-    {{.PascalName}} {{if .IsOptionalObject}}*{{end}}{{template "type" .Entity}} `{{template "field-tag" . }}`{{end}}
+    {{.PascalName}} {{template "type" .Entity}} `{{template "field-tag" . }}`{{end}}
 }
 
 {{end}}
@@ -35,20 +35,23 @@ type {{.PascalName}} struct {
 {{- end -}}
 
 {{- define "type" -}}
-	{{- if not . }}any /* ERROR */
-	{{- else if .IsExternal }}{{.Package.Name}}.{{.PascalName}}
-	{{- else if .IsAny}}any
-	{{- else if .IsEmpty}}{{.PascalName}}
-	{{- else if .IsString}}types.String
-	{{- else if .IsBool}}types.Bool
-	{{- else if .IsInt64}}types.Int64
-	{{- else if .IsFloat64}}types.Float64
-	{{- else if .IsInt}}types.Int64
-	{{- else if .IsByteStream}}io.ReadCloser
-	{{- else if .ArrayValue }}[]{{template "type" .ArrayValue}}
-	{{- else if .MapValue }}map[string]{{template "type" .MapValue}}
-	{{- else if .IsObject }}{{.PascalName}}
-	{{- else if .Enum }}types.String
-	{{- else}}any /* MISSING TYPE */
-	{{- end -}}
+    {{- if not . }}any /* ERROR */
+    {{- else if .IsExternal }}{{.Package.Name}}.{{.PascalName}}
+    {{- else if .IsAny}}any
+    {{- else if .IsEmpty}}[]{{.PascalName}}
+    {{- else if .IsString}}types.String
+    {{- else if .IsBool}}types.Bool
+    {{- else if .IsInt64}}types.Int64
+    {{- else if .IsFloat64}}types.Float64
+    {{- else if .IsInt}}types.Int64
+    {{- else if .IsByteStream}}io.ReadCloser
+    {{- else if .ArrayValue }}
+        {{- if .ArrayValue.IsObject }}{{template "type" .ArrayValue}}
+        {{- else }}[]{{template "type" .ArrayValue}}
+        {{- end }}
+    {{- else if .MapValue }}map[string]{{template "type" .MapValue}}
+    {{- else if .IsObject }}[]{{.PascalName}}
+    {{- else if .Enum }}types.String
+    {{- else}}any /* MISSING TYPE */
+    {{- end -}}
 {{- end -}}
--- a/internal/providers/pluginfw/converters/converters_test.go
+++ b/internal/providers/pluginfw/converters/converters_test.go
@@ -17,7 +17,6 @@ type DummyTfSdk struct {
 	Floats            types.Float64               `tfsdk:"floats" tf:""`
 	Description       types.String                `tfsdk:"description" tf:""`
 	Tasks             types.String                `tfsdk:"task" tf:"optional"`
-	Nested            *DummyNestedTfSdk           `tfsdk:"nested" tf:"optional"`
 	NoPointerNested   DummyNestedTfSdk            `tfsdk:"no_pointer_nested" tf:"optional"`
 	NestedList        []DummyNestedTfSdk          `tfsdk:"nested_list" tf:"optional"`
 	NestedPointerList []*DummyNestedTfSdk         `tfsdk:"nested_pointer_list" tf:"optional"`
@@ -28,6 +27,8 @@ type DummyTfSdk struct {
 	EnumField         types.String                `tfsdk:"enum_field" tf:"optional"`
 	AdditionalField   types.String                `tfsdk:"additional_field" tf:"optional"`
 	DistinctField     types.String                `tfsdk:"distinct_field" tf:"optional"`
+	SliceStruct       []DummyNestedTfSdk          `tfsdk:"slice_struct" tf:"optional"`
+	SliceStructPtr    []DummyNestedTfSdk          `tfsdk:"slice_struct_ptr" tf:"optional"`
 	Irrelevant        types.String                `tfsdk:"-"`
 }
 
@@ -69,7 +70,6 @@ type DummyGoSdk struct {
 	Floats            float64                     `json:"floats"`
 	Description       string                      `json:"description"`
 	Tasks             string                      `json:"tasks"`
-	Nested            *DummyNestedGoSdk           `json:"nested"`
 	NoPointerNested   DummyNestedGoSdk            `json:"no_pointer_nested"`
 	NestedList        []DummyNestedGoSdk          `json:"nested_list"`
 	NestedPointerList []*DummyNestedGoSdk         `json:"nested_pointer_list"`
@@ -80,6 +80,8 @@ type DummyGoSdk struct {
 	EnumField         TestEnum                    `json:"enum_field"`
 	AdditionalField   string                      `json:"additional_field"`
 	DistinctField     string                      `json:"distinct_field"` // distinct field that the tfsdk struct doesn't have
+	SliceStruct       DummyNestedGoSdk            `json:"slice_struct"`
+	SliceStructPtr    *DummyNestedGoSdk           `json:"slice_struct_ptr"`
 	ForceSendFields   []string                    `json:"-"`
 }
 
@@ -187,18 +189,6 @@ var tests = []struct {
 		}},
 	},
 	{
-		"pointer conversion",
-		DummyTfSdk{Nested: &DummyNestedTfSdk{
-			Name:    types.StringValue("def"),
-			Enabled: types.BoolValue(true),
-		}},
-		DummyGoSdk{Nested: &DummyNestedGoSdk{
-			Name:            "def",
-			Enabled:         true,
-			ForceSendFields: []string{"Name", "Enabled"},
-		}},
-	},
-	{
 		"list conversion",
 		DummyTfSdk{Repeated: []types.Int64{types.Int64Value(12), types.Int64Value(34)}},
 		DummyGoSdk{Repeated: []int64{12, 34}},
@@ -256,6 +246,34 @@ var tests = []struct {
 				Enabled:         false,
 				ForceSendFields: []string{"Name", "Enabled"},
 			},
+		}},
+	},
+	{
+		"list representation of struct conversion", // we use list with one element in the tfsdk to represent struct in gosdk
+		DummyTfSdk{SliceStruct: []DummyNestedTfSdk{
+			{
+				Name:    types.StringValue("def"),
+				Enabled: types.BoolValue(true),
+			},
+		}},
+		DummyGoSdk{SliceStruct: DummyNestedGoSdk{
+			Name:            "def",
+			Enabled:         true,
+			ForceSendFields: []string{"Name", "Enabled"},
+		}},
+	},
+	{
+		"list representation of struct pointer conversion", // we use list with one element in the tfsdk to represent struct in gosdk
+		DummyTfSdk{SliceStructPtr: []DummyNestedTfSdk{
+			{
+				Name:    types.StringValue("def"),
+				Enabled: types.BoolValue(true),
+			},
+		}},
+		DummyGoSdk{SliceStructPtr: &DummyNestedGoSdk{
+			Name:            "def",
+			Enabled:         true,
+			ForceSendFields: []string{"Name", "Enabled"},
 		}},
 	},
 }
--- a/internal/providers/pluginfw/converters/go_to_tf.go
+++ b/internal/providers/pluginfw/converters/go_to_tf.go
@@ -26,11 +26,13 @@ const goSdkToTfSdkFieldConversionFailureMessage = "gosdk to tfsdk field conversi
 //
 // NOTE:
 //
-//	If field name doesn't show up in ForceSendFields and the field is zero value, we set the null value on the tfsdk.
-//	types.list and types.map are not supported
-//	map keys should always be a string
-//	tfsdk structs use types.String for all enum values
-//	non-json fields will be omitted
+// # Structs in gosdk are represented as slices of structs in tfsdk, and pointers are removed
+//
+// If field name doesn't show up in ForceSendFields and the field is zero value, we set the null value on the tfsdk.
+// types.list and types.map are not supported
+// map keys should always be a string
+// tfsdk structs use types.String for all enum values
+// non-json fields will be omitted
 func GoSdkToTfSdkStruct(ctx context.Context, gosdk interface{}, tfsdk interface{}) diag.Diagnostics {
 
 	srcVal := reflect.ValueOf(gosdk)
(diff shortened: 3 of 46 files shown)


internal/providers/pluginfw/tfschema/attribute_builder.go
```diff
@@ -11,6 +11,8 @@ import (
 type AttributeBuilder interface {
 	BuildDataSourceAttribute() dataschema.Attribute
 	BuildResourceAttribute() schema.Attribute
+	BuildDataSourceBlock() dataschema.Block
+	BuildResourceBlock() schema.Block
```
edwardfeng-db, line 15: Not sure if this is a good idea or should we have a new interface for `BlockBuilder`

internal/providers/pluginfw/tfschema/struct_to_schema.go
```diff
@@ -128,9 +130,9 @@ func typeToSchema(v reflect.Value) map[string]AttributeBuilder {
 			default:
 				// Nested struct
 				nestedScm := typeToSchema(reflect.New(elemType).Elem())
-				scm[fieldName] = MapNestedAttributeBuilder{
+				scm_attr[fieldName] = MapNestedAttributeBuilder{
 					NestedObject: NestedAttributeObject{
-						Attributes: nestedScm,
+						Attributes: nestedScm.Attributes,
```
edwardfeng-db, line 135: I'm not quite sure why but there's no MapNestedBlock in tf library

internal/providers/pluginfw/tfschema/struct_to_schema.go
```diff
@@ -172,18 +174,18 @@ func typeToSchema(v reflect.Value) map[string]AttributeBuilder {
...
-				scm[fieldName] = SingleNestedAttributeBuilder{
-					Attributes: nestedScm,
-					Optional:   isOptional,
-					Required:   !isOptional,
-					Computed:   isComputed,
+				scm_block[fieldName] = ListNestedBlockBuilder{
+					NestedObject: nestedScm,
+					Optional:     isOptional,
+					Required:     !isOptional,
+					Computed:     isComputed,
```
edwardfeng-db, line 181: TODO: Need to add validator for `MaxItem` = 1

internal/providers/pluginfw/resources/library/resource_library.go
```diff
@@ -70,19 +70,29 @@ func (r *LibraryResource) Metadata(ctx context.Context, req resource.MetadataReq
...
+			}
+		}
+		for field, attribute := range c.ToAttributeMap().Blocks {
+			switch attribute.(type) {
+			case tfschema.StringAttributeBuilder:
+				c.AddPlanModifier(stringplanmodifier.RequiresReplace(), field)
+			case tfschema.ListNestedBlockBuilder:
+				c.AddPlanModifier(objectplanmodifier.RequiresReplace(), field)
+			}
+		}
```
edwardfeng-db, line 89: I don't really like this, should be more elegant

internal/providers/pluginfw/tfschema/struct_to_schema.go
edwardfeng-db: Looks like it's not very feasible right now given that we have updated the tfsdk structs directly to use slices for struct types.

The validation is done in the converter functions now.

edwardfeng-db, line 181: This is probably not necessary here, because in the generated tfsdk structs there will no longer be any struct types, all will be wrapped as a slice type.

internal/providers/pluginfw/resources/library/resource_library.go
```diff
@@ -70,19 +70,29 @@ func (r *LibraryResource) Metadata(ctx context.Context, req resource.MetadataReq
...
+			case tfschema.StringAttributeBuilder:
+				c.AddPlanModifier(stringplanmodifier.RequiresReplace(), field)
+			case tfschema.SingleNestedAttributeBuilder:
+				c.AddPlanModifier(objectplanmodifier.RequiresReplace(), field)
+			}
+		}
+		// for field, block := range c.ToAttributeMap().Blocks {
+		// 	switch block.(type) {
+		// 	case tfschema.ListNestedBlockBuilder:
+		// 		c.AddPlanModifier(listplanmodifier.RequiresReplace(), field)
```
edwardfeng-db, line 85: Wanted to do this but it seems like there's no such thing as `listplanmodifier`. TODO - figure this out

internal/providers/pluginfw/converters/converters_test.go
```diff
@@ -28,6 +28,8 @@ type DummyTfSdk struct {
 	EnumField         types.String                `tfsdk:"enum_field" tf:"optional"`
 	AdditionalField   types.String                `tfsdk:"additional_field" tf:"optional"`
 	DistinctField     types.String                `tfsdk:"distinct_field" tf:"optional"`
+	SliceStruct       []DummyNestedTfSdk          `tfsdk:"slice_struct" tf:"optional"`
+	SliceStructPtr    *[]DummyNestedTfSdk         `tfsdk:"slice_struct_ptr" tf:"optional"`
```
mgyucht, line 32: Let's see if we can remove all pointers now. Optionality is expressed as nil/zero-length slice. Ideally, we don't need to use pointers anywhere or distinguish between pointers and concrete types in tfsdk structs.

internal/providers/pluginfw/converters/go_to_tf.go
```diff
@@ -100,10 +100,26 @@ func goSdkToTfSdkSingleField(ctx context.Context, srcField reflect.Value, destFi
 			// Skip nils
 			return nil
 		}
-		destField.Set(reflect.New(destField.Type().Elem()))
+
+		var fieldToSetInterface any
+
+		if destField.Type().Elem().Kind() == reflect.Slice {
```
mgyucht, line 106: This can be removed after removing extra pointers in tfsdk structs

.codegen/model.go.tmpl
```diff
@@ -35,20 +35,23 @@ type {{.PascalName}} struct {
...
+    {{- else if .IsInt64}}types.Int64
+    {{- else if .IsFloat64}}types.Float64
+    {{- else if .IsInt}}types.Int64
+    {{- else if .IsByteStream}}io.ReadCloser
+    {{- else if .ArrayValue }}
+        {{- if .ArrayValue.IsObject }}{{template "type" .ArrayValue}}
+        {{- else }}[]{{template "type" .ArrayValue}}
+        {{- end }}
+    {{- else if .MapValue }}map[string]{{template "type" .MapValue}}
+    {{- else if .IsObject }}[]{{.PascalName}}
```
mgyucht, line 53: Let's see if we can somehow capture some metadata in struct tag or some other way to capture max_items = 1.

internal/providers/pluginfw/converters/tf_to_go.go
```diff
@@ -94,10 +94,24 @@ func tfSdkToGoSdkSingleField(ctx context.Context, srcField reflect.Value, destFi
...
+		srcFieldValFinal := srcFieldValue
+
+		if srcField.Elem().Kind() == reflect.Slice {
+			assertStructSliceLengthIsOne(srcField.Elem())
+			srcFieldValFinal = srcField.Elem().Index(0).Interface()
+		}
+
 		// Recursively populate the nested struct.
-		if TfSdkToGoSdkStruct(ctx, srcFieldValue, destField.Interface()).HasError() {
+		if TfSdkToGoSdkStruct(ctx, srcFieldValFinal, destField.Interface()).HasError() {
```
mgyucht, line 105: This should go away when we don't have pointers in tfsdk structs

internal/providers/pluginfw/converters/tf_to_go.go
```diff
@@ -94,10 +94,24 @@ func tfSdkToGoSdkSingleField(ctx context.Context, srcField reflect.Value, destFi
...
 			panic(fmt.Sprintf("%s. %s", tfSdkToGoSdkStructConversionFailureMessage, common.TerraformBugErrorMessage))
 		}
+	} else if srcField.Kind() == reflect.Slice && destField.Kind() == reflect.Struct {
+		if srcField.IsNil() {
+			// Skip nils
+			return nil
+		}
+		assertStructSliceLengthIsOne(srcField)
```
mgyucht, line 113: This should be moved to the schema validator

internal/providers/pluginfw/converters/converters_test.go
edwardfeng-db: Removed all of those

.codegen/model.go.tmpl
edwardfeng-db: Yes, trying to figure out how to do this, will talk to Hector about it.

internal/providers/pluginfw/converters/go_to_tf.go
```diff
@@ -151,9 +166,19 @@ func goSdkToTfSdkSingleField(ctx context.Context, srcField reflect.Value, destFi
...
+			if GoSdkToTfSdkStruct(ctx, srcFieldValue, destSlice.Index(0).Addr().Interface()).HasError() {
+				panic(fmt.Sprintf("%s. %s", goSdkToTfSdkStructConversionFailureMessage, common.TerraformBugErrorMessage))
+			}
+			destField.Set(destSlice)
+		} else {
+			// resolve the nested struct by recursively calling the function
+			if GoSdkToTfSdkStruct(ctx, srcFieldValue, destField.Addr().Interface()).HasError() {
+				panic(fmt.Sprintf("%s. %s", goSdkToTfSdkStructConversionFailureMessage, common.TerraformBugErrorMessage))
+			}
 		}
```
mgyucht, line 182: This is parallel to the pointer case and can be structured as such.
```suggestion
		var dest any
		if destField.Kind() == reflect.Slice {
			// allocate a slice first
			destSlice := reflect.MakeSlice(destField.Type(), 1, 1)
			destField.Set(destSlice)
			dest = destSlice.Index(0).Addr().Interface()
		} else {
			dest = destField.Addr().Interface()
		}
		// resolve the nested struct by recursively calling the function
		if GoSdkToTfSdkStruct(ctx, srcFieldValue, dest).HasError() {
			panic(fmt.Sprintf("%s. %s", goSdkToTfSdkStructConversionFailureMessage, common.TerraformBugErrorMessage))
		}
```

internal/providers/pluginfw/converters/tf_to_go.go
```diff
@@ -98,6 +100,26 @@ func tfSdkToGoSdkSingleField(ctx context.Context, srcField reflect.Value, destFi
 		if TfSdkToGoSdkStruct(ctx, srcFieldValue, destField.Interface()).HasError() {
 			panic(fmt.Sprintf("%s. %s", tfSdkToGoSdkStructConversionFailureMessage, common.TerraformBugErrorMessage))
 		}
+	} else if srcField.Kind() == reflect.Slice && destField.Kind() == reflect.Struct {
+		if srcField.IsNil() {
+			// Skip nils
+			return nil
+		}
+		assertStructSliceLengthIsOne(srcField)
+		tfsdkToGoSdkStructField(srcField.Index(0), destField, srcFieldName, forceSendFieldsField, ctx)
```
mgyucht, line 109: For my own understanding, why can we call the internal method here but need to call the public method in the Ptr case?

internal/providers/pluginfw/tfschema/customizable_schema.go
```diff
@@ -11,39 +11,45 @@ import (
...
 // CustomizableSchema is a wrapper struct on top of AttributeBuilder that can be used to navigate through nested schema add customizations.
 type CustomizableSchema struct {
-	attr AttributeBuilder
+	attr BaseSchemaBuilder
 }
 
 // ConstructCustomizableSchema constructs a CustomizableSchema given a map from string to AttributeBuilder.
-func ConstructCustomizableSchema(attributes map[string]AttributeBuilder) *CustomizableSchema {
-	attr := AttributeBuilder(SingleNestedAttributeBuilder{Attributes: attributes})
+func ConstructCustomizableSchema(nestedObject NestedBlockObject) *CustomizableSchema {
```
mgyucht, line 18: Make sure to update the doccomments for all of these methods as they refer to the old types, and also for CustomizableSchema.

internal/providers/pluginfw/tfschema/customizable_schema.go
```diff
@@ -11,39 +11,45 @@ import (
...
 // ToAttributeMap converts CustomizableSchema into a map from string to Attribute.
-func (s *CustomizableSchema) ToAttributeMap() map[string]AttributeBuilder {
+func (s *CustomizableSchema) ToAttributeMap() NestedBlockObject {
 	return attributeToMap(&s.attr)
 }
 
 // attributeToMap converts AttributeBuilder into a map from string to AttributeBuilder.
-func attributeToMap(attr *AttributeBuilder) map[string]AttributeBuilder {
-	var m map[string]AttributeBuilder
+func attributeToMap(attr *BaseSchemaBuilder) NestedBlockObject {
```
mgyucht, line 29: should we rename this to `attributeToNestedBlockObject`?

internal/providers/pluginfw/tfschema/customizable_schema.go
```diff
@@ -166,21 +176,33 @@ func (s *CustomizableSchema) SetReadOnly(path ...string) *CustomizableSchema {
...
 // navigateSchemaWithCallback navigates through schema attributes and executes callback on the target, panics if path does not exist or invalid.
-func navigateSchemaWithCallback(s *AttributeBuilder, cb func(AttributeBuilder) AttributeBuilder, path ...string) (AttributeBuilder, error) {
+func navigateSchemaWithCallback(s *BaseSchemaBuilder, cb func(BaseSchemaBuilder) BaseSchemaBuilder, path ...string) (BaseSchemaBuilder, error) {
 	current_scm := s
 	for i, p := range path {
 		m := attributeToMap(current_scm)
-
-		v, ok := m[p]
-		if !ok {
+		m_attr := m.Attributes
```
mgyucht, line 183: nit: let's follow camelCase naming convention: `mAttr`, `mBlock`, `newV`, `castedV`, `currentScm`, etc.

internal/providers/pluginfw/tfschema/float64_attribute.go
```diff
@@ -40,7 +40,7 @@ func (a Float64AttributeBuilder) BuildResourceAttribute() schema.Attribute {
 	}
 }
 
-func (a Float64AttributeBuilder) SetOptional() AttributeBuilder {
+func (a Float64AttributeBuilder) SetOptional() BaseSchemaBuilder {
```
mgyucht, line 43: Does it work if these functions return `Float64AttributeBuilder`, or does the return type need to be the interface?

internal/providers/pluginfw/tfschema/single_nested_block.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+
+	"github.com/databricks/terraform-provider-databricks/common"
+	dataschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+// SingleNestedBlockBuilder represents a list of complex (non-primitive) types.
```
mgyucht, line 13: Not a list, right? Just a single one?

Is this still used today? Should we even include it if not?

internal/providers/pluginfw/tfschema/struct_to_schema.go
```diff
@@ -12,8 +12,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-func typeToSchema(v reflect.Value) map[string]AttributeBuilder {
-	scm := map[string]AttributeBuilder{}
+func typeToSchema(v reflect.Value) NestedBlockObject {
+	scm_attr := map[string]AttributeBuilder{}
```
mgyucht, line 16: nit: ditto about naming convention.

internal/providers/pluginfw/tfschema/struct_to_schema.go
```diff
@@ -198,36 +200,36 @@ func fieldIsOptional(field reflect.StructField) bool {
...
 
 // DataSourceStructToSchema builds a data source schema from a tfsdk struct, with custoimzations applied.
 func DataSourceStructToSchema(v any, customizeSchema func(CustomizableSchema) CustomizableSchema) dataschema.Schema {
-	attributes := DataSourceStructToSchemaMap(v, customizeSchema)
-	return dataschema.Schema{Attributes: attributes}
+	attributes, blocks := DataSourceStructToSchemaMap(v, customizeSchema)
+	return dataschema.Schema{Attributes: attributes, Blocks: blocks}
 }
 
 // ResourceStructToSchemaMap returns a map from string to resource schema attributes using a tfsdk struct, with custoimzations applied.
```
mgyucht, line 213: Ditto, let's update the doccomments.

mgyucht (approved):
Mostly nits and a couple suggestions, but otherwise this is looking great! Feel free to merge once these are addressed.

internal/providers/pluginfw/tfschema/single_nested_block.go
edwardfeng-db: Yes just a single one and right not used for now

internal/providers/pluginfw/tfschema/float64_attribute.go
edwardfeng-db: That works, but I preferred this to comply with the interface of BaseSchemaBuilder

internal/providers/pluginfw/converters/tf_to_go.go
edwardfeng-db: I think they are kind of equivalent because eventually the internal function will call the public method recursively

PR #4047, opened 2024-09-26
[Internal] Migrate Share Resource to Plugin Framework

rauchy:
## Changes

This PR migrates the share resource to the Plugin framework. The code was largely copied "as is" from the previous implementation of the share resource, with the necessary adaptations made for integration with the Plugin framework.

This implementation utilizes the newly generated Effective fields to provide the functionality that was previously achieved through diff suppression.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+func diff(beforeSi sharing.ShareInfo, afterSi sharing.ShareInfo) []sharing.SharedDataObjectUpdate {
+	beforeMap := resourceShareMap(beforeSi)
+	afterMap := resourceShareMap(afterSi)
+	changes := []sharing.SharedDataObjectUpdate{}
+	// not in after so remove
+	for _, beforeSdo := range beforeSi.Objects {
+		_, exists := afterMap[beforeSdo.Name]
+		if exists {
```
renaudhartert-db, line 85: [nit] inline to make it clear that `exists` is not used outside of the condition. Also, it is more idiomatic to call such variable `ok`:
```suggestion
		if _, ok := afterMap[beforeSdo.Name]; ok {
```

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,396 @@
...
+			Action:     sharing.SharedDataObjectUpdateActionRemove,
+			DataObject: &beforeSdo,
+		})
+	}
+
+	// not in before so add
+	// if in before but diff then update
+	for _, afterSdo := range afterSi.Objects {
+		beforeSdo, exists := beforeMap[afterSdo.Name]
+		if exists {
```
renaudhartert-db, line 98: Ditto 
```suggestion
		if beforeSdo, ok := beforeMap[afterSdo.Name]; ok {
```

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+func shareChanges(si sharing.ShareInfo, action string) sharing.UpdateShare {
+	var changes []sharing.SharedDataObjectUpdate
+	for _, obj := range si.Objects {
+		changes = append(changes, sharing.SharedDataObjectUpdate{
+			Action:     sharing.SharedDataObjectUpdateAction(action),
+			DataObject: &obj,
+		},
+		)
```
renaudhartert-db, line 124: Curious, did this pass `gofmt`?
```suggestion
		changes = append(changes, sharing.SharedDataObjectUpdate{
			Action:     sharing.SharedDataObjectUpdateAction(action),
			DataObject: &obj,
		})
```

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,396 @@
...
+					return
+				}
+			} else {
+				resp.Diagnostics.AddError("failed to roll back", common.OwnerRollbackError(err, rollbackErr, currentShareInfo.Owner, plan.Owner.ValueString()).Error())
+			}
+
+			resp.Diagnostics.AddError("failed to update share", err.Error())
+			return
+		}
+	}
```
renaudhartert-db, line 369: [nit] We can be a little more idiomatic here by only having the error flow at the highest level of indentation.

Ref: https://go.dev/wiki/CodeReviewComments#indent-error-flow

```suggestion
	if len(changes) > 0 {
		// if there are any other changes, update the share with the changes
		updatedShareInfo, err := client.Shares.Update(ctx, sharing.UpdateShare{
			Name:    plan.Name.ValueString(),
			Updates: changes,
		})
		
		if err != nil {
			rollbackShareInfo, rollbackErr := client.Shares.Update(ctx, sharing.UpdateShare{
				Name:  currentShareInfo.Name,
				Owner: currentShareInfo.Owner,
			})
			if rollbackErr == nil {
				resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, rollbackShareInfo, &state)...)
				if resp.Diagnostics.HasError() {
					return
				}
			} else {
				resp.Diagnostics.AddError("failed to roll back", common.OwnerRollbackError(err, rollbackErr, currentShareInfo.Owner, plan.Owner.ValueString()).Error())
			}

			resp.Diagnostics.AddError("failed to update share", err.Error())
			return
		}
		
		resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, updatedShareInfo, &state)...)
		if resp.Diagnostics.HasError() {
			return
		}
	}
```

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,396 @@
...
+			if rollbackErr == nil {
+				resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, rollbackShareInfo, &state)...)
+				if resp.Diagnostics.HasError() {
+					return
+				}
+			} else {
+				resp.Diagnostics.AddError("failed to roll back", common.OwnerRollbackError(err, rollbackErr, currentShareInfo.Owner, plan.Owner.ValueString()).Error())
+			}
+
+			resp.Diagnostics.AddError("failed to update share", err.Error())
```
renaudhartert-db, line 366: [nit] Can this come at the beginning of this `else` block? If so, we can make the code much more readable by reducing the overall level of indentation.

Ref: https://go.dev/wiki/CodeReviewComments#indent-error-flow

```go
resp.Diagnostics.AddError("failed to update share", err.Error())

rollbackShareInfo, rollbackErr := client.Shares.Update(ctx, sharing.UpdateShare{
	Name:  currentShareInfo.Name,
	Owner: currentShareInfo.Owner,
})
if rollbackErr != nil {
	resp.Diagnostics.AddError("failed to roll back", common.OwnerRollbackError(err, rollbackErr, currentShareInfo.Owner, plan.Owner.ValueString()).Error())
	return
}

resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, rollbackShareInfo, &state)...)
	return
}
```

renaudhartert-db (approved):
No major concerns about the business logic (though I'm probably not the best person to review it). I left a few nit comments about style.

rauchy:
Thanks @renaudhartert-db for the review! Most of the issues are inherited from the migrated code, but no reason not to tidy up now :) Appreciate the comments.

internal/providers/pluginfw/resources/sharing/resource_share.go
rauchy: with flying colors 🌈

internal/providers/pluginfw/resources/sharing/resource_acc_test.go
```diff
@@ -0,0 +1,140 @@
...
+			object {
+				name = databricks_table.mytable.id
+				comment = "%s"
+				data_object_type = "TABLE"
+				history_data_sharing_status = "DISABLED"
+			}							
+		}`, owner, comment)
+}
+
+func TestUcAccUpdateShare(t *testing.T) {
```
tanmay-db, line 130: Can you please add an integration test which adds an object in middle in the update step? This was a limitation in SDKv2 as ordering used to matter whereas now it should work with plugin framework.

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,395 @@
...
+
+func ResourceShare() resource.Resource {
+	return &ShareResource{}
+}
+
+type ShareInfoExtended struct {
+	sharing_tf.ShareInfo
+}
+
+func matchOrder[T any, K comparable](target, reference []T, keyFunc func(T) K) {
```
tanmay-db, line 35: Having this covered with unit test might be good.

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,395 @@
...
+		return
+	}
+
+	shareChanges := shareChanges(planGoSDK, string(sharing.SharedDataObjectUpdateActionAdd))
+
+	updatedShareInfo, err := w.Shares.Update(ctx, shareChanges)
+	if err != nil {
+		// delete orphaned share if update fails
+		if d_err := w.Shares.DeleteByName(ctx, shareInfo.Name); d_err != nil {
+			resp.Diagnostics.AddError("failed to delete orphaned share", d_err.Error())
```
tanmay-db, line 202: We should also add err along with d_err in the diagnostics here as we return so line 205 won't execute.

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, newState)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+func (r *ShareResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	logger.SetTfLogger(logger.NewTfLogger(ctx))
```
tanmay-db, line 229: I think we should do this while we configure the client so it gets set for each CRUD in all resources like we do in sdkv2 :https://github.com/databricks/terraform-provider-databricks/blob/main/internal/providers/sdkv2/sdkv2.go#L235. 

It doesn't have to be in this PR so we can remove this and create a ticket to track adding this properly.

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func (d *ShareResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	if d.Client == nil && req.ProviderData != nil {
+		d.Client = pluginfwcommon.ConfigureResource(req, resp)
+	}
+}
+
+func (r *ShareResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	w, diags := r.Client.GetWorkspaceClient()
```
tanmay-db, line 168: Also need to add
`ctx = pluginfwcontext.SetUserAgentInResourceContext(ctx, resourceName)` in each CRUD, ref: https://github.com/databricks/terraform-provider-databricks/blob/main/internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor.go#L101

internal/providers/pluginfw/resources/sharing/resource_share.go
```diff
@@ -0,0 +1,395 @@
...
+			}
+		}
+
+		resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, updatedShareInfo, &state)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+
+	state.SyncEffectiveFieldsDuringCreateOrUpdate(plan.ShareInfo)
```
tanmay-db, line 370: Is there an easy way to test this? Also is this something we do this right after update/create is called? i.e. line 328

tanmay-db (approved):
Thanks for the PR, I had some comments, rest LGTM, stamping to unblock merge once addressed.

internal/providers/pluginfw/resources/sharing/resource_acc_test.go
tanmay-db: I think this would fail currently because object in ShareInfo that we use for schema is a list and it should be set instead. 

````

type ShareInfoExtended struct {
	sharing_tf.ShareInfo
}
````

tanmay-db (commented):
I think we might need to use set instead of list here: https://github.com/databricks/terraform-provider-databricks/pull/4047/files#r1806166473.

internal/providers/pluginfw/resources/sharing/resource_share.go
rauchy: Heh, no this is just sloppy leftovers from debugging. Oops!

internal/providers/pluginfw/resources/sharing/resource_share.go
rauchy: 👍 e3500130

internal/providers/pluginfw/resources/sharing/resource_share.go
tanmay-db: sounds good 👍

Issue #4050, opened 2024-09-27
[Exporter] Make TF exporter compatible with new plugin framework


alexott:
New plugin framework uses other data structures, and as we move new resources to it, we need to add support for these resource

PR #4051, opened 2024-09-27
[Feature] Add `databricks_alert` resource to replace `databricks_sql_alert`

alexott:
## Changes

The new resource uses the [new Alerts API](https://docs.databricks.com/api/workspace/alerts/create) instead of the legacy one that will be deprecated. Since the new resource has a slightly different set of parameters, it was decided to create a new resource and deprecate the old one.

This resource uses old TF SDK to be compatible with TF exporter (until #4050 is implemented).

TODOs:

- Need to discuss how to handle permissions - `sql_alert` permissions look like working, but not sure if we should continue to use that API
- Support in the exporter will be in a separate PR

## Tests

- [x] `make test` run locally
- [x] using Go SDK


internal/acceptance/alert_test.go
```diff
@@ -0,0 +1,76 @@
...
+			query = "SELECT 1 AS p1, 2 as p2"
+		}
+
+        resource "databricks_permissions" "alert_usage" {
+			sql_alert_id = databricks_alert.alert.id
+			access_control {
+              group_name       = "users"
+              permission_level = "CAN_RUN"
+			}
+		}
```
mgyucht, line 22: Let's add this as a separate test case in `permissions_test.go` rather than having it in the resource-specific tests, then we can leverage that testing infra to make sure we cover all cases.

sql/resource_alert.go
```diff
@@ -0,0 +1,121 @@
...
+	"log"
+	"strings"
+
+	"github.com/databricks/databricks-sdk-go/service/sql"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func ResourceAlert() common.Resource {
```
mgyucht, line 14: Did you consider implementing this in the plugin framework?

sql/resource_alert.go
```diff
@@ -0,0 +1,121 @@
...
+	s := common.StructToSchema(sql.Alert{}, func(m map[string]*schema.Schema) map[string]*schema.Schema {
+		common.CustomizeSchemaPath(m, "display_name").SetRequired()
+		common.CustomizeSchemaPath(m, "query_id").SetRequired()
+		common.CustomizeSchemaPath(m, "condition").SetRequired()
+		// TODO: can we automatically generate it from SDK? Or should we avoid validation at all?
+		common.CustomizeSchemaPath(m, "condition", "op").SetRequired().SetValidateFunc(validation.StringInSlice([]string{
```
mgyucht, line 20: Yes, we should generate this in the SDK. We should make this a feature request against the SDK if there isn't already one.

sql/resource_alert.go
```diff
@@ -0,0 +1,121 @@
...
+		// common.CustomizeSchemaPath(m, "condition", "threshold").SetRequired()
+		common.CustomizeSchemaPath(m, "condition", "threshold", "value").SetRequired()
+		alof := []string{
+			"condition.0.threshold.0.value.0.string_value",
+			"condition.0.threshold.0.value.0.double_value",
+			"condition.0.threshold.0.value.0.bool_value",
+		}
+		for _, f := range alof {
+			common.CustomizeSchemaPath(m, "condition", "threshold", "value",
+				strings.TrimPrefix(f, "condition.0.threshold.0.value.0.")).SetAtLeastOneOf(alof)
```
mgyucht, line 39: ExactlyOneOf or AtLeastOneOf?

sql/resource_alert.go
```diff
@@ -0,0 +1,121 @@
...
+			var a sql.CreateAlertRequestAlert
+			common.DataToStructPointer(d, s, &a)
+			apiAlert, err := w.Alerts.Create(ctx, sql.CreateAlertRequest{
+				Alert: &a,
+			})
+			if err != nil {
+				return err
+			}
+			d.SetId(apiAlert.Id)
+			owner := d.Get("owner_user_name").(string)
```
mgyucht, line 66: We can't set Owner on create? :(

sql/resource_alert.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+			apiAlert, err := w.Alerts.GetById(ctx, d.Id())
+			if err != nil {
+				log.Printf("[WARN] error getting alert by ID: %v", err)
+				return err
+			}
+			parentPath := d.Get("parent_path").(string)
+			if parentPath != "" && strings.HasPrefix(apiAlert.ParentPath, "/Workspace") && !strings.HasPrefix(parentPath, "/Workspace") {
+				apiAlert.ParentPath = strings.TrimPrefix(parentPath, "/Workspace")
+			}
```
mgyucht, line 91: If we do this, do we need the custom diff suppression still for parent path?

mgyucht (commented):
Few comments, but otherwise mostly LGTM. Did you try to use the plugin framework, or was it just easier to implement as an SDK resource because the original resource used the SDK?

docs/resources/alert.md
```diff
@@ -0,0 +1,162 @@
...
+      value {
+        double_value = 42
+      }
+    }
+  }
+}
+```
+
+* Remove the old resource from the state with the `terraform state rm databricks_sql_alert.alert` command.
+* Import new resource with the `terraform import databricks_alert.alert <alert-id>` command.
```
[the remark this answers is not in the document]
alexott: ok - let me look into it. We still may need to have migration guide for TF < 1.7.x because some people still on the old versions due to the re-licensing

sql/resource_alert.go
alexott: yes, but we need to have it in the exporter that needs to migrate to the plugin framework as well (until https://github.com/databricks/terraform-provider-databricks/issues/4050 is implemented).

alexott: yes, but we need to have it in the exporter that needs to migrate to the plugin framework as well (until https://github.com/databricks/terraform-provider-databricks/issues/4050 is implemented).

sql/resource_alert.go
alexott: ah, yes - it should be exactly one of

sql/resource_alert.go
alexott: No, I raised it to the Redash team, but they don't support it. We have same situation with UC resources

sql/resource_alert.go
alexott: yes, but we need to have it in the exporter that needs to migrate to the plugin framework as well (until https://github.com/databricks/terraform-provider-databricks/issues/4050 is implemented).

sql/resource_alert.go
alexott: suppress diff is also from empty value

internal/acceptance/alert_test.go
alexott: we can do it there as well.

sql/resource_alert.go
alexott: I didn't find any, filed https://github.com/databricks/databricks-sdk-go/issues/1063 - it's possible to get something with `reflect`, but it's better to have in SDK.

sql/resource_alert.go
```diff
@@ -0,0 +1,125 @@
...
+			return common.StructToData(apiAlert, s, d)
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var a sql.UpdateAlertRequestAlert
+			updateMask := "display_name,query_id,seconds_to_retrigger,condition,custom_body,custom_subject"
```
mgyucht, line 100: Do we use updateMask anywhere else in the TF provider? I wonder if we should always set it for all fields, even if fields don't have any change, since the `*schema.ResourceData` has the desired values for all of those fields. Mainly asking from the perspective of if we need logic around setting update masks or if we can avoid it to make things simpler.

mgyucht (approved):
LGTM, one small question about update mask but otherwise it's great.

PR #4056, opened 2024-09-28
[Fix] Add `no_wait` flag to `databricks_online_table` to avoid timeouts

alexott:
## Changes

When creating an online table from the big Delta table, waiting for full sync could take too long time, and creation timeouts. The new `no_wait` flag allows to skip wait for online table creation.

It would be useful to merge #4048 as well

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_online_table.go
+++ b/catalog/resource_online_table.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"time"
 
 	"github.com/databricks/databricks-sdk-go"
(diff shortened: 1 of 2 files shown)


catalog/resource_online_table.go
```diff
@@ -64,6 +65,16 @@ func ResourceOnlineTable() common.Resource {
 			runTypes := []string{"spec.0.run_triggered", "spec.0.run_continuously"}
 			common.CustomizeSchemaPath(m, "spec", "run_triggered").SetAtLeastOneOf(runTypes).SetSuppressDiff()
 			common.CustomizeSchemaPath(m, "spec", "run_continuously").SetAtLeastOneOf(runTypes).SetSuppressDiff()
+
+			m["no_wait"] = &schema.Schema{
```
tanmay-db, line 69: Curious -- how useful would this be? My understanding is that it would fail most of the times if waiting for online table creation takes a few minutes. But if this helps some use cases, then I think it might be helpful to have this.

tanmay-db (approved):

catalog/resource_online_table.go
alexott: I need to retest - we have a recent change when the product team added a native wait functionality, but it's not clear how much time we'll still spend on waiting

alexott:
We don't need it anymore - with new waiter implementation the creation of the table is fast.

PR #4057, opened 2024-09-30
[Internal] Generate Effective Fields

rauchy:
This PR introduces changes to the TFSDK generator to support “effective fields” for server-provided values. When fields are marked with the new proto annotation `ServerProposedIfEmpty` (name pending final decision, until then the `computed` annotation is used to indicate these fields, while specific fields are excluded), the generator will create an additional computed field (e.g., `Effective<FieldName>`) and add two sync functions to ensure proper handling of user-provided and server-determined values.

### Generated Struct:
```
type ResourceModel struct {
    OriginalField    types.String `tfsdk:"original_field" tf:"optional"`
    EffectiveField   types.String `tfsdk:"effective_field" tf:"computed"`
}
```
### Sync Functions:
```
func (newState *ResourceModel) SyncEffectiveFieldsDuringCreateOrUpdate(plan ResourceModel) {
    newState.EffectiveField = newState.OriginalField
    newState.OriginalField = plan.OriginalField
}

func (newState *ResourceModel) SyncEffectiveFieldsDuringRead(existingState ResourceModel) {
    if existingState.EffectiveField.ValueString() == newState.OriginalField.ValueString() {
        newState.OriginalField = existingState.OriginalField
    }
}
```

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/.codegen/model.go.tmpl
+++ b/.codegen/model.go.tmpl
@@ -18,20 +18,74 @@ import (
 	"github.com/databricks/databricks-sdk-go/marshal"
     "github.com/hashicorp/terraform-plugin-framework/types"
 )
+{{- $excluded := dict "ShareInfo"        (list "CreatedAt" "CreatedBy" "UpdatedAt" "UpdatedBy")
+                      "SharedDataObject" (list "AddedAt" "AddedBy" "Status") -}}
 {{range .Types}}
 {{- if or .Fields .IsEmpty}}
 {{.Comment "// " 80}}
 type {{.PascalName}} struct {
+    {{- $excluded := getOrDefault $excluded .PascalName (list) -}}
     {{- range .Fields}}
     {{.Comment "    // " 80}}
-    {{.PascalName}} {{template "type" .Entity}} `{{template "field-tag" . }}`{{end}}
+	{{- $data := dict "field" . "excluded" $excluded }}
+	{{template "field" $data}}{{if and .Entity.IsComputed (not (in $excluded .PascalName))}}{{ $data := dict "field" . "excluded" $excluded "effective" true }}{{printf "\n"}}{{template "field" $data}}{{end}}{{end}}
+}
+
+func (newState *{{.PascalName}}) SyncEffectiveFieldsDuringCreateOrUpdate(plan {{.PascalName}}) {
+  {{- range .Fields -}}
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  newState.Effective_{{.PascalName}} = newState.{{.PascalName}}
+  newState.{{.PascalName}} = plan.{{.PascalName}}
+  {{end}}
+  {{end}}
+  {{end}}
+}
+
+func (newState *{{.PascalName}}) SyncEffectiveFieldsDuringRead(existingState {{.PascalName}}) {
+  {{- range .Fields -}}
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  {{- $type := "" -}}
+  {{- if .Entity.IsString}}{{$type = "String"}}{{end}}
+  {{- if .Entity.IsBool}}{{$type = "Bool"}}{{end}}
+  {{- if .Entity.IsInt64}}{{$type = "Int64"}}{{end}}
+  {{- if .Entity.IsFloat64}}{{$type = "Float64"}}{{end}}
+  {{- if .Entity.IsInt}}{{$type = "Int64"}}{{end}}
+  {{- if .Entity.Enum}}{{$type = "String"}}{{end}}
+  if existingState.Effective_{{.PascalName}}.Value{{$type}}() == newState.{{.PascalName}}.Value{{$type}}() {
+    newState.{{.PascalName}} = existingState.{{.PascalName}}
+  }
+  {{end}}
+  {{end}}
+  {{end}}
 }
 
 {{end}}
 {{end}}
 
+{{- define "field" -}}
+{{if .effective}}Effective_{{end}}{{.field.PascalName}} {{template "type" .field.Entity}} `{{template "field-tag" . }}`
+{{- end -}}
+
 {{- define "field-tag" -}}
-  {{if .IsJson}}tfsdk:"{{if and (ne .Entity.Terraform nil) (ne .Entity.Terraform.Alias "") }}{{.Entity.Terraform.Alias}}{{else}}{{.Name}}{{end}}" tf:"{{- $first := true -}}{{- if not .Required -}}{{- if not $first -}},{{end}}optional{{- $first = false -}}{{- end -}}{{- if .Entity.IsObject -}}{{- if not $first -}},{{end}}object{{- $first = false -}}{{- end -}}"{{else}}tfsdk:"-"{{end -}}
+    {{- $annotations := "" -}}
+	{{- if in .excluded .field.PascalName -}}
+		{{- $annotations = (printf "%scomputed,optional," $annotations) -}}
+	{{- else if .effective -}}
+		{{- $annotations = (printf "%scomputed,optional," $annotations) -}}
+	{{- else -}}
+		{{- if not .field.Required -}}
+ 	       {{- $annotations = (printf "%soptional," $annotations) -}}
+  	    {{- end -}}
+		{{- if .field.Entity.IsObject -}}
+ 	       {{- $annotations = (printf "%sobject," $annotations) -}}
+  	    {{- end -}}
+	{{- end -}}
+ 	{{- if gt (len $annotations) 0 -}}
+ 	   {{- $annotations = (printf "%s" (trimSuffix "," $annotations)) -}}
+ 	{{- end -}}
+	{{if .field.IsJson}}tfsdk:"{{if and (ne .field.Entity.Terraform nil) (ne .field.Entity.Terraform.Alias "") }}{{.field.Entity.Terraform.Alias}}{{else}}{{if .effective}}effective_{{end}}{{.field.Name}}{{end}}" tf:"{{$annotations}}"{{else}}tfsdk:"-"{{end -}}
 {{- end -}}
 
 {{- define "type" -}}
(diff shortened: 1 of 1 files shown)


hectorcast-db (approved):

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
...
+}
+
+func (newState *{{.PascalName}}) SyncEffectiveFieldsDuringRead(existingState {{.PascalName}}) {
+  {{- range .Fields -}}
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  {{- $type := "" -}}
+  {{- if .Entity.IsString}}{{$type = "String"}}{{end}}
+  {{- if .Entity.IsBool}}{{$type = "Bool"}}{{end}}
+  {{- if .Entity.IsInt64}}{{$type = "Int64"}}{{end}}
```
tanmay-db, line 52: Might be good to also add Int32? (types.Int32)

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
...
+  {{- if .Entity.IsFloat64}}{{$type = "Float64"}}{{end}}
+  {{- if .Entity.IsInt}}{{$type = "Int64"}}{{end}}
+  {{- if .Entity.Enum}}{{$type = "String"}}{{end}}
+  if existingState.Effective_{{.PascalName}}.Value{{$type}}() == newState.{{.PascalName}}.Value{{$type}}() {
```
tanmay-db, line 56: Note: .Value<type> returns zero value (eg: "" string if string is null or unknown), is there a case where we would want to differentiate between these? Right now if one is null / one is unknown, then we will compare them to be the same.

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
...
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  newState.Effective_{{.PascalName}} = newState.{{.PascalName}}
+  newState.{{.PascalName}} = plan.{{.PascalName}}
+  {{end}}
+  {{end}}
+  {{end}}
+}
+
+func (newState *{{.PascalName}}) SyncEffectiveFieldsDuringRead(existingState {{.PascalName}}) {
```
tanmay-db, line 45: For my understanding, where would we be calling these in the read? Asking because in case there is an error in the read, we shouldn't sync.

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
...
+  {{- range .Fields -}}
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  {{- $type := "" -}}
+  {{- if .Entity.IsString}}{{$type = "String"}}{{end}}
+  {{- if .Entity.IsBool}}{{$type = "Bool"}}{{end}}
+  {{- if .Entity.IsInt64}}{{$type = "Int64"}}{{end}}
+  {{- if .Entity.IsFloat64}}{{$type = "Float64"}}{{end}}
```
tanmay-db, line 53: same for float32

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
...
+  {{end}}
+  {{end}}
+}
+
+func (newState *{{.PascalName}}) SyncEffectiveFieldsDuringRead(existingState {{.PascalName}}) {
+  {{- range .Fields -}}
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  {{- $type := "" -}}
+  {{- if .Entity.IsString}}{{$type = "String"}}{{end}}
```
tanmay-db, line 50: I am guessing -- we don't support other types like Object, list, map for now and this is to unblock share migration right?

tanmay-db (approved):
codewise LGTM, had some questions.

internal/service/jobs_tf/model.go
```diff
@@ -22,13 +22,36 @@ type BaseJob struct {
...
+	Effective_EffectiveBudgetPolicyId types.String `tfsdk:"effective_effective_budget_policy_id" tf:"computed,optional"`
 	// The canonical identifier for this job.
 	JobId types.Int64 `tfsdk:"job_id" tf:"optional"`
 	// Settings for this job and all of its runs. These settings can be updated
 	// using the `resetJob` method.
 	Settings []JobSettings `tfsdk:"settings" tf:"optional,object"`
 }
 
+func (newState *BaseJob) SyncEffectiveFieldsDuringCreateOrUpdate(plan BaseJob) {
+
```
tanmay-db, line 41: nit: The space here and below can be removed.

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
...
-    {{.PascalName}} {{template "type" .Entity}} `{{template "field-tag" . }}`{{end}}
+	{{- $data := dict "field" . "excluded" $excluded }}
+	{{template "field" $data}}{{if and .Entity.IsComputed (not (in $excluded .PascalName))}}{{ $data := dict "field" . "excluded" $excluded "effective" true }}{{printf "\n"}}{{template "field" $data}}{{end}}{{end}}
+}
+
+func (newState *{{.PascalName}}) SyncEffectiveFieldsDuringCreateOrUpdate(plan {{.PascalName}}) {
+  {{- range .Fields -}}
+  {{- if and .Entity.IsComputed (or .Entity.IsString .Entity.IsBool .Entity.IsInt64 .Entity.IsFloat64 .Entity.IsInt .Entity.Enum) -}}
+  {{- if not (in $excluded .PascalName) -}}
+  newState.Effective_{{.PascalName}} = newState.{{.PascalName}}
```
tanmay-db, line 38: Do we want the _ in the variable names? Also it seems like some places have 2 Effective, highlighted in another comment.

internal/service/sharing_tf/model.go
```diff
@@ -567,34 +931,66 @@ type RotateRecipientToken struct {
...
-	CreatedAt types.Int64 `tfsdk:"created_at" tf:"optional"`
+	CreatedAt types.Int64 `tfsdk:"created_at" tf:"computed,optional"`
 	// Username of share creator.
-	CreatedBy types.String `tfsdk:"created_by" tf:"optional"`
+	CreatedBy types.String `tfsdk:"created_by" tf:"computed,optional"`
 	// Name of the share.
 	Name types.String `tfsdk:"name" tf:"optional"`
 	// A list of shared data objects within the share.
-	Objects []SharedDataObject `tfsdk:"objects" tf:"optional"`
+	Objects []SharedDataObject `tfsdk:"object" tf:"optional"`
```
tanmay-db, line 967: was this (objects -> object) done upstream?

internal/service/jobs_tf/model.go
```diff
@@ -22,13 +22,36 @@ type BaseJob struct {
...
 	// the user has already been deleted.
 	CreatorUserName types.String `tfsdk:"creator_user_name" tf:"optional"`
+	// The id of the budget policy used by this job for cost attribution
+	// purposes. This may be set through (in order of precedence): 1. Budget
+	// admins through the account or workspace console 2. Jobs UI in the job
+	// details page and Jobs API using `budget_policy_id` 3. Inferred default
+	// based on accessible budget policies of the run_as identity on job
+	// creation or modification.
+	EffectiveBudgetPolicyId           types.String `tfsdk:"effective_budget_policy_id" tf:"optional"`
+	Effective_EffectiveBudgetPolicyId types.String `tfsdk:"effective_effective_budget_policy_id" tf:"computed,optional"`
```
tanmay-db, line 32: Rare case when the field name itself upstream starts with Effective?

.codegen/_openapi_sha
```diff
@@ -1 +1 @@
-0c86ea6dbd9a730c24ff0d4e509603e476955ac5
\ No newline at end of file
+cf9c61453990df0f9453670f2fe68e1b128647a2
```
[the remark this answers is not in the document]
rauchy: We don't really need them as part of this PR, it was more to give an example of the changes. I'll remove that commit.

.codegen/model.go.tmpl
```diff
@@ -18,20 +18,74 @@ import (
 	"github.com/databricks/databricks-sdk-go/marshal"
     "github.com/hashicorp/terraform-plugin-framework/types"
 )
+{{- $excluded := dict "ShareInfo"        (list "CreatedAt" "CreatedBy" "UpdatedAt" "UpdatedBy")
```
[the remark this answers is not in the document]
rauchy: We will soon implement a new annotation (`ServerProposedIfEmpty`) for these effective fields. This annotation will not be applied to the excluded fields, allowing us to eliminate the exclusion list. As a result, the list will not expand to include other resources.

.codegen/model.go.tmpl
rauchy: Possibly, but I was only transferring what was already in place, so I wasn't aware of the consequences of making changes that fall outside the scope of this PR.

.codegen/model.go.tmpl
rauchy: I believe that using an underscore helps differentiate the actual field name, especially in rare edge cases where the field name includes "Effective." Personally, it helps my understanding. However, if using underscores is considered a significant style violation in Golang, we can reconsider our approach.

.codegen/model.go.tmpl
rauchy: `SyncEffectiveFieldsDuringRead` is a function that should be invoked before saving to the state during a Read operation, while `SyncEffectiveFieldsDuringCreateOrUpdate` should be called before saving to the state during a Create or Update operation. We might consider introducing an interface here to allow this syncing to be done automatically by the Resource. However, I would prefer to postpone this refactoring for the time being.

.codegen/model.go.tmpl
rauchy: Correct. For the moment, I chose primitives because I didn't see a use case for other options. We should revisit this decision in the future if we encounter any relevant cases.

internal/service/sharing_tf/model.go
rauchy: Yupp

internal/service/jobs_tf/model.go
rauchy: Wonderful :)

rauchy: Anyway, dropped underscores.

internal/service/jobs_tf/model.go
rauchy: Addressed in [ddc10d2](https://github.com/databricks/terraform-provider-databricks/pull/4057/commits/ddc10d206ab5898c1c8c3a65fbc6477d3c65544d)

Issue #4066, opened 2024-10-02
[ISSUE] Issue with `databricks_system_schema` resource 


Waiwait:
### Configuration

`databricks_system_schema` resource passes silently when a schema has been disabled outside of terraform. This was discovered when we switched the metastore for a workspace and then dependent grants on the `system_schema` started failing after not being able to access the table. 

### Expected Behavior
Terraform correctly recognises that the schema has been disabled outside of the state and acts appropriately.

### Actual Behavior
Terraform silently passes, assuming that the schema is still enabled.

### Terraform and provider versions

`terraform-provider-databricks=1.49.0`
`terraform=v1.55.0`

### Would you like to implement a fix?

I think this is caused by the `Read` function simply returning `nil` at the end of the for loop if no appropriate schema is found. This would be fixed by adding `d.SetId("")` to this, so that on the next run, it'll force a `Create` call. Happy to implement, just wanted to make sure this was the right approach.

alexott:
yes, we're using similar approach in other resources, like, library: https://github.com/databricks/terraform-provider-databricks/blob/main/clusters/resource_library.go#L86

alexott:
Fixed by #4068

PR #4068, opened 2024-10-03
[Fix] Recreate missing system schema 

Waiwait:
## Changes

Addresses https://github.com/databricks/terraform-provider-databricks/issues/4066.
Makes read function more in line with rest of module to ignore non `enabled` schemas.

## Tests

- [x] `make test` run locally
- [ ] ~using Go SDK~


catalog/resource_system_schema.go
```diff
@@ -93,10 +93,21 @@ func ResourceSystemSchema() common.Resource {
 					if err != nil {
 						return err
 					}
+
+					// only track enabled schemas inline with rest of module
+					if schema.State != catalog.SystemSchemaInfoStateEnableCompleted &&
```
alexott, line 98: We also need to handle NotAvailable - these are schemas that you can't enable anymore

Waiwait: What should happen differently in this case? Currently (with this PR), it releases it from state and then assumedly fails when the Create function is called in the next run.

alexott: @Waiwait I think that we need to exclude `UNAVAILABLE` status as well.

Waiwait: Okay, I get it, thanks. Have added.

alexott (approved):

PR #4072, opened 2024-10-04
[Fix] Set ID for online table resource if creation succeeds but it isn't available yet

tanmay-db:
## Changes

We should set the id right after creation and before waiting for online table to be available. This is because in case when online table isn't available, we still should have that resource in the state.

Also the timeout has been increased to 2x (I am going to following up with online tables team for suitable timeout but since we have to do a release, going ahead with small time increase should be good) 

Note: We should add setting id right after creation for similar resources to CONTRIBUTING guide (which I will do in a separate PR)

## Tests

Added unit test to check that pathway, id is set (which wasn't the case before)
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_online_table.go
+++ b/catalog/resource_online_table.go
@@ -81,7 +81,7 @@ func ResourceOnlineTable() common.Resource {
 				return err
 			}
 			// Note: We should set the id right after creation and before waiting for online table to be available.
-			// This is because in case when online table isn't availabe, we still should have that resource in the state.
+			// If the resource creation timeout is exceeded while waiting for the online table to be ready, this ensures the online table is persisted in the state.
 			d.SetId(res.Name)
 			// this should be specified in the API Spec - filed a ticket to add it
 			err = waitForOnlineTableCreation(w, ctx, res.Name)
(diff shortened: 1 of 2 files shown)


alexott:
if it will fail to initialize online table during waiting, it will be in the broken state. We really need to get the definitive answer from the dev team about condition when table is considered successfully created and initialized.

tanmay-db:
@alexott that is true but the resource would still be there in the failed state and we should put this in the state even if the resources is not available (but creation succeeds).

catalog/resource_online_table.go
```diff
@@ -80,13 +80,14 @@ func ResourceOnlineTable() common.Resource {
 			if err != nil {
 				return err
 			}
+			// Note: We should set the id right after creation and before waiting for online table to be available.
+			// This is because in case when online table isn't availabe, we still should have that resource in the state.
```
mgyucht, line 84: nit: slightly more clear/precise way of explaining this
```suggestion
			// If the resource creation timeout is exceeded while waiting for the online table to be ready, this ensures the online table is persisted in the state.
```

mgyucht (approved):
LGTM, just a small suggestion on the comment.

Could you add a note about this in the contributing guide?

tanmay-db:
Hi @mgyucht, yes regarding contributing guide, I will be doing that in a separate PR so we isolate the changes.

PR #4085, opened 2024-10-09
[Internal] Support adding context in resources and data sources

tanmay-db:
## Changes

Set `resource` and `data` context for each resource in their respective CRUD operations. Two methods are defined to make setting resources / datasources as default easier: `GetDatabricksProductionName` and `GetDatabricksStagingName`.

Note: Library update isn't supported and hence we shouldn't inject useragent there. 

Every resource has:

`````
        ctx = pluginfwcommon.SetResourceNameInContext(ctx, resourceName)
	w, diags := r.Client.GetWorkspaceClient()
	resp.Diagnostics.Append(diags...)
		return
	}
`````
We could avoid the repetition by creating a common base struct but since this is going to be autogenerated soon, I think it's fine to have these as is due to return statements / clarity of reading the code. 

## Tests

Unit tests
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/common/common.go
+++ b/internal/providers/pluginfw/common/common.go
@@ -1,8 +1,10 @@
 package common
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/databricks/databricks-sdk-go/useragent"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: internal/providers/pluginfw/common/common.go [hunk 1]
```diff
@@ -45,3 +47,21 @@ func ConfigureResource(req resource.ConfigureRequest, resp *resource.ConfigureRe
...
+// Once a migrated resource is ready to be used as default, the Metadata method for that resource should be updated to use GetDatabricksProductionName.
+func GetDatabricksStagingName(name string) string {
+	return fmt.Sprintf("databricks_%s_pluginframework", name)
+}
+
+func GetDatabricksProductionName(name string) string {
+	return fmt.Sprintf("databricks_%s", name)
+}
+
+func SetResourceNameInContext(ctx context.Context, resourceName string) context.Context {
```
mgyucht, line 61: To prevent `common.go` from being a dumping ground, let's put this into its own file, e.g. `useragent.go` or `context.go`.

mgyucht (approved):
This seems good to me.

When we talk about the rollout of the plugin framework resources, we probably would benefit from adding one more bit to the user agent indicating if the plugin framework is being used. That way, we can detect if users need to roll back to the SDK implementation of any resources. cc @edwardfeng-db

PR #4094, opened 2024-10-10
[Internal] Add maxItem=1 validator for object types in plugin framework schema

edwardfeng-db:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


.codegen/model.go.tmpl
```diff
@@ -31,7 +31,7 @@ type {{.PascalName}} struct {
 {{end}}
 
 {{- define "field-tag" -}}
-	{{if .IsJson}}tfsdk:"{{if and (ne .Entity.Terraform nil) (ne .Entity.Terraform.Alias "") }}{{.Entity.Terraform.Alias}}{{else}}{{.Name}}{{end}}" tf:"{{if not .Required}}optional{{end}}"{{else}}tfsdk:"-"{{end -}}
+  {{if .IsJson}}tfsdk:"{{if and (ne .Entity.Terraform nil) (ne .Entity.Terraform.Alias "") }}{{.Entity.Terraform.Alias}}{{else}}{{.Name}}{{end}}" tf:"{{- $first := true -}}{{- if not .Required -}}{{- if not $first -}},{{end}}optional{{- $first = false -}}{{- end -}}{{- if .Entity.IsObject -}}{{- if not $first -}},{{end}}object{{- $first = false -}}{{- end -}}"{{else}}tfsdk:"-"{{end -}}
 {{- end -}}
```
edwardfeng-db, line 35: This is kinda ugly, mainly because I need to keep track of whether the item is the first one in the `tf` tags to add commas in the proper positions

edwardfeng-db: @hectorcast-db wondering if there's a better way

hectorcast-db: You can create a function in the Go code to return the tfsdk options already formatted. So you can have all this logic in Go, which is more readable.

{{if .IsJson}}tfsdk:".Entity.TFSDKOptions" tf:".Entify.TFOptions"{{end-}}

hectorcast-db (approved):
Added a comment on how to make it more readable. Otherwise, LGTM

internal/providers/pluginfw/tfschema/single_item_list_validator.go
```diff
@@ -0,0 +1,47 @@
+package tfschema
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+// MaxItemsValidator is a custom validator that enforces the maximum number of items in a list.
```
mgyucht, line 10: Nothing against your implementation, but can we use the existing sizeAtMost validator? https://github.com/hashicorp/terraform-plugin-framework-validators/blob/main/listvalidator/size_at_most.go

This just minimizes the amount of logic that we're responsible for maintaining/testing. I think your implementation is identical but it should be OK to depend on these.

.codegen/model.go.tmpl
```diff
@@ -31,7 +31,7 @@ type {{.PascalName}} struct {
 {{end}}
 
 {{- define "field-tag" -}}
-	{{if .IsJson}}tfsdk:"{{if and (ne .Entity.Terraform nil) (ne .Entity.Terraform.Alias "") }}{{.Entity.Terraform.Alias}}{{else}}{{.Name}}{{end}}" tf:"{{if not .Required}}optional{{end}}"{{else}}tfsdk:"-"{{end -}}
+  {{if .IsJson}}tfsdk:"{{if and (ne .Entity.Terraform nil) (ne .Entity.Terraform.Alias "") }}{{.Entity.Terraform.Alias}}{{else}}{{.Name}}{{end}}" tf:"{{- $first := true -}}{{- if not .Required -}}{{- if not $first -}},{{end}}optional{{- $first = false -}}{{- end -}}{{- if .Entity.IsObject -}}{{- if not $first -}},{{end}}object{{- $first = false -}}{{- end -}}"{{else}}tfsdk:"-"{{end -}}
```
mgyucht, line 34: `object` is bit too vague as a name. Maybe `singleObject` is more obvious?

internal/providers/pluginfw/tfschema/struct_to_schema.go
```diff
@@ -32,6 +33,7 @@ func typeToSchema(v reflect.Value) NestedBlockObject {
 		}
 		isOptional := fieldIsOptional(typeField)
 		isComputed := fieldIsComputed(typeField)
+		isObject := fieldIsAnObject(typeField)
```
mgyucht, line 36: Implementation suggestion: we should define a struct that captures all elements of the struct tag:
```go
type structTag struct {
        optional bool
        computed bool
        singleObject bool
}
```
Then we have one method to extract the struct tag. This means less plumbing of the struct attributes.

The other huge advantage of this is that it is a single, natural place where we can document all supported struct tags.

mgyucht (commented):
Behavior seems correct to me, but it would be good to take my points into account before merging.

internal/providers/pluginfw/tfschema/struct_to_schema.go
edwardfeng-db: Great idea, thanks!

.codegen/model.go.tmpl
edwardfeng-db: Sounds good, merging this PR for now to unblock and I'll work on the go function as a follow up

PR #4097, opened 2024-10-11
[Fix] Fix databricks_cluster_pluginframework data source

mgyucht:
## Changes
After https://github.com/databricks/terraform-provider-databricks/commit/bac842d603465ddbf3582e2dab2d9189c17e3113 , the autogenerated structures used for interacting with state/config/plan always use Lists for nested structures, even if there is only ever at most one, for compatibility with older versions of the TF provider. The cluster data source on the plugin framework is handwritten and also needs to be converted to use lists instead of a pointer as well.

## Tests
Ran the `TestAccDataSourceClusterByID` and `TestAccDataSourceClusterByName` integration tests, which now pass.

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/resources/cluster/data_cluster.go
+++ b/internal/providers/pluginfw/resources/cluster/data_cluster.go
@@ -32,9 +32,9 @@ type ClusterDataSource struct {
 }
 
 type ClusterInfo struct {
-	ClusterId   types.String               `tfsdk:"cluster_id" tf:"optional,computed"`
-	Name        types.String               `tfsdk:"cluster_name" tf:"optional,computed"`
-	ClusterInfo *compute_tf.ClusterDetails `tfsdk:"cluster_info" tf:"optional,computed"`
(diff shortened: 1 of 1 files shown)


alexott:
Why can't we do such a conversion automatically? It makes it harder for people to port existing code - I've got the same problem with the registered model data source yesterday...

mgyucht:
That's a good question @alexott. Unfortunately we have to do it this way because the issue arises when calling the `Get()` method on `tfsdk.Value` types like `req.Plan`/`req.Config`/`req.State` in order to preserve compatibility with current resources. SDKv2 only supports list-like data structures for nested structure attributes, but the plugin framework is more flexible. So, any resources that are being migrated need to use lists for nested structure attributes, otherwise it would be a breaking change.

The error we're seeing is what happens when calling `Value.Get()` with a non-list attribute when the schema indicates a list. You could work around this by defining a second data type with a list in place of the pointer, but that seems like overkill to me.

Note that this conversion is handled automatically by the GoToTfSDk/TfSDKToGo converters.

PR #4099, opened 2024-10-12
[Feature] Add `databricks_app` resource

### LINKED ISSUE #4084: [FEATURE] Add support for Databricks Apps
Databricks Apps are now in the [public preview](https://www.databricks.com/blog/introducing-databricks-apps), so it would be useful to add Terraform integration for it based on [published APIs](https://docs.databricks.com/api/workspace/apps).

nkvuong:
## Changes
- Added `databricks_app` resource

Resolves #4084

## Tests

- [x] `make test` run locally
- [x] using Go SDK


apps/resource_app.go
```diff
@@ -0,0 +1,181 @@
...
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			_, err = w.Apps.Update(ctx, update.UpdateAppRequest)
+			if err != nil {
+				return err
+			}
+			if d.HasChanges("source_code_path", "mode") {
+				_, err = w.Apps.Deploy(ctx, apps.CreateAppDeploymentRequest{
```
alexott, line 157: Should we wait here as well?

docs/resources/app.md
```diff
@@ -0,0 +1,80 @@
...
+
+Apps run directly on a customer’s Databricks instance, integrate with their data, use and extend Databricks services, and enable users to interact through single sign-on.
+
+## Example Usage
+
+```hcl
+resource "databricks_app" "this" {
+  name             = "my-custom-app"
+  description      = "My app"
+  source_code_path = "/Workspace/user@test.com/my_custom_app"
```
alexott, line 16: should be `/Workspace/Users/user@test.com/...` ?

docs/resources/app.md
```diff
@@ -0,0 +1,80 @@
...
+
+The following arguments are required:
+
+* `name` - (Required) The name of the app. The name must contain only lowercase alphanumeric characters and hyphens. It must be unique within the workspace.
+* `description` - The description of the app.
+
+One or more `resource` block with the following arguments:
+
+* `name` - Name of the App Resource.
+* `description` - Description of the App Resource.
```
alexott, line 31: I think that it should be removed

alexott (changes requested):
small changes in the doc required, otherwise code looks good

apps/resource_app.go
```diff
@@ -0,0 +1,181 @@
...
+			if err != nil {
+				log.Printf("[ERROR] Error waiting for app to be created: %s", err.Error())
+				_, nestedErr := w.Apps.DeleteByName(ctx, createApp.Name)
+				if nestedErr != nil {
+					log.Printf("[ERROR] Error cleaning up app: %s", nestedErr.Error())
+				}
+				return err
+			}
+			// now deploy the app, using the source code path
+			createAppDeployment.AppName = app.Name
```
andrewnester, line 117: I think we discussed it offline and agreed that the deployment won't be part of TF configuration. Instead, if an users want to deploy, they use CLI to do that. Otherwise it's confusing because we do the deployments only when source_code_path changed but we need to do deployments when source code is changed.

@pietern wdyt?

apps/resource_app.go
```diff
@@ -0,0 +1,181 @@
...
+			common.DataToStructPointer(d, appSchema, &update)
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			_, err = w.Apps.Update(ctx, update.UpdateAppRequest)
+			if err != nil {
+				return err
+			}
+			if d.HasChanges("source_code_path", "mode") {
```
andrewnester, line 156: See my comment above, I think we should remove app deployment altogether for now

apps/resource_app.go
```diff
@@ -0,0 +1,116 @@
...
+	// Required fields & validation
+	s.SchemaPath("name").SetRequired().SetForceNew().SetValidateFunc(validation.StringLenBetween(2, 30))
+
+	// Computed fields
+	for _, p := range []string{"active_deployment", "app_status", "compute_status", "create_time", "creator",
+		"default_source_code_path", "pending_deployment", "service_principal_id", "service_principal_name",
+		"update_time", "updater", "url"} {
+		s.SchemaPath(p).SetComputed()
+	}
+	return s
```
pietern, line 42: The resource types are mutually exclusive per resource element (`secret`, `sql_warehouse`, etc)

apps/resource_app_test.go
```diff
@@ -0,0 +1,284 @@
...
+								Permission: "CAN_MANAGE",
+							},
+							ServingEndpoint: &apps.AppResourceServingEndpoint{
+								Name:       "databricks-meta-llama-3-1-70b-instruct",
+								Permission: "CAN_MANAGE",
+							},
+							Job: &apps.AppResourceJob{
+								Id:         "1234",
+								Permission: "CAN_MANAGE",
+							},
```
pietern, line 151: These are mutually exclusive AFAICT.

docs/resources/app.md
```diff
@@ -0,0 +1,74 @@
...
+
+Apps run directly on a customer’s Databricks instance, integrate with their data, use and extend Databricks services, and enable users to interact through single sign-on.
+
+## Example Usage
+
+```hcl
+resource "databricks_app" "this" {
+  name             = "my-custom-app"
+  description      = "My app"
+}
```
pietern, line 16: This can include a warehouse reference (should be common).

docs/resources/app.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+## Argument Reference
+
+The following arguments are required:
+
+* `name` - (Required) The name of the app. The name must contain only lowercase alphanumeric characters and hyphens. It must be unique within the workspace.
+* `description` - The description of the app.
+
+One or more `resource` block with the following arguments:
```
pietern, line 26: The `name` is required and `description` optional.

apps/resource_app.go
```diff
@@ -0,0 +1,116 @@
...
+	"time"
+
+	"github.com/databricks/databricks-sdk-go/service/apps"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+const defaultAppProvisionTimeout = 10 * time.Minute
+const deleteCallTimeout = 10 * time.Second
```
pietern, line 15: Nit:
```
const (
  ...
)
```

pietern (commented):
Thanks, @nkvuong !

As this doesn't include deployments yet we should figure that out next.

For the casual reader, the difficulty lies in TF being unable to detect whether or not the source code for an app has changed and therefore if an app must be redeployed or not.

apps/resource_app.go
nkvuong: you can add multiple resources to an app

apps/resource_app.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+
+			// create the app
+			wait, err := w.Apps.Create(ctx, apps.CreateAppRequest{
+				App: &createApp.App,
+			})
+			if err != nil {
+				return err
+			}
+			app, err := wait.GetWithTimeout(d.Timeout(schema.TimeoutCreate) - deleteCallTimeout)
```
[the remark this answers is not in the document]
nkvuong: my understanding is that it is enforced by Terraform itself, so we need to avoid dangling apps by never let the resource hits the timeout

apps/resource_app.go
pietern: Yes, but within a resource you should only have one of (`secret`, `sql_warehouse`, etc).

alexott (changes requested):
code looks good, changes required for docs

apps/resource_app.go
```diff
@@ -0,0 +1,132 @@
...
+func (appStruct) Aliases() map[string]map[string]string {
+	return map[string]map[string]string{
+		"apps.appStruct": appAliasMap,
+	}
+}
+
+func (appStruct) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
+
+	// Required fields & validation
+	s.SchemaPath("name").SetRequired().SetForceNew().SetValidateFunc(validation.StringMatch(regexp.MustCompile("^[a-z-]{2,30}$"), "name must contain only lowercase alphanumeric characters and hyphens, and be between 2 and 30 characters long"))
```
pietern, line 37: Style: this line is very long

apps/resource_app.go
```diff
@@ -0,0 +1,132 @@
...
+	// Resources should be a set
+	s.SchemaPath("resource").SetSliceSet()
```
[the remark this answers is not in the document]
pietern: Do we gain much from treating this as a set vs a list?

AFAICT this only allows the user to shuffle the order if they want.

apps/resource_app.go
pietern: Could we add this constraint back, where you cannot have more than one defined per resource?

apps/resource_app.go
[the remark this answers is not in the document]
pietern: Thanks!

docs/resources/app.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+terraform import databricks_app.this <app_name>
+```
+
+## Related Resources
+
+The following resources are used in the same context:
+
+* [databricks_cluster](cluster.md) to create [Databricks Clusters](https://docs.databricks.com/clusters/index.html).
+* [databricks_library](library.md) to install a [library](https://docs.databricks.com/libraries/index.html) on [databricks_cluster](cluster.md).
```
pietern, line 94: These don't seem relevant.

docs/resources/app.md
```diff
@@ -0,0 +1,94 @@
...
+    permission = "CAN_MANAGE"
+  }
+  serving_endpoint {
+    name = "databricks-meta-llama-3-1-70b-instruct"
+    permission = "CAN_MANAGE"
+  }
+  job {
+    id = "1234"
+    permission = "CAN_MANAGE"
+  }
```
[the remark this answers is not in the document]
nkvuong: updated

apps/resource_app.go
nkvuong: because the type is nested within `resource`, we cannot add constraints with sdkv2 - would need to rewrite this to framework to be able to add this validation rule

apps/resource_app.go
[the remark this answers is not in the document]
nkvuong: the diff is also cleaner if the user is removing/adding resources, it will only show the resource being added/removed, rather than the entire list changing.
@alexott is the issue that `slice_set` has some edge cases for reflection? I did some manual testing and it works okay (removing, adding resources)

docs/resources/app.md
nkvuong: updated doc

apps/resource_app.go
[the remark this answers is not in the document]
alexott: yes, there was a behavior of SDK when doing a diff on structs that contained complex objects, like, arrays or structs, etc. This was caused the way the hash was calculated, etc. Maybe it's not a problem in this case, just wanted to make sure

alexott (approved):

andrewnester (commented):
I've just realised that there's an important piece is missing: we need to supporting setting permissions for app resource as we do for other resources (see `ResourcePermissions`). Apps do support this API https://docs.databricks.com/api/workspace/apps/setpermissions

alexott:
@andrewnester maybe we can do it in a separate PR?

andrewnester:
@alexott we could, but having this permissions management for app is critical for DABs anyway, so I'd rather have it in this PR too. @pietern wdyt?

nkvuong:
@andrewnester makes sense - I've added `apps` to the permission resource and corresponding tests

docs/resources/app.md
```diff
@@ -0,0 +1,105 @@
+---
+subcategory: "Apps"
+---
+# databricks_app Resource
+
+-> This feature is in [Public Preview](https://docs.databricks.com/release-notes/release-types.html).
+
+Apps run directly on a customer’s Databricks instance, integrate with their data, use and extend Databricks services, and enable users to interact through single sign-on.
```
mgyucht, line 8: Can we link to public docs about Apps?

mgyucht, line 8: We should probably mention that this only creates the application and doesn't do anything with the app deployment.

mgyucht (commented):
Behavior-wise, this seems OK to me. Can we use the plugin framework to implement this instead of the SDKv2? We're trying to move away from SDkv2 to plugin framework for all new resources.

nihil0:
@alexott : Are there plans to merge this PR and release it in a new version of the Terraform provider. We are looking for ways to create and deploy Databricks Apps using infrastructure-as-code.

alexott:
@mgyucht wdyt?

internal/providers/pluginfw/products/app/resource_app.go
```diff
@@ -0,0 +1,226 @@
...
+		return
+	}
+
+	// Update the app
+	var appGoSdk apps.App
+	resp.Diagnostics.Append(converters.TfSdkToGoSdkStruct(ctx, app, &appGoSdk)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	_, err := w.Apps.Update(ctx, apps.UpdateAppRequest{App: &appGoSdk, Name: app.Name.ValueString()})
```
rauchy, line 186: Why is it necessary to explicitly mention `Name` if it’s already included in `App`?

internal/providers/pluginfw/products/app/resource_app.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	var app apps_tf.App
+	resp.Diagnostics.Append(req.State.Get(ctx, &app)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Delete the app
+	_, err := w.Apps.DeleteByName(ctx, app.Name.ValueString())
```
rauchy, line 219: Going forward, should we consider using the more general `w.Apps.Delete`? I understand it achieves the same result, but with auto-generated resources might want to avoid specific identifier names. Instead, they’d probably lean on tf-to-go and the `DeleteRequest` structures to align things correctly. This way, we ensure a bit more consistency across different resources. What do you think?

rauchy (approved):
Approved 🎉 @mgyucht this PR really ties together all the effort from your earlier ones. Great to see it all come together 👏

mgyucht (approved):

docs/resources/app.md
```diff
@@ -0,0 +1,114 @@
...
+
+[Databricks Apps](https://docs.databricks.com/en/dev-tools/databricks-apps/index.html) run directly on a customer’s Databricks instance, integrate with their data, use and extend Databricks services, and enable users to interact through single sign-on. This resource creates the application but does not handle app deployment, which should be handled separately as part of your CI/CD pipeline.
+
+## Example Usage
+
+```hcl
+resource "databricks_app" "this" {
+  name             = "my-custom-app"
+  description      = "My app"
+  resources = [{
```
alexott, line 16: Having resources as array isn't really nice IMHO - individual `resource` blocks are easier to use

PR #4101, opened 2024-10-14
[Internal][Draft] Plugin Framework Rollout

edwardfeng-db:
## Changes

- Made it possible to use environment variable `USE_SDK_V2` to control rollout
- Bumped resource and data source names from staging names to production names (can debate on whether we want to do that now or in a more gradual fashion)
- Proposed mechanism: https://docs.google.com/document/d/1zfTp8YesMe4GxkbIt9Sbwd5nQL7Y4mXed8dPDGmAoxU/edit?tab=t.0

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


internal/providers/pluginfw/pluginfw.go
```diff
@@ -43,17 +44,27 @@ type DatabricksProviderPluginFramework struct {
...
+	if os.Getenv("USE_SDK_V2") == "true" {
+		return []func() datasource.DataSource{
+			registered_model.DataSourceRegisteredModel,
+		}
+	} else {
+		return []func() datasource.DataSource{
+			cluster.DataSourceCluster,
+			volume.DataSourceVolumes,
+			registered_model.DataSourceRegisteredModel,
+		}
```
edwardfeng-db, line 67: Debatable decision - do we want to have more fine grained control such that we can use different environment variable values to control specific resources / data sources?

internal/providers/pluginfw/resources/volume/data_volumes.go
```diff
@@ -35,7 +35,7 @@ type VolumesList struct {
 }
 
 func (d *VolumesDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
-	resp.TypeName = pluginfwcommon.GetDatabricksStagingName(dataSourceName)
+	resp.TypeName = pluginfwcommon.GetDatabricksProductionName(dataSourceName)
```
edwardfeng-db, line 38: Debatable decision 2 - do we want to roll them all out at once or in a more gradual fashion?

internal/providers/pluginfw/pluginfw.go
tanmay-db: As discussed, mentioning for visibility -- it would be better to have more finer grained control per resource so people can pass the resources they want to use sdkv2 for individually as otherwise they would have to use sdkv2 for all resources if there is a bug in one resource.

internal/providers/pluginfw/resources/volume/data_volumes.go
tanmay-db: As discussed, mentioning for visibility -- we could roll them out individually in different releases (this would help if one resource contains a bug and people can pin the other version but won't help if the resource with bug is the first one that would be released as other following releases will have that resource anyway.

As the effort to do a release is low right after one successful release, we can create separate releases just making other resource use ProductionName to minimise blast radius but usually people don't pin the tf version so they would be using the latest version anyway.

I think both approaches have same value but given the effort is low for staggered releases, I say we can make them default in different release.

edwardfeng-db: Updated

internal/providers/pluginfw/pluginfw_rollout_utils.go
```diff
@@ -0,0 +1,123 @@
...
+var migratedResourceMap = map[string]func() resource.Resource{
+	"databricks_qualitymonitor": qualitymonitor.ResourceQualityMonitor,
+	"databricks_library":        library.ResourceLibrary,
+}
+
+// Map of data sources that have been migrated from SDK V2 to plugin framework
+var migratedDataSourceMap = map[string]func() datasource.DataSource{
+	"databricks_cluster":          cluster.DataSourceCluster,
+	"databricks_volumes":          volume.DataSourceVolumes,
+	"databricks_registered_model": registered_model.DataSourceRegisteredModel,
```
mgyucht, line 32: This was not migrated IIUC. It should only exist in the plugin framework.

internal/providers/pluginfw/resources/volume/data_volumes.go
```diff
@@ -35,7 +35,7 @@ type VolumesList struct {
 }
 
 func (d *VolumesDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
-	resp.TypeName = pluginfwcommon.GetDatabricksStagingName(dataSourceName)
```
mgyucht, line 38: Can we remove `GetDatabricksStagingName` now?

internal/providers/pluginfw/pluginfw_rollout_utils.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/resources/qualitymonitor"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/resources/registered_model"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/resources/volume"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+// Map of resources that have been migrated from SDK V2 to plugin framework
+var migratedResourceMap = map[string]func() resource.Resource{
+	"databricks_qualitymonitor": qualitymonitor.ResourceQualityMonitor,
```
mgyucht, line 24: Instead of hard-coding the names here, we can get them by calling the `Metadata()` method for each resource/data source.

mgyucht (approved):
Overall, approach seems fine. It's a bit strange to have the SDKv2 implementation depend on the pluginfw one, but I think that's an OK direction (the other way around would be problematic).

Can we make sure to update the documentation for each of these migrated resources with instructions for customers to fall-back to the original implementation?

Also, let's split migrated and non-migrated resources, since non-migrated resources should not be possible to roll-back to the original implementation.

internal/providers/pluginfw/resources/volume/data_volumes.go
edwardfeng-db: I think it is probably still useful for resources to be migrated in the future. So I'd suggest keeping it for now

mgyucht:
Stale, closing.

PR #4103, opened 2024-10-14
[Feature] Add `databricks_query` resource instead of `databricks_sql_query`

alexott:
## Changes

This PR is built on top of #4051 which should be merged first.

The new resource uses the new [Queries API](https://docs.databricks.com/api/workspace/queries/create) instead of the legacy one that will be deprecated. Since the new resource has a slightly different set of parameters, it was decided to create a new resource and deprecate the old one.

This resource uses old TF SDK to be compatible with TF exporter (until #4050 is implemented).

TODOs:

- Need to discuss how to handle permissions - `sql_query` permissions look like working, but not sure if we should continue to use that API
- Support in the exporter will be in a separate PR

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/docs/resources/query.md
+++ b/docs/resources/query.md
@@ -1,0 +1,162 @@
+---
+subcategory: "Databricks SQL"
+---
+# databricks_query Resource
+
+This resource allows you to manage [Databricks SQL Queries](https://docs.databricks.com/en/sql/user/queries/index.html).  It supersedes [databricks_sql_query](sql_query.md) resource - see migration guide below for more details.
+
+## Example Usage
+
+```hcl
+resource "databricks_directory" "shared_dir" {
+  path = "/Shared/Queries"
+}
+
+# This will be replaced with new databricks_query resource
+resource "databricks_query" "this" {
+  warehouse_id = databricks_sql_endpoint.example.id
+  display_name = "My Query Name"
+  query_text   = "SELECT 42 as value"
+  parent_path  = databricks_directory.shared_dir.path
+}
+```
+
+## Argument Reference
+
+The following arguments are available:
+
+* `query_text` - (Required, String) Text of SQL query.
+* `display_name` - (Required, String) Name of the query.
+* `warehouse_id` - (Required, String) ID of a SQL warehouse which will be used to execute this query.
+* `parent_path` - (Optional, String) The path to a workspace folder containing the query. The default is the user's home folder.  If changed, the query will be recreated.
+* `owner_user_name` - (Optional, String) Query owner's username.
+* `apply_auto_limit` - (Optional, Boolean) Whether to apply a 1000 row limit to the query result.
+* `catalog` - (Optional, String) Name of the catalog where this query will be executed.
+* `schema` - (Optional, String) Name of the schema where this query will be executed.
+* `description` - (Optional, String) General description that conveys additional information about this query such as usage notes.
+* `run_as_mode` - (Optional, String) Sets the "Run as" role for the object.
+* `tags` - (Optional, List of strings) Tags that will be added to the query.
+* `parameter` - (Optional, Block) Query parameter definition.  Consists of following attributes (one of `*_value` is required):
+  * `name` - (Required, String) Literal parameter marker that appears between double curly braces in the query text.
+  * `title` - (Optional, String) Text displayed in the user-facing parameter widget in the UI.
+  * `text_value` - (Block) Text parameter value. Consists of following attributes:
+    * `value` - (Required, String) - actual text value.
+  * `numeric_value` -  (Block) Numeric parameter value. Consists of following attributes:
+    * `value` - (Required, Double) - actual numeric value.
+  * `date_value` - (Block) Date query parameter value. Consists of following attributes (Can only specify one of `dynamic_date_value` or `date_value`):
+    * `date_value` - (String) Manually specified date-time value
+    * `dynamic_date_value` - (String) Dynamic date-time value based on current date-time.  Possible values are `NOW`, `YESTERDAY`.
+    * `precision` - (Optional, String) Date-time precision to format the value into when the query is run.  Possible values are `DAY_PRECISION`, `MINUTE_PRECISION`, `SECOND_PRECISION`.  Defaults to `DAY_PRECISION` (`YYYY-MM-DD`).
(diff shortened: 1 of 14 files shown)


docs/resources/query.md
```diff
@@ -0,0 +1,162 @@
...
+* `warehouse_id` - (Required, String) ID of a SQL warehouse which will be used to execute this query.
+* `parent_path` - (Optional, String) The path to a workspace folder containing the query. The default is the user's home folder.  If changed, the query will be recreated.
+* `owner_user_name` - (Optional, String) Query owner's username.
+* `apply_auto_limit` - (Optional, Boolean) Whether to apply a 1000 row limit to the query result.
+* `catalog` - (Optional, String) Name of the catalog where this query will be executed.
+* `schema` - (Optional, String) Name of the schema where this query will be executed.
+* `description` - (Optional, String) General description that conveys additional information about this query such as usage notes.
+* `run_as_mode` - (Optional, String) Sets the "Run as" role for the object.
+* `tags` - (Optional, List of strings) Tags that will be added to the query.
+* `parameter` - (Optional, Block) Query parameter definition.  Consists of following attributes (one of `*_value` is required):
```
mgyucht, line 39: Just curious: in other docs, these blocks are often documented as their own headings. Is there any need for consistency around these aspects?

docs/resources/query.md
```diff
@@ -0,0 +1,162 @@
...
+    text_value {
+      value = "default"
+    }
+  }
+}
+```
+
+* Remove the old resource from the state with the `terraform state rm databricks_sql_query.query` command.
+* Import new resource with the `terraform import databricks_query.query <query-id>` command.
+* Run the `terraform plan` command to check possible changes, like, value type change, etc.
```
mgyucht, line 132: Ditto on these w.r.t. using import/removed blocks

docs/resources/query.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Access Control
+
+[databricks_permissions](permissions.md#sql-query-usage) can control which groups or individual users can *Manage*, *Edit*, *Run* or *View* individual queries.
+
+```hcl
+resource "databricks_permissions" "query_usage" {
+  sql_query_id = databricks_query.query.id
```
mgyucht, line 140: Great that we can use this as-is again.

internal/acceptance/query_test.go
```diff
@@ -0,0 +1,48 @@
...
+func TestAccQuery(t *testing.T) {
+	WorkspaceLevel(t, Step{
+		Template: `
+		resource "databricks_query" "this" {
+			warehouse_id = "{env.TEST_DEFAULT_WAREHOUSE_ID}"
+			display_name = "tf-{var.RANDOM}"
+			query_text = "SELECT 1 AS p1, 2 as p2"
+		}
+
+        resource "databricks_permissions" "query_usage" {
```
mgyucht, line 16: Ditto, let's move this to permissions_test.go.

sql/resource_query.go
```diff
@@ -0,0 +1,169 @@
...
+			})
+			if err != nil {
+				return err
+			}
+			d.SetId(apiQuery.Id)
+			owner := d.Get("owner_user_name").(string)
+			if owner != "" {
+				_, err = w.Queries.Update(ctx, sql.UpdateQueryRequest{
+					Query: &sql.UpdateQueryRequestQuery{
+						OwnerUserName: owner,
```
mgyucht, line 115: is this another set of APIs where the owner cannot be set on create?

mgyucht (commented):
Overall seems good to me. I'll wait to approve until databricks_alert is merged.

sql/resource_query.go
alexott: yep - same as for alerts

docs/resources/query.md
alexott: that's a good question - sometimes, moving to a separate block will add an unnecessary context switch. But agree, we need to unify this

sql/resource_query.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			var q queryUpdateStruct
+			common.DataToStructPointer(d, s, &q)
+			updateMask := "display_name,query_text,warehouse_id,parameters"
+			for _, f := range []string{"run_as_mode", "owner_user_name", "description", "tags",
+				"apply_auto_limit", "catalog", "schema"} {
+				if d.HasChange(f) {
+					updateMask += "," + f
+				}
+			}
```
mgyucht, line 152: For Terraform, because DataToStructPointer has values for all fields, can we just use `*` for updateMask?

mgyucht (approved):
Overall LGTM, just a couple of smaller questions.

sql/resource_alert.go
```diff
@@ -19,7 +19,6 @@ func ResourceAlert() common.Resource {
 		// TODO: can we automatically generate it from SDK? Or should we avoid validation at all?
 		common.CustomizeSchemaPath(m, "condition", "op").SetRequired().SetValidateFunc(validation.StringInSlice([]string{
 			"GREATER_THAN", "GREATER_THAN_OR_EQUAL", "LESS_THAN", "LESS_THAN_OR_EQUAL", "EQUAL", "NOT_EQUAL", "IS_NULL"}, true))
-		common.CustomizeSchemaPath(m, "condition", "op").SetRequired()
```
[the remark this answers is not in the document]
alexott: Because it's covered by previous line

sql/resource_alert.go
```diff
@@ -39,6 +38,7 @@ func ResourceAlert() common.Resource {
 				strings.TrimPrefix(f, "condition.0.threshold.0.value.0.")).SetExactlyOneOf(alof)
 		}
 		common.CustomizeSchemaPath(m, "owner_user_name").SetSuppressDiff()
+		common.CustomizeSchemaPath(m, "notify_on_ok").SetDefault(true)
```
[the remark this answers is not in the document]
alexott: We get default from the API, and if we use `suppress_diff` here, then we can't set it to the `false`

sql/resource_query.go
alexott: API spec says:
> To specify multiple fields in the field mask, use a comma as the separator (no space).

I'm not sure that `*` is supported.

Especially for `owner_user_name` if we don't have change, it will decrease a pressure on backend, as changing the owner is heavy operation (at least as I understood from the explanation of the corresponding team).

rauchy:
Merging for now to unblock releases.

PR #4116, opened 2024-10-16
[Fix] Mark unity_catalog_provisioning_state as ReadOnly

rauchy:
## Changes

In a recent update to the GoSDK, a new field called `unity_catalog_provisioning_state` was added. This field changes values based on server decisions, so it's supposed to be marked as computed. This pull request sorts out that issue.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_online_table.go
+++ b/catalog/resource_online_table.go
@@ -59,6 +59,7 @@ func ResourceOnlineTable() common.Resource {
 			common.CustomizeSchemaPath(m, "spec", "source_table_full_name").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
 			common.CustomizeSchemaPath(m, "name").SetRequired().SetForceNew()
 			common.CustomizeSchemaPath(m, "status").SetReadOnly()
+			common.CustomizeSchemaPath(m, "unity_catalog_provisioning_state").SetReadOnly()
 			common.CustomizeSchemaPath(m, "table_serving_url").SetReadOnly()
 			common.CustomizeSchemaPath(m, "spec", "pipeline_id").SetReadOnly()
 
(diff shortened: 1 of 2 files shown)


hectorcast-db (approved):

catalog/resource_online_table.go
```diff
@@ -59,6 +59,7 @@ func ResourceOnlineTable() common.Resource {
 			common.CustomizeSchemaPath(m, "spec", "source_table_full_name").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
 			common.CustomizeSchemaPath(m, "name").SetRequired().SetForceNew()
 			common.CustomizeSchemaPath(m, "status").SetReadOnly()
+			common.CustomizeSchemaPath(m, "unity_catalog_provisioning_state").SetComputed()
```
alexott, line 62: it's better to use `.SetReadOnly()` because it will set optional and required into correct states

alexott: we also need to add it to the docs?

rauchy: Thanks Alex. Could you explain when it makes sense to use computed properties versus readonly properties?

alexott: The `SetReadOnly` function does a few things:

* set `Computed` - same as `SetComputed`
* It sets `Required` to `false` and `Optional` to `true` because `Computed` can't be used with `Required`.
* Sets `MinItems` to 0...

In general, if the field is already `Optional`, both functions do the same, but `SetReadOnly` is just safer here because you don't need to check if the field already marked as optional or not.

alexott (approved):

PR #4118, opened 2024-10-17
[Exporter] **Breaking change**: Move `databricks_workspace_file` to a separate service

alexott:
## Changes

Move `databricks_workspace_file` to a separate service `wsfiles`, so we can list and export them separately from notebooks.  If you used `notebooks` in `-listing` or `-services` options, then you need to append `wsfiles` to these options.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -374,14 +374,30 @@ func (ic *importContext) Run() error {
 	ic.startImportChannels()
 
 	// Start listing of objects
+	listWorkspaceObjectsAlreadyRunning := false
 	for rnLoop, irLoop := range ic.Importables {
 		resourceName := rnLoop
 		ir := irLoop
+		// TODO: extend this to other services?  Like, Git Folders
+		if !ic.accountLevel && (ir.Service == "notebooks" || ir.Service == "wsfiles" || (ir.Service == "directories" && !ic.incremental)) {
+			if _, exists := ic.listing[ir.Service]; exists && !listWorkspaceObjectsAlreadyRunning {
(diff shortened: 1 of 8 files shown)


exporter/context.go
```diff
@@ -374,14 +374,30 @@ func (ic *importContext) Run() error {
...
 		resourceName := rnLoop
 		ir := irLoop
+		// TODO: extend this to other services?  Like, Git Folders
+		if !ic.accountLevel && (ir.Service == "notebooks" || ir.Service == "wsfiles" || (ir.Service == "directories" && !ic.incremental)) {
+			if _, exists := ic.listing[ir.Service]; exists && !listWorkspaceObjectsAlreadyRunning {
+				ic.waitGroup.Add(1)
+				log.Printf("[DEBUG] Starting listing of workspace objects")
+				go func() {
+					if err := listWorkspaceObjects(ic); err != nil {
+						log.Printf("[ERROR] listing of workspace objects failed %s", err)
```
tanmay-db, line 388: should we return after this?

exporter/context.go
```diff
@@ -374,14 +374,30 @@ func (ic *importContext) Run() error {
...
+					}
+					log.Print("[DEBUG] Finished listing of workspace objects")
+					ic.waitGroup.Done()
+				}()
+				listWorkspaceObjectsAlreadyRunning = true
```
tanmay-db, line 393: In case of err on line 388, this would still be set to true right? which shouldn't be the case.

tanmay-db (approved):

PR #4123, opened 2024-10-18
[Exporter] Use `List` + iteration instead of call to `ListAll`

### LINKED ISSUE #4119: [Exporter] Use `List` from Go SDK instead of `ListAll` for high volume resources
Some resources, like queries, alerts, dashboards, jobs, DLT pipelines, etc., may have a huge number of resources. Right now, most listing implementations use `ListAll` from Go SDK, but this leads to a situation when we need to wait until the listing finishes instead of starting processing resources page by page. Switching to the `List` from Go SDK will allow us to start processing individual resources as soon as we get a page from the backend.

alexott:
## Changes

This change significantly improve performance of export for resources with big number of objects because we're starting to export objects as soon as we get first page with list of objects.  I.e., for Lakeview dashboards, the export time for ~10k dashboards went from 47 minutes down to 22 minutes.

Resolves #4119

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/exporter/context.go
+++ b/exporter/context.go
@@ -200,6 +200,7 @@ var goroutinesNumber = map[string]int{
 	"databricks_dbfs_file":         3,
 	"databricks_user":              1,
 	"databricks_service_principal": 1,
+	"databricks_dashboard":         4,
 	"databricks_sql_dashboard":     3,
 	"databricks_sql_widget":        4,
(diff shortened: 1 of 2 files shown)


mgyucht (approved):
Awesome!

PR #4153, opened 2024-10-25
[Fix] Handle edge case for `effective_properties` in `databricks_sql_table`

### LINKED ISSUE #4098: [ISSUE] Issue with `databricks_sql_table` resource perpetual diff on options
### Configuration

```hcl
resource "databricks_sql_table" "my_table" {
  name               = "foo"
  catalog_name       = "bar"
  schema_name        ="baz"
  storage_location   = "abfss://some-location"
  data_source_format = "JSON"
  options = {
    recursiveFileLookup = true
    multiLine           = true
  }
}
```

### Expected Behavior
The options should be applied and not trigger a change on the next plan

### Actual Behavior
We get a perpetual diff like this:
```
  ~ resource "databricks_sql_table" "my_table" {
      ~ effective_properties = {
          + "option.multiLine"                                = "true"
          + "option.recursiveFileLookup"                      = "true"
            # (2 unchanged elements hidden)
        }
```

### Steps to Reproduce
1. `terraform apply`
2. `terraform apply` again
3. Diff appears, even though we didn't change any code

Terraform version: `1.8.5`
Provider version: `1.52`

### Is it a regression?
Yes, I think this is en edge case that wasn't covered by the great PR that solved a previous similar issue:
https://github.com/databricks/terraform-provider-databricks/pull/3925
I would assume that since the `options` in this case end up like this in the state:
```json
     "effective_properties": {
              "spark.sql.dataSourceOptions.multiLine": "true",
              "spark.sql.dataSourceOptions.recursiveFileLookup": "true"
            },
```
And not
```json
     "effective_properties": {
              "options.multiLine": "true",
              "options.recursiveFileLookup": "true"
            },
```
we will get that diff. I think we need to handle the edge case where the options are of type `spark.sql.dataSourceOptions`.

alexott:
## Changes

It was reported in #4098 that some of the specified options, like, `multiLine`, `recursiveFileLookup` and potentially more, aren't returned as `option.multiLine`, etc., but instead are expanded into full names, like, `spark.sql.dataSourceOptions.multiLine`.

This PR changes lookup logic a bit, and if we can't find `option.something`, then we're looking for all options ending with `.something` (only if there are no `.` in the name).

Resolves #4098

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -21,6 +21,7 @@ import (
 )
 
 var MaxSqlExecWaitTimeout = 50
+var optionPrefix = "option."
 
 type SqlColumnInfo struct {
 	Name     string         `json:"name"`
@@ -67,7 +68,6 @@ type SqlTableInfo struct {
 }
 
 func (ti SqlTableInfo) CustomizeSchema(s *common.CustomizableSchema) *common.CustomizableSchema {
-
 	caseInsensitiveFields := []string{"name", "catalog_name", "schema_name"}
 	for _, field := range caseInsensitiveFields {
 		s.SchemaPath(field).SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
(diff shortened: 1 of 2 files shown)


catalog/resource_sql_table.go
```diff
@@ -655,6 +669,10 @@ func ResourceSqlTable() common.Resource {
 			if err != nil {
 				return err
 			}
+			// opts := d.Get("options").(map[string]any)
+			// for k, v := range opts {
+			// 	ti.EffectiveProperties[optionPrefix+k] = v.(string)
+			// }
```
mgyucht, line 675: Can you remove this unused code?

catalog/resource_sql_table.go
```diff
@@ -598,18 +598,32 @@ func ResourceSqlTable() common.Resource {
...
-					diff["option."+k] = userSpecifiedValue
+			for userOptName, userSpecifiedValue := range userSpecifiedOptions {
+				effectOptName := optionPrefix + userOptName
+				var found bool
+				var effectiveValue any
+				// If the option is not found, check if the user specified the option without the prefix
+				// i.e. if user specified `multiLine` for JSON, then backend returns `spark.sql.dataSourceOptions.multiLine`
+				if effectiveValue, found = effectiveProperties[effectOptName]; !found && !strings.Contains(userOptName, ".") {
+					for k, v := range effectiveProperties {
+						if strings.HasSuffix(k, "."+userOptName) {
```
mgyucht, line 618: Just checking the suffix feels a bit brittle to me. At least you're checking for `"." + userOptName`, which should be safer. Is `spark.sql.dataSourceOptions.` a well-known prefix for effective options? Would it work if we did something like:
```go
var prefixes = []string{"option.", "spark.sql.dataSourceOptions."}
...
for _, prefix := range prefixes {
        if v, ok := effectiveProperties[prefix + userOptName]; ok {
                found = true
                effectiveValue = v
                break
        }
}
```
This way, we are explicit about what effective property prefixes should be considered as options.

mgyucht (approved):
Suggestion to be explicit about the allowed prefixes, rather than allowing any prefix. Otherwise this LGTM.

catalog/resource_sql_table.go
alexott: sure

catalog/resource_sql_table.go
alexott: it's not clear, really - we need to ask UC team about it.

alexott: Got confirmation that it's only for data source options so the current version should handle it.

PR #4161, opened 2024-10-28
[Internal] Migrate Share Data Source to Plugin Framework

rauchy:
## Changes

This PR migrates the share/shares data sources to the Plugin framework. The code was largely copied "as is" from the previous implementation of the share data source, with the necessary adaptations made for integration with the Plugin framework.

## Tests

Note: current tests create shares using the SDKv2 resource, but fetch them using the new plugin framework data source. Once the resource migration will be merged, I will amend this.

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/resources/share/data_share.go
+++ b/internal/providers/pluginfw/resources/share/data_share.go
@@ -1,0 +1,74 @@
+package share
+
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/service/sharing"
+	"github.com/databricks/terraform-provider-databricks/common"
+	pluginfwcommon "github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/common"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/converters"
+	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/tfschema"
+	"github.com/databricks/terraform-provider-databricks/internal/service/sharing_tf"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func DataSourceShare() datasource.DataSource {
+	return &ShareDataSource{}
+}
+
+var _ datasource.DataSourceWithConfigure = &ShareDataSource{}
+
+type ShareDataSource struct {
+	Client *common.DatabricksClient
+}
+
+func (d *ShareDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = pluginfwcommon.GetDatabricksStagingName("share")
+}
+
+func (d *ShareDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	attrs, blocks := tfschema.DataSourceStructToSchemaMap(sharing_tf.ShareInfo{}, nil)
+	resp.Schema = schema.Schema{
+		Attributes: attrs,
+		Blocks:     blocks,
+	}
+}
+
+func (d *ShareDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if d.Client == nil {
+		d.Client = pluginfwcommon.ConfigureDataSource(req, resp)
+	}
+}
+
+func (d *ShareDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	w, diags := d.Client.GetWorkspaceClient()
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var config sharing_tf.ShareInfo
+	diags = req.Config.Get(ctx, &config)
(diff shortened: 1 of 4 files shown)


internal/providers/pluginfw/resources/share/data_share.go
```diff
@@ -0,0 +1,74 @@
...
+func (d *ShareDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = pluginfwcommon.GetDatabricksStagingName("share")
+}
+
+func (d *ShareDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	attrs, blocks := tfschema.DataSourceStructToSchemaMap(sharing_tf.ShareInfo{}, nil)
+	resp.Schema = schema.Schema{
+		Attributes: attrs,
+		Blocks:     blocks,
+	}
```
alexott, line 35: It's not a direct problem, but such code is repeated everywhere. Should we add a helper function that we can use instead?

alexott (commented):
Why do we need to do staging for data sources? Imho it's less critical than resources.

but the main issue IMHO is that we modify the acceptance tests to work only with the staged resources, and leaving a hole in acceptance testing of "normal" data sources

alexott:
@rauchy I think that it makes sense just to duplicate tests, not to replace resources/data sources, so we keep tests for SDK-based resources as well, and remove the duplicate tests when we replace SDK-based with plugin framework-based

rauchy:
@alexott that was exactly what I was about to propose :)

internal/providers/pluginfw/resources/sharing/data_share.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+var _ datasource.DataSourceWithConfigure = &ShareDataSource{}
+
+type ShareDataSource struct {
+	Client *common.DatabricksClient
+}
+
+func (d *ShareDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = pluginfwcommon.GetDatabricksStagingName("share")
```
tanmay-db, line 27: nit: We should abstract this out above like: https://github.com/databricks/terraform-provider-databricks/blob/main/internal/providers/pluginfw/resources/volume/data_volumes.go#L19

internal/providers/pluginfw/resources/sharing/data_share.go
```diff
@@ -0,0 +1,74 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	share, err := w.Shares.Get(ctx, sharing.GetShareRequest{
+		Name:              config.Name.ValueString(),
+		IncludeSharedData: true,
+	})
+	if err != nil {
```
tanmay-db, line 62: We should remove from state if missing:
````

		if apierr.IsMissing(err) {
			resp.State.RemoveResource(ctx)
		}
````

File: internal/providers/pluginfw/resources/sharing/data_shares.go [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+}
+
+var _ datasource.DataSourceWithConfigure = &SharesDataSource{}
+
+type SharesDataSource struct {
+	Client *common.DatabricksClient
+}
+
+func (d *SharesDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = pluginfwcommon.GetDatabricksStagingName("shares")
```
tanmay-db, line 31: Same here

tanmay-db (approved):

internal/providers/pluginfw/resources/sharing/data_shares.go
rauchy: Initially, I planned to use a `dataSourceName` constant for both the _share_ and _shares_ data sources. However, since they share the same namespace, that wasn’t an option. So, I just included the constant within the `Metadata` function, particularly since it’s only used once in each data source. Alternatively, I could assign a `dataSourceName` constant to one data source and rename it for the other, but that didn't feel quite right.

Let me know if you have a preferred approach.

internal/providers/pluginfw/resources/sharing/data_share.go
rauchy: Good to know that's needed :) [b7c32b7](https://github.com/databricks/terraform-provider-databricks/pull/4161/commits/b7c32b76687fe4a476b37cb9eb9b323d21072dda)

internal/providers/pluginfw/resources/sharing/data_share.go
rauchy: See my comment below.

internal/providers/pluginfw/resources/sharing/data_shares.go
tanmay-db: I see, yeah we could use dataSourceNameShare, dataSourceNameShares, mentioning this because we would also need to set the useragent (eg: https://github.com/databricks/terraform-provider-databricks/blob/main/internal/providers/pluginfw/resources/volume/data_volumes.go#L56) and use the same name there as well so might be better to abstract this in a single variable.

PR #4165, opened 2024-10-28
[Internal] Skip integration tests on docs-only changes

mgyucht:
## Changes
Currently, integration tests run on every PR. This takes a considerable amount of time, up to around an hour and 15 minutes. When a change only affects the `docs/` directory, there is no need to run integration tests, as that directory only includes the public documentation and user guides. Thus, integration tests can be skipped.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -9,8 +9,27 @@ on:
   
 
 jobs:
-  trigger-tests:
+  check-integration-test-changes:
     if: github.event_name == 'pull_request'
+    name: Check if changes require integration tests to be triggered
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Check if all changes are in docs/ directory
+      id: check-integration-tests-required
+      run: |
+        changed_files=$(git diff --name-only ${{ github.event.before }} ${{ github.sha }})
+        if echo "$changed_files" | grep -qv '^docs/'; then
+          echo "integration_tests_required=true" >> $GITHUB_ENV
+        else
+          echo "integration_tests_required=false" >> $GITHUB_ENV
+        fi
+
+  trigger-tests:
+    needs: check-integration-test-changes
+    if: '{{ env.integration_tests_required == "true" }}'
     name: Trigger Tests
     runs-on: ubuntu-latest
     environment: "test-trigger-is"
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -9,8 +9,27 @@ on:
...
-  trigger-tests:
+  check-integration-test-changes:
     if: github.event_name == 'pull_request'
+    name: Check if changes require integration tests to be triggered
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Check if all changes are in docs/ directory
```
hectorcast-db, line 20: There is an action to that does the grep for you.
```
      - uses: dorny/paths-filter@v3
        id: changes
        with:
          filters: |
            integration_tests_required:
              - '**/*'
              - '!docs/**'
 ```

```
needs: [changes]
if: '{{ needs.changes.outputs.integration_tests_required == "true" }}'
```

File: .github/workflows/integration-tests.yml [hunk 2]
```diff
@@ -36,24 +52,29 @@ jobs:
...
   # In case of multiple commits in a single PR, the hash of the squashed commit 
   # will not match the one for the latest (approved) commit in the PR.
   # We auto approve the check for the merge queue for two reasons:
   # * Queue times out due to duration of tests.
   # * Avoid running integration tests twice, since it was already run at the tip of the branch before squashing.
+  # Additionally, integration tests are not run on docs-only changes.
   auto-approve:
-    if: github.event_name == 'merge_group'
+    needs: check-integration-test-changes
+    if: '{{ github.event_name == "merge_group" || env.integration_tests_required == "false" }}'
```
hectorcast-db, line 65: Actually, this won't work in the merge queue. If the "need" job is skipped, this will be skipped too (even with the "if" condition). You can use always() to ensure it runs. Also, use the output
```if: always() && (github.event_name == 'merge_group' || needs.check-integration-test-changes.outputs.integration_tests_required == 'true')```

File: .github/workflows/integration-tests.yml [hunk 3]
```diff
@@ -9,8 +9,24 @@ on:
...
+      with:
+        # integration_tests_required should be true if integration tests should be run.
+        # - Any PRs that only modify documentation should not require integration tests to run.
+        filters: |
+          integration_tests_required:
+            - '!**/*.md'
+
+  trigger-tests:
+    needs: check-integration-test-changes
+    if: '{{ env.integration_tests_required == "true" }}'
```
hectorcast-db, line 29: It's an output, not an env. 
Define the output and set an id for the job
```
jobs:
  changes:
    runs-on: ubuntu-latest
    outputs:  <----
      integration_tests_required: ${{ steps.filter.outputs.integration_tests_required }}
    steps:
      - uses: actions/checkout@v4
      
      - uses: dorny/paths-filter@v3
        id: filter <----
        with:
          filters: |
            integration_tests_required:
              - '!**/*.md'
```

PR #4168, opened 2024-10-29
[Fix] Provide more prescriptive error when users fail to create a single node cluster

shreyas-goenka:
## Changes
A better error message is warranted because many DABs customers have reportedly run into this. Original issue: https://github.com/databricks/cli/issues/1546

## Tests
Unit test

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -38,6 +38,10 @@ cluster please include the following configuration in your cluster configuration
     "ResourceClass" = "SingleNode"
   }
 
+Please note that the Databricks terraform provider cannot detect if the above configuration
+is defined in a policy used by the cluster. Please define this in the cluster configuration
+itself to create a single node cluster.
+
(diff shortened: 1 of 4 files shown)


clusters/resource_cluster.go
```diff
@@ -26,7 +26,26 @@ var clusterSchema = resourceClusterSchema()
...
+    "ResourceClass" = "SingleNode"
+  }
+
+Please note that the Databricks Terraform provider cannot detect if the above configuration
+is defined in a policy used by the cluster. Please define this in the cluster configuration
+itself to create a single node cluster.
+
+For more details please see:
+  1. https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/cluster#fixed-size-or-autoscaling-cluster
+  2. https://docs.databricks.com/clusters/single-node.html`
```
shreyas-goenka, line 47: TODO:Followup on better docs.

PR #4173, opened 2024-10-30
[Feature] Skip calling Read after Create/Update operations for notebooks

alexott:
## Changes

It's based on #4190.

It was found that the `import` API returns `object_id` as a result of its execution, so we don't really need to call the `Read` operation that will call `get-status` to fill all attributes (when we create a notebook using the `SOURCE` format).

The other formats (`DBC` and `JUPYTER`) will still call the `Read` to fill in the missing parameters, such as `language`.  

It also changed the `language` to `optional,computed` so we can safely set the value and don't trigger the diff.

This should help when we import a large number of notebooks, i.e., when applying exported resources.

## Tests

- [x] `make test` run locally
- [ ] ~using Go SDK~


workspace/resource_notebook.go
```diff
@@ -60,6 +60,10 @@ type ObjectStatus struct {
 	Size                  int64                  `json:"size,omitempty"`
 }
 
+type ImportResponce struct {
```
mgyucht, line 63:
```suggestion
type ImportResponse struct {
```

workspace/resource_notebook.go
```diff
@@ -291,16 +306,24 @@ func ResourceNotebook() common.Resource {
...
 					}
-					err = notebooksAPI.Create(createNotebook)
+					resp, err = notebooksAPI.Create(createNotebook)
 				}
 				if err != nil {
 					return err
 				}
 			}
+			d.Set("object_id", resp.ObjectID)
+			d.Set("skip_get_status", true)
```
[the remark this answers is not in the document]
alexott: let me look

alexott:
@mgyucht I've opened #4190 with changes in `common` so it will be easier to review. This PR will be used only for notebook changes.

workspace/resource_notebook.go
```diff
@@ -98,12 +102,14 @@ type NotebooksAPI struct {
...
-func (a NotebooksAPI) Create(r ImportPath) error {
+func (a NotebooksAPI) Create(r ImportPath) (ImportResponse, error) {
 	if r.Format == "DBC" {
 		mtx.Lock()
 		defer mtx.Unlock()
 	}
-	return a.client.Post(a.context, "/workspace/import", r, nil)
+	var responce ImportResponse
+	err := a.client.Post(a.context, "/workspace/import", r, &responce)
+	return responce, err
```
rauchy, line 112:
```suggestion
	var response ImportResponse
	err := a.client.Post(a.context, "/workspace/import", r, &response)
	return response, err
```

rauchy (approved):
I think this PR is doing a bit more than what the title suggests, especially now that #4190 lets us skip calling Read, and this PR is using that. I'm missing some specific Notebook context, so it's tough for me to fully judge if the change is correct. My review is mainly focused on the tests and code, and I don't see any issues there. @mgyucht, do you think you should have another look at it as well?

alexott:
@rauchy I've added more context to the PR description

workspace/resource_notebook.go
```diff
@@ -291,16 +303,32 @@ func ResourceNotebook() common.Resource {
...
 					}
-					err = notebooksAPI.Create(createNotebook)
+					resp, err = notebooksAPI.Create(createNotebook)
 				}
 				if err != nil {
 					return err
 				}
 			}
+			if d.Get("object_type").(string) == "" {
+				d.Set("object_type", Notebook)
```
mgyucht, line 313: Why do we need to know the object type here? Isn't it always notebook for a databricks_notebook? Or is it because this can be a DBC file which can contain other stuff as well?

I am also missing when it would be set before (when this `if` block would be skipped). It's marked as `deprecated` and should always be notebook, I guess?

workspace/resource_notebook.go
```diff
@@ -272,6 +279,10 @@ func ResourceNotebook() common.Resource {
 				Path:      path,
 				Overwrite: true,
 			}
+			if createNotebook.Format == "" && createNotebook.Language != "" {
```
mgyucht, line 282: Can users specify format? It's not documented on the public docs: https://registry.terraform.io/providers/databricks/databricks/1.61.0/docs/resources/notebook. If not, we can remove the first branch of the `if` since it is always true.

workspace/resource_notebook.go
```diff
@@ -272,6 +279,10 @@ func ResourceNotebook() common.Resource {
...
+				createNotebook.Format = "SOURCE"
+				d.Set("format", createNotebook.Format)
+			}
```
mgyucht, line 285: To make this easier to follow, it seems like we should do something like the following:
```go
if createNotebook.Language == "" {
	// Users must not specify `content_base64` without a language, so
	// `source` must be set.
	ext := strings.ToLower(filepath.Ext(d.Get("source").(string)))
	createNotebook.Language = extMap[ext].Language
	createNotebook.Format = extMap[ext].Format
	// Overwrite cannot be used for Dbc format
	createNotebook.Overwrite = extMap[ext].Overwrite
	d.Set("language", createNotebook.Language)
} else {
	createNotebook.Format = "SOURCE"
}
d.Set("format", createNotebook.Format)
```
Is this logic correct?

workspace/resource_notebook_test.go
```diff
@@ -36,6 +36,42 @@ func TestResourceNotebookRead(t *testing.T) {
...
+					Language:   "PYTHON",
+				},
+			},
+		},
+		Resource: ResourceNotebook(),
+		Read:     true,
+		New:      true,
+		State: map[string]any{
+			"source": "acceptance/testdata/tf-test-jupyter.ipynb",
+			"format": "JUPYTER",
```
mgyucht, line 64: I don't think these will normally be part of the resource state during read for a newly imported resource. What is this test meant to cover?

workspace/resource_notebook_test.go
```diff
@@ -119,22 +155,8 @@ func TestResourceNotebookCreate_DirectoryExist(t *testing.T) {
 					Overwrite: true,
 					Format:    "SOURCE",
 				},
-			},
-			{
```
mgyucht, line 123: Very nice that we save multiple API calls here!

workspace/acceptance/testdata/acc-test-update1.ipynb
```diff
@@ -0,0 +1,42 @@
+{
```
mgyucht, line 1: Can you add a readme for the `.dbc` files? It's not clear what is in those.

workspace/resource_notebook.go
```diff
@@ -203,31 +209,29 @@ func (a NotebooksAPI) Delete(path string, recursive bool) error {
...
+	d.Set("workspace_path", "/Workspace"+d.Id())
+}
+
 // ResourceNotebook manages notebooks
 func ResourceNotebook() common.Resource {
 	s := FileContentSchema(map[string]*schema.Schema{
 		"language": {
 			Type:     schema.TypeString,
 			Optional: true,
+			Computed: true, // we need it because it will be filled by the provider or backend
```
mgyucht, line 223: 👍  thank you for correcting this

workspace/resource_notebook.go
```diff
@@ -203,31 +209,29 @@ func (a NotebooksAPI) Delete(path string, recursive bool) error {
...
-				if source == "" {
-					return false
-				}
-				ext := strings.ToLower(filepath.Ext(source))
-				return old == extMap[ext].Language
-			},
 		},
 		"format": {
 			Type:     schema.TypeString,
 			Optional: true,
```
mgyucht, line 233: You don't need to change it in this PR, but is this really optional? From the docs, users can't specify this, and it seems like we always set it in Create based on the language/source rather than accepting the user-provided value.

mgyucht (changes requested):
I might be misunderstanding parts of this, as `databricks_notebook` is a pretty complex resource, but I think there are some concerns we need to address before merging. LMK what you think.

workspace/resource_notebook.go
alexott: it's because DBC could create a directory

workspace/resource_notebook.go
alexott: The `language` needs to be set for notebooks. When we use the file content then we can guess it from the file extension, but if we provide content as base64, then we don't have this information and user must provide it.

workspace/resource_notebook.go
```diff
@@ -291,16 +303,32 @@ func ResourceNotebook() common.Resource {
...
+			if d.Get("object_type").(string) == "" {
+				d.Set("object_type", Notebook)
+			}
+			d.Set("object_id", resp.ObjectID)
 			d.SetId(path)
+			setComputedProperties(d, c)
 			return nil
 		},
 		Read: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			oldFormat := d.Get("format").(string)
```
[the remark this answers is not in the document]
alexott: Yes, this is for importing, specifically for DBC files - If I remember correctly, the problem is that when no `format` is set and the DBC file is the source, then the Update will delete the original content...

workspace/resource_notebook.go
alexott: yes, I can encode my `.ipynb` or `.dbc` file and submit as `content_base64` in terraform.

workspace/resource_notebook.go
alexott: `format` is default to `SOURCE` automatically. We need to change it only if we specify DBC file, but it's rare

workspace/resource_notebook_test.go
alexott: it will be a "state" read by terraform during the read, and then it's also populated from Read

workspace/acceptance/testdata/acc-test-update1.ipynb
alexott: Added README

mgyucht (approved):

PR #4177, opened 2024-10-31
[Exporter] Allow to match resource names by regular expression

alexott:
## Changes

In addition to the existing `-match` option, this PR allows the matching of names by regex during the listing operation.  There are new options:

- `-matchRegex` - checks if name matches a regex - this could be useful for exporting notebooks for only specific users, or something like that.
- `-excludeRegex` - checks if name matches a regex, and skips processing of that object. For example, it could be used to exclude `databricks_automl` directories.  This parameter has higher priority than the `-match` and `-matchRegex`.
- `filterDirectoriesDuringWorkspaceWalking` - if we should apply match logic to directory names when we're performing workspace tree walking.  *Note: be careful with it as it will be applied to all entries, so if you want to filter only specific users, then you will need to specify the condition for `/Users` as well, so regex will be `^(/Users|/Users/[a-c].*)$`*

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


mgyucht (approved):

docs/guides/experimental-exporter.md
```diff
@@ -61,6 +61,9 @@ All arguments are optional, and they tune what code is being generated.
...
 * `-services` - Comma-separated list of services to import. By default, all services are imported.
 * `-match` - Match resource names during listing operation. This filter applies to all resources that are getting listed, so if you want to import all dependencies of just one cluster, specify `-match=autoscaling -listing=compute`. By default, it is empty, which matches everything.
+* `-matchRegex` - Match resource names against a given regex during listing operation. Applicable to all resources selected for listing.
+* `-excludeRegex` - Exclude resource names matching a given regex. Applied during the listing operation and has higher priority than `-match` and `-matchRegex`.  Applicable to all resources selected for listing.  Could be used to exclude things like `databricks_automl` notebooks, etc.  
+* `filterDirectoriesDuringWorkspaceWalking` - if we should apply match logic to directory names when we're performing workspace tree walking.  *Note: be careful with it as it will be applied to all entries, so if you want to filter only specific users, then you will need to specify condition for `/Users` as well, so regex will be `^(/Users|/Users/[a-c].*)$`*.
```
mgyucht, line 66:
```suggestion
* `-filterDirectoriesDuringWorkspaceWalking` - if we should apply match logic to directory names when we're performing workspace tree walking.  *Note: be careful with it as it will be applied to all entries, so if you want to filter only specific users, then you will need to specify condition for `/Users` as well, so regex will be `^(/Users|/Users/[a-c].*)$`*.
```

mgyucht (approved):

PR #4190, opened 2024-11-03
[Internal] Make `Read` after `Create`/`Update` configurable

alexott:
## Changes

This PR adds the ability for a resource to specify that it may not need to call `Read` after `Create` and `Update` operations so we can avoid performing another API call(s). The resource may implement `CanSkipReadAfterCreateAndUpdate` function that can decide if the `Read` operation should be skipped.

I decided to move common part from #4173 to make it easier to review

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/resource_test.go
+++ b/common/resource_test.go
@@ -39,7 +39,7 @@ func TestImportingCallsRead(t *testing.T) {
 	assert.Equal(t, 1, d.Get("foo"))
 }
 
-func TestCreateSkipRead(t *testing.T) {
+func createTestResourceForSkipRead(skipRead bool) Resource {
 	res := Resource{
 		Create: func(ctx context.Context,
 			d *schema.ResourceData,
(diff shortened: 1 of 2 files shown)


common/resource_test.go
```diff
@@ -38,6 +39,100 @@ func TestImportingCallsRead(t *testing.T) {
...
+			},
+		},
+		CanSkipReadAfterCreateAndUpdate: func(d *schema.ResourceData) bool {
+			return true
+		},
+	}
+
+	client := &DatabricksClient{}
+	ctx := context.Background()
+	// Test with skipping read
```
mgyucht, line 70: Let's separate these into separate test cases, as they cover two different paths.

mgyucht (approved):
LGTM, one suggestion on the test.

PR #4196, opened 2024-11-06
Suppress equal fold diff for DLT pipeline resource

shreyas-goenka:
## Changes
Fixes https://github.com/databricks/cli/issues/1763. During creation the `catalog` name for a DLT pipeline is normalized to small case causing a persistent drift.

## Tests
Manually with the following configuration:
```
resource "databricks_pipeline" "this" {
  name = "testing caps"
  catalog = "MaiN"
  library {
    notebook {
      path = "/a/b/c"
    }
  }
}
```

Before:
There'd be a persistent drift where terraform would try to convert "MaiN" -> "main"

After:
No diff detected.

### DIFF
--- a/pipelines/resource_pipeline.go
+++ b/pipelines/resource_pipeline.go
@@ -230,6 +230,11 @@ func (Pipeline) CustomizeSchema(s *common.CustomizableSchema) *common.Customizab
 	s.SchemaPath("edition").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
 	s.SchemaPath("storage").SetCustomSuppressDiff(suppressStorageDiff)
 
+	// As of 6th Nov 2024, the DLT API only normalizes the catalog name when creating
+	// a pipeline. So we only ignore the equal fold diff for the catalog name and not other
+	// UC resources like target or ingestion_definition.connection_name.
+	s.SchemaPath("catalog").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
+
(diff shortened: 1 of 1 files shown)


pipelines/resource_pipeline.go
```diff
@@ -230,6 +230,11 @@ func (Pipeline) CustomizeSchema(s *common.CustomizableSchema) *common.Customizab
 	s.SchemaPath("edition").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
 	s.SchemaPath("storage").SetCustomSuppressDiff(suppressStorageDiff)
 
+	// As of 6th Nov 2024, the DLT API only normalizes the catalog name when creating
+	// a pipeline. So we only ignore the equal fold diff for the catalog name and not other
+	// UC resources like target or ingestion_definition.connection_name.
+	s.SchemaPath("catalog").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
```
alexott, line 236: Can you check if it's also done for the `schema` field?

alexott (commented):
lgtm, but let check for the `schema` field as well

pipelines/resource_pipeline.go
shreyas-goenka: Thanks, done and updated the comment. `schema` is not normalized.

alexott: it's really interesting that schema isn't normalized...

alexott (approved):

PR #4200, opened 2024-11-07
[Fix] Upload content `databricks_workspace_file` using raw format

alexott:
## Changes

This fixes a problem with uploading zip-based files with the `databricks_workspace_file` resource.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


edwardfeng-db (commented):
Do we wanna check whether the file extension is `.zip` and conditionally use `RAW`?

alexott:
But why we should bring this logic into TF? File is supposed to be used as-is. For notebooks we have a separate resource

parthban-db (approved):

PR #4155, opened 2024-10-25
[Feature] Add support for Primary and Foreign Key Constraints in databricks_sql_table

hshahconsulting:
## Changes

This pull request aims to enable support for Primary Key and Foreign Key constraints that are generally available in Databricks.

Reference: https://www.databricks.com/blog/primary-key-and-foreign-key-constraints-are-ga-and-now-enable-faster-queries

## Tests

- [ ] `make test` run locally
- [ ] relevant change in `docs/` folder
- [x] covered with integration tests in `internal/acceptance`
- [ ] relevant acceptance tests are passing
- [x] using Go SDK

PR #4205, opened 2024-11-09
[Feature] Add support for constraints in databricks_sql_table resource

wiatrak2:
## Changes
Hi!

I'd like to extend the `databricks_sql_table` resource with [constraints](https://docs.databricks.com/en/tables/constraints.html) feature. 

Within this pull request I extend the table's resource by allowing user to specify `PRIMARY KEY` and `FOREIGN KEY` constraints. I haven't implemented the `CHECK` constraint, as this one is more tricky and is not allowed during table creation - it can only be added later in the table's lifecycle. If the proposed changed are accepted, I can try to add the implementation for `CHECK`, as well as cover other constraint options.

I'm aware that the amount of implemented tests may not be satisfying. If this is the case - please let me know and guide my, what kind of tests should be added.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott:
there is a #4155 in progress - can you coordinate with that user to avoid duplications?

wiatrak2:
oh, indeed! I see the other user decided on slightly different approach to configuration of the constraint. And seems that the syntax is not compliant with Databricks SQL, maybe that's why there are some test failures. I will ask in that PR if we can somehow cooperate :) 

otherwise, is there something I could do to proceed with the feature?

wiatrak2:
> there is a #4155 in progress - can you coordinate with that user to avoid duplications?

seems the feature should be developed within this PR - I'm open to review and suggestions

hshahconsulting:
Could you add functionality for the "RELY" optimization of Primary Keys?

`ALTER TABLE customer ADD PRIMARY KEY (c_customer_sk) RELY`

Source: https://www.databricks.com/blog/primary-key-and-foreign-key-constraints-are-ga-and-now-enable-faster-queries

hshahconsulting (commented):
Would like the RELY optimization support for primary keys, please

wiatrak2:
Sure, optional configuration of a constraint with `RELY` added

catalog/resource_sql_table.go
```diff
@@ -89,6 +99,9 @@ func (ti SqlTableInfo) CustomizeSchema(s *common.CustomizableSchema) *common.Cus
 	s.SchemaPath("column", "type").SetCustomSuppressDiff(func(k, old, new string, d *schema.ResourceData) bool {
 		return getColumnType(old) == getColumnType(new)
 	})
+	s.SchemaPath("constraint", "type").SetCustomSuppressDiff(func(k, old, new string, d *schema.ResourceData) bool {
+		return getColumnType(old) == getColumnType(new)
```
tanmay-db, line 103: This would work because we are essentially checking for case insensitivity and ignoring the mapping here:
```
	if alias, ok := columnTypeAliases[caseInsensitiveColumnType]; ok {
		return alias
	}
```

We shouldn't use getColumnType here.

catalog/resource_sql_table.go
```diff
@@ -242,6 +255,39 @@ func (ti *SqlTableInfo) serializeColumnInfos() string {
 	return strings.Join(columnFragments[:], ", ") // id INT NOT NULL, name STRING, age INT
 }
 
+func serializePrimaryKeyConstraint(constraint ConstraintInfo) string {
```
tanmay-db, line 258: Should we make this as part of the ConstraintInfo struct that we do for getWrappedConstraintName as it will only be used for each constraint info and internally uses ConstraintInfo. getWrappedConstraintName, getWrappedKeyColumnNames and Rely? 

We do the same for SqlTableInfo. serializeColumnInfos, serializeColumnInfo

catalog/resource_sql_table.go
```diff
@@ -242,6 +255,39 @@ func (ti *SqlTableInfo) serializeColumnInfos() string {
...
+	return constraint_clause
+}
+
+func (ti *SqlTableInfo) serializeConstraints() string {
+	constraintFragments := make([]string, len(ti.Constraints))
+
+	for i, constraint := range ti.Constraints {
+		if constraint.Type == "PRIMARY KEY" {
+			constraintFragments[i] = serializePrimaryKeyConstraint(constraint)
+		} else if constraint.Type == "FOREIGN KEY" {
```
tanmay-db, line 283: Since we only support `PRIMARY KEY` and `FOREIGN KEY`, we should fail client side if some other value is used.

tanmay-db (commented):
Thank you @wiatrak2 for the PR, it looks good, I left some comments, can you please take a look?

hshahconsulting (approved):

wiatrak2:
Hi @tanmay-db 

Thank you for your review. I adjusted the implementation to your comments:
- regarding `CustomizeSchema` method - I misunderstood the functionality. Now, I decided to remove any entry for `constraints`. If you have any suggestions - please let me know
- I changed the `serialize` functions to be part of `ConstraintInfo` struct
- I strongly agree, that providing any other Constraint type than primary/foreign key on table creation phase should trigger some error, hence I decided to rearrange the statement composing function a bit to handle some `error`. Moreover, validation on constraint type during table creation allowed to implement the `CHECK` type as well, that would be consistent do Databricks way of work - i.e. it's not allowed for `CREATE TABLE`, but you can add it later.

Please let me know what you think on these recent changes

tanmay-db:
Hi @wiatrak2, thanks a lot for the contribution and sorry for the delay -- missed following up on it due to other items. There was a typo in the tests due to which unit tests were failing so added a small change. The change overall looks good, however integration tests are failing since the plan isn't empty. I will take a look at fixing this tomorrow. 

Hi @nkvuong, I agree that we should have a resource to create them but for this change I think it should be fine for it to go as is and we create a ticket to add support for PK/FK table constraint later on depending upon bandwidth of the team.

vsluc:
@wiatrak2 Thanks for taking the initiative to add support for constraints. I am looking for this as well. I see that this PR has been open for a while now, is the effort being tracked elsewhere now? Any idea when the support will be rolled out?

jreddehas-gb:
Hi @alexott , @tanmay-db ,
assuming you are the right ones to address this to. Do you know if this is still being worked on or covered somewhere else?
We would love to be able to use the databricks provider to manage primary and foreign key constraints.

Thanks in advance.

alexott:
@jreddehas-gb unfortunately current implementation is very fragile. We need new REST API instead.

tanmay-db (commented):
Hi, please re-request review if this is still valid.

PR #4206, opened 2024-11-10
[Fix] Do not validate single node clusters if a policy is configured

shreyas-goenka:
## Changes
Addresses another problem reported in https://github.com/databricks/cli/issues/1546, where customers were not able to configure and use cluster policies for single node clusters.

## Tests
Unit tests, and manually verified that cluster is successfully created without spark_conf even when `num_workers` is 0.

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -435,22 +435,6 @@ type Cluster struct {
 }
 
 // TODO: Remove this once all the resources using clusters are migrated to Go SDK.
-// They would then be using Validate(cluster compute.CreateCluster) defined in resource_cluster.go that is a duplicate of this method but uses Go SDK.
-func (cluster Cluster) Validate() error {
-	// TODO: rewrite with CustomizeDiff
-	if cluster.NumWorkers > 0 || cluster.Autoscale != nil {
-		return nil
-	}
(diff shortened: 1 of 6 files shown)


clusters/clusters_api.go
```diff
@@ -434,22 +434,6 @@ type Cluster struct {
 	ClusterMounts    []MountInfo   `json:"cluster_mount_infos,omitempty" tf:"alias:cluster_mount_info"`
 }
 
-// TODO: Remove this once all the resources using clusters are migrated to Go SDK.
-// They would then be using Validate(cluster compute.CreateCluster) defined in resource_cluster.go that is a duplicate of this method but uses Go SDK.
-func (cluster Cluster) Validate() error {
```
shreyas-goenka, line 439: This function was not being used anywhere.

alexott:
@shreyas-goenka if required attributes will be filled via cluster policy, then there will be permanent configuration drift. See note in the [doc](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/cluster#policy_id-1)

shreyas-goenka:
Closing this PR in favour of https://github.com/databricks/terraform-provider-databricks/pull/4222

PR #4216, opened 2024-11-12
[Fix] Remove single-node validation from jobs clusters

shreyas-goenka:
## Changes
Fixes https://github.com/databricks/cli/issues/1896. Introducing this validation has caused a regression for both DABs and TF customers. This PR removes the validation for job clusters.

## Tests
Unit tests.

### DIFF
--- a/clusters/clusters_api.go
+++ b/clusters/clusters_api.go
@@ -435,22 +435,6 @@ type Cluster struct {
 }
 
 // TODO: Remove this once all the resources using clusters are migrated to Go SDK.
-// They would then be using Validate(cluster compute.CreateCluster) defined in resource_cluster.go that is a duplicate of this method but uses Go SDK.
-func (cluster Cluster) Validate() error {
-	// TODO: rewrite with CustomizeDiff
-	if cluster.NumWorkers > 0 || cluster.Autoscale != nil {
-		return nil
-	}
-	profile := cluster.SparkConf["spark.databricks.cluster.profile"]
-	master := cluster.SparkConf["spark.master"]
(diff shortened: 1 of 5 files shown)


clusters/clusters_api.go
```diff
@@ -434,22 +434,6 @@ type Cluster struct {
 	ClusterMounts    []MountInfo   `json:"cluster_mount_infos,omitempty" tf:"alias:cluster_mount_info"`
 }
 
-// TODO: Remove this once all the resources using clusters are migrated to Go SDK.
-// They would then be using Validate(cluster compute.CreateCluster) defined in resource_cluster.go that is a duplicate of this method but uses Go SDK.
-func (cluster Cluster) Validate() error {
```
shreyas-goenka, line 439: this function is not used anywhere.

alexott (approved):

alexott:
Do we still need #4206?

shreyas-goenka:
@alexott Yeah we still need https://github.com/databricks/terraform-provider-databricks/pull/4206 since it tackles the issue of policy not working from clusters. I need to update that PR however.

tanmay-db:
> @shreyas-goenka Please update this as well:
> 
> https://github.com/databricks/terraform-provider-databricks/blob/e3b25617d9c9357eccc10fba2eacbd26ac3e7118/clusters/resource_cluster.go#L133-L135
> 
> @tanmay-db Can you confirm if this aligns with the direction you're taking for these resources?

Hi @pietern, yes this would be implemented a bit differently in plugin framework so it should be good. Eventually we will remove the cluster resource in SDKv2 so there would only be one place where validation would happen.

PR #4222, opened 2024-11-13
Remove single-node validation from interactive clusters

shreyas-goenka:
## Changes
There were reports in https://github.com/databricks/cli/issues/1546 of customers unable to use cluster policies for single-node clusters due to Terraform performing this validation. This PR removes the validation altogether that a single-node cluster is correctly configured.

This PR removes this validation altogether to enable those use cases. Meanwhile, the clusters team is working on API-side improvements to improve the API interface for creating single-node clusters, which will make it much less likely for users to misconfigure their single-node clusters.

## Tests
Unit test and manually.

### DIFF
--- a/clusters/resource_cluster.go
+++ b/clusters/resource_cluster.go
@@ -26,26 +26,6 @@ var clusterSchema = resourceClusterSchema()
 var clusterSchemaVersion = 4
 
 const (
-	numWorkerErr = `num_workers may be 0 only for single-node clusters. To create a single node
-cluster please include the following configuration in your cluster configuration:
-
-  spark_conf = {
-    "spark.databricks.cluster.profile" : "singleNode"
-    "spark.master" : "local[*]"
-  }
-
-  custom_tags = {
-    "ResourceClass" = "SingleNode"
-  }
-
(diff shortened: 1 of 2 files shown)


alexott:
@tanmay-db what is your opinion giving that you're porting clusters to the plugin framework?

pietern:
@shreyas-goenka The PR you reference talks about job clusters. Am I missing the ref to interactive clusters?

alexott (approved):
ok from the code perspective. Let's wait for Tanmay's opinion

shreyas-goenka:
@pietern, you are right. The issue only discusses job clusters. There's no reference for interactive clusters, but now that we support interactive clusters via DABs, we should solve this issue here as well and keep them consistent. 

Updated the PR description to say as much.

tanmay-db:
> @tanmay-db what is your opinion giving that you're porting clusters to the plugin framework?

Hi @alexott, it looks good, I will update the plugin framework implementation to not have this validation.

tanmay-db (approved):

shreyas-goenka:
@tanmay-db We'll be adding this validation to DABs as a warning instead of a warning error. Maybe that's something we should do in the plugin framework as well if supported.

PR #4224, opened 2024-11-13
[Fix] remove pause_status for lakehouse monitoring to fix updatemonitor error 

xinyijiang-db:
## Changes

remove the pause_status from the terraform 
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


catalog/resource_quality_monitor.go
```diff
@@ -53,7 +53,6 @@ func ResourceQualityMonitor() common.Resource {
 			common.CustomizeSchemaPath(m, "profile_metrics_table_name").SetReadOnly()
 			common.CustomizeSchemaPath(m, "status").SetReadOnly()
 			common.CustomizeSchemaPath(m, "dashboard_id").SetReadOnly()
-			common.CustomizeSchemaPath(m, "schedule", "pause_status").SetReadOnly()
```
tanmay-db, line 56: Hi @xinyijiang-db , thanks for the PR, this doesn't remove the pause_status from schema, only removes it not being set to read only. Also, isn't this a breaking change? 

If this is needed to be done, then it would also need to be done here: https://github.com/databricks/terraform-provider-databricks/blob/main/internal/providers/pluginfw/resources/qualitymonitor/resource_quality_monitor.go.

alexott (changes requested):
I think that we need to go standard deprecation route in this case - if the field is removed from the corresponding API struct, then we need to add it to schema manually and mark it as deprecated, and remove in 2-3 releases after that.

xinyijiang-db:
> I think that we need to go standard deprecation route in this case - if the field is removed from the corresponding API struct, then we need to add it to schema manually and mark it as deprecated, and remove in 2-3 releases after that.

hi @alexott this is a fix because the pause_status is not supposed to be in the schema. 
when we update, we do a read before we do a read then update, and having the pause_status causes the update function to fail because our service don't allow pause_status to be passed in the request

also cc @arpitjasa-db who was in the initial discussion with the LHM team and suggested the fix

alexott:
@xinyijiang-db pause status is still in the API spec: https://docs.databricks.com/api/workspace/qualitymonitors/create#schedule-pause_status, so it's automatically pulled into Go SDK and used. Also, Update REST API has it: https://docs.databricks.com/api/workspace/qualitymonitors/update#schedule-pause_status

I agree with @tanmay-db that it doesn't remove this field from the schema... If you need to remove it completely - make sure you deprecate it in the API spec

xinyijiang-db:
@alexott let me clarify a little bit more. we still want it in the response so that users can see it but we don't want it in the update function request to our service

it seems that terraform does a read before it calls our update function, which causes our logic to break

arpitjasa-db:
> @xinyijiang-db pause status is still in the API spec: https://docs.databricks.com/api/workspace/qualitymonitors/create#schedule-pause_status, so it's automatically pulled into Go SDK and used. Also, Update REST API has it: https://docs.databricks.com/api/workspace/qualitymonitors/update#schedule-pause_status
> 
> I agree with @tanmay-db that it doesn't remove this field from the schema... If you need to remove it completely - make sure you deprecate it in the API spec

I think the idea is that this field doesn't need to be managed by Terraform, and LHM team has had some folks report issues because this field is being sent back in the Update API, so it's easier to just have Terraform ignore this field. However, this field is already marked as `ReadOnly` in the code so not sure why is this happening? Is this an issue with `ReadOnly` then?

In terms of a breaking change, folks shouldn't be able to set this field today in their HCL config right? Curious if it does break existing users then?

catalog/resource_quality_monitor.go
arpitjasa-db: > @alexott let me clarify a little bit more. we still want it in the response so that users can see it but we don't want it in the update function request to our service

> it seems that terraform does a read before it calls our update function, which causes our service to reject the update request

@xinyijiang-db yeah if you want to remove it from schema, you need to delete it explicitly, not just remove the read only status. Alternatively, in the `Update` section of the resource, you could also empty the field there so it's not sent in the request

alexott:
We need to add or modify integration tests with the `schedule` block defined, so we can make sure that change actually works

PR #4235, opened 2024-11-15
[Fix] Fix deletion of dashboard if it was trashed out of band

pietern:
## Changes

If a dashboard created by TF was trashed out of band, it could no longer be deleted by TF.

Example TF configuration:
```hcl
terraform {
  required_providers {
    databricks = {
      version = "1.58.0"
    }
  }
}

data "databricks_current_user" "me" {
  // This data source is used to get the current user's username
}

resource "databricks_dashboard" "this" {
  display_name = "Terraform test"
  parent_path = "/Workspace/Users/${data.databricks_current_user.me.user_name}/tf-dashboard-oob-deletion"
  warehouse_id = "58aa1b363649e722"
  serialized_dashboard = file("nyc_taxi_trip_analysis.lvdash.json")
}
```

If you apply this, and then run:
```
databricks lakeview trash <dashboard id>
```

The subsequent apply would error out with:
```
Error: cannot delete dashboard: dashboard [01efa39479a1128aae017018e434a747] lifecycle state [TRASHED] is not among the accepted states [Set(ACTIVE)]
```

It turns out trashing an already trashed dashboard returns an error.

This change guards against this condition returns success if the underlying dashboard has already been trashed.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/dashboards/resource_dashboard.go
+++ b/dashboards/resource_dashboard.go
@@ -2,9 +2,11 @@ package dashboards
 
 import (
 	"context"
+	"errors"
 	"log"
 	"strings"
 
+	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/service/dashboards"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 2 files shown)


dashboards/resource_dashboard.go
```diff
@@ -175,9 +177,32 @@ func ResourceDashboard() common.Resource {
...
+
+			// Attempt to trash the dashboard.
+			err = w.Lakeview.Trash(ctx, dashboards.TrashDashboardRequest{
 				DashboardId: d.Id(),
 			})
+
+			// If the dashboard was already trashed, we'll get a 403 (Permission Denied) error.
+			// There may be other cases where we get a 403, so we first confirm that the
+			// dashboard state is actually trashed, and if so, return success.
+			if errors.Is(err, apierr.ErrPermissionDenied) {
```
tanmay-db, line 189: Thanks for the PR. Curious -- should we also check for 404 or if a dashboard is trashed already then the only response is 403? Mentioning since the api doc https://docs.databricks.com/api/workspace/lakeview/trash has 404 as one of the responses.

tanmay-db (approved):

dashboards/resource_dashboard.go
pietern: It will return 404 if it really doesn't exist (not even in the trash).

Good point, though. This fix works around the issue at deletion time, but we should also treat trashed as a 404 at read time so that if it was trashed, the next plan will show **creation** instead of **recreation**.

PR #4265, opened 2024-11-22
[Fix] Forced send `auto_stop_mins` for `databricks_sql_endpoint` resource

nkvuong:
## Changes
- Forced send `auto_stop_mins` for `databricks_sql_endpoint` resource. This allows creating SQL warehouses with auto stop turned off

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/sql/resource_sql_endpoint.go
+++ b/sql/resource_sql_endpoint.go
@@ -95,7 +95,7 @@ func ResourceSqlEndpoint() common.Resource {
 			}
 			var se sql.CreateWarehouseRequest
 			common.DataToStructPointer(d, s, &se)
-			common.SetForceSendFields(&se, d, []string{"enable_serverless_compute", "enable_photon"})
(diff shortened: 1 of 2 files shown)


sql/resource_sql_endpoint.go
```diff
@@ -95,7 +95,7 @@ func ResourceSqlEndpoint() common.Resource {
 			}
 			var se sql.CreateWarehouseRequest
 			common.DataToStructPointer(d, s, &se)
-			common.SetForceSendFields(&se, d, []string{"enable_serverless_compute", "enable_photon"})
+			common.SetForceSendFields(&se, d, []string{"enable_serverless_compute", "enable_photon", "auto_stop_mins"})
```
alexott, line 98: We need to send it for update as well... it will be better to defined a single constant array and use it for both update & create

alexott (changes requested):

alexott (approved):

PR #4281, opened 2024-11-29
[Doc] Update documentation for importing some MWS resources

mgyucht:
## Changes
Some resources are documented as not being importable, but some actually are. This PR adds docs to a few that I was able to test out myself.

## Tests
Imported existing resources into a local TF project, which worked.


alexott (commented):
Doc looks good, but let's put some warning about a potential danger that changes may cause destroying of resources. With either `~>` or `!>` callout

alexott (approved):

PR #4283, opened 2024-11-29
[Internal] Remove unused configuration from blocks

mgyucht:
## Changes
Currently, our schema builder elements in the plugin framework components allow you to set Computed/Optional/Required/Sensitive fields on Block schema elements. However, these are simply dropped when converting to actual schema elements: only attributes have these fields. This PR moves the SetComputed/SetOptional/SetRequired/SetSensitive/SetReadOnly methods to the AttributeBuilder interface and panics if you attempt to set them on a block.

## Tests
Added a unit test to verify that these methods panic when called on an element that is converted into a block.

### DIFF
--- a/internal/providers/pluginfw/tfschema/attribute_builder.go
+++ b/internal/providers/pluginfw/tfschema/attribute_builder.go
@@ -10,6 +10,25 @@ import (
 // This common interface prevents us from keeping two copies of StructToSchema and CustomizableSchema.
 type AttributeBuilder interface {
 	BaseSchemaBuilder
+
+	// SetOptional sets the attribute as optional in the schema. This does not affect whether the attribute is computed.
+	// It fails if the attribute is already optional.
+	SetOptional() BaseSchemaBuilder
+
+	// SetRequired sets the attribute as required in the schema. This does not affect whether the attribute is computed.
+	// It fails if the attribute is already required.
+	SetRequired() BaseSchemaBuilder
+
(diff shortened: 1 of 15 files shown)


mgyucht:
@alexott That's a great point. Going forward, we're going to follow the advice from https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed to migrate those blocks to nested attributes. Once this is done, we can mark the entire nested attribute with ReadOnly(). I will make another PR to handle that conversion.

internal/providers/pluginfw/tfschema/attribute_builder.go
```diff
@@ -10,6 +10,25 @@ import (
...
+	// SetSensitive sets the attribute as sensitive in the schema. It fails if the attribute is already sensitive.
+	SetSensitive() BaseSchemaBuilder
+
+	// SetComputed sets the attribute as computed in the schema. It fails if the attribute is already computed.
+	SetComputed() BaseSchemaBuilder
+
+	// Sets the attribute as read-only in the schema, i.e. computed and neither optional or required. It fails if the
+	// attribute is already read-only.
+	SetReadOnly() BaseSchemaBuilder
+
```
rauchy, line 31: Should these return `BaseSchemaBuilder` or `AttributeBuilder`?

rauchy (approved):

PR #4284, opened 2024-12-02
[Internal] Add ConvertToAttribute() to convert blocks in a resource/data source schema to attributes

mgyucht:
## Changes
Blocks in the Plugin Framework cannot be computed. Terraform checks that the number of blocks in the plan matches the number in the config (https://github.com/hashicorp/terraform/blob/81441564dd29b4aa5fb9576323ef90b9787d1967/internal/plans/objchange/plan_valid.go#L115). Today, for compatibility with existing resources in the TF provider implemented with the SDKv2, we treat nested structures as lists with a single element. As a result, this precludes any nested structures from being read-only (since those would be converted into ListNestedBlocks, which cannot be computed). This blocks databricks_app from being implemented on the plugin framework, as its app_status and compute_status fields 
are objects and not primitives. Note that TF recommends computing computed-only blocks to attributes [here](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed). Other useful discussions can be found [here](https://discuss.hashicorp.com/t/set-list-attribute-migration-from-sdkv2-to-framework/56472) and [here](https://discuss.hashicorp.com/t/optional-computed-block-handling-in-plugin-framework/56337). The migration guide for blocks can be found [here](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks).

As an intermediate step, we add `ConvertToAttribute()` in `CustomizableSchema`. This allows resource maintainers to convert specific, known read-only blocks to attributes in accordance with [the migration guide](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed), which then can be marked as read-only.

As blocks can contain other blocks, this method also recursively converts nested blocks into nested attributes.

This is based on https://github.com/databricks/terraform-provider-databricks/pull/4283, so we should wait for that PR to merge & rebase this before merging this.

## Tests
Unit tests validate that ListNestedBlocks are converted to ListNestedAttributes and that SingleNestedBlocks are converted to SingleNestedAttributes.

### DIFF
--- a/internal/providers/pluginfw/tfschema/customizable_schema.go
+++ b/internal/providers/pluginfw/tfschema/customizable_schema.go
@@ -211,34 +211,8 @@ func (s *CustomizableSchema) ConvertToAttribute(path ...string) *CustomizableSch
 
 	cb := func(attr BaseSchemaBuilder) BaseSchemaBuilder {
 		switch a := attr.(type) {
-		case ListNestedBlockBuilder:
-			elem, ok := a.NestedObject.Blocks[field]
-			if !ok {
-				panic(fmt.Errorf("field %s does not exist in nested block", field))
-			}
-			if a.NestedObject.Attributes == nil {
-				a.NestedObject.Attributes = make(map[string]AttributeBuilder)
-			}
-			a.NestedObject.Attributes[field] = elem.ToAttribute()
-			delete(a.NestedObject.Blocks, field)
-			if len(a.NestedObject.Blocks) == 0 {
-				a.NestedObject.Blocks = nil
-			}
-			return a
-		case SingleNestedBlockBuilder:
-			elem, ok := a.NestedObject.Blocks[field]
-			if !ok {
-				panic(fmt.Errorf("field %s does not exist in nested block", field))
-			}
-			if a.NestedObject.Attributes == nil {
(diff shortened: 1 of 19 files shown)


internal/providers/pluginfw/tfschema/customizable_schema.go
```diff
@@ -199,8 +200,63 @@ func (s *CustomizableSchema) SetReadOnly(path ...string) *CustomizableSchema {
 	return s
 }
 
+// ConvertToAttribute converts the last element of the path from a block to an attribute.
+// It panics if the path is empty, if the path does not exist in the schema, or if the path
+// points to an attribute, not a block.
+func (s *CustomizableSchema) ConvertToAttribute(path ...string) *CustomizableSchema {
```
rauchy, line 206: This method is a bit complex. Can we simplify it somehow?

Just an idea: it looks like the parts related to `CustomizableSchema` are mostly outside the callback definition, in lines 207-210 and 247-255. What if we move the callback outside of this method? Would that make things easier to follow? Maybe it would be better to implement the callback within the builders and split it between `ListNestedBlockBuilder` and `SingleNestedBlockBuilder`?

rauchy (approved):

PR #4289, opened 2024-12-03
[Fix] Add client side validation for `volume_type`

shreyas-goenka:
## Changes
This PR adds validation that the value provided for `volume_type` is one of the correct values. The server today provides an incorrect error message:

```
10:30:10 DEBUG POST /api/2.1/unity-catalog/volumes
> {
>   "catalog_name": "main",
>   "name": "cli-volume",
>   "schema_name": "schema-dec-dabs",
>   "volume_type": "managed"
> }
< HTTP/2.0 400 Bad Request
< {
<   "details": [
<     {
<       "@type": "type.googleapis.com/google.rpc.ErrorInfo",
<       "domain": "unity-catalog.databricks.com",
<       "metadata": {
<         "field_name": "volume_type"
<       },
<       "reason": "INVALID_FIELD"
<     },
<     {
<       "@type": "type.googleapis.com/google.rpc.RequestInfo",
<       "request_id": "2ca7e630-ce06-4c85-ad95-9b3b52987009",
<       "serving_data": ""
<     }
<   ],
<   "message": "CreateVolume Missing required field: volume_type"
< }
```

## Tests
Unit tests

### DIFF
--- a/catalog/resource_volume.go
+++ b/catalog/resource_volume.go
@@ -57,8 +57,8 @@ func ResourceVolume() common.Resource {
 			// error message "CreateVolume Missing required field: volume_type"
 			// if you specify an invalid value for volume_type (i.e. not one of "MANAGED" or "EXTERNAL").
 			//
-			// If server side validation is added in the future, this validation function 
-			// can be removed. 
+			// If server side validation is added in the future, this validation function
+			// can be removed.
 			m["volume_type"].ValidateDiagFunc = func(i interface{}, p cty.Path) diag.Diagnostics {
 				s, ok := i.(string)
 				if !ok {
--- a/catalog/resource_volume.go
+++ b/catalog/resource_volume.go
@@ -53,6 +53,12 @@ func ResourceVolume() common.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			}
+			// As of 3rd December 2024, the Volumes create API returns an incorrect
+			// error message "CreateVolume Missing required field: volume_type"
+			// if you entire an invalid value for volume_type (i.e. not one of "MANAGED" or "EXTERNAL").
+			//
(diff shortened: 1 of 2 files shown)


catalog/resource_volume.go
```diff
@@ -50,6 +53,25 @@ func ResourceVolume() common.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			}
+			// As of 3rd December 2024, the Volumes create API returns an incorrect
+			// error message "CreateVolume Missing required field: volume_type"
+			// if you specify an invalid value for volume_type (i.e. not one of "MANAGED" or "EXTERNAL").
+			//
+			// If server side validation is added in the future, this validation function
+			// can be removed.
+			m["volume_type"].ValidateDiagFunc = func(i interface{}, p cty.Path) diag.Diagnostics {
```
alexott, line 62: why do we need complex validation function instead of `.SetValidateFunc(validation.StringInSlice`? like here: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_workspace_binding.go#L46

shreyas-goenka: I'm fine with either approach. The advantage is if a new volume type is added in the future (unlikely, but who knows), then this code will automatically be updated.

Let me know which approach you prefer though.

alexott: We need to generalize it - I discussed it with @mgyucht before - we have other places where we need to check against enum

shreyas-goenka: Thanks, switched to the simpler validation function here. This can be removed once we have the general mechanism.

alexott (approved):

alexott:
@shreyas-goenka I've opened a DECO Jira to add a generation of all possible constants, so it could then be automatically updated when the new SDK introduces new constants.

PR #4300, opened 2024-12-06
[Internal] Bump Go SDK and generate TF structs

tanmay-db:
## Changes

Bump Go SDK to latest and generate TF structs to same OpenAPI spec as Go SDK 

## Tests

N/A
- [ ] `make test` run locally
- [ ] using Go SDK


internal/service/cleanrooms_tf/model.go
```diff
@@ -0,0 +1,608 @@
...
+
+func (newState *CleanRoomAssetTable) SyncEffectiveFieldsDuringRead(existingState CleanRoomAssetTable) {
+}
+
+type CleanRoomAssetTableLocalDetails struct {
+	// The fully qualified name of the table in its owner's local metastore, in
+	// the format of *catalog*.*schema*.*table_name*
+	LocalName types.String `tfsdk:"local_name" tf:"optional"`
+	// Partition filtering specification for a shared table.
+	Partitions sharing.PartitionSpecificationPartition `tfsdk:"partitions" tf:"optional"`
```
tanmay-db, line 163: This should use sharing_tf instead of sharing.

alexott (approved):

internal/service/cleanrooms_tf/model.go
tanmay-db: This most likely is a bug in generation, we would need to fix this before we can merge.

alexott: @tanmay-db is it really a blocker? We don't have any CleanRoom-related resource yet...
Imho it could be fixed in the subsequent SDK updates

PR #4309, opened 2024-12-11
[Internal] Panic if the provided path is invalid

mgyucht:
## Changes
Callers of `CustomizableSchema`'s methods can supply invalid paths. Currently, this returns an error, but this is never OK and should always be fixed before making any changes to a resource. Here, I change `navigateSchemaWithCallback` to panic if the supplied path does not exist.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


rauchy (approved):
A very welcome change!

PR #4315, opened 2024-12-12
[Internal] Use Attributes by default for List Objects

mgyucht:
## Changes
Resources implemented on the SDKv2 used blocks to represent two kinds of fields: traditional lists, and nested complex types (which were treated as a list of length at most 1). The Plugin Framework now has support for list attributes, and it is recommended to use this in general, except for places where compatibility with SDKv2 is needed.

To achieve this, this change ensures that *StructToSchema methods in the plugin framework generate attributes for lists by default. Then, for resources migrated from SDKv2, the implementer must call `ConfigureForSdkV2Migration()` in the customize callback. This will convert any list attributes into list blocks, preserving compatibility with SDKv2-implemented resources.

Additionally, this PR improves the handling of computed, non-optional fields. For lists annotated as `tf:"computed"`, the resulting schema is invalid: it is marked as computed and as required because of the absence of the `"optional"` tag. When a field is labeled as computed, it can be marked as optional but should never be marked as required; that is addressed here as well.

Finally, this PR also includes some autogenerated changes modifying Apps to use `types.Object` in place of `types.List` for nested complex types.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


internal/providers/pluginfw/tfschema/customizable_schema.go
```diff
@@ -202,32 +202,12 @@ func (s *CustomizableSchema) SetReadOnly(path ...string) *CustomizableSchema {
...
-	if len(path) == 1 {
-		s.attr = cb(s.attr)
-	} else {
-		navigateSchemaWithCallback(&s.attr, cb, path[0:len(path)-1]...)
-	}
-
+// ConfigureForSdkV2Migration modifies the underlying schema to be compatible with SDKv2. This method must
+// be called on all resources that were originally implemented using the SDKv2 and are migrated to the plugin
+// framework.
+func (s *CustomizableSchema) ConfigureForSdkV2Migration() *CustomizableSchema {
```
rauchy, line 208: I don't know why, but for me `ConfigureForSdkV2Migration` sends the message that it prepares it for a migration _towards_ SDKv2, how about something with "compatibility", like `MakeCompatibleWithSdkV2` or `ConfigureAsSdkV2Compatible`?

rauchy (approved):

mgyucht:
Note: the diff in schema on this PR is purely in data sources, converting block types to attributes for purely computed fields, in line with https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed. There are no diffs in resource schemas.

PR #4316, opened 2024-12-12
[Internal] Generate models with correctly computed annotations

rauchy:
## Changes

In this PR, we've tackled an issue with our plugin framework models. Previously, fields that were meant to be computed were mistakenly generated with "computed,optional" annotations instead of just "computed."
Also, if any fields were indicated as having effective values, the effective fields also ended up with the wrong "computed,optional" tag rather than the correct "computed." We've regenerated the models to correct these annotations.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


internal/service/apps_tf/model.go
```diff
@@ -418,11 +418,11 @@ func (o *AppAccessControlResponse) SetAllPermissions(ctx context.Context, v []Ap
...
 type AppDeployment struct {
 	// The creation time of the deployment. Formatted timestamp in ISO 6801.
-	CreateTime types.String `tfsdk:"create_time" tf:"computed,optional"`
+	CreateTime types.String `tfsdk:"create_time" tf:"computed"`
 	// The email of the user creates the deployment.
-	Creator types.String `tfsdk:"creator" tf:"computed,optional"`
+	Creator types.String `tfsdk:"creator" tf:"computed"`
 	// The deployment artifacts for an app.
-	DeploymentArtifacts types.Object `tfsdk:"deployment_artifacts" tf:"optional,object"`
+	DeploymentArtifacts types.Object `tfsdk:"deployment_artifacts" tf:"computed,object"`
```
mgyucht, line 425: This was optional before but now is computed non-optional. Is that expected?

mgyucht (approved):
LGTM, one question.

PR #4332, opened 2024-12-18
[Internal] Generate both SdkV2-compatible and Plugin Framework-compatible structures

mgyucht:
## Changes
To make it possible to implement new resources following the Plugin Framework's recommendation to use attributes while allowing resources to be migrated from SDKv2 to the Plugin Framework without breaking, we will create two copies of each code-generated structure in the Terraform provider. The structures with no suffix should be used for new resources. The structures with the `_SdkV2` suffix should be used for resources in the plugin framework that are migrated from SDKv2.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


rauchy (approved):

internal/service/sharing_tf/model.go
```diff
@@ -2289,7 +2289,7 @@ type ShareInfo struct {
 	// A list of shared data objects within the share.
 	Objects types.List `tfsdk:"object" tf:"optional"`
 	// Username of current owner of share.
-	Owner types.String `tfsdk:"owner" tf:"computed,optional"`
+	Owner types.String `tfsdk:"owner" tf:"computed"`
```
rauchy, line 2292:
```suggestion
	Owner types.String `tfsdk:"owner" tf:"computed,optional"`
```

mgyucht:
Only diff is in that databricks_share_pluginframework's `owner` attribute is no longer optional, but the whole resource is computed anyways and users only specify the share name.

PR #4344, opened 2024-12-23
[Doc] apply `make fmt-docs` to all docs

alexott:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/resources/app.md
+++ b/docs/resources/app.md
@@ -11,28 +11,28 @@ subcategory: "Apps"
 
 ```hcl
 resource "databricks_app" "this" {
-  name             = "my-custom-app"
-  description      = "My app"
+  name        = "my-custom-app"
+  description = "My app"
   resources = [{
     name = "sql-warehouse"
     sql_warehouse = {
-      id = "e9ca293f79a74b5c"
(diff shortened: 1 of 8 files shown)


docs/resources/app.md
```diff
@@ -11,28 +11,28 @@ subcategory: "Apps"
 
 ```hcl
 resource "databricks_app" "this" {
-  name             = "my-custom-app"
-  description      = "My app"
+  name        = "my-custom-app"
+  description = "My app"
   resources = [{
```
mgyucht, line 16: Interesting. This first resource is less indented than the others, maybe due to some strange logic with the `[` and `{` on the same line. Can you put a newline there and run it once more?

mgyucht (approved):
Thanks for this! One small suggestion, otherwise LGTM.

docs/resources/app.md
alexott: adding a new line doesn't change the behavior. Also, I tested with putting the code into a terraform file and running `terraform fmt` on it - it's formatted it the same way as with `fmt-docs`

PR #4355, opened 2024-12-27
[Internal] Retry on 504 when calling the permission API

renaudhartert-db:
## Changes

This PR adds logic to retry failed get calls to the permissions API when the error is a 504. This solution is meant to be temporary and will be removed as soon as such retries are handled natively in the Databricks Go SDK. 

## Tests

Complete coverage of the added retrier.

### DIFF
--- a/common/retry.go
+++ b/common/retry.go
@@ -2,9 +2,11 @@ package common
 
 import (
 	"context"
-	"log"
+	"errors"
 	"regexp"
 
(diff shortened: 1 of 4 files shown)


rauchy (approved):
LGTM. Do we want to stub out a 504 in `resource_permissions_test.go` and verify the retry is kicking in there?

PR #4364, opened 2024-12-30
[WIP][Feature] add no_compute support

mgyucht:
## Changes
Example PR of how to support `no_compute` flag for Apps. The flag is only used on create to determine whether compute should be provisioned as part of the Create App RPC. Changes to this flag for existing resources are no-ops and simply propagated from plan to state.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/products/app/resource_app.go
+++ b/internal/providers/pluginfw/products/app/resource_app.go
@@ -15,12 +15,18 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
 const (
 	resourceName       = "app"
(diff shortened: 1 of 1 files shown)


internal/providers/pluginfw/products/app/resource_app.go
```diff
@@ -128,7 +189,7 @@ func (a *resourceApp) Read(ctx context.Context, req resource.ReadRequest, resp *
 		return
 	}
 
-	var newApp apps_tf.App
+	var newApp appResource
 	resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, appGoSdk, &newApp)...)
```
alexott, line 193: When we read it back, won't it be reset to `false`?

alexott (commented):
codewise looks good

internal/providers/pluginfw/products/app/resource_app.go
mgyucht: Excellent eye. This is causing integration tests to fail. I think it is set to `null` but otherwise you're exactly right. I need to set this explicitly in the `Read()` method.

alexott (approved):

Issue #4353, opened 2024-12-27
[ISSUE] Issue with `databricks_library` resource not starting the cluster before library installation and as a result getting failed


sudhir-garg:
### Configuration

```hcl
resource "databricks_library" "pyyaml" {
  cluster_id = databricks_cluster.my_cluster_with_libs.id
  pypi {
    package = "PyYAML==6.0.2"
  }
}
```

### Expected Behavior

As per the documentation the resource will first start the cluster if it is not in 'start' state and then install the required packages onto the cluster.

### Actual Behavior

The resource is not starting my cluster though it is in 'terminated' state and keep on waiting for the library installation for 15 minutes and then gets failed.

```
Plan: 1 to add, 0 to change, 0 to destroy.
databricks_library.pyyaml: Creating...
databricks_library.pyyaml: Still creating... [10s elapsed]
databricks_library.pyyaml: Still creating... [20s elapsed]
databricks_library.pyyaml: Still creating... [30s elapsed]
databricks_library.pyyaml: Still creating... [40s elapsed]
databricks_library.pyyaml: Still creating... [50s elapsed]
databricks_library.pyyaml: Still creating... [1m0s elapsed]
databricks_library.pyyaml: Still creating... [1m10s elapsed]
databricks_library.pyyaml: Still creating... [1m20s elapsed]
databricks_library.pyyaml: Still creating... [1m30s elapsed]
databricks_library.pyyaml: Still creating... [1m40s elapsed]
databricks_library.pyyaml: Still creating... [1m50s elapsed]
databricks_library.pyyaml: Still creating... [2m0s elapsed]
databricks_library.pyyaml: Still creating... [2m10s elapsed]
databricks_library.pyyaml: Still creating... [2m20s elapsed]
databricks_library.pyyaml: Still creating... [2m30s elapsed]
databricks_library.pyyaml: Still creating... [2m40s elapsed]
databricks_library.pyyaml: Still creating... [2m50s elapsed]
databricks_library.pyyaml: Still creating... [3m0s elapsed]
databricks_library.pyyaml: Still creating... [3m10s elapsed]
databricks_library.pyyaml: Still creating... [3m20s elapsed]
databricks_library.pyyaml: Still creating... [3m30s elapsed]
databricks_library.pyyaml: Still creating... [3m40s elapsed]
databricks_library.pyyaml: Still creating... [3m50s elapsed]
databricks_library.pyyaml: Still creating... [4m0s elapsed]
[...]

Issue #4036, opened 2024-09-20
[FEATURE] Add external libraries without starting clusters


pgrandjean:
### Use-cases

Deployment of external libraries in clusters takes a lot of time as clusters need to be started, even if they already exist. Once a cluster is started, then adding external libraries is fast. However, when accounting for the start-up time, it actually is quite slow. It looks like starting the cluster is a mandatory step to install an external library. However, it is possible to install external libraries from the UI, manually, without starting a cluster.

The aim of this feature request is to implement a solution to install external libraries on existing clusters without starting them, as it is possible to do it manually from the UI.

PR #4384, opened 2025-01-08
[Internal] Restart Cluster before Library Installation

parthban-db:
## Changes

Library resource with SDK_V2 implementation waits for the cluster to get into the running state before starting installation. This PR mimics this behavior in the plugin framework implementation of the resource. 

Fixes https://github.com/databricks/terraform-provider-databricks/issues/4353
## Tests

Added a integration test.
- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/products/library/resource_library_acc_test.go
+++ b/internal/providers/pluginfw/products/library/resource_library_acc_test.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/retries"
+	"github.com/databricks/databricks-sdk-go/service/compute"
 	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
 	"github.com/databricks/terraform-provider-databricks/internal/providers"
 	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw"
@@ -63,7 +65,7 @@ func TestAccLibraryReinstalledIfClusterDeleted(t *testing.T) {
 				return nil
 			},
 		},
-		// If the cluster is terminated before apply, it should be restarted and the library reinstalled.
+		// If the cluster is deleted before apply, it should be recreated and the library reinstalled on the new cluster.
 		acceptance.Step{
 			PreConfig: func() {
 				// Delete the created cluster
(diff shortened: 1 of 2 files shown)


alexott:
@parthban-db we have https://github.com/databricks/terraform-provider-databricks/issues/4036 that mentions that we can install libraries in UI without starting the cluster - the question is if it works via API the same way or not...

alexott (approved):
In general is good, see my comment about existing issue about installing on terminated clusters via UI

parthban-db:
>@parthban-db we have https://github.com/databricks/terraform-provider-databricks/issues/4036 that mentions that we can install libraries in UI without starting the cluster - the question is if it works via API the same way or not...

I will ask the team about this.

parthban-db:
> @parthban-db we have https://github.com/databricks/terraform-provider-databricks/issues/4036 that mentions that we can install libraries in UI without starting the cluster - the question is if it works via API the same way or not...

@alexott I talked with the team, and according to them, `/api/2.0/libraries/install` should not need to start the cluster, but the actual installation on the dataplane will require the cluster to be live. 
I'm not sure how we want Terraform to behave in this case. We can either start the cluster and wait for it to go live or just create the library, and the library will be installed once the cluster goes live.

alexott:
@parthban-db, We can make it controllable by the user - we start the cluster and install the library just to check if there are no errors. But if we can allow users to specify that they don't want to wait for it, then it could be a solution.

@mgyucht wdyt?

mgyucht:
LGTM: we revert this back to the original & documented behavior of starting the cluster if it isn't already running.

Longer term, we can revisit how Terraform users will think about cluster and library lifecycles. I can see this as very useful in some use-cases (e.g. DABs), but there may also be users who don't want a cluster to be started for the sake of creating/updating a library.

internal/providers/pluginfw/products/library/resource_library_acc_test.go
```diff
@@ -92,6 +94,42 @@ func TestAccLibraryReinstalledIfClusterDeleted(t *testing.T) {
...
+				clusterId = s.RootModule().Resources["databricks_cluster.this"].Primary.ID
+				return nil
+			},
+		},
+		// If the cluster is Terminated before apply, it should be restarted before installing library.
+		acceptance.Step{
+			PreConfig: func() {
+				// Delete the created cluster
+				w := databricks.Must(databricks.NewWorkspaceClient())
+				getter, err := w.Clusters.Delete(context.Background(), compute.DeleteCluster{
```
mgyucht, line 112: What a confusing name for this method... @renaudhartert-db maybe we want to revisit this in the next version of the SDK. It should really be "terminate cluster" to align with basically all other terminology.

mgyucht (approved):

PR #4406, opened 2025-01-16
[Feature] Updated AWS UC storage credential to include permissions for file events

borremosch-db:
## Changes
[Databricks documentation for storage credentials](https://docs.databricks.com/en/connect/unity-catalog/cloud-storage/storage-credentials.html) contains instructions to add permissions for file events, but as of yet these are missing from the terraform provider. This PR adds them for AWS. PRs for Azure and GCP will follow soon

## Tests
Updated test: `aws/data_aws_unity_catalog_policy_test.go`

- [x] `make test` run locally
- [ ] using Go SDK


aws/data_aws_unity_catalog_policy.go
```diff
@@ -60,6 +60,59 @@ func generateReadContext(ctx context.Context, d *schema.ResourceData, m *common.
...
+			"sqs:GetQueueAttributes",
+			"sqs:SetQueueAttributes",
+			"sqs:TagQueue",
+			"sqs:ChangeMessageVisibility",
+			"sqs:PurgeQueue",
+		},
+		Resources: []string{
+			fmt.Sprintf("arn:%s:s3:::%s", awsPartition, bucket),
+			"arn:aws:sqs:*:*:*",
+			"arn:aws:sns:*:*:*",
```
mgyucht, line 90: Are customers comfortable with granting us permission on all SQS queues and SNS destinations? It may be sensible as a default but I expect they will want to be more selective. I wonder if e.g. there is a specific prefix that Databricks always uses so we can restrict this somewhat?

aws/data_aws_unity_catalog_policy.go
```diff
@@ -60,6 +60,59 @@ func generateReadContext(ctx context.Context, d *schema.ResourceData, m *common.
...
+	},
+		&awsIamPolicyStatement{
+			Sid:    "ManagedFileEventsListStatement",
+			Effect: "Allow",
+			Actions: []string{
+				"sqs:ListQueues",
+				"sqs:ListQueueTags",
+				"sns:ListTopics",
+			},
+			Resources: "*",
```
mgyucht, line 101: Ditto here.

aws/data_aws_unity_catalog_policy.go
```diff
@@ -60,6 +60,59 @@ func generateReadContext(ctx context.Context, d *schema.ResourceData, m *common.
...
+			Sid:    "ManagedFileEventsTeardownStatement",
+			Effect: "Allow",
+			Actions: []string{
+				"sns:Unsubscribe",
+				"sns:DeleteTopic",
+				"sqs:DeleteQueue",
+			},
+			Resources: []string{
+				"arn:aws:sqs:*:*:*",
+				"arn:aws:sns:*:*:*",
```
mgyucht, line 113: Ditto here.

mgyucht (commented):
Would be good to get @alexott or @nkvuong's thoughts on this change. Would we want to allow users to opt out of this?

aws/data_aws_unity_catalog_policy.go
nkvuong: these sns/sqs will be created by Databricks, and will follow the pattern `arn:aws:sqs:<region>:<aws_account_id>:csms-*`

aws/data_aws_unity_catalog_policy.go
```diff
@@ -60,6 +60,59 @@ func generateReadContext(ctx context.Context, d *schema.ResourceData, m *common.
 			Resources: []string{kmsArn},
 		})
 	}
+	policy.Statements = append(policy.Statements, &awsIamPolicyStatement{
```
nkvuong, line 63: agree with @mgyucht - these can be left as opt-out/in, as our official documentation mentioned this is optional but strongly recommended

borremosch-db: We're moving from opt-in/out to mandatory (see [PRD: Maximizing coverage of managed file events](https://docs.google.com/document/d/1drQV58fMtINnlEm1TeA1oAOR1kIjHyAQW_xxNRTToxw))

aws/data_aws_unity_catalog_policy.go
borremosch-db: I added ARN resource ID prefix. LMK if this works.

borremosch-db:
@mgyucht thanks for the review. Ideally we would not make these changes opt-in/out as we're moving towards making file events mandatory (see [PRD: Maximizing coverage of managed file events](https://docs.google.com/document/d/1drQV58fMtINnlEm1TeA1oAOR1kIjHyAQW_xxNRTToxw))

alexott (approved):

borremosch-db:
@mgyucht @nkvuong could you have another look at this?

aws/data_aws_unity_catalog_policy.go
```diff
@@ -60,6 +60,62 @@ func generateReadContext(ctx context.Context, d *schema.ResourceData, m *common.
...
+			"sqs:GetQueueAttributes",
+			"sqs:SetQueueAttributes",
+			"sqs:TagQueue",
+			"sqs:ChangeMessageVisibility",
+			"sqs:PurgeQueue",
+		},
+		Resources: []string{
+			fmt.Sprintf("arn:%s:s3:::%s", awsPartition, bucket),
+			fmt.Sprintf("arn:%s:sqs:*:*:csms-*", awsPartition),
+			fmt.Sprintf("arn:%s:sns:*:*:csms-*", awsPartition),
```
mgyucht, line 90: In the other policies, we restrict the resource templates to a single account. Can we do that here as well for consistency, or do we need to allow access to all accounts? (Understanding that this would also require cross-account policy on the target account, but with defence in depth in mind, we shouldn't force people to give broader access than necessary.) Ditto for the other resources. We already accept `awsAccountId` as a parameter.

mgyucht (commented):
I think we can make the requested resources slightly narrower, which would be marginally better from a security stance. Otherwise, seems OK to me.

aws/data_aws_unity_catalog_policy.go
borremosch-db: @mgyucht I considered scoping down to the account, but this is not the case for the S3 permissions in this specific file and I believe we should align the new permissions because they're targeting resources for the same use case.

mgyucht (approved):

PR #4407, opened 2025-01-16
[Internal] Remove tag signing in tagging.py script

mgyucht:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/.github/workflows/tagging.yml
+++ b/.github/workflows/tagging.yml
@@ -35,7 +35,6 @@ jobs:
       - name: Set up Git configuration
         run: |
           git config user.name "Databricks SDK Release Bot"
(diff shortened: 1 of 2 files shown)


.github/workflows/tagging.yml
```diff
@@ -35,7 +35,6 @@ jobs:
       - name: Set up Git configuration
         run: |
           git config user.name "Databricks SDK Release Bot"
-          git config --unset-all http.https://github.com/.extraheader
```
mgyucht, line 38: This was likely unnecessary. We will see when running this workflow again on main.

hectorcast-db (approved):

PR #4410, opened 2025-01-16
[Internal] Use github API to create tag in tagging workflow

mgyucht:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


mgyucht:
This is not needed: the original approach will work once app permissions are updated.

PR #4422, opened 2025-01-21
[Fix] type issue in `databricks_sql_table`

### LINKED ISSUE #4421: [ISSUE] Issue with `databricks_XXX` resource
### Configuration

```hcl
resource "databricks_sql_table" "example_table" {
  name         = local.example_table_name
  catalog_name = var.example_catalog.name
  schema_name  = example_schema.tenant_lookup.name
  table_type   = "MANAGED"
  warehouse_id = data.databricks_sql_warehouse.warehouse.id
  comment      = "Example resource"
  owner        = var.resource_owner_name

  properties = {
    "delta.feature.allowColumnDefaults" = "supported"
  }

  column {
    name     = "column_a"
    type     = "STRING"
    comment  = ""
    nullable = false
  }
  column {
    name     = "column_b"
    type     = "TIMESTAMP DEFAULT current_timestamp()"
    comment  = ""
    nullable = false
  }
}
```

### Expected Behavior

When I apply the first time the table is created successfully. When I plan/apply a second time with the same configuration no changes should be detected.
### Actual Behavior

When I apply the first time the table is created successfully. However, a follow up plan/apply will yield the error `changing the 'type' of an existing column is not supported`.
### Steps to Reproduce

```
terraform version
Terraform v1.10.4
on darwin_arm64
+ provider registry.terraform.io/databricks/databricks v1.63.0
```

### Is it a regression?

Nope

Sure, I'll link a PR in the comments.

ian-norris:
## Changes

Updated type check to only compare the first word in the type string.

Resolves #4421

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


catalog/resource_sql_table.go
```diff
@@ -540,10 +540,11 @@ var columnTypeAliases = map[string]string{
 
 func getColumnType(columnType string) string {
 	caseInsensitiveColumnType := strings.ToLower(columnType)
-	if alias, ok := columnTypeAliases[caseInsensitiveColumnType]; ok {
+	normalizedColumnType := strings.Split(caseInsensitiveColumnType, " ")[0]
```
alexott, line 543: Can we strip spaces from the string? Just to avoid the potential problem when a user puts a space into type?

Can we also add a unit test for this function? It will detect code problems earlier, without need to run a full integration test suite.

alexott (changes requested):

catalog/resource_sql_table.go
ian-norris-ncino: The case I was trying to solve was when the type is something like `TIMESTAMP DEFAULT current_timestamp()`. If I strip the spaces I won't be able to split the type from the rest of the string. Unless I use some sort of string search?

Also, I thought I added unit test in the `resource_sql_table_test.go` file. I am new to Go so if that is the wrong place or wrong kind of test could you give direction on where I should add more?

ian-norris-ncino: After reviewing the tests I see that the test is dependent on cloud credentials. Could you point me to a good location to add unit tests for this?

alexott: To test a signle function that receives string and returns string you don't need to have any credentials. You just need to add a test here: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_sql_table_test.go that can be executed locally without any credential.  You added a test into the acceptance test suite that is executed in a special environment.

alexott:
You also need to think how to handle cases like, if you have a decimal type declared as `DECIMAL(3, 2)` - it will be converted into `DECIMAL(3,` and this is not correct.  So we'll need to have a bit more sophisticated approach.

ian-norris-ncino:
> You also need to think how to handle cases like, if you have a decimal type declared as `DECIMAL(3, 2)` - it will be converted into `DECIMAL(3,` and this is not correct. So we'll need to have a bit more sophisticated approach.

I used regex instead. It looked at this [document](https://docs.databricks.com/en/sql/language-manual/sql-ref-datatypes.html) and it seems all the types use parentheses to signify arguments.

catalog/resource_sql_table.go
alexott: if you add your test to `resource_sql_table_test.go` then that function will be visible as it's inside the same package.

But you need to have two tests:

- unit test of that function alone, in the `resource_sql_table_test.go` file
- integration test as you added it

ian-norris-ncino: Ahh, that is much clearer! I just pushed up new tests and stripped the string of spaces after regex matching the type and possible arguments.

alexott:
Integration tests has failed with the following error:

```
cannot create sql table: cannot execute CREATE TABLE `main`.`sheghbiihgiaj`.`fjdhiaajkejh` (`name` TIMESTAMP DEFAULT current_timestamp() COMMENT 'comment')
  | COMMENT 'this table is managed by terraform'
  | TBLPROPERTIES ('delta.minWriterVersion'='7', 'this'='that', 'delta.feature.columnMapping'='supported', 'delta.feature.invariants'='supported', 'something'='else', 'delta.columnMapping.mode'='name', 'delta.minReaderVersion'='3');: [WRONG_COLUMN_DEFAULTS_FOR_DELTA_FEATURE_NOT_ENABLED] Failed to execute CREATE TABLE command because it assigned a column DEFAULT value,
  | but the corresponding table feature was not enabled. Please retry the command again
  | after executing ALTER TABLE tableName SET
  | TBLPROPERTIES('delta.feature.allowColumnDefaults' = 'supported'). SQLSTATE: 0AKDE
```

ian-norris-ncino:
> Integration tests has failed with the following error:
> 
> ```
> cannot create sql table: cannot execute CREATE TABLE `main`.`sheghbiihgiaj`.`fjdhiaajkejh` (`name` TIMESTAMP DEFAULT current_timestamp() COMMENT 'comment')
>   | COMMENT 'this table is managed by terraform'
>   | TBLPROPERTIES ('delta.minWriterVersion'='7', 'this'='that', 'delta.feature.columnMapping'='supported', 'delta.feature.invariants'='supported', 'something'='else', 'delta.columnMapping.mode'='name', 'delta.minReaderVersion'='3');: [WRONG_COLUMN_DEFAULTS_FOR_DELTA_FEATURE_NOT_ENABLED] Failed to execute CREATE TABLE command because it assigned a column DEFAULT value,
>   | but the corresponding table feature was not enabled. Please retry the command again
>   | after executing ALTER TABLE tableName SET
>   | TBLPROPERTIES('delta.feature.allowColumnDefaults' = 'supported'). SQLSTATE: 0AKDE
> ```

I updated the template generation to include the needed delta flag

ian-norris-ncino:
@alexott Anything else I can do?

alexott:
It looks like there is some syntax error - integration tests are failing with:

```
Error: Missing key/value separator
  | 
  |   on terraform_plugin_test.tf line 23, in resource "databricks_sql_table" "this":
  |   20:     properties         = {
  |   21:         "this"                        = "that"
  |   22:         "something"                   = "else"
  |   23:         "delta.feature.allowColumnDefaults = "supported"
  | 
  | Expected an equals sign ("=") to mark the beginning of the attribute value.
  | If you intended to given an attribute name containing periods or spaces,
  | write the name in quotes to create a string literal.
  | 
  | Error: Invalid multi-line string
  | 
  |   on terraform_plugin_test.tf line 23, in resource "databricks_sql_table" "this":
  |   23:         "delta.feature.allowColumnDefaults = "supported"
  |   24:         "delta.feature.columnMapping" = "supported"
  | 
  | Quoted strings may not be split over multiple lines. To produce a multi-line
  | string, either use the \n escape to represent a newline character or use the
  | "heredoc" multi-line template syntax.
  | 
  | Error: Invalid multi-line string
  | 
  |   on terraform_plugin_test.tf line 24, in resource "databricks_sql_table" "this":
  |   24:         "delta.feature.columnMapping" = "supported"
  |   25:         "delta.feature.invariants"    = "supported"
  | 
```

ian-norris-ncino:
> It looks like there is some syntax error - integration tests are failing with:
> 
> ```
> Error: Missing key/value separator
>   | 
>   |   on terraform_plugin_test.tf line 23, in resource "databricks_sql_table" "this":
>   |   20:     properties         = {
>   |   21:         "this"                        = "that"
>   |   22:         "something"                   = "else"
>   |   23:         "delta.feature.allowColumnDefaults = "supported"
>   | 
>   | Expected an equals sign ("=") to mark the beginning of the attribute value.
>   | If you intended to given an attribute name containing periods or spaces,
>   | write the name in quotes to create a string literal.
>   | 
>   | Error: Invalid multi-line string
>   | 
>   |   on terraform_plugin_test.tf line 23, in resource "databricks_sql_table" "this":
>   |   23:         "delta.feature.allowColumnDefaults = "supported"
>   |   24:         "delta.feature.columnMapping" = "supported"
>   | 
>   | Quoted strings may not be split over multiple lines. To produce a multi-line
>   | string, either use the \n escape to represent a newline character or use the
>   | "heredoc" multi-line template syntax.
>   | 
>   | Error: Invalid multi-line string
>   | 
>   |   on terraform_plugin_test.tf line 24, in resource "databricks_sql_table" "this":
>   |   24:         "delta.feature.columnMapping" = "supported"
>   |   25:         "delta.feature.invariants"    = "supported"
>   | 
> ```

🤦‍♂️ just updated.

alexott (commented):
in general looks good, but need to think on how the change of the type may affect apply/read (I'm not sure if it won't lead to the configuration drift).

integration test is passing, but we need to fix our build to make it overall green

catalog/resource_sql_table.go
```diff
@@ -540,10 +541,14 @@ var columnTypeAliases = map[string]string{
...
-	if alias, ok := columnTypeAliases[caseInsensitiveColumnType]; ok {
+	pattern := regexp.MustCompile(`^\w+(\s*\(.*?\))?`)
+	patternMatched := pattern.FindString(caseInsensitiveColumnType)
+	normalizedColumnType := strings.ReplaceAll(patternMatched, " ", "")
+
+	if alias, ok := columnTypeAliases[normalizedColumnType]; ok {
 		return alias
 	}
-	return caseInsensitiveColumnType
+	return normalizedColumnType
```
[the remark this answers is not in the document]
ian-norris-ncino: Yeah, let me know what you think the best approach is and I'm happy to implement.

ian-norris-ncino:
@alexott Any thoughts on how we can proceed?

mgyucht:
Hi @ian-norris-ncino, thank you for this contribution.

I've gone over this change carefully with @alexott. While I recognize it solves your immediate issue with being able to specify a timestamp column's default value, I am concerned about merging the change as it is.

We need to be able to accurately compare the type of a column as returned by the Tables API with what is specified in a user's config. I have two concerns about this:
1. Using regular expressions to parse the type provided by users has certain downsides. For instance: there are many attributes of columns that can be expressed in the `type`: how can we use regexs to properly identify those?
2. Currently, this PR actually ignores any of those other attributes specified by the user. This causes the diff checking to be incorrect: the provider only checks whether the type matches the type of the column, but it doesn't verify that other annotations are set properly. For example, if the default condition were removed or changed to some other value, the provider would not detect that.

The current implementation is far from ideal, and I'll be the first to admit that, but I don't think this is the right direction to start addressing this. To better support this, I would propose the following pathway.
1. Define a struct that captures a column's definition, including all attributes. It's OK if this struct doesn't include everything from the start, but it needs to be easily extensible for new attributes.
2. Implement methods that deserialize the TF config and the Tables API response into this type.
3. Define a struct that captures the difference between the existing and configured type.
4. Define a method to compute this difference given the existing and configured type.
5. Implement a method to convert a given difference into a set of SQL statements that alter the column to achieve the configured type.

I don't know how easy this will be, as I don't know if there is a complete reference for `type_json`. https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/sql-ref-syntax-aux-describe-table is an initial point, but I think there are many fields that are not documented here (such as a `metadata` field which includes the default values).

ian-norris-ncino:
> Hi @ian-norris-ncino, thank you for this contribution.
> 
> I've gone over this change carefully with @alexott. While I recognize it solves your immediate issue with being able to specify a timestamp column's default value, I am concerned about merging the change as it is.
> 
> We need to be able to accurately compare the type of a column as returned by the Tables API with what is specified in a user's config. I have two concerns about this:
> 
> 1. Using regular expressions to parse the type provided by users has certain downsides. For instance: there are many attributes of columns that can be expressed in the `type`: how can we use regexs to properly identify those?
> 2. Currently, this PR actually ignores any of those other attributes specified by the user. This causes the diff checking to be incorrect: the provider only checks whether the type matches the type of the column, but it doesn't verify that other annotations are set properly. For example, if the default condition were removed or changed to some other value, the provider would not detect that.
> 
> The current implementation is far from ideal, and I'll be the first to admit that, but I don't think this is the right direction to start addressing this. To better support this, I would propose the following pathway.
> 
> 1. Define a struct that captures a column's definition, including all attributes. It's OK if this struct doesn't include everything from the start, but it needs to be easily extensible for new attributes.
> 2. Implement methods that deserialize the TF config and the Tables API response into this type.
> 3. Define a struct that captures the difference between the existing and configured type.
> 4. Define a method to compute this difference given the existing and configured type.
> 5. Implement a method to convert a given difference into a set of SQL statements that alter the column to achieve the configured type.
> 
> I don't know how easy this will be, as I don't know if there is a complete reference for `type_json`. https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/sql-ref-syntax-aux-describe-table is an initial point, but I think there are many fields that are not documented here (such as a `metadata` field which includes the default values).

Thanks @mgyucht and @alexott for thinking deeply about this. I've starting getting deeper into an implementation and it comes with many complexities. Mainly around parsing the user provided type string into a common structure with the get table API structure. I wonder if it makes sense to add additional keys to the resource for default values, precision, scales, and interval types as to match with the get api? An example would the look like
```tf
column {
  name     = "updated_at"
  type     = "timestamp"
  default  = "current_timestamp()"
  comment  = ""
  nullable = false
}
column {
  name           = "value"
  type           = "decimal"
  type_precision = 10
  type_scale     = 0
  comment        = ""
  nullable       = false
}
```
There would be some considerations on parsing the type_json metadata but this would make handling the user input much easier. What do you think?

svyotov-lg:
This continues to be a problem, is there an ETA?

ian-norris:
> This continues to be a problem, is there an ETA?

I stopped working on this after I didn't receive a reply

PR #4433, opened 2025-01-24
Update config to be optional and computed

arpitjasa-db:
## Changes

Update config to be optional and computed

## Tests

- [X] `make test` run locally
- [X] using Go SDK
- [X] using TF Plugin Framework


mgyucht:
Thanks @arpitjasa-db, I'm incorporating this into the next Go SDK bump: https://github.com/databricks/terraform-provider-databricks/pull/4430.

Note that I'm taking a slightly different approach with it: we won't treat it as a computed field, but rather recreate the model serving endpoint if the user removes the config. This ensures that the value for the config attribute completely determines the model's configuration; otherwise, an empty config could mean actually no config or that there is a config but the customer simply removed it from their TF template. Furthermore, it wouldn't be possible in a single apply to reset a model serving endpoint to a config-less state.

PR #4446, opened 2025-01-29
[Fix] Make removing `config` a no-op for `databricks_model_serving`

mgyucht:
## Changes
After discussing with @arpitjasa-db, we need to allow users to remove a config specified in the model serving endpoint without recreating the endpoint. To achieve this, the config will be considered computed: if the config is specified, it will be used when creating or updating. If not, the config is allowed to have any value and will not result in a diff, even if the state version is different than the live model serving endpoint.

## Tests

Modified the newly added unit test to verify that removing config did not result in any diff.


tanmay-db (approved):

arpitjasa-db (approved):
Thanks @mgyucht !

alexott (approved):

PR #4448, opened 2025-01-29
[Internal] Use gofumpt instead of gofmt

mgyucht:
## Changes
Gofumpt is a stricter version of gofmt. This is used in the `cli` repo, and we would like to adopt it in the TF provider as well.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott:
It would be better to open a new PR given a big number of conflicts

PR #4456, opened 2025-01-31
[Feature] Support `run_as` in pipelines

vsamoilov:
## Changes

Resolves https://github.com/databricks/terraform-provider-databricks/issues/4398. No updates to docs or changelog, because feature is in private preview
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

NO_CHANGELOG=true

### DIFF
--- a/internal/acceptance/init.go
+++ b/internal/acceptance/init.go
@@ -448,3 +448,28 @@ func LoadDebugEnvIfRunsFromIDE(t *testing.T, key string) {
 		os.Setenv(k, v)
 	}
 }
+
+func isAuthedAsWorkspaceServicePrincipal(ctx context.Context) (bool, error) {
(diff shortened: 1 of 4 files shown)


alexott:
What is the difference from https://github.com/databricks/terraform-provider-databricks/pull/4445 ?

alexott:
Ok. So we can close #4445?

vsamoilov:
> Ok. So we can close #4445?

Yes, I think the current PR is the one, that will be merged

mgyucht (approved):

cameronattard:
FYI this appears to cause issues in accounts that haven't been signed up to the private preview:

```
Error: cannot update pipeline: Configuring run as in pipelines is not supported.
```

PR #4499, opened 2025-02-13
[Internal] Don't panic if a schema attribute is already set to desired value

alexott:
## Changes

Imo, It's not critical to set the same parameter of an attribute to the same value

NO_CHANGELOG=true

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/tfschema/single_nested_attribute.go
+++ b/internal/providers/pluginfw/tfschema/single_nested_attribute.go
@@ -2,6 +2,7 @@ package tfschema
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/databricks/terraform-provider-databricks/common"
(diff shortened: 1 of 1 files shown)


nkvuong (approved):
@alexott do we need to update this for other attribute builder as well?

alexott:
Let discuss it with @mgyucht

mgyucht:
Sorry I missed this. The reason we added panics here is to make sure that the extra customizations added to the schema are purely additive. Otherwise, you may think you are undoing a customization by removing it from ApplySchemaCustomizations (e.g. removing SetRequired would make something optional), but actually it is still there because it is the default behavior (e.g. in that case, the field may still be marked as required). This prevents this case from happening.

alexott:
Let's close it then.

PR #4505, opened 2025-02-14
[Fix] Remove configuration drift when configuring `databricks_connection` to built-in Hive Metastore

alexott:
## Changes

When configuring built-in Hive Metastore, the backend returns a number of additional options.  But it's hard to suppress them as they are valid when `builtin = "false"`, and each option is evaluated separately.  So it's easier just to remove them before saving to the state.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


catalog/resource_connection.go
```diff
@@ -100,13 +100,21 @@ func ResourceConnection() common.Resource {
 			if err != nil {
 				return err
 			}
+			if conn.Options == nil {
+				conn.Options = map[string]string{}
+			}
+			// remove not necessary parameters for builtin HMS
+			if val, exists := conn.Options["builtin"]; exists && val == "true" {
```
mgyucht, line 107: Where is this `builtin` field documented? I don't see it on the linked website.

mgyucht (commented):
Is this behavior documented anywhere? Trying to see where this is specified.

catalog/resource_connection.go
alexott: yeah, that's a major problem with API spec - all these fields are buried under the docs for specific connection type.  For built-in hive metastore it's: https://docs.databricks.com/en/data-governance/unity-catalog/hms-federation/hms-federation-internal.html#language-SQL

Frankly speaking, I would prefer to have something like it's done for model serving and external models - each model type has its own structures with well defined what is computed/required/optional

catalog/resource_connection.go
```diff
@@ -87,13 +87,21 @@ func ResourceConnection() common.Resource {
...
+				log.Printf("[DEBUG] Removing not necessary parameters for builtin HMS")
+				delete(conn.Options, "host")
+				delete(conn.Options, "port")
+				delete(conn.Options, "home_workspace_id")
+				delete(conn.Options, "database")
+			}
 			// We need to preserve original sensitive options as API doesn't return them
 			var cOrig catalog.CreateConnection
 			common.DataToStructPointer(d, s, &cOrig)
 			// If there are no options returned, need to initialize the map
```
nkvuong, line 104: this comment should move with the code block as well

docs/resources/connection.md
```diff
@@ -61,12 +61,25 @@ resource "databricks_connection" "bigquery" {
...
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
 - `name` - Name of the Connection.
-- `connection_type` - Connection type. `BIGQUERY` `MYSQL` `POSTGRESQL` `SNOWFLAKE` `REDSHIFT` `SQLDW` `SQLSERVER`, `SALESFORCE` or `DATABRICKS` are supported. [Up-to-date list of connection type supported](https://docs.databricks.com/query-federation/index.html#supported-data-sources)
+- `connection_type` - Connection type. `BIGQUERY` `MYSQL` `POSTGRESQL` `SNOWFLAKE` `REDSHIFT` `SQLDW` `SQLSERVER`, `SALESFORCE`, `HIVE_METASTORE`, `GLUE`, `TERADATA`, `ORACLE` or `DATABRICKS` are supported. [Up-to-date list of connection type supported is in the documentation](https://docs.databricks.com/query-federation/index.html#supported-data-sources)
```
nkvuong, line 82: the link is too long otherwise?
```suggestion
- `connection_type` - Connection type. `BIGQUERY` `MYSQL` `POSTGRESQL` `SNOWFLAKE` `REDSHIFT` `SQLDW` `SQLSERVER`, `SALESFORCE`, `HIVE_METASTORE`, `GLUE`, `TERADATA`, `ORACLE` or `DATABRICKS` are supported. Up-to-date list of connection type supported is in the [documentation](https://docs.databricks.com/query-federation/index.html#supported-data-sources)
```

nkvuong (approved):
just a couple of comments on doc

catalog/resource_connection.go
nkvuong: should we add it to the doc as well? a general complaint is that customers don't know which `option` to provide

PR #4506, opened 2025-02-17
[Doc] Improve documentation for databricks_cluster and databricks_clusters

mgyucht:
## Changes
We get periodic questions about how to use `databricks_cluster` and `databricks_clusters`, both around performance and errors with multiple clusters with the same name. This PR adds some examples and extra documentation about how to use these resources.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Documentation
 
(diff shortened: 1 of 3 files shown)


nkvuong (approved):

mgyucht:
As this is docs-only, I will merge now.

PR #4508, opened 2025-02-17
[Dependency] Bump to 1.24.0

mgyucht:
## Changes
Let's upgrade the version of go we use to latest, so we can take advantages of new language features like range expressions and the iter library.

## Tests

Existing tests should cover this.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -174,7 +174,7 @@ func (a CommandsAPI) waitForCommandFinished(commandID, contextID, clusterID stri
 			return nil
 		}
 		log.Printf("[DEBUG] Command is in %s state", commandInfo.Status)
-		return resource.RetryableError(fmt.Errorf(commandInfo.Status))
+		return resource.RetryableError(fmt.Errorf("%s", commandInfo.Status))
 	})
 }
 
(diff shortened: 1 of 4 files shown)


alexott:
Should we update corresponding github actions?

mgyucht:
We shouldn't need to: I recently updated them to use the version defined in the `go.mod`, so they will automatically update.

commands/commands.go
```diff
@@ -174,7 +174,7 @@ func (a CommandsAPI) waitForCommandFinished(commandID, contextID, clusterID stri
 			return nil
 		}
 		log.Printf("[DEBUG] Command is in %s state", commandInfo.Status)
-		return resource.RetryableError(fmt.Errorf(commandInfo.Status))
+		return resource.RetryableError(fmt.Errorf("%s", commandInfo.Status))
```
alexott, line 177: it's better to use `errors.New`

mws/resource_mws_workspaces.go
```diff
@@ -217,7 +217,7 @@ func (a WorkspacesAPI) verifyWorkspaceReachable(ws Workspace) *resource.RetryErr
 
 func (a WorkspacesAPI) explainWorkspaceFailure(ws Workspace) error {
 	if ws.NetworkID == "" {
-		return fmt.Errorf(ws.WorkspaceStatusMessage)
+		return fmt.Errorf("%s", ws.WorkspaceStatusMessage)
```
alexott, line 220: Same here

alexott (approved):
Small nits

PR #4520, opened 2025-02-24
[Fix] Delete `databricks_sql_endpoint` that failed to start

### LINKED ISSUE #4496: [ISSUE] Issue with `databricks_sql_endpoint` resource ("CLASSIC" type)
### Configuration

We just want to deploy the "CLASSIC" sql warehouse by running `terraform apply` the following:
```tf
  provider                  = databricks.workspace
  name                      = "${var.databricks_workspace_name}-default-sql"
  cluster_size              = "2X-Small"
  auto_stop_mins            = 10
  enable_serverless_compute = "false"
  warehouse_type            = "CLASSIC"
}

```

### Expected Behavior

The  "CLASSIC" sql warehouse is created normally. ("Serverless" SQL warehouse can be created normally.)

### Actual Behavior

1st apply fails after 20 mins:

```tf
  ╷
  │ Error: cannot create sql endpoint: failed waiting for warehouse to start: timed out: Clusters are failing to launch. Cluster launch will be retried.
  │ 
  │   with databricks_sql_endpoint.this,
  │   on main.tf line 1, in resource "databricks_sql_endpoint" "this":
  │    1: resource "databricks_sql_endpoint" "this" {
  │ 
```

2nd apply fails with the following error:

```log
 ╷
 │ Error: cannot create sql endpoint: failed creating warehouse: SQL warehouse with name `<my-workspace>-default-sql` already exists
 │ 
 │   with databricks_sql_endpoint.this,
 │   on main.tf line 1, in resource "databricks_sql_endpoint" "this":
 │    1: resource "databricks_sql_endpoint" "this" {
 │ 
```

### Steps to Reproduce

just run `terraform apply` for the TF file above.

- 1.65.1

nkvuong:
## Changes
- If SQL warehouse is created but failed to start, we need to delete it to avoid failure in the next apply. Resolves #4496

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/sql/resource_sql_endpoint.go
+++ b/sql/resource_sql_endpoint.go
@@ -103,6 +103,11 @@ func ResourceSqlEndpoint() common.Resource {
 			}
 			resp, err := wait.Get()
 			if err != nil {
+				// Rollback by deleting the warehouse
+				rollbackErr := w.Warehouses.DeleteById(ctx, wait.Id)
+				if rollbackErr != nil {
+					return fmt.Errorf("failed deleting warehouse: %w", rollbackErr)
+				}
 				return fmt.Errorf("failed waiting for warehouse to start: %w", err)
 			}
 			d.SetId(resp.Id)
--- a/sql/resource_sql_endpoint.go
+++ b/sql/resource_sql_endpoint.go
@@ -106,7 +106,7 @@ func ResourceSqlEndpoint() common.Resource {
 				// Rollback by deleting the warehouse
 				rollbackErr := w.Warehouses.DeleteById(ctx, wait.Id)
 				if rollbackErr != nil {
-					return fmt.Errorf("failed deleting warehouse: %w", rollbackErr)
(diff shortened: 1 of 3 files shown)


sql/resource_sql_endpoint.go
```diff
@@ -103,6 +103,11 @@ func ResourceSqlEndpoint() common.Resource {
 			}
 			resp, err := wait.Get()
 			if err != nil {
+				// Rollback by deleting the warehouse
+				rollbackErr := w.Warehouses.DeleteById(ctx, wait.Id)
+				if rollbackErr != nil {
+					return fmt.Errorf("failed deleting warehouse: %w", rollbackErr)
```
alexott, line 109: This will hide an original error. Maybe we can put it here as well?

alexott (approved):

sql/resource_sql_endpoint.go
nkvuong: agreed, updated the error message :)

PR #4531, opened 2025-02-26
[Feature] Mark GKE-related fields for `databricks_mws_workspaces` and `databricks_mws_networks` as deprecated

mgyucht:
## Changes
Databricks in GCP is transitioning from using GKE-based clusters to using GCE-based clusters. In this transition, the GKE-related fields for all resources are being deprecated and will be removed at a later date. The affected fields are:

* `databricks_mws_networks`: `gcp_network_info.pod_ip_range_name`, `gcp_network_info.service_ip_range_name`
* `databricks_mws_workspaces`: `gke_config`, `gcp_managed_network_config.gke_cluster_pod_ip_range`, `gcp_managed_network_config.gke_cluster_service_ip_range`

These fields are marked as deprecated, and users will see a warning message when their configuration includes any of these fields. They will still be included in API requests, but they will be ignored in the API itself for new networks and workspaces. The underlying networks and workspaces will be GCE-compatible.

Existing resources can simply remove these deprecated fields. This will not trigger any recreation of either resource.

As part of this, I have refactored the `internal/docs` module to simplify references to the Terraform Registry documentation pages in our deprecation and log messages. I've introduced a `DocOptions` struct that encapsulates how to construct a link to a specific reference page and, optionally, to an anchor/fragment in that page. We use this in this PR to reference the GCP network and workspace setup guide.

## Tests

Added integration tests for `databricks_mws_networks` and `databricks_mws_workspaces` to verify that it is safe to remove GKE-based configs.

## To-do

- [ ] Get the documentation link for the GCE transition project. (I'll add this to the guide as a follow-up PR.)
- [x] Verify that it is possible to create networks/workspaces without GKE fields. Add to the start of this test a step that tests create network without these fields.


docs/resources/mws_networks.md
```diff
@@ -26,9 +26,9 @@ Please follow this [complete runnable example](../guides/aws-workspace.md) with
 
 Use this resource to [configure VPC](https://docs.gcp.databricks.com/administration-guide/cloud-configurations/gcp/customer-managed-vpc.html) & subnet for new workspaces within GCP. It is essential to understand that this will require you to configure your provider separately for the multiple workspaces resources.
 
-* Databricks must have access to a subnet in the same region as the workspace, of which IP range will be used to allocate your workspace’s GKE cluster nodes.
+* Databricks must have access to a subnet in the same region as the workspace, of which IP range will be used to allocate your workspace’s GCE cluster nodes.
 * The subnet must have a netmask between /29 and /9.
-* Databricks must have access to 2 secondary IP ranges, one between /21 to /9 for workspace’s GKE cluster pods, and one between /27 to /16 for workspace’s GKE cluster services.
+* Databricks must have access to 2 secondary IP ranges, one between /21 to /9 for cluster pods, and one between /27 to /16 for cluster services.
```
alexott, line 31: We need only one subnet, no secondary IP ranges

alexott (changes requested):
in general looks good, we need to adjust docs

docs/resources/mws_networks.md
mgyucht: Ah OK, thanks, will remove this.

alexott (approved):

PR #4547, opened 2025-03-04
[Fix] handle auto-enabled errors with`databricks_system_schema`

### LINKED ISSUE #4539: [ISSUE] Failures for Auto-Enabled System Tables
While using Terraform with Databricks, we noticed state changes related to system tables during the terraform plan step. The following system tables are affected:

- databricks_system_schema.access_logs is tainted and requires replacement.
- databricks_system_schema.billing_logs is marked for creation.
- databricks_system_schema.compute_logs is tainted and requires replacement.
- databricks_system_schema.lakeflow_logs is tainted and requires replacement.

However, during the `terraform apply` step, encounter the following error, which prevents the deployment of new changes:

```
Error: cannot create system schema: billing system schema can only be enabled by Databricks.

 with databricks_system_schema.billing_logs,
 on main.tf line 226, in resource "databricks_system_schema" "billing_logs":
 226: resource "databricks_system_schema" "billing_logs" {

Error: Process completed with exit code 1.
```

### Root Cause:

The `billing` schema is now enabled by default in Databricks, meaning that any resource block attempting to create it manually will result in an error.

### Proposed Solution:

Documentation Update: Clearly state in the documentation that certain system schemas (such as `billing`) may be auto-enabled and should not be manually declared in Terraform configurations.

Error Handling Improvement: Instead of failing the terraform apply process, introduce logic that detects and ignores errors when certain system tables are already enabled by default.

Terraform Provider Enhancement: Modify the Databricks Terraform provider to automatically detect auto-enabled system schemas and exclude them from the state file or ignore changes during `terraform apply`.

nkvuong:
## Changes
- Handle new error message `billing system schema can only be enabled by Databricks`. Resolves #4539
- Update documentation with warnings about auto-enabled system schemas.
- Enable system schema integration test on GCP

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/docs/resources/system_schema.md
+++ b/docs/resources/system_schema.md
@@ -9,6 +9,8 @@ subcategory: "Unity Catalog"
 
 Manages system tables enablement. System tables are a Databricks-hosted analytical store of your account’s operational data. System tables can be used for historical observability across your account. System tables must be enabled by an account admin.
 
+Certain system schemas (such as `billing`) may be auto-enabled once GA and should not be manually declared in Terraform configurations.
+
 ## Example Usage
 
 Enable the system schema `access`
--- a/docs/resources/system_schema.md
+++ b/docs/resources/system_schema.md
@@ -7,9 +7,9 @@ subcategory: "Unity Catalog"
 
 -> This resource can only be used with a workspace-level provider!
 
-Manages system tables enablement. System tables are a Databricks-hosted analytical store of your account’s operational data. System tables can be used for historical observability across your account. System tables must be enabled by an account admin.
+-> Certain system schemas (such as `billing`) may be auto-enabled once GA and should not be manually declared in Terraform configurations.
 
-Certain system schemas (such as `billing`) may be auto-enabled once GA and should not be manually declared in Terraform configurations.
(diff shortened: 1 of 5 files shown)


alexott:
@nkvuong how it will handle destroy? Will it fail trying to disable the schema?

docs/resources/system_schema.md
```diff
@@ -9,6 +9,8 @@ subcategory: "Unity Catalog"
 
 Manages system tables enablement. System tables are a Databricks-hosted analytical store of your account’s operational data. System tables can be used for historical observability across your account. System tables must be enabled by an account admin.
 
+Certain system schemas (such as `billing`) may be auto-enabled once GA and should not be manually declared in Terraform configurations.
```
alexott, line 12: Should we put this into the callout?

nkvuong:
> @nkvuong how it will handle destroy? Will it fail trying to disable the schema?

@alexott our API actually succeeds 😕

docs/resources/system_schema.md
nkvuong: done

alexott:
@nkvuong maybe we can add a computed flag that specifies that it's auto-enabled? And then in the `Delete` skip the deletion?

alexott (approved):

PR #4569, opened 2025-03-11
[Fix] Update `databricks_mlflow_experiment` resource to include tags

tanmay-db:
## Changes

This resource is outdated as it is maintained manually. PR adds the tags field which is present in OpenAPI spec and removes the description field which isn't present. 

The right way to solve this is to migrate this to Go SDK however since the bandwidth is low for this quarter to include that work and this is required in CLI, adding a quick fix until this is migrated to autogeneration framework. 

## Tests

Added unit test
- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New Features and Improvements
 
 ### Bug Fixes
+* [Fix] Update databricks_mlflow_experiment resource to include tags ([#4569](https://github.com/databricks/terraform-provider-databricks/pull/4569))
 
 ### Documentation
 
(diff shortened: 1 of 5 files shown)


NEXT_CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New Features and Improvements
 
 ### Bug Fixes
+* [Fix] Update databricks_mlflow_experiment resource to include tags ([#4569](https://github.com/databricks/terraform-provider-databricks/pull/4569))
```
alexott, line 8: I would say that this "new feature" not a fix

alexott (approved):
in general good, just not sure that this is a fix

nkvuong:
Superseded by #4606

PR #4572, opened 2025-03-12
[Fix] Recreate `databricks_access_control_rule_set` when the `name` changes

### LINKED ISSUE #4565: [ISSUE] databricks_access_control_rule_set doesn't update name and id in state when name is changed
### Configuration

```hcl
resource "databricks_access_control_rule_set" "this" {

  name = "accounts/${data.databricks_current_config.mws.account_id}/servicePrincipals/<service principal>/ruleSets/default"
  grant_rules {
    principals = [databricks_service_principal.this.acl_principal_id]
    role       = "roles/servicePrincipal.user"
  }
}
```
When changing the `name` argument terraform shows the resource being updated but both the `name` as well as the `id` attributes in the state remain unchanged resulting in errors in case the resource referenced in `name` no longer exists
```
  │ Error: cannot read access control rule set: ServicePrincipal <old service principal id> not found
  │ 
  │   with databricks_access_control_rule_set.this,
  │   on databricks.tf line 52, in resource "databricks_access_control_rule_set" "this":
  │   52: resource "databricks_access_control_rule_set" "this" {
```

### Expected Behavior
The state should be correctly updated

### Actual Behavior
The state isn't updated correctly, resulting in breakage.

```
OpenTofu v1.9.0-dev
on linux_amd64
+ provider registry.opentofu.org/databricks/databricks v1.68.0
```

It seems like the same issue was described here https://github.com/databricks/terraform-provider-databricks/issues/3369 but incorrectly dismissed.

Looking at the code, `update` doesn't seem to update/set the `id` https://github.com/databricks/terraform-provider-databricks/blob/f50e02dfeb6541445d2b9a7887d9e646fcf7bce6/permissions/resource_access_control_rule_set.go#L111 like `create` does https://github.com/databricks/terraform-provider-databricks/blob/f50e02dfeb6541445d2b9a7887d9e646fcf7bce6/permissions/resource_access_control_rule_set.go#L98
Also `read` seems to be using the `id` as the value for the `name` https://github.com/databricks/terraform-provider-databricks/blob/f50e02dfeb6541445d2b9a7887d9e646fcf7bce6/permissions/resource_access_control_rule_set.go#L103 which would align with the issue we're seeing.

alexott:
## Changes

The corresponding API doesn't allow changes in the `name`, so we need to recreate the resource.

Resolves #4565

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


nkvuong (approved):

kamliolfa1:
@alexott When you say:
"Update for the name isn't supported.."
Does this mean that if we want to rename a Service Principal or delete it, we need to proceed in the following order?

1. Delete the databricks_access_control_rule_set
2. Delete the Service Principal with the old name
3. Create the Service Principal with the new name
4. Recreate the databricks_access_control_rule_set with the new Service Principal
--
Also, would adding a depends_on resolve the issue ?

PR #4575, opened 2025-03-12
[Fix] Support updating all attributes of `databricks_model_serving`

### LINKED ISSUE #4038: [ISSUE] Issue with `databricks_model_serving` resource (tags not updating)
### Configuration

Using TF to manage serving endpoints, and the tags are somehow being changed on the Databricks side. Hence, every time you run a plan/apply, the state of the tags changes. 
<!--```hcl
```-->

### Expected Behavior
<!-- What should have happened? -->
Tags should update as per the changes pushed/applied via the TF provider every time

### Actual Behavior
<!-- What actually happened? -->
Tags look normal when applied but if you try to add a tag or delete a tag - you can no longer manage it via Terraform after first ‘apply’. It just acts as a one-way setting.

### Steps to Reproduce
<!-- Please list the steps required to reproduce the issue, for example:
1. `terraform apply`-->
Create a ‘databricks_model_serving.generic_endpoint' and try updating the tags for the same

<!-- Please paste the output of `terraform version`. If version of `databricks` provider is not the latest (https://github.com/databricks/terraform-provider-databricks/releases), please make sure to use the latest one. -->
The version being used is (1.51.0)

### LINKED ISSUE #4564: [ISSUE] Issue with `databricks_model_serving` resource
The Tags don't show up in the plan when added from UI or when added from code then don't show up in UI

### Configuration

resource "databricks_model_serving" "thyroid_cancer_risk_model" {
  name     = "thyroid_cancer_risk_model"
  config {
    served_entities {
      name           = "thyroid_cancer_risk"
      entity_name    = "ml_catalog.model_serving.thyroid_cancer_risk"
      entity_version = "22"
      workload_size  = "Small"
      workload_type  = "CPU"
      scale_to_zero_enabled = true
      environment_vars = {
        "ENABLE_MLFLOW_TRACING" : "true"
      }
    }
    traffic_config {
      routes {
        served_model_name  = "thyroid_cancer_risk"
        traffic_percentage = 100
      }
    }
    auto_capture_config {
      catalog_name      = "ml_catalog"
      schema_name       = "model_serving"
      table_name_prefix = "thyroid_cancer_risk"
      enabled           = true
    }
  }
  route_optimized = true

  tags {
      key   = tags.key
      value = tags.value
  }
}

### Expected Behavior
tags should have been added on UI

### Actual Behavior

Image

### Steps to Reproduce
Either add some tags on UI and run the terraform speculative plan
Or add tags in the TF config and apply the config

TF provider - 1.7.5
Databricks provider - 1.68.0

### Is it a regression?
Not sure

### LINKED ISSUE #4589: [ISSUE] Managing foundation models with `databricks_model_serving` resource
### Configuration

```hcl
resource "databricks_model_serving" "databricks_foundatation_model_endpoint" {
  name = "databricks-meta-llama-3-3-70b-instruct"

  rate_limits {
    key            = "endpoint"
    calls          = 2
    renewal_period = "minute"
  }

  rate_limits {
    key            = "users"
    calls          = 1
    renewal_period = "minute"
  }
}
```

### Expected Behavior
- `databricks_model_serving` to be able to manage rate_limits of the foundational model

### Actual Behavior
- Cannot detect configuration drift
- Setting values in rate_limits do not take effect

### Steps to Reproduce

1. Create the configuration above
2. Import the resource because we cannot create these precreated models
3. Perform terraform plan or terraform apply
4. Plan and apply are successful but rate_limits not applied

Tested with `1.65.1` and `1.70.0`

### Is it a regression?

I assume no. I couldn't find more information how to manage foundation models in an appropriate manner.

### Debug Output
- Not applicable 

### Important Factoids
- According to documentation `rate_limits` should be supported.

- I can't do it.

mgyucht:
## Changes
Currently, it is only possible to update the config for `databricks_model_serving` resources. The Serving Endpoints API exposes separate APIs for updating different elements of serving endpoints, including the config, tags, rate limits, and AI gateway configuration. Because of this, we need to call each API separately when requested by a user.

This PR adds support for updating these attributes. The provider will use `ResourceData`'s `HasChanges()` method to determine whether each corresponding update method needs to be called.

From what I can tell, it is not possible to update the route-optimized setting for model serving endpoints, so I've marked it as `ForceNew` to ensure endpoints are recreated appropriately if a user requests to enable or disable route optimization.

Resolves #4564, Resolves #4038, Resolves #4589

## Tests
- [x] Added tags to the `TestAccModelServing` test, verifying that tags can be provided at creation, and that in update tags can be removed, updated, and added.
- [ ] Added AI gateway to the above test.
- [ ] Added rate limits to the above test.


serving/resource_model_serving.go
```diff
@@ -13,6 +14,85 @@ import (
...
+		}
+	}
+	if _, err := w.ServingEndpoints.Patch(ctx, req); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Update the rate limit configuration for a model serving endpoint.
+func updateRateLimits(ctx context.Context, w *databricks.WorkspaceClient, name string, newRateLimits []serving.RateLimit, d *schema.ResourceData) error {
```
nkvuong, line 76: this endpoint is now deprecated, you can remove it
`This API is deprecated for your Foundation Model endpoint. Please use AI Gateway to manage rate limits.`

renaudhartert-db (approved):

serving/resource_model_serving.go
alexott: @mgyucht - see Vuong's comment

VOVELEE:
@mgyucht, @alexott : Do you think you can merge this so eventually the bug is not there with the next release :)

PR #4578, opened 2025-03-14
[Feature] Add `databricks_disable_legacy_access_setting` resource to disable legacy access methods

willwwt:
## Changes

Add the TF support for setting DisableLegacyAccess.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


settings/resource_disable_legacy_access_setting.go
```diff
@@ -0,0 +1,39 @@
...
+		t.SettingName = "disable_legacy_access"
+		res, err := w.Settings.DisableLegacyAccess().Update(ctx, settings.UpdateDisableLegacyAccessRequest{
+			AllowMissing: true,
+			Setting:      t,
+			FieldMask:    "disable_legacy_access.value",
+		})
+		if err != nil {
+			return "", err
+		}
+		return res.Etag, err
```
alexott, line 28: it would be easier to understand if we write this as `return res.Etag, nil`

settings/resource_disable_legacy_access_setting_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/service/settings"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWorkDisableLegacyAccessSetting(t *testing.T) {
```
alexott, line 17: This is really an acceptance test - it should be named `TestAcc`.

alexott (changes requested):
We need to make changes:

1. Add an entry into the `NEXT_CHANGELOG.md` file
2. Add unit test that will check that API is called in the correct order (like this: https://github.com/databricks/terraform-provider-databricks/blob/main/settings/resource_aibi_dashboard_embedding_access_policy_setting_test.go)
3. Rename existing `settings/resource_disable_legacy_access_setting_test.go` into `settings/disable_legacy_access_setting_test.go` to make it acceptance test (and fix the function name)c

alexott:
Can you please add an integration test as well?

alexott:
also, please run `make fmt` on the source code

willwwt:
> Can you please add an integration test as well?

could you please point me an example? Is it identical to my previous accept test?

PR #4579, opened 2025-03-14
fix: support updating catalog name

### LINKED ISSUE #3818: [ISSUE] Issue with `databricks_catalog` resource
### Configuration

before running the code, we had someone manually create a catalog from a delta share called "prod2_share". the goal here is 1) import that catalog into terraform then (2) rename it then (3) create a workspace binding for it

```hcl
resource "databricks_catalog" "prod2_share" {
  name           = "prod2"
  share_name     = "prod-share"
  isolation_mode = "ISOLATED"
  provider_name  = "aws:us-east-1:xxxxx-xxxx-xxx-xxx-xxxxxxxxx"
}

import {

  to = databricks_catalog.prod2_share
  id = "prod2_share"
}

resource "databricks_catalog_workspace_binding" "prod2_shared" {

  securable_name = databricks_catalog.prod2_share.name
  workspace_id   = "XXXXXX"
}
```

plan from the above showed....

```
  # databricks_catalog.prod will be updated in-place
  # (imported from "prod2_share")
!   resource "databricks_catalog" "prod2_share" {
+       force_destroy  = false
        id             = "prod2_share"
        isolation_mode = "ISOLATED"
        metastore_id   = "xxxxxxx-xxx-xxx-xxxx-xxxxxx"
!       name           = "prod2_share" -> "prod2"
        owner          = "xxxxxxx-xxxx-xxxx-xxxx-xxxx"
        provider_name  = "aws:us-east-1:xxxxx-xxxx-xxx-xxx-xxxxxxxxx"
        share_name     ="prod-share"
    }

  # databricks_catalog_workspace_binding.prod2_shared will be created
  + resource "databricks_catalog_workspace_binding" "prod2_shared" {
      + binding_type   = "BINDING_TYPE_READ_WRITE"
      + id             = (known after apply)
      + securable_name = "prod2"
      + securable_type = "catalog"
      + workspace_id   = xxxxxxxxx
    }

```

### Expected Behavior
- on terraform apply, the above plan works as intended, it renames the catalog and all is good

### Actual Behavior
- can see in apply it thinks it did the modification
```
databricks_catalog.prod2_share[0]: Modifying... [id=prod2_share]
databricks_catalog.prod2_share[0]: Modifications complete after 0s [id=prod2_share]
```
- then throws this error...
```
│ Error: Provider produced inconsistent final plan
│ 
╷
│ When expanding the plan for
│ databricks_catalog_workspace_binding.prod2_shared[0] to include new values
│ learned so far during apply, provider
│ "registry.terraform.io/databricks/databricks" produced an invalid new value
│ for .securable_name: was cty.StringVal("prod2"), but now
│ cty.StringVal("prod2_share").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.

```
- after words i checked if the imported catalog was renamed, and it was NOT. This i assume is the root issue here, which is renaming it did nothing. 
- obviously i should of imported first, then done follow up PRs to rename/workspace bindings, but not sure if that would fix, considering it didnt rename the catalog despite it thinking it did. This also resulted in a bad tf state, where it thinks the catalog has a name now that it does not.

### Steps to Reproduce

- in one apply, import a delta share catalog, rename it, and create a workspace binding from it 

`1.49`

eric-carlsson:
## Changes

Fixes #3818

This PR adds missing logic for handling the renaming of Catalogs.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

Have tested against live Databricks deployment and works as expected.


alexott:
The problem is not in the renaming itself (btw, you forgot to set `id` in the update method. The problem is with the downstream objects that need to be updated as well, and not all of them support renames - like, Vector Search indexes will be tainted and this will lead to their recreation.

To support renaming, UC catalogs (and schemas, etc.) must use some stable ID (like, UUID generated during the creation).

eric-carlsson:
> btw, you forgot to set `id` in the update method.

`id` was already set from the returned object, am I missing it somewhere else?

https://github.com/databricks/terraform-provider-databricks/blob/afbaf075f7dade1ab7a94bef79881c43009ab3ca/catalog/resource_catalog.go#L181-L183

> The problem is with the downstream objects that need to be updated as well, and not all of them support renames - like, Vector Search indexes will be tainted and this will lead to their recreation.

I'm aware that renaming catalogs can have side-effects, but I think that's a concern for the user of the provider and not the provider itself. By not supporting this behavior, we are effectively forced to manually manipulate the state if we want to move catalogs, which is arguably even worse.

Most of the downstream objects already mark the `catalog_name` field with `forceNew`. Shouldn't this be sufficient to inform the user that renaming the catalog can be a destructive action? If the user is not ok with this, they can add  `removed` and `import` blocks for the downstream objects to "manually" migrate their IDs. Regardless, it's left up to the user to decide.

PR #4581, opened 2025-03-17
Aggressive caching for installations with large number of users and groups

rusabd:
## Changes

On the installations with large number of users the provider creates a large number of API calls to the Databricks API leading to slow deployments. Specifically, it runs a search on for each user as a separate API call. The provider also requests all members for each group membership check.

In this change users data provider and group member resource use expiring caches - in our case we halved the default deployment times on existing installation and eliminated Databricks API throttling.

I am also looking for some guidance on how to make this change more generic.

## Tests

We run this provider in production with about 800 users and ~100 groups.

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/scim/data_user.go
+++ b/scim/data_user.go
@@ -3,25 +3,80 @@ package scim
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
+
+
+func NewUsersCache(expiration time.Duration) *UsersCache {
+	return &UsersCache {
+		cache: map[string]User{},
+		mutex: sync.Mutex{},
+		cacheExpiration: expiration,
+		// Make sure the cache is populated on the first read
+		populateTime: time.Time{},
+	}
+}
+
+type UsersCache struct {
+	mutex sync.Mutex
+	cache map[string]User
+	cacheExpiration time.Duration
+	populateTime time.Time
+}
+
+var usersCache *UsersCache = NewUsersCache(time.Second * 60)
+
+func (c *UsersCache) populate(api UsersAPI) error {
+	var users UserList
+	req := map[string]string{}
+	req["excludedAttributes"] = "roles"
+	err := api.client.Scim(api.context, http.MethodGet, "/preview/scim/v2/Users", req, &users)
+	if err != nil {
+		return err
+	}
+	u := users.Resources
+	for _, user := range u {
+		tflog.Info(api.context, fmt.Sprintf("Caching user %s", user.UserName))
+		c.cache[strings.ToLower(user.UserName)] = user
+	}
+	c.populateTime = time.Now()
+	return nil
+}
+
+func (c *UsersCache) Get(api UsersAPI, userName string) (User, error) {
+	// Databricks' search is case-insensitive
+	userName = strings.ToLower(userName)
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	if time.Since(c.populateTime) > c.cacheExpiration {
+		if err := c.populate(api); err != nil {
+			return User{}, err
+		}
+	}
+
+	tflog.Debug(api.context, fmt.Sprintf("Getting user %s from cache", userName))
+	user, ok := c.cache[userName]
+	tflog.Debug(api.context, fmt.Sprintf("User %s found in cache: %t", userName, ok))
+	if ok {
+		return user, nil
+	}
+
+	return User{}, fmt.Errorf("cannot find user %s", userName)
+}
 
 func getUser(usersAPI UsersAPI, id, name string) (user User, err error) {
 	if id != "" {
 		return usersAPI.Read(id, "userName,displayName,externalId,applicationId")
 	}
-	userList, err := usersAPI.Filter(fmt.Sprintf(`userName eq "%s"`, name), true)
-	if err != nil {
-		return
-	}
-	if len(userList) == 0 {
-		err = fmt.Errorf("cannot find user %s", name)
-		return
-	}
-	user = userList[0]
+	user, err = usersCache.Get(usersAPI, name)
 	return
 }
 
--- a/scim/data_user.go
+++ b/scim/data_user.go
@@ -13,11 +13,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-
 func NewUsersCache(expiration time.Duration) *UsersCache {
-	return &UsersCache {
-		cache: map[string]User{},
-		mutex: sync.Mutex{},
+	return &UsersCache{
+		cache:           map[string]User{},
+		mutex:           sync.Mutex{},
 		cacheExpiration: expiration,
 		// Make sure the cache is populated on the first read
 		populateTime: time.Time{},
@@ -25,10 +24,10 @@ func NewUsersCache(expiration time.Duration) *UsersCache {
 }
 
 type UsersCache struct {
-	mutex sync.Mutex
-	cache map[string]User
+	mutex           sync.Mutex
+	cache           map[string]User
 	cacheExpiration time.Duration
-	populateTime time.Time
+	populateTime    time.Time
 }
 
 var usersCache *UsersCache = NewUsersCache(time.Second * 60)
--- a/scim/resource_group_member.go
+++ b/scim/resource_group_member.go
@@ -13,11 +13,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
-
 type GroupMembersInfo struct {
 	initialized bool
 	members     map[string]struct{}
-	lock 	    *sync.Mutex
+	lock        *sync.Mutex
 }
 
 type GroupCache struct {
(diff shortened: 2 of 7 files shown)


alexott (changes requested):
I'm really not sure if the user cache will drastically improve the performance, at least until it's combined with the `databricks_user` resource.  It makes complete sense a caching for group membership.

You need to add an entry to the `NEXT_CHANGELOG.md` file describing your change.

rusabd:
> I'm really not sure if the user cache will drastically improve the performance, at least until it's combined with the `databricks_user` resource. It makes complete sense a caching for group membership.

It turned out to be surprisingly effective in our case — over 70% of the performance improvement came from the user cache. With our large user base, it appears this is primarily due to avoiding throttling.

 
> You need to add an entry to the `NEXT_CHANGELOG.md` file describing your change.
done

rusabd:
@alexott is there anything left to do get this PR accepted? We would prefer to not to use our fork of the provider internally.

alexott:
Do we really need a new dependency if simple map + mutex will work? The run of TF usually isn't so long to justify the need for LRU policies, etc.?

rusabd:
@alexott if it is not desirable, I can remove it. It is handy, but not absolutely necessary.

scim/data_user.go
```diff
@@ -3,25 +3,79 @@ package scim
...
+		mutex:           sync.Mutex{},
+		cacheExpiration: expiration,
+		// Make sure the cache is populated on the first read
+		populateTime: time.Time{},
 	}
-	userList, err := usersAPI.Filter(fmt.Sprintf(`userName eq "%s"`, name), true)
+}
+
+type UsersCache struct {
+	mutex           sync.Mutex
```
alexott: Won't RWMutex be better here?

scim/data_user.go
```diff
@@ -3,25 +3,79 @@ package scim
...
+	cache           map[string]User
+	cacheExpiration time.Duration
```
alexott: Do we really need expiration time? I'm thinking about the lifetime of the cache entries - it should be relatively short. I would also not worry about the size of the cache - even if we have 100s of thousands of objects it's not so huge amount of memory

scim/data_user.go
rusabd: It will. I just went for simpler approach, will change that.

scim/data_user.go
rusabd: TBH, I don't have a good justification for this. We have quite a lot of users, but still not that much - will remove

scim/resource_group_member.go
```diff
@@ -3,21 +3,141 @@ package scim
...
+		tflog.Debug(api.context, fmt.Sprintf("Group %s has %d members", groupID, len(group.Members)))
+		for _, member := range group.Members {
+			memberKey := strings.ToLower(member.Value)
+			groupInfo.members[memberKey] = struct{}{}
+		}
+		groupInfo.initialized = true
+		return groupInfo.members, nil
+	}
+	tflog.Debug(api.context, fmt.Sprintf("Group %s has %d members (cached)", groupID, len(groupInfo.members)))
+	return groupInfo.members, nil
```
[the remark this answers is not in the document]
rusabd: done

scim/resource_group_member.go
```diff
@@ -3,21 +3,141 @@ package scim
...
 
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
+type GroupMembersInfo struct {
+	initialized bool
+	members     map[string]struct{}
+	lock        *sync.Mutex
```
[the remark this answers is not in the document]
rusabd: done

scim/data_user.go
rusabd: done

scim/data_user.go
rusabd: done

alexott (changes requested):
The user caching doesn't work correctly when we have both data source and users creation/update in the same plan as we populate the cache once, on the first use...

Also, we may need to make the caching customizable - if I only need a few of users, listing of all users could be very expensive in big organizations. In TF exporter I do a listing of user names/user IDs only, and then fetch all necessary details on demand.

To proceed with this PR I suggest to take out the user caching part and leave only group membership caching, and make users caching in a separate PR.

rusabd:
@alexott removed user cache - will make a separate PR for it

alexott (approved):

scim/resource_group_member.go
```diff
@@ -3,21 +3,127 @@ package scim
...
 )
 
+type GroupMembersInfo struct {
+	initialized bool
+	members     map[string]struct{}
+	lock        sync.Mutex
+}
+
+type GroupCache struct {
+	cache map[string]*GroupMembersInfo
```
mgyucht: We are working on a project to support managing workspace-level resources while authenticated at the account-level. When this happens, we will need to make a change here to add the workspace ID to the cache key because group IDs could be duplicated across workspaces. However, this isn't supported yet, so for now we can mark this as such.
```suggestion
	// The mapping of cached members for this group. The cache key is the group ID.
	// TODO: add workspace ID to the cache key when account-level and workspace-level providers are unified.
	cache map[string]*GroupMembersInfo
```
cc @tanmay-db

scim/resource_group_member.go
```diff
@@ -3,21 +3,127 @@ package scim
...
+	}
+	if groupInfo.initialized {
+		memberKey := strings.ToLower(memberID)
+		groupInfo.members[memberKey] = struct{}{}
+	}
+	return err
+}
+
+func hasMember(members map[string]struct{}, memberID string) bool {
+	memberKey := strings.ToLower(memberID)
```
mgyucht: Why do we need to lower-case the member ID? I think this is always a number (though in a string), so I don't think there is a reason to lower-case it here or elsewhere in this file. @alexott does this match your understanding?

scim/resource_group_member.go
```diff
@@ -3,21 +3,127 @@ package scim
...
 	"fmt"
+	"strings"
+	"sync"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
+type GroupMembersInfo struct {
```
mgyucht: Do these types need to be exported?
```suggestion
// groupMembersInfo is the set of members that belong to a group. Read and write access
// to the underlying members map must be protected by the lock.
type groupMembersInfo struct {
```

scim/resource_group_member.go
```diff
@@ -3,21 +3,127 @@ package scim
...
+	initialized bool
+	members     map[string]struct{}
+	lock        sync.Mutex
+}
+
+type GroupCache struct {
```
mgyucht: Ditto:
```suggestion
// groupCache is a cache of groups to their members by their ID. Access to the cache
// must be protected by the lock.
type groupCache struct {
```

scim/resource_group_member.go
```diff
@@ -3,21 +3,127 @@ package scim
...
+			initialized: false,
+			members:     make(map[string]struct{}),
+			lock:        sync.Mutex{},
+		}
+		gc.cache[groupID] = groupInfo
+	}
+	return groupInfo
+}
+
+func (gc *GroupCache) GetMembers(api GroupsAPI, groupID string) (map[string]struct{}, error) {
```
mgyucht: same, does this need to be exported?
```suggestion
func (gc *GroupCache) getMembers(api GroupsAPI, groupID string) (map[string]struct{}, error) {
```

alexott:
@rusabd please ping me when all necessary changes are done, so I can rerun integration tests

scim/resource_group_member.go
rusabd: I am pretty sure it comes sometimes as email, sure @alexott  can confirm

rusabd:
> @rusabd please ping me when all necessary changes are done, so I can rerun integration tests

@alexott should be fine now

scim/resource_group_member.go
alexott: @rusabd no, the `value` returned by API would be always a number (SCIM ID) as string, like this:

```
        {
          "$ref":"ServicePrincipals/8279988778900481",
          "display":"test-mi",
          "value":"8279988778900481"
        },
        {
          "$ref":"Users/4273874807742774",
          "display":"User",
          "value":"4273874807742774"
        },
```

rusabd: @alexott removed

alexott:
@mgyucht all comments were addressed

scim/resource_group_member.go
```diff
@@ -3,21 +3,128 @@ package scim
...
+	}
+	return err
+}
+
+func hasMember(members map[string]struct{}, memberID string) bool {
+	_, ok := members[memberID]
+	return ok
+}
+
+var globalGroupsCache = newGroupCache()
```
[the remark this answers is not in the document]
rusabd: @alexott would api.client.Config.Host be a sufficient proxy for workspace id? There is no workspace id in this context,  and at least in our setup api.client.Config.Host corresponds to workspace. If not, what would be a good alternative?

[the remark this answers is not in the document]
alexott: it's more a TODO for the future. This change should be made in the client to make sure that the value is cached and we don't call respective REST API on each call... So leave it as-is right now

rusabd:
@mgyucht is there anything else left to do in this PR?

scim/resource_group_member.go
```diff
@@ -3,21 +3,128 @@ package scim
...
+	initialized bool
+	members     map[string]struct{}
+	lock        sync.Mutex
+}
+
+// groupCache is a cache of groups to their members by their ID. Access to the cache
+// must be protected by the lock.
+type groupCache struct {
+	// The mapping of cached members for this group. The cache key is the group ID.
+	// TODO: add workspace ID to the cache key when account-level and workspace-level providers are unified.
```
mgyucht, line 25: cc @tanmay-db as we migrate this non-Go SDK resource to unified provider, we can handle this TODO.

PR #4590, opened 2025-03-20
databricks_mount: remove redundant listing of mounted directories.

KoenDeCausmaecker-TomTom:
Directory listings can take very long when there are lots of files in the mounted directory. This slows down both the state refresh and the actual mounting up to the point that the operation times out.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott:
The `ls` operation was used to make sure that the mount works - there were cases when the mount was defined, but it was actually not working because of expired credentials, etc. So it should be discussed how we can reliable detect that mount works

alexott (approved):

alexott:
Thank you! 

Out of curiosity - do you also use the exporter, or did you just make the change to be uniform across all places?

KoenDeCausmaecker-TomTom:
Oh no, I'm not using the exporter - so I haven't tried it either.
Thanks for pointing this out because I was not aware.
I have removed the change from the exporter code

PR #4597, opened 2025-03-25
Increase HTTP timeout in tests to 5m

mgyucht:
## Changes
Currently the TF provider uses the Go SDK's default HTTP timeout of 60 seconds to wait for a response from the API. This is coupled with a 60 second server-side timeout for the API proxy. However, backend services that expect to take longer to handle requests can have an increased server-side timeout. In this case, the client may need to wait for longer than 60 seconds to receive a response. As a result, some integration tests for the provider fail with errors like "request timed out after 1m0s".

This PR adjusts this by configuring the SDK used by the provider to have a 10-minute timeout during integration tests. This is longer than any individual server's server-side timeout. Long-term, the SDK should be responsible for selecting the appropriate client-side timeout for a given API, but in the meantime we'll make this adjustment.

This only affects tests. The default value for http_timeout_seconds is still 60 seconds.

To-do:
- [ ] Add doccomments to new exported methods/functions/types.

## Tests
- [ ] Run integration tests, see that they pass.

NO_CHANGELOG=true


renaudhartert-db (approved):

tanmay-db (approved):

alexott:
I would say that it's either something on backend, or it's something with the environment (i.e., firewall?)

> cannot create permissions: The service at /api/2.0/permissions/instance-pools/0325-130841-warps228-pool-mr7luntc is taking too long to process your request.

PR #4620, opened 2025-04-07
Ensure release process automatically publishes non-draft release

mgyucht:
## Changes
With the updated release automation based on the `tagging` workflow, to complete the process of automating the full release lifecycle, the `release` workflow should use the latest changelog entry and publish a non-draft release. This is a small change to read the release notes from the most recent changelog entry and supply that to goreleaser.

## Tests

- [x] Added a small unit test for the extract_release_notes.py utility that is run as part of unit tests for the repo.

I can't easily manually test the workflow changes, so we will need to do a small release to verify. I'll wait until a single change is merged to test this out.

NO_CHANGELOG=true


.github/scripts/extract_release_notes.py
```diff
@@ -0,0 +1,30 @@
...
+        
+        if match:
+            release_notes = match.group(1).strip()
+            print(release_notes)
+            return 0
+        else:
+            print("No release notes found", file=sys.stderr)
+            return 1
+            
+    except Exception as e:
```
tanmay-db, line 24: nit: Usually it's better to catch specific exceptions rather than catching all, might be good to check for: `FileNotFoundError`

tanmay-db (approved):

hectorcast-db:
Why not extracting the changelog from the tag?
It should be the full content

hectorcast-db:
Why not extracting it from the tag? Seems easier and no parsing is needed.

```
      # Extract the tag message
      - name: Get tag message
        id: tag_message
        run: |
          TAG_MESSAGE=$(git for-each-ref --format '%(subject)' refs/tags/${GITHUB_REF#refs/tags/})
          echo "TAG_MESSAGE=${TAG_MESSAGE}" >> $GITHUB_ENV

      # Run GoReleaser
      - name: Run GoReleaser
        uses: goreleaser/goreleaser-action@v6
        with:
          version: ~> v2
          args: release --clean ${{ !startsWith(github.ref, 'refs/tags/v') && '--snapshot' || '' }} --release-notes "$TAG_MESSAGE"
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
          TAG_MESSAGE: ${{ env.TAG_MESSAGE }}
```

mgyucht:
Good point, I forgot the changelog was also in the tag.

PR #4634, opened 2025-04-11
[Feature] Support for `lifetime` attribute in `databricks_service_principal_secret` resource

### LINKED ISSUE #4631: [FEATURE] databricks_service_principal_secret resource missing lifetime parameter
Hi There is a databricks terraform resource **databricks_service_principal_secret** 
(https://cf-registry.tf-registry-prod-use1.terraform.io/providers/databricks/databricks/1.72.0/docs/resources/service_principal_secret)
but it is missing '**lifetime**' parameter.

According to API call: https://docs.databricks.com/api/account/serviceprincipalsecrets/create
There should be lifetime parameter which will set lifetime of the secret in seconds.

alexott:
## Changes

Also refactored it to use Go SDK to expose more attributes.

Resolves #4631 

## Tests

- [x] `make test` run locally
- [x] using Go SDK


docs/resources/service_principal_secret.md
```diff
@@ -25,14 +25,20 @@ resource "databricks_service_principal_secret" "terraform_sp" {
...
 
 In addition to all arguments above, the following attributes are exported:
 
 * `id` - ID of the secret
-* `secret` - Generated secret for the service principal
+* `secret` - Generated secret for the service principal.
+* `create_time` - UTC time when the secret was created.
+* `expire_time` - UTC time when the secret will expire. If the field is not present, the secret does not expire.
+* `secret_hash` - Secret Hash.
+* `status`  - Status of the secret (i.e., `ACTIVE`).
```
tanmay-db, line 40: Might be good to list all the enum values here the `status` can take

tokens/resource_service_principal_secret.go
```diff
@@ -2,125 +2,128 @@ package tokens
...
-			if c.Config.AccountID == "" {
-				return errors.New("must have `account_id` on provider")
+			ac, err := c.AccountClient()
+			if err != nil {
+				return err
+			}
+			spId := d.Get("service_principal_id").(string)
+			spIdNumeric, err := strconv.ParseInt(spId, 10, 64)
+			if err != nil {
+				return fmt.Errorf("failed to convert service principal ID to numeric: %w", err)
```
tanmay-db, line 120: We have this error message appear thrice, would be good to abstract this out so we reduce redundancy.

tanmay-db (approved):

docs/resources/service_principal_secret.md
alexott: Right now it's only `ACTIVE`

PR #4643, opened 2025-04-14
[Fix] Make `spark_version` field optional to work with defaults in policies

pietern:
## Changes

A cluster policy can enforce a specific `spark_version` field and set it as the default. This mechanism allows for a centralized choice of the Databricks Runtime version across all jobs. To allow a job to inherit this field from a policy, it must be configured as optional in the schema.

The job resource referred to `JobSettings` and `JobSettingsResource` with a mix of `js` and `jsr` variable names. This PR updates references to `JobSettingsResource` to be called `jsr`.

## Tests

- [x] `make test` run locally


pietern:
This fails with:
```
Error: cannot create job: panic: job_cluster: new_cluster: inconsistency: spark_version is optional, default is empty, but has no omitempty
```

I suspect the field needs to be marked as optional in the OpenAPI spec for clusters to make this work.

clusters/clusters_api.go
```diff
@@ -399,7 +399,7 @@ type Cluster struct {
 	ClusterID   string `json:"cluster_id,omitempty"`
 	ClusterName string `json:"cluster_name,omitempty"`
 
-	SparkVersion              string     `json:"spark_version"`
+	SparkVersion              string     `json:"spark_version,omitempty"`
```
pietern, line 402: Without this change, the jobs unit tests failed with:
```
panic: new_cluster: inconsistency: spark_version is optional, default is empty, but has no omitempty
```

Reflection on the 2.0 structs flagged the incompatibility between an optional field in the schema without a default and a mandatory struct field. This struct is no longer used directly by a resource.

PR #4657, opened 2025-04-17
[Doc] Unify/reorganize notes in docs for resources/data sources

alexott:
## Changes

Changes include:

* Remove not necessary `**Note**` in `->`, etc.
* Put notes about account or workspace level provider after the first paragraph, so it's easier to understand what resource/data source is doing, and then read the note.
* Added missing notes about account or workspace level provider
* Removed repeated notes about `authentication is not configured for provider errors` that is more about troubleshooting than about specific resource/data source.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


docs/data-sources/metastores.md
```diff
@@ -3,11 +3,11 @@ subcategory: "Unity Catalog"
...
 # databricks_metastores Data Source
 
--> **Note** This data source can only be used with an account-level provider!
-
 Retrieves a mapping of name to id of [databricks_metastore](../resources/metastore.md) objects, that were created by Terraform or manually, so that special handling could be applied.
 
--> **Note** [`account_id`](../index.md#account_id) provider configuration property is required for this resource to work. Data resource will error in case of metastores with duplicate names. This data source is only available for users & service principals with account admin status
+-> This data source can only be used with an account-level provider!
+
+-> [`account_id`](../index.md#account_id) provider configuration property is required for this resource to work. Data resource will error in case of metastores with duplicate names. This data source is only available for users & service principals with account admin status
```
nkvuong, line 10: > This data source is only available for users & service principals with account admin status

this note applies to most of account-level resources right?

docs/data-sources/aws_crossaccount_policy.md
```diff
@@ -3,10 +3,10 @@ subcategory: "Deployment"
 ---
 # databricks_aws_crossaccount_policy Data Source
 
--> **Note** This data source can only be used with an account-level provider!
-
 This data source constructs necessary AWS cross-account policy for you, which is based on [official documentation](https://docs.databricks.com/administration-guide/account-api/iam-role.html#language-Your%C2%A0VPC,%C2%A0default).
 
+-> This data source can be used with an account or workspace-level provider.
```
nkvuong, line 8:
```suggestion
-> This data source can be used with an account-level provider.
```
this data source only makes sense for account-level context

docs/data-sources/aws_bucket_policy.md
```diff
@@ -5,6 +5,8 @@ subcategory: "Deployment"
 
 This datasource configures a simple access policy for AWS S3 buckets, so that Databricks can access data in it.
 
+-> This data source can be used with an account or workspace-level provider.
```
nkvuong, line 8:
```suggestion
-> This data source can be used with an account-level provider.
```
this data source only makes sense for account-level context

docs/data-sources/aws_assume_role_policy.md
```diff
@@ -6,6 +6,8 @@ subcategory: "Deployment"
 
 This data source constructs necessary AWS STS assume role policy for you.
 
+-> This data source can be used with an account or workspace-level provider.
```
nkvuong, line 9:
```suggestion
-> This data source can be used with an account-level provider.
```
this data source only makes sense for account-level context

docs/resources/system_schema.md
```diff
@@ -3,14 +3,12 @@ subcategory: "Unity Catalog"
 ---
 # databricks_system_schema Resource
 
--> This feature is in [Public Preview](https://docs.databricks.com/release-notes/release-types.html).
```
nkvuong, line 6: need to add this back

nkvuong (changes requested):

docs/data-sources/metastores.md
alexott: yes. I'm really not sure that we should have it. I was just moving the comment around

docs/data-sources/aws_crossaccount_policy.md
alexott: Technically it works with both: https://github.com/databricks/terraform-provider-databricks/blob/main/aws/data_aws_crossaccount_policy.go#L26 - we changed it to account only, and it broke people's work: https://github.com/databricks/terraform-provider-databricks/issues/3505

docs/resources/system_schema.md
alexott: it depends on the table. In general they are GA: https://learn.microsoft.com/en-us/azure/databricks/release-notes/product/2024/september#system-tables-are-now-generally-available

docs/data-sources/aws_bucket_policy.md
alexott: It really doesn't require any provider. We already tried to make it account-only and got a pushback from users.

PR #4670, opened 2025-04-23
Support serverless workspaces in `databricks_mws_workspaces`

mgyucht:
## Changes
A new field `compute_mode` has been added to `databricks_mws_workspaces`. This can be set to `SERVERLESS` to indicate that a workspace should be serverless. Serverless workspaces are documented at https://docs.databricks.com/aws/en/admin/workspace/serverless-workspaces. These workspaces need neither `credentials_id` or `storage_configuration_id`, so the setup should be much easier.

## Tests
Added an integration test for creating serverless workspaces in AWS.

## Todos
- [x] Need to verify if serverless workspaces work in GCP. EDIT: not yet supported.
- [x] Need to verify which regions users can specify for serverless workspaces. EDIT: this will be documented shortly. I'll include a link.
- [x] The test does not work right now, I think our E2 account still needs to be onboarded. EDIT: our account was onboarded and we verified that this works.


mws/resource_mws_workspaces.go
```diff
@@ -577,7 +583,7 @@ func ResourceMwsWorkspaces() common.Resource {
 			var workspace Workspace
 			workspacesAPI := NewWorkspacesAPI(ctx, c)
 			common.DataToStructPointer(d, workspaceSchema, &workspace)
-			if err := requireFields(c.IsAws(), d, "aws_region", "credentials_id", "storage_configuration_id"); err != nil {
```
alexott, line 580: Should we have two cases - when compute_mode is serverless, then use this new condition, and when isn't, then use the old condition?

NEXT_CHANGELOG.md
```diff
@@ -5,7 +5,11 @@
 ### New Features and Improvements
 
  * Add support for `power_bi_task` in jobs ([#4647](https://github.com/databricks/terraform-provider-databricks/pull/4647))
+<<<<<<< HEAD
```
alexott, line 8: we need to review merge conflict

mgyucht: Ah new laptop/editor, need to setup autosave.

alexott (approved):

Issue #3980, opened 2024-09-02
[ISSUE] Issue with `databricks_sql_table` resource import


levysaar:
Hey all,
When importing a databricks_sql_table resource, partitions do not get imported.
Therefore, when running plan/apply Terraform mark the table as "must be replaced" and will drop and recreate it.

### Configuration
```hcl
resource "databricks_sql_table" "import_test" {
  catalog_name = "catalog_dev"
  name         = "import_test_table"
  schema_name  = "default"
  table_type   = "MANAGED"
  warehouse_id = "abcd1234"
  data_source_format = "DELTA"
  partitions = ["dt"]

  column {
    name = "id"
    type = "int"
  }
  column {
    name = "dt"
    type = "date"
  }
}
```

Import command:
```
terraform import databricks_sql_table.import_test catalog_dev.default.import_test_table
```

output of Terraform state show:
```
# databricks_sql_table.import_test:
resource "databricks_sql_table" "import_test" {
    catalog_name       = "catalog_dev"
    data_source_format = "DELTA"
    id                 = "catalog_dev.default.import_test_table"
    name               = "import_test_table"
    owner              = "me@me.com"
    properties         = {
        "delta.checkpoint.writeStatsAsJson"   = "false"
        "delta.checkpoint.writeStatsAsStruct" = "true"
        "delta.enableDeletionVectors"         = "true"
        "delta.feature.deletionVectors"       = "supported"
        "delta.lastCommitTimestamp"           = "1725279945000"
        "delta.lastUpdateVersion"             = "0"
        "delta.minReaderVersion"              = "3"
        "delta.minWriterVersion"              = "7"
    }
    schema_name        = "default"
    storage_location   = "s3://....."
    table_type         = "MANAGED"

    column {
        name     = "id"
        nullable = true
        type     = "int"
    }
    column {
        name     = "dt"
        nullable = true
        type     = "date"
    }
}
```

### Expected Behavior
When importing the resource, include existing partitions

### Actual Behavior
Partitions do not gets imported, which leads to the table being recreated by TF

### Steps to Reproduce
1. table creation:
```
create table catalog_dev.default.import_test_table (id int, dt date) partitioned by (dt);
```

2. Import the table resource:
```
terraform import databricks_sql_table.import_test catalog_dev.default.import_test_table
```

3. output of Terraform state show:
```
terraform state show databricks_sql_table.import_test

# databricks_sql_table.import_test:
[...]

PR #4486, opened 2025-02-12
[Doc] Clarifying limitations when importing `databricks_sql_table`

nkvuong:
## Changes
- Clarify which attributes are not read when importing `databricks_sql_table`
resolves #3980
## Tests

- [x] relevant change in `docs/` folder

PR #4674, opened 2025-04-25
[Fix] Populate partitions when reading `databricks_sql_table`

nkvuong:
## Changes
- Redo #4486, marking `partitions` column as optional computed to avoid unexpected diff.
- UC returns `partition_index` in the column struct, so we can calculate partitions value based on that
- Resolves https://github.com/databricks/terraform-provider-databricks/issues/3980

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_sql_table.go
+++ b/catalog/resource_sql_table.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"reflect"
 	"slices"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


catalog/resource_sql_table.go
```diff
@@ -675,13 +676,42 @@ func ResourceSqlTable() common.Resource {
...
+					partitionIndexes[c.PartitionIndex] = c.Name
+				}
+			}
+			indexes := []int{}
+			partitions := []string{}
+
+			for index := range partitionIndexes {
+				indexes = append(indexes, index)
+			}
+			sort.Ints(indexes)
```
alexott, line 708: we can simplify it to 

```
indexes := slices.Sorted(maps.Keys(partitionIndexes))
```

alexott (approved):

PR #4133, opened 2024-10-21
WIP: [Feature] Add `databricks_git_folder` resource to replace `databricks_repo`

840:
## Changes
According to [Databricks - What happened to Databricks Repos?](https://docs.databricks.com/en/repos/what-happened-repos.html), Databricks Repos is being deprecated in favour of Git Folders. This PR aims to add `databricks_git_folder` resource and deprecate `databricks_repo` resource.

Resolves #3877 

## Tests
- [ ] `make test` run locally
- [ ] relevant change in `docs/` folder
- [ ] covered with integration tests in `internal/acceptance`
- [ ] relevant acceptance tests are passing
- [ ] using Go SDK

PR #4678, opened 2025-04-28
[FEATURE] Rename databricks_repo into databricks_git_folder

WeiJun428:
## Changes

This PR attempts to fix the issue #3448. It
- Replaces all usage of `databricks_repo` to `databricks_git_folder`
- Renames `docs/resources/repo.md` to `docs/resources/git_folder.md` and update all the references in the docs folder accordingly.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott (changes requested):
We can't just rename a resource - it will be a breaking change for a huge number of customers. Instead, the existing resource should be marked as deprecated

See #4133 (WIP). You can see example in `catalog/resource_catalog_workspace_binding.go`, but it's a bit different as git folder should have a bit different schema (the `path` is required), etc.

alexott:
Closing as stale

PR #4684, opened 2025-05-01
Update storage_credential docs to reference name not id

karayount:
## Changes
Updates docs/resources/storage_credential.md examples of creating databricks_grants resources that reference a storage_credential. Previously it had indicated referencing the credential by id. In my case, the name and id are not identical, and referencing credential by id caused errors, where referencing by name did not.

It appears that this package may be written with the assumption that name and id have the same value.
On line 117 of this same storage_credential.md it says
`- 'id' - ID of this storage credential - same as the 'name'.`
And there is a test `catalog/external_location_test.go` that indicates storage_credential name and id are interchangeable
```
func externalLocationTemplateWithOwner(comment string, owner string) string {
	return fmt.Sprintf(`
		resource "databricks_external_location" "some" {
			name            = "external-{var.STICKY_RANDOM}"
			url             = "s3://{env.TEST_BUCKET}/some{var.STICKY_RANDOM}"
			credential_name = databricks_storage_credential.external.id
			isolation_mode  = "ISOLATION_MODE_ISOLATED"
			comment         = "%s"
			owner = "%s"
		}
	`, comment, owner)
}
```

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/docs/resources/storage_credential.md
+++ b/docs/resources/storage_credential.md
@@ -26,7 +26,7 @@ resource "databricks_storage_credential" "external" {
 }
 
 resource "databricks_grants" "external_creds" {
-  storage_credential = databricks_storage_credential.external.id
+  storage_credential = databricks_storage_credential.external.name
   grant {
     principal  = "Data Engineers"
     privileges = ["CREATE_EXTERNAL_TABLE"]
@@ -46,7 +46,7 @@ resource "databricks_storage_credential" "external_mi" {
 }
 
 resource "databricks_grants" "external_creds" {
-  storage_credential = databricks_storage_credential.external_mi.id
+  storage_credential = databricks_storage_credential.external_mi.name
   grant {
     principal  = "Data Engineers"
     privileges = ["CREATE_EXTERNAL_TABLE"]
@@ -63,7 +63,7 @@ resource "databricks_storage_credential" "external" {
 }
 
 resource "databricks_grants" "external_creds" {
-  storage_credential = databricks_storage_credential.external.id
+  storage_credential = databricks_storage_credential.external.name
   grant {
     principal  = "Data Engineers"
     privileges = ["CREATE_EXTERNAL_TABLE"]


alexott:
Show the error that you get - Storage credentials and External locations have `.id` the same as `.name`: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_storage_credential.go#L90

We're using `.id` in the example to make sure that object is created...

karayount:
> Show the error that you get - Storage credentials and External locations have `.id` the same as `.name`: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_storage_credential.go#L90
> 
> We're using `.id` in the example to make sure that object is created...

The terraform apply failed with this error, when I used storage_credential's `.id`. The screenshot shows the name it's searching for is a uuid which is our id, rather than the name ("databricks-csat-data-access")
Screenshot 2025-05-07 at 8 45 02 AM

The apply succeeded when I changed `.id` to `.name` in this code
Screenshot 2025-05-07 at 8 53 09 AM

alexott:
Hmmm, works fine for me:

```
resource "databricks_storage_credential" "aott_db_access" {
  name    = "test-access-tmp"
  comment = "test"
  azure_managed_identity {
    access_connector_id = "/subscriptions/1111/resourceGroups/test-rg/providers/Microsoft.Databricks/accessConnectors/test-access"
  }
}

resource "databricks_grants" "storage_credential_aott_db_access" {
  storage_credential = databricks_storage_credential.aott_db_access.id
  grant {
    privileges = ["ALL_PRIVILEGES"]
    principal  = "user1@domain.com"
  }
  grant {
    privileges = ["ALL_PRIVILEGES"]
    principal  = "user1@domain.com"
  }
}
```

alexott:
Can you share the debug log as per troubleshooting guide?

mgyucht:
@alexott the difference is that @karayount is using the databricks_storage_credentials data source, which exposes the UUID ID on the `id` field. However, for the resource, the `id` is indeed the same as the `name`. Since you need to use the name and not the ID for the `databricks_grants` resource, I think this is OK.

In this case, this is a bit of technical debt about the definition of the `databricks_storage_credentials` resource/datasource. They probably should use the same value for `id` but don't.

mgyucht:
Yeah, but it will work with `name` if you use either, as the `id` of the resource is set from the `name` today, and the API expects the `name`.

alexott:
@karayount I'm wondering why do we need a data source in your case if you just know the name? Instead of changing the documentation to use `name` I would just improve the description that say that it's a name that could be used directly for existing resources (for new resources we need to use `.id` to make sure that resource is created, before granting an access).

karayount:
> @karayount I'm wondering why do we need a data source in your case if you just know the name? Instead of changing the documentation to use `name` I would just improve the description that say that it's a name that could be used directly for existing resources (for new resources we need to use `.id` to make sure that resource is created, before granting an access).

Ah. I will close this. I missed that the docs were for resource and I was using data, and I found the error surprising. Might be nice to add a data example to the docs too, if folks agree.

PR #4694, opened 2025-05-07
Use latest Databricks SDK

mgyucht:
## Changes
Bump the TF provider to use latest Databricks SDK, enabling OIDC support.

## Tests
Covered by existing tests.


pipelines/resource_pipeline.go
```diff
@@ -274,7 +274,6 @@ func (Pipeline) CustomizeSchema(s *common.CustomizableSchema) *common.Customizab
 	s.SchemaPath("edition").SetValidateFunc(validation.StringInSlice([]string{"pro", "core", "advanced"}, true))
 
 	// RequiredWith fields
-	s.SchemaPath("gateway_definition").SetRequiredWith([]string{"gateway_definition.0.gateway_storage_name", "gateway_definition.0.gateway_storage_catalog", "gateway_definition.0.gateway_storage_schema"})
```
mgyucht, line 277: After completing the OpenAPI generator migration, these nested fields became required; thus, it is not necessary to specify them as RequiredWith (because they are always required). This causes failures during validation in Terraform, so we remove it here.

PR #4697, opened 2025-05-08
[Fix] Fail when creating `databricks_query` and `databricks_alert` with already existing names

alexott:
## Changes

Right now, when an object with the same name already exists in a workspace, the Query and Alert APIs just create an object with a different name (by appending `(1)` or something like that). This leads to permanent configuration drift as TF will try to rename the object to the name specified in the HCL.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


nkvuong (approved):

tanmay-db:
PR to Bump Go SDK: https://github.com/databricks/terraform-provider-databricks/pull/4735.

PR #4701, opened 2025-05-08
Add provisioned throughput provider

Bafoleco:
## Changes

Adds new databricks_model_serving_provisioned_throughput resource heavily based on the databricks_model_serving
resource. This is to enable creation of provisioned throughput foundation models through our new /serving-endpoints/pt API. We separate this type of endpoint creation in our API, and in our terraform provider, because it enables a much simpler user experience, then having a single overloaded approach to creating many different model types

## Tests

- [x] `make test` run locally
- [x] using Go SDK


alexott:
I'm thinking if we can add a flag to the existing model serving resource to specify that it will be used for automatically provisioned models?

Bafoleco:
> I'm thinking if we can add a flag to the existing model serving resource to specify that it will be used for automatically provisioned models?

I think we could do this, but not sure if it's the best long term approach. We've created a new API for our provisioned throughput foundation models so that we can simplify the API, and I think similar reasoning applies to the terraform provider. The new docs for the provisioned throughput provider are much simpler, because we remove a lot of the complexity from combining very different endpoint types into one provider.

Bafoleco:
> > I'm thinking if we can add a flag to the existing model serving resource to specify that it will be used for automatically provisioned models?
> 
> I think we could do this, but not sure if it's the best long term approach. We've created a new API for our provisioned throughput foundation models so that we can simplify the API, and I think similar reasoning applies to the terraform provider. The new docs for the provisioned throughput provider are much simpler, because we remove a lot of the complexity from combining very different endpoint types into one provider.

Probably the deeper problem is that the user shouldn't have to consider this flag in the long term, eventually all users will just use the new /serving-endpoints/pt api for all new foundation models. And we can't autodetect the correct value for this flag without making an additional API call first

serving/resource_model_serving_provisioned_throughput.go
```diff
@@ -0,0 +1,116 @@
...
+	s := common.StructToSchema(
+		serving.CreatePtEndpointRequest{},
+		func(m map[string]*schema.Schema) map[string]*schema.Schema {
+			m["name"].ForceNew = true
+			common.MustSchemaPath(m, "config", "traffic_config").Computed = true
+			common.MustSchemaPath(m, "config", "served_entities", "name").Computed = true
+			common.MustSchemaPath(m, "config", "served_entities", "provisioned_model_units").Required = true
+			common.MustSchemaPath(m, "config", "served_entities", "entity_name").Required = true
+			common.MustSchemaPath(m, "config", "served_entities", "entity_version").Required = true
+			common.MustSchemaPath(m, "config", "served_entities", "entity_version").Optional = false
```
alexott, line 27: let use `common.CustomizeSchemaPath(s, "<attr>").SetComputed()`, `common.CustomizeSchemaPath(s, "<attr>").SetRequired()`, etc. - it will remove the need to have changes two times like:

```
common.MustSchemaPath(m, "config", "served_entities", "entity_version").Required = true
common.MustSchemaPath(m, "config", "served_entities", "entity_version").Optional = false
```

alexott (commented):
In general, looks good to me

serving/resource_model_serving_provisioned_throughput.go
```diff
@@ -0,0 +1,115 @@
...
+			return nil
+		},
+		Update: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			var e serving.CreatePtEndpointRequest
+			common.DataToStructPointer(d, s, &e)
+			if d.HasChange("config") {
```
alexott, line 84: What about changes for budget policy, tags, and ai_gateway?

serving/resource_model_serving_provisioned_throughput.go
```diff
@@ -0,0 +1,115 @@
...
+			return nil
+		},
+		Delete: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
+			w, err := c.WorkspaceClient()
+			if err != nil {
+				return err
+			}
+			return w.ServingEndpoints.DeleteByName(ctx, d.Id())
+		},
+		StateUpgraders: []schema.StateUpgrader{},
```
alexott, line 107: `StateUpgraders` is not required, `SchemaVersion` as well - they both will be initialized to zero values anyway

alexott (commented):
In general is ok - we need to discuss how to update fields outside of `config`

docs/resources/model_serving.md
```diff
@@ -294,18 +294,6 @@ terraform import databricks_model_serving.this <model-serving-endpoint-name>
...
-The following resources are often used in the same context:
-
-* [databricks_registered_model](registered_model.md) to create [Models in Unity Catalog](https://docs.databricks.com/en/mlflow/models-in-uc.html) in Databricks.
-* [End to end workspace management](../guides/workspace-management.md) guide.
-* [databricks_directory](directory.md) to manage directories in [Databricks Workspace](https://docs.databricks.com/workspace/workspace-objects.html).
-* [databricks_mlflow_model](mlflow_model.md) to create models in the [workspace model registry](https://docs.databricks.com/en/mlflow/model-registry.html) in Databricks.
-* [databricks_notebook](notebook.md) to manage [Databricks Notebooks](https://docs.databricks.com/notebooks/index.html).
-* [databricks_notebook](../data-sources/notebook.md) data to export a notebook from Databricks Workspace.
-* [databricks_repo](repo.md) to manage [Databricks Repos](https://docs.databricks.com/repos.html).
-
```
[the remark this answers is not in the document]
Bafoleco: This section is duplicated below, was just cleaning this up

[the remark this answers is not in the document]
alexott: No - it's duplicate in a separate file...

serving/resource_model_serving_provisioned_throughput.go
alexott: We have already a PR that just needs to be merged after small changes, so it makes sense to do this resource correctly from the beginning

docs/resources/model_serving.md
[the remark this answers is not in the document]
Bafoleco: current docs:

image

PR #4704, opened 2025-05-10
Fix use of `force` option for service principals in `databricks_git_credential`

alexott:
## Changes

When we're trying to set git authentication for service principal, the error message is slightly different from that for users, so I tweaked it to work with both.  Also adjusted the docs `git_username` and `personal_access_token` aren't required for all Git providers (i.e., for Entra ID Auth).

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


rauchy (approved):
Is there an easy way we could add a test for this?

PR #4712, opened 2025-05-14
Fix reading of `external_id` for `databricks_service_principal`

alexott:
## Changes

During work on exporter I found that `external_id` wasn't filled for service principals and this lead to a problem with generation of correct code.  The `databricks_user` had the correct code, so I unified the things

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


scim/resource_user.go
```diff
@@ -101,12 +109,8 @@ func ResourceUser() common.Resource {
 			if err != nil {
 				return err
 			}
+			setCommonUserFields(d, user, user.UserName)
```
tanmay-db, line 112: We can remove passing user.UserName since we are passing user so inside setCommonUserFields, we can use user.UserName directly.

alexott: it's required because we need to pass different attributes (UserName for user and ApplicationId for service principal)

PR #4727, opened 2025-05-20
[Fix] Don't fail delete when `databricks_system_schema` can be disabled only by Databricks

alexott:
## Changes

Also rewrote corresponding tests to use the mock framework from Go SDK

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_system_schema.go
+++ b/catalog/resource_system_schema.go
@@ -143,10 +143,18 @@ func ResourceSystemSchema() common.Resource {
 			if err != nil {
 				return err
 			}
-			return w.SystemSchemas.Disable(ctx, catalog.DisableRequest{
+			err = w.SystemSchemas.Disable(ctx, catalog.DisableRequest{
 				MetastoreId: metastoreSummary.MetastoreId,
 				SchemaName:  schemaName,
 			})
+			if err != nil {
+				//ignore "<schema-name> system schema can only be disabled by Databricks" error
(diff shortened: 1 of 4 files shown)


catalog/resource_system_schema.go
```diff
@@ -143,10 +143,18 @@ func ResourceSystemSchema() common.Resource {
...
 				MetastoreId: metastoreSummary.MetastoreId,
 				SchemaName:  schemaName,
 			})
+			if err != nil {
+				//ignore "<schema-name> system schema can only be disabled by Databricks" error
+				if !strings.Contains(err.Error(), "can only be disabled by Databricks") {
+					return err
+				}
+				log.Printf("[WARN] %s can be disabled only by Databricks, ignoring it", schemaName)
+			}
```
tanmay-db, line 156: Should we also do this above in createOrUpdate? 
```
			err = w.SystemSchemas.Disable(ctx, catalog.DisableRequest{
				MetastoreId: metastoreSummary.MetastoreId,
				SchemaName:  old,
			})
			if err != nil {
				return err
			}
```

tanmay-db (approved):

catalog/resource_system_schema.go
tanmay-db: If yes, then might be better to abstract this out in a method to remove redundancy.

alexott: ah, you're right - missed this piece there...

nkvuong (approved):

PR #4728, opened 2025-05-20
Unify client construction pathways and logging between SDKv2 and Plugin Framework implementations

mgyucht:
## Changes
Currently, the SDKv2 and Plugin Framework provider implementations have slightly different behavior in how the provider is configured. The SDKv2's schema.Schema is constructed using `DefaultFunc`, which allows setting provider-level configuration attributes to have default values based on environment variables. The specific environment variables are taken from struct tags in the Go SDK's Config object. On the other hand, the Plugin Framework's provider.Schema has no support for default values. Eventually, when the config.Config from the Databricks Go SDK is resolved, those attributes will be populated from the environment anyways. Thus, it should be safe to remove these environment variable defaults from the SDKv2 implementation, ensuring a uniform treatment of provider configuration between both implementations.

Beyond this, the code to construct the `DatabricksClient` used by each provider was copied, resulting in a small, unexpected divergence between the two. This common code has now been refactored into `internal/providers/client/databricks_client.go` and is used by both implementations.

Lastly, the Plugin Framework implementation incorrectly logged which attributes were configured by the user. This makes it challenging to review debug logs and interpret what attributes were actually set by a user in their provider configuration. This PR addresses this, with a small refactor of the logic to set an individual attribute refactored into a function.

## Tests
This should pass existing integration tests.

Additionally, when running integration tests locally, I can see the following log output:
```
2025-05-20T10:41:14.603Z [INFO]  databricks: (sdkv2) No attributes specified in provider configuration: tf_provider_addr=registry.terraform.io/hashicorp/databricks tf_rpc=ConfigureProvider tf_mux_provider=tf5to6server.v5tov6Server tf_req_id=b6fd5713-ac58-ef1e-ec8e-10ca5ecb8aa4
2025-05-20T10:41:14.604Z [INFO]  databricks: (plugin framework) No attributes specified in provider configuration: tf_provider_addr=registry.terraform.io/hashicorp/databricks tf_rpc=ConfigureProvider tf_mux_provider="*proto6server.Server" tf_req_id=b6fd5713-ac58-ef1e-ec8e-10ca5ecb8aa4
```

which is now accurate as the provider configuration is empty for integration tests.

### DIFF
--- a/internal/providers/client/databricks_client.go
+++ b/internal/providers/client/databricks_client.go
@@ -1,0 +1,57 @@
+package client
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/databricks/databricks-sdk-go/client"
+	"github.com/databricks/databricks-sdk-go/config"
+	"github.com/databricks/terraform-provider-databricks/commands"
+	"github.com/databricks/terraform-provider-databricks/common"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+func PrepareDatabricksClient(ctx context.Context, cfg *config.Config, configCustomizer func(*config.Config) error) (*common.DatabricksClient, error) {
+	if cfg.AuthType != "" {
+		// mapping from previous Google authentication types
+		// and current authentication types from Databricks Go SDK
+		oldToNewerAuthType := map[string]string{
+			"google-creds":     "google-credentials",
+			"google-accounts":  "google-id",
+			"google-workspace": "google-id",
+		}
+		newer, ok := oldToNewerAuthType[cfg.AuthType]
+		if ok {
+			tflog.Info(ctx, fmt.Sprintf("Changing required auth_type from %s to %s", cfg.AuthType, newer))
+			cfg.AuthType = newer
+		}
+	}
+	cfg.EnsureResolved()
+	// Unless set explicitly, the provider will retry indefinitely until context is cancelled
+	// by either a timeout or interrupt.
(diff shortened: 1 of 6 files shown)


internal/providers/client/databricks_client.go
```diff
@@ -0,0 +1,67 @@
...
+		// mapping from previous Google authentication types
+		// and current authentication types from Databricks Go SDK
+		oldToNewerAuthType := map[string]string{
+			"google-creds":     "google-credentials",
+			"google-accounts":  "google-id",
+			"google-workspace": "google-id",
+		}
+		newer, ok := oldToNewerAuthType[cfg.AuthType]
+		if ok {
+			tflog.Info(ctx, fmt.Sprintf("Changing required auth_type from %s to %s", cfg.AuthType, newer))
```
rauchy, line 35: Is there any value in specifying sdkv2/pluginfw here? If so, maybe our tflog could be parameterized according to the provider?

rauchy (approved):
This is awesome, thank you!

internal/providers/client/databricks_client.go
mgyucht: This is just to distinguish the log message printed from the plugin framework provider implementation vs the SDKv2 implementation. The same log message is printed on both sides, this way we can tell which log message came from which implementation. We could do something like this with the `additionalFields` parameter of the log message, but it is about the same.

PR #4744, opened 2025-05-23
[Feature] Improve support for new fields in `databricks_pipeline`

alexott:
## Changes

New Go SDK brought new fields `root_path` and `library.glob`, so added doc for them and also made `library.file.path`, `library.glob.include` and `library.notebook.path` required.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/pipelines/resource_pipeline.go
+++ b/pipelines/resource_pipeline.go
@@ -212,6 +212,11 @@ func (Pipeline) CustomizeSchema(s *common.CustomizableSchema) *common.Customizab
 	s.SchemaPath("ingestion_definition", "connection_name").SetForceNew()
 	s.SchemaPath("ingestion_definition", "ingestion_gateway_id").SetForceNew()
 
+	// Required fields
+	s.SchemaPath("library", "glob", "include").SetRequired()
+	s.SchemaPath("library", "notebook", "path").SetRequired()
+	s.SchemaPath("library", "file", "path").SetRequired()
+
 	// Computed fields
 	s.SchemaPath("cluster", "node_type_id").SetComputed()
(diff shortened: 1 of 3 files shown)


pipelines/resource_pipeline.go
```diff
@@ -212,6 +212,11 @@ func (Pipeline) CustomizeSchema(s *common.CustomizableSchema) *common.Customizab
 	s.SchemaPath("ingestion_definition", "connection_name").SetForceNew()
 	s.SchemaPath("ingestion_definition", "ingestion_gateway_id").SetForceNew()
 
+	// Required fields
+	s.SchemaPath("library", "glob", "include").SetRequired()
+	s.SchemaPath("library", "notebook", "path").SetRequired()
+	s.SchemaPath("library", "file", "path").SetRequired()
+
```
tanmay-db, line 219: > `glob`: This field cannot be used together with `notebook` or `file`.

Are all of these required?

mgyucht (approved):

pipelines/resource_pipeline.go
mgyucht:
```
	// Only absolute attribute paths, ones starting with top level attribute
	// names, are supported. Attribute paths cannot be accurately declared
	// for TypeList (if MaxItems is greater than 1), TypeMap, or TypeSet
	// attributes. To reference an attribute under a single configuration block
	// (TypeList with Elem of *Resource and MaxItems of 1), the syntax is
	// "parent_block_name.0.child_attribute_name".
```
I don't think you can do this because a list of libraries would fall under the "TypeList" exception for ConflictsWith.

PR #4763, opened 2025-06-02
Make `databricks_entitlements` forward-compatible with new entitlements

mgyucht:
## Changes
The current implementation of `databricks_entitlements` crashes if an unexpected entitlement is returned from the API. This makes it impossible for new entitlements to be introduced and available by default. This is different from how most resources behave, where if new fields are added to resources in the API response, they are simply dropped rather than causing the provider to crash.

This PR is a clean-up of the handling of entitlements in the provider, both in the `databricks_entitlements` resource and the user, group and SP resources and data sources that expose entitlements. The entitlements are defined as a single `entitlements` structure which can be embedded in every SCIM structure. This makes it possible to use the standard StructToSchema, DataToStructPointer and StructToData methods that are ubiquitous throughout the SCIM API.

I removed the `entitlements` type alias and just use `[]ComplexValue` in all places where that is used to minimize confusion with the new `entitlements` type. The methods on that type are no longer needed because the corresponding interactions are handled by the aforementioned methods.

As part of this cleanup, I have also ported the extra fields for the User, SP, Group and Entitlements resources to the struct definition, relying on StructToSchema to handle the transformation. This keeps the treatment of these APIs more uniform w.r.t. other resources on the plugin framework. 

## Tests
This should be a no-op change, so no new tests are needed.

The schema of the provider should not change as a result of this PR.


scim/resource_user.go
```diff
@@ -23,82 +23,47 @@ const (
 	userAttributes = "userName,displayName,active,externalId,entitlements"
 )
 
-func setCommonUserFields(d *schema.ResourceData, user User, username string) {
-	d.Set("display_name", user.DisplayName)
-	d.Set("active", user.Active)
-	d.Set("external_id", user.ExternalID)
-	d.Set("home", fmt.Sprintf("/Users/%s", username))
```
alexott, line 30: We need to make sure that all these fields are set in both user and SP. I recently fixed the problem in exporter that was caused by missing `external_id` when you import the resource

scim/entitlements.go
```diff
@@ -0,0 +1,59 @@
...
+		case "allow-cluster-create":
+			e.AllowClusterCreate = true
+		case "allow-instance-pool-create":
+			e.AllowInstancePoolCreate = true
+		case "databricks-sql-access":
+			e.DatabricksSQLAccess = true
+		case "workspace-access":
+			e.WorkspaceAccess = true
+		default:
+			tflog.Info(ctx, fmt.Sprintf("Ignoring unknown entitlement: %s", c.Value))
```
alexott: should it be warning?

alexott (approved):
in general, lgtm. just small comments

alexott:
@mgyucht can you look with my changes?

PR #4768, opened 2025-06-05
Update SDK to latest OpenAPI spec

parthban-db:
## Changes

This PR Bumps Go SDK and Update terraform to latest OpenAPI Spec.

## Tests

Existing CI.

### DIFF
--- a/exporter/util_test.go
+++ b/exporter/util_test.go
@@ -58,7 +58,7 @@ func TestAddAwsMounts(t *testing.T) {
 var (
 	userListIdUsernameFixture = qa.HTTPFixture{
 		Method:   "GET",
-		Resource: "/api/2.0/preview/scim/v2/Users?attributes=id%2CuserName&count=100&startIndex=1",
+		Resource: "/api/2.0/preview/scim/v2/Users?attributes=id%2CuserName&count=10000&startIndex=1",
 		Response: iam.ListUsersResponse{
 			Resources: []iam.User{
 				{
@@ -73,7 +73,7 @@ var (
 	}
 	userListIdUsernameFixture2 = qa.HTTPFixture{
 		Method:   "GET",
-		Resource: "/api/2.0/preview/scim/v2/Users?attributes=id%2CuserName&count=100&startIndex=2",
+		Resource: "/api/2.0/preview/scim/v2/Users?attributes=id%2CuserName&count=10000&startIndex=2",
 		Response: iam.ListUsersResponse{
 			Resources:    []iam.User{},
 			TotalResults: 1,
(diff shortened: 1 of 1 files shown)


catalog/permissions/permissions.go
```diff
@@ -26,15 +26,15 @@ func NewUnityCatalogPermissionsAPI(ctx context.Context, m any) UnityCatalogPermi
 	return UnityCatalogPermissionsAPI{client, ctx}
 }
 
-func (a UnityCatalogPermissionsAPI) GetPermissions(securable catalog.SecurableType, name string) (list *catalog.PermissionsList, err error) {
```
parthban-db, line 29: @hectorcast-db , @rauchy, can you check the changes from PermissionsList to GetPermissionsResponse? This is necessary because the Grants Get Method interface has changed in the latest version of the SDK. After the change, we started returning next-page-token in the get API, but we already had page-token in this method's request. So, this could have happened because of the migration. So, this change should not change any behavior.

https://github.com/databricks/databricks-sdk-go/blob/56af964ae7d5f86bd454dfdba4edac42aac4d19f/service/catalog/api.go#L1033

hectorcast-db: Just a rename.

The reason is UpdatePermission and GetPermission had two different proto message as responses, but a single s schema in OpenAPI. So we renamed them to decouple them and avoid future issues.

docs/resources/quality_monitor_v2.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+-> **Note** This resource can only be used with an workspace-level provider!
+
+
+## Example Usage
+```hcl
+resource "databricks_quality_monitor_v2" "this" {
+  object_type = "schema"
+  object_id = "ecb4f03c-f6c9-4d84-8934-e50e087d2435
```
alexott, line 15: it's not very user-friendly API imho...

parthban-db: @tanmay-db, I think this is generated documentation. Can you take a look?

parthban-db:
@alexott, can you look at the test failure in the exporter? Thanks!

docs/resources/quality_monitor_v2.md
tanmay-db: Hi @alexott , is the concern regarding object_id specified to some exact value? the examples are added by the team so if any changes are required here then it might be good to start a thread with the team.

tanmay-db: started a thread

alexott: It's not only the example - it's the whole resource.  When user works with UC, it works with catalog/schema/table names, even if they use object_id under the hood.

alexott:
Superced by #4777

PR #4773, opened 2025-06-06
Add `schema_id` to the output of `databricks_schema` resource

mgyucht:
## Changes
Schema now exports a stable UUID identifier in the `schema_id` field. This PR adds this to the model for schema as a computed attribute.

## Tests
Manually verified that the schema_id was populated in state in GCP. I don't want to add a test for this per se since it's possible that it isn't populated in all environments (yet).


alexott:
We also need to add `table_id` to `databricks_sql_table`: https://docs.databricks.com/api/workspace/tables/get#table_id

mgyucht:
@alexott Yeah, I will do that in a follow-up PR.

PR #4778, opened 2025-06-11
Use single-node USER_ISOLATION cluster by default when managing `databricks_sql_permissions` resources

mgyucht:
## Changes
In https://github.com/databricks/terraform-provider-databricks/commit/27ad78de516e5baf98c64ae92e7815faecc9e153, we modified the default cluster launched by the `databricks_sql_permissions` resource to use a cluster with a worker node. This was done to mitigate an issue where it was no longer possible to create single-node clusters with legacy Table ACLs. Since the start of May, single-node `USER_ISOLATION` clusters are supported and can be used to manage Table ACLs. This PR changes this resource to use such a cluster going forward.

## Tests
Covered by existing integration tests.


access/resource_sql_permissions.go
```diff
@@ -281,22 +281,14 @@ func (ta *SqlPermissions) getOrCreateCluster(clustersAPI clusters.ClustersAPI) (
...
-			DataSecurityMode:       "LEGACY_TABLE_ACL",
-			// TODO: return back after backend fix is rolled out
-			NumWorkers: 1,
-			// SparkConf: map[string]string{
-			// 	"spark.databricks.cluster.profile": "singleNode",
-			// 	"spark.master":                     "local[*]",
-			// },
-			// CustomTags: map[string]string{
-			// 	"ResourceClass": "SingleNode",
-			// },
```
alexott, line 299: Shouldn't we still specify the settings for single-node clusters?

access/resource_sql_permissions.go
```diff
@@ -281,22 +281,15 @@ func (ta *SqlPermissions) getOrCreateCluster(clustersAPI clusters.ClustersAPI) (
...
+			DataSecurityMode:       "USER_ISOLATION",
+			IsSingleNode:           true,
```
alexott, line 292: IsSingleNode should work only if you specify `kind` that is in preview... I think that it's better to stuck with the "traditional way" via spark conf, etc.  Also, right now there is a configuration drift (we have PR for it)

alexott:
Closing in favor of #4813

PR #4796, opened 2025-06-16
[Feature] Add `key` argument to `databricks_jobs` data source to support duplicate job names

### LINKED ISSUE #4266: [FEATURE] A `databricks_jobs` data source that allows duplicate names
### Use-cases

The data source [databricks_jobs API](https://github.com/databricks/terraform-provider-databricks/blob/main/jobs/data_jobs.go) currently raises an exception if jobs with duplicate names are deteced.

This seems a bit confusing since the Databricks API, asset bundles, and other SDKs in multiple languages allow creation of jobs with duplicate names. 

**This request is to expose a flag to this source that allows duplicate names**, or provide another data source that allows retrieval of all `job_id` values in a workspace. In our use case, we only care about the list of IDs of jobs in our workspace.

### Attempted Solutions

Our current workaround is to delete any jobs with duplicate names in our workspace. But we can't prevent new ones accidentally being prevented by developers by accident, given this limitation isn't in other Databricks SDKs.

AFAIK there otherwise isn't a solution using the existing provider to retrieve a list of `job_id` values in a workspace if the workspace has jobs with duplicated names.

### Proposal

Given that the return value of the current [databricks_jobs](https://github.com/databricks/terraform-provider-databricks/blob/main/jobs/data_jobs.go) data source is a map of `name -> Id`, a new data source might be a better option to ensure backwards compatibility of the existing data source, otherwise the key/value structure of the existing map would need to be reversed.

It'd be similar to the existing implementation except the map would use the `Id` as a key and remove the exception. If this request is something that the maintainers would be happy with, I'd be happy to raise a PR.

### References

_I've not found any duplicate or similar issues._

zcking:
## Changes
Added support for handling jobs with duplicate names in the `databricks_jobs` data source by introducing a new `key` parameter that allows users to choose between using job names (default) or job IDs as the mapping keys. This maintains backward compatibility while providing flexibility for workspaces with duplicate job names.

Key changes:
- Added `key` parameter with options "name" (default) or "id"
- When `key = "id"`, the data source uses job IDs as keys, allowing duplicate job names
- When `key = "name"` (default), maintains existing behavior of erroring on duplicate names
- Added test coverage for new functionality

Close #4266 

## Tests

- [X] `make test` run locally

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking Changes
 
 ### New Features and Improvements
+* Added `key` argument to `databricks_jobs` data source to enable mapping by job ID and allow duplicate job names ([#4796](https://github.com/databricks/terraform-provider-databricks/pull/4796)).
 
 ### Bug Fixes
 


mgyucht (approved):
Thank you for making this PR!

alexott (approved):

PR #4797, opened 2025-06-16
Fix Model Serving Endpoint Tests

mgyucht:
## Changes
A behavior change since Friday in the Serving Endpoints API causes Terraform tests to fail. The API now returns min and max provisioned concurrency attributes which are considered optional parameters, even if a user doesn't specify their value. As a result, there is a diff between state with those attributes present and config with those attributes absent.

To fix this, I've just added these attributes with their default values in our test cases, which eliminates the diff.

## Tests
This is tests.

NO_CHANGELOG=true

### DIFF
--- a/internal/providers/pluginfw/products/app/resource_app_acc_test.go
+++ b/internal/providers/pluginfw/products/app/resource_app_acc_test.go
@@ -43,8 +43,8 @@ const baseResources = `
 		config {
 			served_entities {
 				name = "prod_model"
-				model_name = "experiment-fixture-model"
-				model_version = "1"
+				entity_name = "experiment-fixture-model"
+				entity_version = "1"
 				workload_size = "Small"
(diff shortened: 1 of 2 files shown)


alexott:
Shouldn't we mark them as computed as well? Otherwise it will cause the configuration drift for users

mgyucht:
@alexott I'm going to follow-up with the team first to see if they can modify their API's behavior to not return default values when unset by user. Otherwise we'll be in the position where people can't reset to the default value when removing these arguments from their resource definition.

alexott:
> Error: cannot create model serving: Only one of provisioned concurrency or workload size should be defined.

alexott:
@mgyucht it looks like it works now without any code changes

mgyucht:
Behavior change was reverted, so we can close this.

PR #4802, opened 2025-06-18
[Feature] Enable the option to add `comment` when defining a new `databricks_share`

rudolfsberzins:
## Changes
Added the option for the user to define a `comment` when creating `databricks_share` which will be displayed in Databricks UI for defined share.

Additionally, made `share.md` documentation more consistent with Terraform style.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


sharing/resource_share.go
```diff
@@ -228,7 +230,17 @@ func ResourceShare() common.Resource {
...
-			if !d.HasChangeExcept("owner") {
+			if d.HasChange("comment") {
+				_, err = client.Shares.Update(ctx, sharing.UpdateShare{
+					Name:    afterSi.Name,
+					Comment: afterSi.Comment,
+				})
+				if err != nil {
+					return err
+				}
+			}
```
alexott: Why can't we do this update together with Update call below or with update of `owner`?

rudolfsberzins: I was trying to emulate how it was being done for `owner`, because this is my first time working in Golang. But after some investigation, it's a fair point that it could be one call. I have reworked it so that both `owner` and `comment` updates are done together with `changes` update call below.

rudolfsberzins:
@alexott Thanks for the help on this PR 👍

PR #4807, opened 2025-06-20
[DOCS] added example of granting Account Admin role to a service principal

zcking:
## Changes

This PR makes changes to docs only, no code changes. The `databricks_service_principal_role` is capable of giving account admin role to a service principal; however, this was not very clear in the docs whereas we do have an example given in `databricks_user_role` for users. 

I added an example demonstrating this capability of setting `role = "account_admin"` and a note that it can only be used with an account-level provider.

## Tests

### DIFF
--- a/docs/resources/service_principal_role.md
+++ b/docs/resources/service_principal_role.md
@@ -46,7 +46,7 @@ resource "databricks_service_principal_role" "tf_admin_account" {
 The following arguments are supported:
 
 * `service_principal_id` - (Required) This is the id of the [service principal](service_principal.md) resource.
-* `role` -  (Required) This is the id of the role or [instance profile](instance_profile.md) resource.
+* `role` -  (Required) This is the role name, role id, or [instance profile](instance_profile.md) resource.
 
 ## Attribute Reference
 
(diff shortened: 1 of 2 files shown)


alexott (approved):

docs/resources/service_principal_role.md
```diff
@@ -26,6 +26,21 @@ resource "databricks_service_principal_role" "my_service_principal_instance_prof
...
+resource "databricks_service_principal" "tf_admin" {
+  display_name = "Terraform Admin"
+}
+
+resource "databricks_service_principal_role" "tf_admin_account" {
+  service_principal_id = databricks_service_principal.tf_admin.id
+  role                 = "account_admin"
+}
+```
+
```
alexott, line 43: It would be very useful to add a callout pointing to https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/access_control_rule_set#role-1 and saying that other roles could be set with `databricks_access_control_rule_set`

alexott (commented):
also, it would be useful to modify description of the `role` argument to say that it's role name, role id or instance_profile.

Also, it would be useful to add `databricks_access_control_rule_set` to the list of related resources

PR #4805, opened 2025-06-19
[Internal] Bump Go SDK to v0.73 & Generate Latest TF Structs

rauchy:
## Tests

- [ ] `make test` run locally
- [ ] relevant change in `docs/` folder
- [ ] covered with integration tests in `internal/acceptance`
- [ ] using Go SDK
- [ ] using TF Plugin Framework

PR #4809, opened 2025-06-23
[Internal] Bump Go SDK to v0.73.1 & Generate Latest TF Structs

rauchy:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,7 +9,7 @@
 * Added `time_rotating` argument to `databricks_service_principal_secret` to enable secret rotation ([#4789](https://github.com/databricks/terraform-provider-databricks/pull/4789)).
 * Add AWS support for `databricks_mws_ncc_private_endpoint_rule` ([#4804](https://github.com/databricks/terraform-provider-databricks/pull/4804)).
 * Added `key` argument to `databricks_jobs` data source to enable mapping by job ID and allow duplicate job names ([#4796](https://github.com/databricks/terraform-provider-databricks/pull/4796)).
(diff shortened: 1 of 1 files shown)


NEXT_CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 * Added `time_rotating` argument to `databricks_service_principal_secret` to enable secret rotation ([#4789](https://github.com/databricks/terraform-provider-databricks/pull/4789)).
 * Add AWS support for `databricks_mws_ncc_private_endpoint_rule` ([#4804](https://github.com/databricks/terraform-provider-databricks/pull/4804)).
 * Added `key` argument to `databricks_jobs` data source to enable mapping by job ID and allow duplicate job names ([#4796](https://github.com/databricks/terraform-provider-databricks/pull/4796)).
+* Added online_store resource and data sources ([#4805](https://github.com/databricks/terraform-provider-databricks/pull/4805)).
```
mgyucht: Will we close that PR? Should we update this to point to this PR?
```suggestion
* Added online_store resource and data sources ([#4809](https://github.com/databricks/terraform-provider-databricks/pull/4809)).
```

mgyucht (approved):
One small nit on the changelog, otherwise LGTM.

I notice that some fields for autogenerated resources are listed in the documentation, but their description is empty. Sometimes, this is because the description hasn't been provided in the API definition, and we should follow-up with teams on this (I know @tanmay-db is working on doc enforcement for TF as part of private preview launches for resources). Separately, it can also be because there is no cloud-agnostic description of a field. Will follow-up with him offline about addressing this gap.

NEXT_CHANGELOG.md
rauchy: #4805 will close.

PR #4817, opened 2025-06-26
Add documentation for single-node clusters in `databricks_cluster`

mgyucht:
## Changes
The `is_single_node` field was recently introduced for clusters to simplify creation of single-node clusters. This PR updates the Terraform documentation for `databricks_cluster` to show how to configure the cluster resource using this new flag. This simplifies the process from before, which required a number of settings to be configured simultaneously.

## Tests
Manually tested by creating a cluster with the documented template.

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -98,16 +98,7 @@ When you [create a Databricks cluster](https://docs.databricks.com/clusters/conf
 * `min_workers` - (Optional) The minimum number of workers to which the cluster can scale down when underutilized. It is also the initial number of workers the cluster will have after creation.
 * `max_workers` - (Optional) The maximum number of workers to which the cluster can scale up when overloaded. max_workers must be strictly greater than min_workers.
 
-When using a [Single Node cluster](https://docs.databricks.com/clusters/single-node.html), `num_workers` needs to be `0`. It can be set to `0` explicitly, or simply not specified, as it defaults to `0`.  When `num_workers` is `0`, provider checks for presence of the required Spark configurations:
-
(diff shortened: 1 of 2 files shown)


docs/resources/cluster.md
```diff
@@ -98,16 +98,7 @@ When you [create a Databricks cluster](https://docs.databricks.com/clusters/conf
...
-
-* `spark.master` must have prefix `local`, like `local[*]`
-* `spark.databricks.cluster.profile` must have value `singleNode`
-
-and also `custom_tag` entry:
-
-* `"ResourceClass" = "SingleNode"`
-
-The following example demonstrates how to create an single node cluster:
+To create a [single node cluster](https://docs.databricks.com/clusters/single-node.html), set `is_single_node = true` and `kind = "CLASSIC_PREVIEW"` for the cluster. Single-node clusters are suitable for small, non-distributed workloads like single-node machine learning use-cases.
```
alexott, line 101: I don't remember if `kind` still requires explicit enablement in previews or not.
Also, `kind` may not be supported in GovCloud, etc. So it's best to leave both examples (new and old way of doing it).

mgyucht: Let me check with the team, I'll get back to you. I think the feature has been released for a while now, so I suspect it has been rolled out everywhere, but I don't know for sure.

kunyaoliu: @alexott 

kind is available via API for all doesn't require enablement.

- simple cluster form (currently on stage 3-2 + enabled for all group cluster customers) default having kind.
- classic cluster form doesn't submit with kind
- kind and w/o are both supported from API

alexott (approved):

PR #4825, opened 2025-06-30
[Internal] Enable Apps integration tests on GCP

alexott:
## Changes

Apps are now available on GCP, so it makes sense to enable tests there as well

NO_CHANGELOG=true

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott:
App test fails on GCP with the error `Error: cannot create model serving: Workspace Model Registry serving is deprecated. Please use Unity Catalog models for serving.` We need to register `experiment-fixture-model` in UC as well, not only in the model registry.

tanmay-db (approved):

nkvuong (approved):

alexott:
@nkvuong we need a registered UC model to make it working on GCP

PR #4834, opened 2025-07-04
Add synthetic field to control `apply_policy_default_values`

pietern:
## Changes

This commit introduces a new synthetic field, `__apply_policy_default_values_allow_list`, to the job resource schema. This field is a list of strings representing the cluster config attributes that the user explicitly wants to manage. During create and update operations, only the fields listed in this attribute will be included in the API payload for the job's cluster configuration. All other fields will be omitted, allowing the backend to apply or update cluster policy defaults as needed (when `apply_policy_default_values = true`).

This approach works around a limitation in the legacy Terraform Plugin SDK, where it is not possible to distinguish between user-specified and server-defaulted fields after the initial create. By requiring users to explicitly list the fields they wish to control, we enable them to reset cluster config fields to new policy defaults by removing them from both the config and the synthetic list.

- Adds `__apply_policy_default_values_allow_list` to the job resource schema.
- Filters cluster config fields in the API payload based on this list.
- Documents the new behavior and usage pattern for users.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


jobs/jobs_api_go_sdk.go
```diff
@@ -174,21 +218,21 @@ func updateJobClusterSpec(clusterSpec *compute.ClusterSpec, d *schema.ResourceDa
...
+		if js.Tasks[i].NewCluster != nil {
+			err := updateJobClusterSpec(d, fmt.Sprintf("task.%d.new_cluster.0", i), js.Tasks[i].NewCluster)
 			if err != nil {
 				return err
 			}
 		}
 	}
 	for i := range js.JobClusters {
-		err := updateJobClusterSpec(&js.JobClusters[i].NewCluster, d)
+		err := updateJobClusterSpec(d, fmt.Sprintf("job_cluster.%d.new_cluster.0", i), &js.JobClusters[i].NewCluster)
```
pietern, line 235: These changes were necessary for adding a `schema.ResourceData` prefix argument.

jobs/jobs_api_go_sdk.go
```diff
@@ -199,16 +243,16 @@ func prepareJobSettingsForUpdateGoSdk(d *schema.ResourceData, js *JobSettingsRes
...
+		if jc.Tasks[i].NewCluster != nil {
+			err := updateJobClusterSpec(d, fmt.Sprintf("task.%d.new_cluster.0", i), jc.Tasks[i].NewCluster)
 			if err != nil {
 				return err
 			}
 		}
 	}
 	for i := range jc.JobClusters {
-		err := updateJobClusterSpec(&jc.JobClusters[i].NewCluster, d)
+		err := updateJobClusterSpec(d, fmt.Sprintf("job_cluster.%d.new_cluster.0", i), &jc.JobClusters[i].NewCluster)
```
pietern, line 255: These changes were necessary for adding a `schema.ResourceData` prefix argument.

jobs/jobs_api_go_sdk.go
```diff
@@ -156,8 +156,52 @@ func (c controlRunStateLifecycleManagerGoSdk) OnUpdate(ctx context.Context) erro
...
+	v, ok := d.GetOk(prefix + "." + applyPolicyDefaultValuesAllowListField)
+	if !ok {
+		return nil
+	}
+
+	// Load as a list of strings.
+	var allowList []string
+	for _, v := range v.([]any) {
+		allowList = append(allowList, v.(string))
+	}
```
tanmay-db, line 186: Should we have a check for `v.(string)` to avoid panic?

pietern: It is guaranteed to be a string per the schema below. I looked for precedent in the codebase and it appears there are more places where we do a hard type assertion for values pulled from `schema.ResourceData`.

I agree it's better to be defensive, but we should update all call sites to use type assertions then.

pietern (commented):
Will follow up with an integration test.

I think skipping docs here is better because I rather not encourage people to use this unless they really really need to use it. The field is only necessary until the migration to the plugin framework completes.

Issue #4836, opened 2025-07-08
[DOC] Incorrect resource name in example for `databricks_alert_v2`


samwhitehall:
### Affected Resource(s)
[databricks_alert_v2 basic alert example](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/alert_v2#basic-alert-example)

### Expected Details
Expecting to see correct resource type in example code. It seems it should be `databricks_alert_v2` instead of `databricks_sql_alert `.

<img width="686" height="301" alt="Image" src="https://github.com/user-attachments/assets/5c5a07f7-895f-4c9a-b73e-2a28dc373cd7" />

### List of things to potentially add/remove
See above.

### Important Factoids

N/A

### References

N/A

nagaarjun-p:
That's a good catch, @samwhitehall . I have created a PR to fix this #4838

PR #4838, opened 2025-07-08
{MINOR][DOCS} Update alert_v2.md

nagaarjun-p:
Replaced the `databricks_sql_alert` with the `databricks_alert_v2` resource in the sample code.

## Changes

This change resolves the issue #4836 . We are just fixing the name of the resource used in the sample code.

## Tests
Ensured that the markdown file displays the sample code properly in my local machine.
I need some support if any further testing is required.

- [ ] `make test` run locally
- [ ] using Go SDK


alexott:
The documentation for this resource is generated automatically, so it should be fixed in the API spec and regenerated...

Cc: @tanmay-db

tanmay-db:
Hi @nagaarjun-p, thanks for reaching out. Yes @alexott, you are right. I will follow up with the API owning team.

Issue #4195, opened 2024-11-06
[ISSUE] Issue with `databricks_secret_acl` resource - Cannot work with account level groups


seblatre:
### Configuration

```hcl
terraform {
  required_providers {
    databricks = {
      source  = "databricks/databricks"
      version = "~> 1.56.0"
    }
  }
}
provider "databricks" {
  alias      = "account"
  host       = "https://accounts.azuredatabricks.net"
  account_id = "<databricksAccountId>"
}
provider "databricks" {
  alias                       = "workspace"
  host                        = "adb-<databricksWorkspaceId>.10.azuredatabricks.net"
  azure_workspace_resource_id = "/subscriptions/<subId>/resourceGroups/<rgName>/providers/Microsoft.Databricks/workspaces/<databricksWorkspaceName>"
}
resource "databricks_group" "account_group" {
  display_name = "GROUP-A45"

  provider = databricks.account
}
resource "databricks_mws_permission_assignment" "workspace_group" {
  workspace_id = "<databricksWorkspaceId>"
  principal_id = databricks_group.account_group.id
  permissions  = ["USER"]

  provider = databricks.account
}
resource "databricks_secret_scope" "team" {
  name = "Scope_A45_team"

  keyvault_metadata {
    resource_id = "/subscriptions/<subId>/resourceGroups/<rgName>/providers/Microsoft.KeyVault/vaults/keyvaulttesta45"
    dns_name    = "https://keyvaulttesta45.vault.azure.net/"
  }

  provider = databricks.workspace
}
resource "databricks_secret_acl" "team_acl" {
  principal  = databricks_group.account_group.display_name
  permission = "READ"
  scope      = databricks_secret_scope.team.name

  provider = databricks.workspace
}
```

### Expected Behavior
This configuration should work without issue

### Actual Behavior
When running {{terraform apply}}, we run into the following error
```
databricks_group.account_group: Creating...
databricks_secret_scope.team: Creating...
databricks_group.account_group: Creation complete after 2s [id=980426357238593]
databricks_mws_permission_assignment.workspace_group: Creating...
databricks_secret_scope.team: Creation complete after 4s [id=Scope_A45_team]
databricks_secret_acl.team_acl: Creating...
databricks_mws_permission_assignment.workspace_group: Creation complete after 5s [id=25518447772330|980426357238593]
╷
│ Error: cannot create secret acl: User or Group GROUP-A45 does not exist.
│
[...]

PR #4839, opened 2025-07-08
[Fix] Add check and  retry logic for secret ACL creation because of API inconsistency for edge cases

elghali97:
## Changes
@alexott has pushed a fix (#4548) for Secret ACL assignment (fails when it happens right after a principal assignment to a workspace). We tried first to ignore Read after resource creation as no additional data is returned by GET Secret Scope ACL API but the API is inconsistent and sometimes the ACL is not applied even if the API returns status code 200 (it's also the case when we recreate a group from the workspace to the account level with the same name, the legacy group id is retained in the cache and the assignment is then not consistent). I propose a simple check and retry mechanism  while the eng team are fixing the API inconsistency issue.

This has been requested by Michelin team (see issue [4195](https://github.com/databricks/terraform-provider-databricks/issues/4195))

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


secrets/resource_secret_acl_test.go
```diff
@@ -135,15 +174,211 @@ func TestResourceSecretACLCreate_ScopeWithSlash(t *testing.T) {
 }
 
 func TestResourceSecretACLCreate_Error(t *testing.T) {
+	// Set retry interval to 1ms for faster testing
+	os.Setenv("DATABRICKS_SECRET_ACL_TEST_RETRY_INTERVAL_MS", "1")
+	defer os.Unsetenv("DATABRICKS_SECRET_ACL_TEST_RETRY_INTERVAL_MS")
+	
+	// Add 3 failures to test that all 3 retries are attempted
+	fixtures := []qa.HTTPFixture{}
+	for i := 0; i < 3; i++ {
```
alexott, line 183: You don't need loop - just add `ReuseRequest: true,` to a fixture

secrets/resource_secret_acl.go
```diff
@@ -33,23 +37,66 @@ func ResourceSecretACL() common.Resource {
 	}
 	return common.Resource{
 		Schema: s,
-		CanSkipReadAfterCreateAndUpdate: func(_ *schema.ResourceData) bool {
-			return true
-		},
```
alexott: We still need to avoid calling `Read` after a successful `Create`.

secrets/resource_secret_acl.go
```diff
@@ -33,23 +37,66 @@ func ResourceSecretACL() common.Resource {
...
+			retryInterval := 90 * time.Second
+			
+			// Allow tests to override retry interval
+			if testInterval := os.Getenv("DATABRICKS_SECRET_ACL_TEST_RETRY_INTERVAL_MS"); testInterval != "" {
+				if ms, err := strconv.Atoi(testInterval); err == nil {
+					retryInterval = time.Duration(ms) * time.Millisecond
+				}
+			}
+			
+			for attempt := 1; attempt <= maxRetries; attempt++ {
```
alexott: we can use `retry` package for it: https://github.com/databricks/terraform-provider-databricks/blob/main/vectorsearch/resource_vector_search_index.go#L22

secrets/resource_secret_acl.go
```diff
@@ -33,23 +37,66 @@ func ResourceSecretACL() common.Resource {
...
+				// Attempt to create the ACL
+				err = w.Secrets.PutAcl(ctx, req)
```
alexott, line 65: Shouldn't we do `PutAcl` once and then read it multiple times?

secrets/resource_secret_acl_test.go
elghali97: Will do thanks !

secrets/resource_secret_acl.go
elghali97: ok will delete this part.

secrets/resource_secret_acl.go
elghali97: will do thanks

secrets/resource_secret_acl.go
elghali97: Nop sometimes the `PutAcl` can send back a 200 status to when attaching the Acl to the wrong (cached) principal. That's why I'm updating the ACL and see if it's assigned to the right principal and not the one in cache (mainly because we attach the ACL using the principal name and not the principal id).

alexott:
@elghali97 please run `make fmt; make lint`

elghali97:
Reopenning it.

elghali97:
@renaudhartert-db can you please have a look at this PR ?

alexott:
Supersed by #4885

alexott:
Supersed by #4885

PR #4785, opened 2025-06-12
Disable queueing by default in databricks_job

mgyucht:
## Changes
`databricks_job` disables queueing by default so as not to cause a behavior change for existing `databricks_job` instances when upgrading the provider to later versions that support queueing. Before version 1.71.0 of the provider, the `databricks_job` resource was backed by the 2.1 version of the Jobs API, which supports queueing but was disabled by default. If queue was not specified, it was also not returned by the API:

```
2025-06-12T11:55:12.994Z [DEBUG] provider.terraform-provider-databricks_v1.70.0: POST /api/2.1/jobs/create
> {
>   "max_concurrent_runs": 1,
>   "name": "test-job",
>   "tasks": [
>     {
>       "notebook_task": {
>         "notebook_path": "<PATH>"
>       },
>       "task_key": "task-1"
>     }
>   ]
> }
< HTTP/2.0 200 OK
< {
<   "job_id": <id>
< }
2025-06-12T11:55:13.250Z [DEBUG] provider.terraform-provider-databricks_v1.70.0: GET /api/2.1/jobs/get?job_id=<ID>
< HTTP/2.0 200 OK
< {
<   ...
<   "settings": {
<     "email_notifications": {},
<     "format": "MULTI_TASK",
<     "max_concurrent_runs": 1,
<     "name": "test-job",
<     "tasks": [
<       {
<         "email_notifications": {},
<         "notebook_task": {
<           "notebook_path": "<PATH>",
<           "source": "WORKSPACE"
<         },
<         "run_if": "ALL_SUCCESS",
<         "task_key": "task-1",
<         "timeout_seconds": 0
<       }
<     ],
<     "timeout_seconds": 0,
<     "webhook_notifications": {}
<   }
< }
```

Starting in 1.71.0, this resource is now backed by the 2.2 API, which enables queueing by default. Because of the old behavior, the disabled queue is not stored in the state and not sent back in the Reset API call. With the new endpoint, the Jobs service suddenly enables queueing for the job:

```
2025-06-12T11:56:43.247Z [DEBUG] provider.terraform-provider-databricks_v1.71.0: POST /api/2.2/jobs/reset
> {
>   "job_id": <ID>,
>   "new_settings": {
>     "email_notifications": {},
>     "format": "MULTI_TASK",
>     "max_concurrent_runs": 1,
>     "name": "test-job",
>     "run_as": {
>       "user_name": "<USERNAME>"
>     },
>     "tasks": [
>       {
>         "email_notifications": {},
[...]

PR #4840, opened 2025-07-09
Disable queue by default when updating `databricks_job` 

hectorcast-db:
## Changes
Disable queue by default when updating `databricks_job`. This is done for new resources, but needs to be done also during update to support exhisting resources created before Terraform 1.71.0

## Tests

- [x] `make test` run locally
- [x] using Go SDK


alexott:
what is the difference from #4785

hectorcast-db:
> what is the difference from #4785

Same thing. @mgyucht is on a leave and this ticket was redirected to me. I missed that he opened a PR. I will close this and merge the other.

PR #4846, opened 2025-07-10
Promote Plugin Framework Share Resource to Production

rauchy:
## Changes

This PR promotes the plugin framework implementation of the `databricks_share` resource to be the default implementation, replacing the SDK v2 version. You can now use `databricks_share` to access the plugin framework implementation, with an opt-out mechanism available for those who wish to continue using the legacy SDK v2 version.

### Rollback Instructions
If you encounter issues with the new plugin framework implementation, you can temporarily revert to the SDK v2 implementation by setting an environment variable:

```bash
export USE_SDK_V2_RESOURCES=databricks_share
```
For data sources, use:
```bash
export USE_SDK_V2_DATA_SOURCES=databricks_share,databricks_shares
```
## Tests

Updated all test files to use `databricks_share` instead of `databricks_share_pluginframework`.

- [x] `make test` run locally
- [x] using Go SDK


tanmay-db (commented):
Hi @rauchy, thanks for the PR, could you please add test to test the migration between SDKv2 and Plugin framework? 

Since the resource is used by a lot of customers, we should make sure there are no behavior changes when migrating to the plugin framework.
https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#testing-migration
> During migration, we recommend writing tests to verify that switching from SDKv2 to the Framework has not affected your provider's behavior. These tests use identical configuration. The first test step applies a plan and generates state with the SDKv2 version of the provider. The second test step generates a plan with the Framework version of the provider and verifies that the plan is a no-op, indicating that migrating to the framework has not altered behaviour.

rauchy:
Good idea @tanmay-db! Added in 21cd6ae.

NEXT_CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 * Added output attribute `endpoint_url` in `databricks_model_serving` ([#4877](https://github.com/databricks/terraform-provider-databricks/pull/4877)).
 * Deprecate `egg` library type in `databricks_cluster`, `databricks_job`, and `databricks_library` ([#4881](https://github.com/databricks/terraform-provider-databricks/pull/4881)).
+* Added output attribute `endpoint_url` in `databricks_model_serving`([#4877](https://github.com/databricks/terraform-provider-databricks/pull/4877)).
```
alexott: This is a duplicate of the line number 9

exporter/importables.go
```diff
@@ -1990,31 +1989,36 @@ var resourcesMap map[string]importable = map[string]importable{
...
-			common.DataToStructPointer(r.Data, s, &share)
-			// TODO: how to link recipients to share?
+			// Handle share objects by reading them directly from the data
+			// since the resource has been migrated to pluginfw
 			ic.emitUCGrantsWithOwner("share/"+r.ID, r)
-			for _, obj := range share.Objects {
-				switch obj.DataObjectType {
+
+			// Get objects array from the resource data
+			objectsList := r.Data.Get("object").([]any)
```
alexott: `r.Data` will still be using the SDKv2 implementation under the hood, so I'm not sure if there is a benefit of using `.Get` to get raw data instead of doing `common.DataToStructPointer`

tanmay-db (approved):

PR #4859, opened 2025-07-15
[Doc] Document additional custom properties for resource `workspace_conf`

abdulmoemoh:
## Changes

Document additional custom properties for resource `workspace_conf`. These custom properties are currently supported, but not available in the current docs. 

## Tests

- [ ] `make test` run locally
- [ X] relevant change in `docs/` folder
- [ ] using Go SDK
- [ X] has entry in `NEXT_CHANGELOG.md` file

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Documentation
 
-* Document additional custom configuration properties for resource `workspace_conf`
+* Document additional custom configuration properties for resource `workspace_conf` ([#4859](https://github.com/databricks/terraform-provider-databricks/pull/4859))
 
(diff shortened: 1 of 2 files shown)


alexott:
@abdulmoemoh, Thank you for your PR, but these configurations aren't officially documented in the main Databricks docs, and they are subject to potential breaking changes; that's why they aren't documented in the Terraform docs. We had similar situation recently with one of the settings.

There is a work in progress on exposing such settings correctly, like this: https://docs.databricks.com/api/azure/workspace/enableexportnotebook/getenableexportnotebook. After that, existing settings could be decommissioned, but it's a decision of the corresponding development team.

Please reach your Databricks account team for further discussion.

PR #4873, opened 2025-07-24
Fix handling of `force` option in `databricks_git_credential`

### LINKED ISSUE #4871: [ISSUE] Issue with `databricks_git_credential` resource
### Configuration

```

resource "databricks_git_credential" "ado" {
    force                 = true
    git_provider          = "azureDevOpsServices"
    git_username          = "USERNAME"
    personal_access_token = (sensitive value)
}

```

### Expected Behavior

The above code shoudl set the credential without any further errors. 

### Actual Behavior

It's failing with the error: "Error: cannot create git credential: A Git credential already exists for provider azureDevOpsServices. Only one credential per provider is allowed for service principals."

v1.85.0
v1.84.0

### Is it a regression?

Was working on v1.84.0 until upgrading the version to v1.85.0

### Debug Output

2025-07-16T23:01:13.8769139Z Error: cannot create git credential: A Git credential already exists for provider azureDevOpsServices. Only one credential per provider is allowed for service principals.
2025-07-16T23:01:13.8769260Z
2025-07-16T23:01:13.8769412Z   with module.repo.databricks_git_credential.ado,
2025-07-16T23:01:13.8769567Z   on .terraform/modules/repo/main.tf line 15, in resource "databricks_git_credential" "ado":
2025-07-16T23:01:13.8769717Z   15: resource "databricks_git_credential" "ado" {

alexott:
With introduction of support for multiple Git credentials the error message did change a bit, so the `force` option may not work in all cases.

Resolves #4871

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


repos/resource_git_credential.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func isOnlyOneGitCredentialError(err error) bool {
+	errStr := err.Error()
+	return (strings.Contains(errStr, "Only one Git credential is supported ") && strings.Contains(errStr, " at this time")) ||
+		strings.Contains(errStr, "Only one credential per provider is allowed")
+}
```
renaudhartert-db: Is there any chance we can depend on the error code instead? We should ideally avoid parsing error messages as these are not expected to remain stable.

alexott: Added. I kept the old logic just in case - IIRC at that time they didn't return 409, that's why it was implemented as check for specific error messages

repos/resource_git_credential.go
```diff
@@ -2,15 +2,23 @@ package repos
...
+	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/service/workspace"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func isOnlyOneGitCredentialError(err error) bool {
+	errStr := err.Error()
+	return errors.Is(err, apierr.ErrResourceConflict) ||
+		(strings.Contains(errStr, "Only one Git credential is supported ") && strings.Contains(errStr, " at this time"))
```
GrantIsEaton, line 19: this is the old error message. Are you intending to parse the new one for SPs + multiple git creds? https://sourcegraph.prod.databricks-corp.com/databricks-eng/universe/-/blob/projects/main/scala/credentials/messages/GitCredentialsMessages.scala?L10

repos/resource_git_credential.go
```diff
@@ -2,15 +2,23 @@ package repos
...
 	"strconv"
 	"strings"
 
+	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/service/workspace"
 	"github.com/databricks/terraform-provider-databricks/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func isOnlyOneGitCredentialError(err error) bool {
```
GrantIsEaton: did you want this error or the old one?
```suggestion
func isOnlyOneGitCredentialForProviderError(err error) bool {
```

repos/resource_git_credential.go
alexott: The main question here is - can we rely only on HTTP 409, or do we still need to rely on error messages?

alexott: The case for single credential for SP is handled now via `errors.Is(err, apierr.ErrResourceConflict)`

rauchy (approved):

PR #4885, opened 2025-07-28
Add workaround for internal caching issue in `databricks_secret_acl`. 

renaudhartert-db:
## Changes

This PR improves the robustness of the `databricks_secret_acl` by working around an internal caching issue that leads to the wrong permissions being applied, see issue #4195. This PR is based on PR #4839.

Resolves: #4195

## Tests

- [x] `make test` run locally
- [x] using Go SDK


tanmay-db (approved):
I think, we should be able to replicate the issue https://github.com/databricks/terraform-provider-databricks/issues/4195 as part of acceptance test: https://github.com/databricks/terraform-provider-databricks/blob/main/secrets/secret_acl_test.go and then over this PR the test should pass.

alexott (approved):

PR #4886, opened 2025-07-28
Do not set `run_as` from `run_as_user_name` in DLT pipelines

### LINKED ISSUE #4897: [ISSUE] Issue with `terraform-provider-databricks version >= 1.65.1` due to run_as feature
### Configuration

```hcl
resource "databricks_pipeline" "pipeline" {
  name = "${var.pipeline_name}_pipeline"

  catalog    = var.catalog_name
  target     = var.schema_name
  serverless = var.use_serverless

  run_as {
    service_principal_name = var.service_principal_application_id
  }

  continuous  = var.continuous
  development = var.development
  channel     = var.channel

}
```
Changes tested:

Adding or updating any pipeline parametres, for example Changing pipeline compute type (serverless → custom cluster or vice versa, Changing channel type, ... ) 

### Expected Behavior
- When running terraform apply after making changes (channel, compute type, etc.), Terraform should:
- Apply all changes successfully on first attempt without error.
- Update the pipeline in Databricks to match the Terraform state.
- Keep Terraform state and Databricks pipeline configuration consistent (no drift).

### Actual Behavior
PS: service principal application ID passed to the run_as isn't null 
- On first terraform apply, the following error is raised:
```hcl
Error: cannot update pipeline: Once configured, run_as of the pipeline can not be set to null. 
Please set run_as to 7d6b363e-6e7c-4522-97e7-********** to preserve the pipeline's run_as
```
- On a second apply, the error no longer appears, but the changes (channel updates, compute type modifications) are not applied to the Databricks pipeline.

- Terraform state is updated as if changes were applied, creating a drift between Databricks' actual pipeline configuration and Terraform state

### Steps to Reproduce
- Deploy a DLT pipeline with Terraform (via databricks_pipeline).

- Update the resource (e.g., add tags or change compute type).

- Run terraform apply.

  - Observe the run_as error.

- Run terraform apply a second time.

  - No error, but Databricks pipeline remains unchanged while Terraform state shows updated values.

- If you manually change the pipeline owner in the Databricks UI before running Terraform, the apply succeeds and applies all changes correctly.

```hcl
Terraform v1.10.x
provider "databricks" v1.65.1
provider "aws" v5.x
```
Affects multiple recent provider versions, since v1.65.1 (e,g,. the v1.65.0, v1.62.0 works fine)

### Is it a regression?
yes, Affects multiple recent provider versions, since v1.65.1 : 
- the v1.65.0, v1.62.0 works fine
- the v1.65.1, v1.66.0 doesn't work

### Debug Output
```hcl
Error: cannot update pipeline: Once configured, run_as of the pipeline can not be set to null. 
Please set run_as to 7d6b363e-6e7c-4522-97e7-************ to preserve the pipeline's run_as
```
Second apply succeeds but pipeline is not updated.

### Important Factoids

- The issue seems related to how pipeline ownership (run_as) is handled during updates.
- Both preview and current channels are affected
- Manual owner change in UI before the tf apply, allows Terraform to apply changes normally.

The issue causes false positives in Terraform state (state says tags are applied but they aren’t).
No, Reporting the bug for maintainers to investigate.

shreyas-goenka:
## Changes

This PR removes custom handling that we have for run_as for DLT:
1. Remove the computed annotation for run_as in Terraform. It's not a server owned field and is instead meant to be set by the client. 
2. Do not read server-side `run_as_user_name` values and set it in run_as. 

This PR reverts some code that was originally introduced in:
1. https://github.com/databricks/terraform-provider-databricks/pull/4456
2. https://github.com/databricks/terraform-provider-databricks/pull/4470

This custom code was added for `run_as` because originally the API never returned values for the run_as field in the GET API call, so Terraform had to depend on values from `run_as_user_name`. Since the backend was patched to return values for run_as, we no longer need the custom code to read `run_as_user_name`.

Removing this custom code fixes the caching of run_as in Terraform state. Before, removing run_as from the TF config would have been a noop. Now, TF will correctly return an error when a user tries to do so.

NO_CHANGELOG=true (The feature is in PrPr)

Resolves #4897

## Tests
Manually created a pipeline with an old version of the provider and checked that upgrading to this PR does not cause Terraform apply to fail. Both when run_as is set and when run is not set.

Existing acceptance tests: TestAccPipelineRunAsUser, TestUcAccPipelineRunAsServicePrincipal, TestUcAccPipelineRunAsMutations

Updated TestUcAccPipelineRunAsMutations to include two new cases:
1. Setting run_as to a pipeline that did not have one configured before works. 
2. Removing run_as causes an error.

-------------------------

Note: This is a breaking change. If a user had a run_as configured on their DLT pipeline, but then removed it at some point they'll start to see the following error when they update their Terraform version:
```
databricks_pipeline.basic_pipeline: Modifying... [id=6f832064-c8b8-4a78-8f6f-f8da5014913c]
╷
│ Error: cannot update pipeline: Once configured, run_as of the pipeline can not be set to null. Please set run_as to fa95fb60-f59b-4b78-832d-83f41da6f284 to preserve the pipeline's run_as
│ 
│   with databricks_pipeline.basic_pipeline,
│   on main.tf line 13, in resource "databricks_pipeline" "basic_pipeline":
│   13: resource "databricks_pipeline" "basic_pipeline" {
│ 
```

To fix they can simply set run_as back again on their pipelines. This is fine because run_as is still in private preview.


pipelines/resource_pipeline.go
```diff
@@ -71,13 +71,6 @@ func Update(w *databricks.WorkspaceClient, ctx context.Context, d *schema.Resour
 	common.DataToStructPointer(d, pipelineSchema, &updatePipelineRequest)
 	updatePipelineRequest.EditPipeline.PipelineId = d.Id()
 	adjustForceSendFields(&updatePipelineRequest.Clusters)
-	// Workspaces not enrolled in the private preview must not send run_as in the update request.
-	// If run_as was persisted in state because of a `terraform refresh`, there will not be a planned change
-	// as long as the user hasn't specified a value.
-	if !d.HasChange("run_as") {
```
shreyas-goenka, line 77: This is only relevant if server-side values are read during "Read". Since this PR no longer sets the `run_as_user_name` value in the `run_as` field, this is no longer necessary:

https://github.com/databricks/terraform-provider-databricks/pull/4470

pipelines/resource_pipeline.go
```diff
@@ -174,7 +167,7 @@ type Pipeline struct {
 	Health               pipelines.GetPipelineResponseHealth `json:"health,omitempty"`
 	LastModified         int64                               `json:"last_modified,omitempty"`
 	LatestUpdates        []pipelines.UpdateStateInfo         `json:"latest_updates,omitempty"`
-	RunAs                pipelines.RunAs                     `json:"run_as,omitempty"`
+	RunAs                *pipelines.RunAs                    `json:"run_as,omitempty"`
```
shreyas-goenka, line 170: This needs to be a pointer to avoid serializing an empty `run_as {}` in generated terraform plans.

andrewnester (approved):

shreyas-goenka:
Integration test failures are unrelated.

alexott (approved):
looks good to me. One q: should we mark `run_as` as `force_new` if we want to go from existing `run_as` to no `run_as` ?

Also, should we update the docs that `run_as` can't be changed to `null` from existing value?

alexott:
@shreyas-goenka even if this feature in PrPr, it makes sense to add a changelog item for it

PR #4914, opened 2025-08-07
[Exporter] Improve handling of new dependencies in jobs, pipelines, model serving

alexott:
## Changes

Changes include:

- handling of email notifications in model serving
- handling of root path, run as, environment in DLT pipelines
- handling of environments in jobs

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/exporter/impl_jobs.go
+++ b/exporter/impl_jobs.go
@@ -268,8 +268,24 @@ func importJob(ic *importContext, r *resource) error {
 		ic.emitIfWsfsFile(param.Default)
 		ic.emitIfVolumeFile(param.Default)
 	}
+	for _, env := range job.Environments {
+		if env.Spec != nil {
+			for _, dep := range env.Spec.Dependencies {
+				emitEnvironmentDependency(ic, dep)
+			}
+		}
+	}
 
 	return ic.importLibraries(r.Data, s)
+}
+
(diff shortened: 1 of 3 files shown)


exporter/impl_jobs.go
```diff
@@ -488,6 +504,12 @@ var (
 		{Path: "parameter.default", Resource: "databricks_workspace_file", Match: "workspace_path"},
 		{Path: "parameter.default", Resource: "databricks_workspace_file", Match: "path"},
 		{Path: "parameter.default", Resource: "databricks_file", Match: "path"},
+		{Path: "environments.spec.dependencies", Resource: "databricks_workspace_file", Match: "workspace_path"},
+		{Path: "environments.spec.dependencies", Resource: "databricks_file"},
```
nkvuong, line 508: should we update the previous line to be consistent with this?
`{Path: "parameter.default", Resource: "databricks_file", Match: "path"},`

alexott: We can, but it's the same behaviour as `path` == `id` for `databricks_file`. I need to go through all dependencies and remove such occurrences, otherwise we can get race conditions

alexott: I will do a follow up PR today with additional changes and will fix it there

PR #4930, opened 2025-08-13
Add `catalog_id` to the output of `databricks_catalog` resource

tanmay-db:
## Changes

Catalog now exports a stable UUID identifier in the catalog_id field. This PR adds this to the model for schema as a computed attribute.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


alexott:
This should be superced by #4976, but for some reason, `catalog_id` isn't a part of `CatalogInfo` struct. We need to make change in the API spec to include it

PR #4938, opened 2025-08-15
[Exporter] Correct support for `library.glob` in `databricks_pipeline`

alexott:
## Changes

The `library.glob` block, recently introduced, allows specifying not only notebooks and files but also all objects in a directory via `directory/**` syntax. This PR adds support for the `library.glob` block, which handles that.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


docs/guides/experimental-exporter.md
```diff
@@ -69,9 +69,9 @@ All arguments are optional, and they tune what code is being generated.
...
-* `-matchRegex` - Match resource names against a given regex during listing operation. Applicable to all resources selected for listing.
-* `-excludeRegex` - Exclude resource names matching a given regex. Applied during the listing operation and has higher priority than `-match` and `-matchRegex`.  Applicable to all resources selected for listing.  Could be used to exclude things like `databricks_automl` notebooks, etc.
+* `-match` - Match resource names during listing operation. This filter applies to many (*but not all!*) resources that are getting listed, so if you want to import all dependencies of just one cluster, specify `-match=autoscaling -listing=compute`. By default, it is empty, which matches everything.
+* `-matchRegex` - Match resource names against a given regex during listing operation. Applicable to many (*but not all!*) resources selected for listing.
+* `-excludeRegex` - Exclude resource names matching a given regex. Applied during the listing operation and has higher priority than `-match` and `-matchRegex`.  Applicable to  to many (*but not all!*) resources selected for listing.  Could be used to exclude things like `databricks_automl` notebooks, etc.
```
nkvuong:
```suggestion
* `-match` - Match resource names during listing operation. This filter applies to many (*but not all!*) resources that are getting listed, so if you want to import all dependencies of just one cluster, specify `-match=autoscaling -listing=compute`. By default, it is empty, which matches everything.
* `-matchRegex` - Match resource names against a given regex during listing operation. Applicable to many (*but not all!*) resources selected for listing.
* `-excludeRegex` - Exclude resource names matching a given regex. Applied during the listing operation and has higher priority than `-match` and `-matchRegex`.  Applicable to  to many (*but not all!*) resources selected for listing.  Could be used to exclude things like `databricks_automl` notebooks, etc.
```

shouldn't this be part of #4939

alexott: moved to correct PR

PR #4944, opened 2025-08-18
[Exporter] Resolve references also for map values

### LINKED ISSUE #3131: Exporter: resolve references in maps
Right now, references are resolved only for attributes of string, int, and list types. We need to support resolving references in maps as well.

alexott:
## Changes

We now resolve references in map values.  Initial support is done for secrets/secret scope references. Also added references for secrets/secret scopes in other places, like, model serving attributes, etc.

Also fixed not-needed generation of `served_model_name` in the traffic config

Resolves #3131

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


exporter/codegen.go
```diff
@@ -519,7 +536,7 @@ func (ic *importContext) generateVariables() error {
 				for _, block := range tbody.Blocks() {
 					typ := block.Type()
 					labels := block.Labels()
-					log.Printf("[DEBUG] blockBody: %v %v\n", typ, labels)
+					// log.Printf("[DEBUG] blockBody: %v %v\n", typ, labels)
```
nkvuong, line 539:
```suggestion
```

nkvuong (approved):

PR #4965, opened 2025-08-22
[Internal] Make `databricks_share_pluginframework` SDKv2 compatible

tanmay-db:
## Changes

Plugin framework implementation of share resource is not compatible with SDKv2. Compatibility difference arises because of how these libraries apply the changes internally:

Plugin Framework checks if the state is null for creation: https://github.com/hashicorp/terraform-plugin-framework/blob/main/internal/fwserver/server_applyresourcechange.go#L47
````go
// If PriorState is missing/null, its a Create request.
if req.PriorState == nil || req.PriorState.Raw.IsNull() {...}
````

Whereas SDKv2 checks the ID: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource.go#L976 
````go
if data.Id() == "" {
// We're creating, it is a new resource.
...
}
````

Because of the incompatibility, SDKv2 re-creates the share resource after it has been created by plugin framework. The PR adds the ID field as part of schema. This change makes the plugin framework implementation compatible with SDKv2

## Tests

- Integration tests
- Also E2E tested for applying share resource in SDKv2 after it's been applied by plugin framework in: https://github.com/databricks/terraform-provider-databricks/pull/4967

### CODE REVIEW

**Comment by tanmay-db** [comment]
Failing tests are unrelated, they are failing on main

File: internal/providers/pluginfw/products/sharing/resource_share.go [hunk 1]
```diff
@@ -390,6 +396,7 @@ func (r *ShareResource) Update(ctx context.Context, req resource.UpdateRequest,
 		return
 	}
 
+	state.ID = state.Name
```
mgyucht: This is a no-op because if Name and ID are not the same, Terraform will plan this resource to be destroyed and then created, rather than calling this Update() method.

internal/providers/pluginfw/products/sharing/resource_acc_test.go
```diff
@@ -230,3 +231,41 @@ func TestUcAccUpdateShareReorderObject(t *testing.T) {
 		}`,
 	})
 }
+
+func shareUpdateWithName(name string) string {
+	return fmt.Sprintf(`resource "databricks_share_pluginframework" "myshare" {
+			name  = "%s"
+			owner = "account users"
+			object {
+				name = databricks_table.mytable.id
```
nkvuong:
```suggestion
				name = databricks_sql_table.mytable.id
```

PR #4969, opened 2025-08-22
[Fix] Fix syncing effective fields in share resource

tanmay-db:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -446,8 +446,13 @@ func (r *ShareResource) syncEffectiveFields(ctx context.Context, plan, state Sha
 	stateObjects, _ := state.GetObjects(ctx)
 	finalObjects := []sharing_tf.SharedDataObject_SdkV2{}
 	for i := range stateObjects {
-		mode.objectLevel(ctx, &stateObjects[i], planObjects[i])
-		finalObjects = append(finalObjects, stateObjects[i])
+		for j := range planObjects {
+			if stateObjects[i].Name == planObjects[j].Name {
+				mode.objectLevel(ctx, &stateObjects[i], planObjects[j])
+				finalObjects = append(finalObjects, stateObjects[i])
```
hectorcast-db: what happens if the plan as an object which is not in the state? It seems that we are ignoring it.

tanmay-db: Yes, we ignore it because in terraform we want to create the infra from the config

hectorcast-db: Can you explain a bit more what is what? I think I am mixing what stateObjects, planObjects and finalObjects are. 
As it is, it feels that we are storing (state) a different set of objects that we are creating (plan).

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -446,8 +446,13 @@ func (r *ShareResource) syncEffectiveFields(ctx context.Context, plan, state Sha
 	stateObjects, _ := state.GetObjects(ctx)
```
hectorcast-db: Are we suppressing the error? Can this fail?

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -446,8 +446,13 @@ func (r *ShareResource) syncEffectiveFields(ctx context.Context, plan, state Sha
...
 	finalObjects := []sharing_tf.SharedDataObject_SdkV2{}
 	for i := range stateObjects {
-		mode.objectLevel(ctx, &stateObjects[i], planObjects[i])
-		finalObjects = append(finalObjects, stateObjects[i])
+		for j := range planObjects {
```
mgyucht: This deserves a comment. Comment here will both explain the code below as well as serve as a secondary check that this code works as specified.

internal/providers/pluginfw/products/sharing/resource_share.go
tanmay-db:
```go
 	if d.HasError() {
		panic(pluginfwcommon.DiagToString(d))
	}
```
The error isn't suppressed: https://github.com/databricks/terraform-provider-databricks/blob/main/internal/service/sharing_tf/legacy_model.go#L4440. 

This was added in: https://github.com/databricks/terraform-provider-databricks/pull/4332/files#diff-5f47d9d2d5e2fcf6b377c528600d81c37f4526da6ce6aba43818bdb0f7039152R2365-R2375

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -439,16 +439,30 @@ func (effectiveFieldsActionRead) objectLevel(ctx context.Context, state *sharing
...
-	for i := range stateObjects {
-		mode.objectLevel(ctx, &stateObjects[i], planObjects[i])
-		finalObjects = append(finalObjects, stateObjects[i])
+	for i := range newStateObjects {
+		for j := range existingStateObjects {
+			if newStateObjects[i].Name == existingStateObjects[j].Name {
+				mode.objectLevel(ctx, &newStateObjects[i], existingStateObjects[j])
+				finalObjects = append(finalObjects, newStateObjects[i])
+				break
+			}
```
mgyucht, line 471: I think we need to handle objects added to the share since last apply here.

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -439,16 +439,33 @@ func (effectiveFieldsActionRead) objectLevel(ctx context.Context, state *sharing
...
+// to the infrastructure and then the newState values are recorded in the terraform state.
+// Hence we ignore the values in existingState which are not present in newState.
+func (r *ShareResource) syncEffectiveFields(ctx context.Context, existingState, newState ShareInfoExtended, mode effectiveFieldsAction) (ShareInfoExtended, diag.Diagnostics) {
 	var d diag.Diagnostics
-	mode.resourceLevel(ctx, &state, plan.ShareInfo_SdkV2)
-	planObjects, _ := plan.GetObjects(ctx)
-	stateObjects, _ := state.GetObjects(ctx)
+	mode.resourceLevel(ctx, &newState, existingState.ShareInfo_SdkV2)
+	existingStateObjects, _ := existingState.GetObjects(ctx)
+	newStateObjects, _ := newState.GetObjects(ctx)
```
mgyucht, line 461: Let's not drop these errors here. I know this was like this before, but let's try to eliminate this tech debt.

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -439,16 +439,33 @@ func (effectiveFieldsActionRead) objectLevel(ctx context.Context, state *sharing
...
+		// For each object in the new state, we check if it exists in the existing state
+		// and if it does, we sync the effective fields.
+		// If it does not exist, we keep the new state object as is.
+		for j := range existingStateObjects {
+			if newStateObjects[i].Name == existingStateObjects[j].Name {
+				mode.objectLevel(ctx, &newStateObjects[i], existingStateObjects[j])
+				break
+			}
+		}
+		finalObjects = append(finalObjects, newStateObjects[i])
```
mgyucht, line 473: Just to clarify, the order is the server-returned order. What happens if the order of shares in the config is changed?

tanmay-db:
Yes, I plan to add the tests but didn't get to them yet as was working on other things, will send the PR in review once they are added.

internal/providers/pluginfw/products/sharing/resource_share.go
tanmay-db: We aren't dropping the errors: https://github.com/databricks/terraform-provider-databricks/blob/main/internal/service/sharing_tf/legacy_model.go#L4440

```go
	if d.HasError() {
		panic(pluginfwcommon.DiagToString(d))
	}
```

internal/providers/pluginfw/products/sharing/resource_share.go
tanmay-db: Change in order of shares in config leads to non empty plan. This is the same behaviour that we have in SDKv2 implementation. Verified: https://github.com/databricks/terraform-provider-databricks/pull/5085

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -445,16 +445,33 @@ func (effectiveFieldsActionRead) objectLevel(ctx context.Context, state *sharing
 	state.SyncFieldsDuringRead(ctx, plan)
 }
 
-func (r *ShareResource) syncEffectiveFields(ctx context.Context, plan, state ShareInfoExtended, mode effectiveFieldsAction) (ShareInfoExtended, diag.Diagnostics) {
+// syncEffectiveFields syncs the effective fields between existingState and newState
+// and returns the newState
+//
+// existingState: infrastructure values that are recorded in the existing terraform state.
```
mgyucht, line 451: This corresponds to the state in Read method and plan in the Create/Update methods.

internal/providers/pluginfw/products/sharing/resource_share.go
mgyucht: note: these are not errors, they are booleans indicating whether the field value is null or unknown.

internal/providers/pluginfw/products/sharing/resource_share.go
mgyucht: This is actually fine since 1) changing the order of objects in the share will always result in a non-empty plan today because this is a list and not a set, and 2) CustomizeDiff already handles the case where state object order and config object order don't match.

PR #4980, opened 2025-08-27
[Fix] Fix metastore ID interpolation for account-level storage credential imports

mgyucht:
## Changes

When importing account-level storage credentials, the metastore ID was not being interpolated correctly, causing API calls to fail with malformed URLs containing double slashes (`/api/2.0/accounts/.../metastores//storage-credentials/...`). This occurred because during import, the Terraform state is empty and `d.Get("metastore_id")` returned an empty string.

Added a `parseStorageCredentialId()` function that handles composite IDs in the format `metastore_id|storage_credential_name` for account-level imports, extracting the metastore ID and setting it in the state before making API calls. This maintains backward compatibility with workspace-level imports using simple credential names and includes comprehensive unit tests and updated documentation.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

Fixes: ES-1561726

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,6 @@
     "files.trimFinalNewlines": true,
     "files.insertFinalNewline": true,
     "go.useCodeSnippetsOnFunctionSuggest": true,
-    "go.useCodeSnippetsOnFunctionSuggestWithoutType": true
+    "go.useCodeSnippetsOnFunctionSuggestWithoutType": true,
+    "makefile.configureOnOpen": false
 }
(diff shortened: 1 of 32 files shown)


.vscode/settings.json
```diff
@@ -10,5 +10,6 @@
     "files.trimFinalNewlines": true,
     "files.insertFinalNewline": true,
     "go.useCodeSnippetsOnFunctionSuggest": true,
-    "go.useCodeSnippetsOnFunctionSuggestWithoutType": true
+    "go.useCodeSnippetsOnFunctionSuggestWithoutType": true,
+    "makefile.configureOnOpen": false
```
nkvuong: I think this is unintentional change

nkvuong (approved):

PR #4981, opened 2025-08-27
Use databricks_sql_table Instead of databricks_table in Sharing Tests

rauchy:
## Changes

Replace deprecated `databricks_table` with `databricks_sql_table` in sharing tests.

This PR migrates all share-related acceptance tests from the deprecated `databricks_table` resource to `databricks_sql_table to` resolve test failures caused by:

  1. Partition spec validation errors when sharing tables created with the deprecated resource
  2. Delta deletion vectors conflicts that prevent table sharing without full history
 
## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/products/sharing/data_shares_acc_test.go
+++ b/internal/providers/pluginfw/products/sharing/data_shares_acc_test.go
@@ -45,9 +45,6 @@ func TestUcAccDataSourceShares(t *testing.T) {
 			name = "bar"
 			table_type = "MANAGED"
 			warehouse_id = "{env.TEST_DEFAULT_WAREHOUSE_ID}"
-			properties = {
-				"delta.enableDeletionVectors" = "false"
-			}
 
 			column {
 				name = "id"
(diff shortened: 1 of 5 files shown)


hectorcast-db:
Is this what we want? We are just ignoring the issues with the old provider.

nkvuong:
`databricks_table` uses unsupported API, so switching to `databricks_sql_table` is the correct approach.

internal/providers/pluginfw/products/sharing/data_shares_acc_test.go
```diff
@@ -39,47 +39,47 @@ func TestUcAccDataSourceShares(t *testing.T) {
...
+		resource "databricks_sql_table" "mytable" {
 			catalog_name = databricks_catalog.sandbox.id
 			schema_name = databricks_schema.things.name
 			name = "bar"
 			table_type = "MANAGED"
-			data_source_format = "DELTA"
+			warehouse_id = "{env.TEST_DEFAULT_WAREHOUSE_ID}"
+			properties = {
+				"delta.enableDeletionVectors" = "false"
+			}
```
nkvuong: is this because deletion vectors are being enabled automatically for workspaces? if so, would it be better to enable history sharing on delta sharing instead?

rauchy: Thanks, makes sense! [962c03c](https://github.com/databricks/terraform-provider-databricks/pull/4981/commits/962c03c54c87f6ba00d730b3bcab509bf7fc87e5)

mgyucht (approved):
LGTM, thanks for doing this!

PR #4986, opened 2025-08-31
[Internal] Bump Go SDK to v0.82.0 & Generate Latest TF Structs 

rauchy:
## Changes

Bump Go SDK to v0.82.0 & Generate Latest TF Structs.


.github/workflows/next-changelog.yml
alexott, line 1: We don't need these changes

docs/data-sources/database_database_catalog.md
alexott, line 1: can we avoid name repetition `database_database`?

docs/data-sources/database_synced_database_table.md
alexott, line 1: I would prefer to use `database_synced_table` instead of this name

docs/data-sources/materialized_features_feature_tags.md
alexott, line 1: here is the same - why we need `features_feature`?

docs/data-sources/rfa_access_request_destinations.md
alexott, line 1: what is `rfa` here? if request for access, then it's duplicating...

docs/data-sources/workspace_settings_v2_setting.md
alexott, line 1: again - naming isn't good. Why not `worspace_setting_v2`?

tanmay-db (approved):

PR #4995, opened 2025-09-03
Fix regression with `databricks_group` data source introduced by a recent change

### LINKED ISSUE #4994: [ISSUE] Error: cannot read group starting from v1.88.0
We've started getting the following error after v1.88.0 

Error: cannot read group https://xxxx/api/2.0/preview/scim/v2/Groups/yyyy?attributes=members,roles,entitlements,externalId is only accessible by admins.

from the brief look it seems to be connected to this change 
https://github.com/databricks/terraform-provider-databricks/pull/4947

As error suggests is this API accessible only by admin user / token ?  
Everything works fine with  1.87.1  version. 

v1.88.0

### Is it a regression?
yes

alexott:
## Changes

It looks like some users are using the `databricks_group` data source as non-admin users (in some cases, incorrectly).  Recent change in SCIM API required to change from List to List for `id` + Get to retrieve group members, but it had an unexpected effect that Get for non-admin users throws an error as we request for `externalId` attribute.

I got clarification from the SCIM team that changes in the `members` retrieval are affecting only the account-level API, and not the workspace-level API.  So this PR reimplements the required changes as follows:

- on account level we do `List` for group ID by name, followed by `Get` to retrieve `members`, `externalId`, etc.
- on workspace level we use the old implementation of `List` + all attributes

Resolves #4994

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/scim/data_group_test.go
+++ b/scim/data_group_test.go
@@ -18,50 +18,41 @@ func TestDataSourceGroup(t *testing.T) {
 		Fixtures: []qa.HTTPFixture{
 			{
 				Method:   "GET",
-				Resource: `/api/2.0/preview/scim/v2/Groups?attributes=id&filter=displayName%20eq%20%22ds%22`,
+				Resource: `/api/2.0/preview/scim/v2/Groups?attributes=members%2Croles%2Centitlements%2CexternalId&filter=displayName%20eq%20%22ds%22`,
 				Response: GroupList{
 					Resources: []Group{
 						{
 							DisplayName: "ds",
 							ID:          "eerste",
-						},
-					},
-				},
-			},
-			{
-				Method:   "GET",
-				Resource: "/api/2.0/preview/scim/v2/Groups/eerste?attributes=members,roles,entitlements,externalId",
-
-				Response: Group{
-					DisplayName: "ds",
-					ID:          "eerste",
-					Entitlements: []ComplexValue{
-						{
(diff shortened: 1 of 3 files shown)


scim/data_group_test.go
```diff
@@ -18,28 +18,114 @@ func TestDataSourceGroup(t *testing.T) {
...
+	assertContains(t, d.Get("instance_profiles"), "b")
+	assertContains(t, d.Get("members"), "1112")
+	assertContains(t, d.Get("members"), "1113")
+	assertContains(t, d.Get("groups"), "abc")
+	assert.Equal(t, true, d.Get("allow_instance_pool_create"))
+	assert.Equal(t, true, d.Get("allow_cluster_create"))
+
+	assertContains(t, d.Get("users"), "1112")
+	assertContains(t, d.Get("service_principals"), "1113")
+	assertContains(t, d.Get("child_groups"), "1114")
```
nkvuong: could we rewrite this with `ApplyAndExpectData`?

rauchy (approved):

alexott:
Please don't merge it yet - I need to make one more change to restore
previous behavior

-- 
Alex Ott, Lead Specialist Solutions Architect, Databricks
***@***.***, +49 176 50439409

scim/data_group_test.go
alexott: done

Issue #4998, opened 2025-09-03
[FEATURE] Add `task_retry_mode` to databricks jobs that run in continous mode


dcrezee:
### Use-cases

We're using Databricks jobs that we have configured to run continuously. We're seeing that some task are failing occasionally and we would like to be able to retry them on failure.

### Attempted Solutions

None.

### Proposal

Currently, the Databricks Terraform provider only allows you to set the `pause_status` within the [continous configuration block](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/job#continuous-configuration-block). The Databricks API allows you to set a field called [task_retry_mode](https://docs.databricks.com/api/workspace/jobs/create#continuous-task_retry_mode), which seems to define the necessary behaviour that would allow you to retry individual tasks on failure. Please add this option in the Terraform provider as well.

derkcrezee:
Can be closed. See remark in https://github.com/databricks/terraform-provider-databricks/pull/4999

alexott:
it's already there in 1.88.0 (after Go SDK upgrade). Documentation is also updated in #4993

PR #4999, opened 2025-09-03
Add support for continuous.task_retry_mode in databricks_job

derkcrezee:
## Changes

Added missing property to the continuous configuration block of the databricks_job resource and should resolve issue [#4998](https://github.com/databricks/terraform-provider-databricks/issues/4998).

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott:
@derkcrezee we don't need this PR - this setting was automatically added with Go SDK upgrade, and it was just undocumented (but it was fixed in #4993

derkcrezee:
Ok, thanks for the info. I wasn't aware that the Go SDK upgrade would solve this issue. We just ran into the lack of support and in the code this seemed like the right place to make a change.

PR #5019, opened 2025-09-11
Add `expected_workspace_status` in `databricks_mws_workspaces` to support least-privileged workspaces

tinglin-db:
## Changes
In GCP, workspace creation triggers the creation of a service account in the customer's project with heightened permissions. This is necessary for classic workspaces for setting up resources in the customer's project, like virtual networks, storage buckets, etc. Security-conscious customers require that such service accounts never have heightened permissions in order for them to make global statements about the security of their GCP projects. Such workspaces are called "least-privileged workspaces."

To support least privileged workspaces on GCP, we add a new field `expected_workspace_status`, which will be translated to `workspace_state` in the API request. Users should create a workspace by setting this value first to `"PROVISIONING"`. After applying their template, they can make any needed changes, such as granting the minimal permissions to this service account in their project. Finally, they can remove the `expected_workspace_status` field or set it explicitly to `"RUNNING"`.

## Tests

Added a unit test `TestResourceWorkspaceCreateGcpWithExpectedProvisioning` and an integration test `TestMwsAccGcpWorkspacesWithExpectedProvisioning`

- [x] `make test` run locally
- [x] using Go SDK

Tested E2E Locally 
```

  # databricks_mws_workspaces.this will be created
  + resource "databricks_mws_workspaces" "this" {
      + account_id                = (sensitive value)
      + cloud                     = (known after apply)
      + creation_time             = (known after apply)
      + effective_compute_mode    = (known after apply)
      + expected_workspace_status = "PROVISIONING"
      + gcp_workspace_sa          = (known after apply)
      + id                        = (known after apply)
      + is_no_public_ip_enabled   = true
      + location                  = "us-central1"
      + pricing_tier              = (known after apply)
      + workspace_id              = (known after apply)
      + workspace_name            = "tlin-classic-test-tf-1"
      + workspace_status          = (known after apply)
      + workspace_status_message  = (known after apply)
      + workspace_url             = (known after apply)

      + cloud_resource_container {
          + gcp {
              + project_id = "databricks-cal-dev-testing"
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

databricks_mws_workspaces.this: Creating...
databricks_mws_workspaces.this: Creation complete after 2s [id=9fcbb245-7c44-4522-9870-e38324104cf8/2181571221551671]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

  # databricks_mws_workspaces.this will be updated in-place
  ~ resource "databricks_mws_workspaces" "this" {
      - expected_workspace_status = "PROVISIONING" -> null
        id                        = "9fcbb245-7c44-4522-9870-e38324104cf8/2181571221551671"
      + network_id                = "a124db3a-0928-4698-97eb-70e751c87934"

        # (3 unchanged blocks hidden)
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

databricks_mws_workspaces.this: Modifying... [id=9fcbb245-7c44-4522-9870-e38324104cf8/2181571221551671]
databricks_mws_workspaces.this: Modifications complete after 6s [id=9fcbb245-7c44-4522-9870-e38324104cf8/2181571221551671]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

### DIFF
--- a/mws/mws_workspaces_test.go
+++ b/mws/mws_workspaces_test.go
@@ -681,9 +681,21 @@ func grantGcpWorkspacePermissions(t *testing.T, serviceAccountEmail string) {
 	})
 }
 
+func getCloudResourceManagerClient(ctx context.Context) (*cloudresourcemanager.Service, error) {
+	options := []option.ClientOption{
+		option.WithScopes(cloudresourcemanager.CloudPlatformScope),
+	}
+	// In integration tests, the GOOGLE_CREDENTIALS environment variable specifies the credentials
+	credentialsJSON := os.Getenv("GOOGLE_CREDENTIALS")
+	if credentialsJSON != "" {
+		options = append(options, option.WithCredentialsJSON([]byte(credentialsJSON)))
+	}
+	return cloudresourcemanager.NewService(ctx, options...)
+}
+
 // assignGcpRoleToServiceAccount assigns a custom role to a service account on a GCP project
 func assignGcpRoleToServiceAccount(ctx context.Context, projectID, serviceAccountEmail, roleID string) error {
-	crmService, err := cloudresourcemanager.NewService(ctx, option.WithScopes(cloudresourcemanager.CloudPlatformScope))
+	crmService, err := getCloudResourceManagerClient(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to create Cloud Resource Manager client: %w", err)
 	}
@@ -741,7 +753,7 @@ func assignGcpRoleToServiceAccount(ctx context.Context, projectID, serviceAccoun
 
 // removeGcpRoleFromServiceAccount removes a custom role from a service account on a GCP project
 func removeGcpRoleFromServiceAccount(ctx context.Context, projectID, serviceAccountEmail, roleName string) error {
-	crmService, err := cloudresourcemanager.NewService(ctx, option.WithScopes(cloudresourcemanager.CloudPlatformScope))
+	crmService, err := getCloudResourceManagerClient(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to create Cloud Resource Manager client: %w", err)
 	}


mws/resource_mws_workspaces.go
```diff
@@ -576,7 +591,10 @@ func ResourceMwsWorkspaces() common.Resource {
 		Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
 			var workspace Workspace
 			workspacesAPI := NewWorkspacesAPI(ctx, c)
-			common.DataToStructPointer(d, workspaceSchema, &workspace)
+			if v, ok := d.GetOk("expected_workspace_status"); ok {
+				workspace.ExpectedWorkspaceStatus = v.(string)
+			}
+
```
nkvuong: we don't need this, `DataToStructPointer` should take care of it

mws/resource_mws_workspaces.go
```diff
@@ -239,22 +244,32 @@ func (a WorkspacesAPI) explainWorkspaceFailure(ws Workspace) error {
 		ws.WorkspaceStatusMessage, strBuffer.String())
 }
 
-// WaitForRunning will wait until workspace is running, otherwise will try to explain why it failed
+// If specified an expected_workspace_status, WaitForRunning will wait until workspace is in the expected status.
+// If not, it will wait until workspace is running, and otherwise will try to explain why it failed
 func (a WorkspacesAPI) WaitForRunning(ws Workspace, timeout time.Duration) error {
```
nkvuong: we should rename the function to `WaitForExpectedStatus` then

mws/resource_mws_workspaces.go
```diff
@@ -97,6 +97,7 @@ type Workspace struct {
 	WorkspaceURL                        string                   `json:"workspace_url,omitempty" tf:"computed"`
 	WorkspaceStatus                     string                   `json:"workspace_status,omitempty" tf:"computed"`
 	WorkspaceStatusMessage              string                   `json:"workspace_status_message,omitempty" tf:"computed"`
+	ExpectedWorkspaceStatus				string					 `json:"expected_workspace_status,omitempty" tf:"optional"`
```
nkvuong: if user does not provide this, will the server populate this?

panchenxue-databricks: No, `ExpectedWorkspaceStatus` is client-owned. `WorkspaceStatus` is what server will populate, no matter whether `ExpectedWorkspaceStatus` is specified or not.

nkvuong: makes sense. Then `tf:"optional"` is not needed as `omitempty` does the job

tinglin-db: updated

mws/resource_mws_workspaces.go
tinglin-db: agreed - updated

alexott:
please run `make fmt ws`

tinglin-db:
jenkins trigger all

mws/mws_workspaces_test.go
```diff
@@ -258,6 +258,44 @@ func TestMwsAccGcpWorkspaces(t *testing.T) {
 	})
 }
 
+func TestMwsAccGcpWorkspacesWithExpectedProvisioning(t *testing.T) {
```
tinglin-db: Integration test run ([link](https://github.com/databricks-eng/eng-dev-ecosystem/actions/workflows/terraform-isolated-pr.yml))
On its last run, this integration test failed during the post-test destroy due to not being able to delete a still PROVISIONING workspace. Error message:
```
Error: -26T17:50:17.498Z [ERROR] sdk.helper_resource: Error running post-test destroy, there may be dangling resources: test_terraform_path=/tmp/plugintest-terraform1770350163/terraform test_step_number=1
  | Error: cannot delete mws workspaces: INVALID_STATE: workspace 49063535405170 cannot be deleted because it is in status PROVISIONING but not must be in one of: PROVISIONING
```
I pushed a new commit to catch `INVALID_STATE_TRANSITION` errors in `DELETE`, but this is probably not the best approach - not sure if there is a way to have the integration test instead run through both CREATE and UPDATE so that the workspace is in a `RUNNING` status? cc @panchenxue-databricks

alexott:
@tinglin-db Now it fails with:

```
Error: cannot create mws workspaces: Workspace limit exceeded, only 20 workspaces are allowed for your account
```

that's why I think that deletion in the provisioning state should be handled on backend, otherwise there will be dangling resources.

mws/mws_workspaces_test.go
mgyucht: You probably need to support this case of deleting workspaces in PROVISIONING, since users could attempt to delete a partially created workspace, or they could modify an attribute that requires recreation (like deployment name).

For this test, you could also just add an additional step to put the workspace in a running state by updating `expected_workspace_status`.

mws/mws_workspaces_test.go
```diff
@@ -258,6 +258,44 @@ func TestMwsAccGcpWorkspaces(t *testing.T) {
...
+			}
+
+			if expectedAttr := rs.Primary.Attributes["expected_workspace_status"]; expectedAttr != expectedStatus {
+				return fmt.Errorf("expected expected_workspace_status to be %s, got %s", expectedStatus, expectedAttr)
+			}
+
+			return nil
+		},
+	})
+}
```
mgyucht, line 441: We should absolutely add a test here for the transition from PROVISIONING to RUNNING.

mws/mws_workspaces_test.go
```diff
@@ -258,6 +258,44 @@ func TestMwsAccGcpWorkspaces(t *testing.T) {
...
+			}
+
+			expectedStatus := "PROVISIONING"
+			if status := rs.Primary.Attributes["workspace_status"]; status != expectedStatus {
+				return fmt.Errorf("expected workspace_status to be %s, got %s", expectedStatus, status)
+			}
+
+			if expectedAttr := rs.Primary.Attributes["expected_workspace_status"]; expectedAttr != expectedStatus {
+				return fmt.Errorf("expected expected_workspace_status to be %s, got %s", expectedStatus, expectedAttr)
+			}
```
mgyucht: I'm fairly certain this is tested automatically by the framework itself; if the final state does not. match the config, it should fail. You can verify this by modifying your implementation to set the expected_workspace_status to a different value on read.

mws/resource_mws_workspaces.go
```diff
@@ -97,6 +97,7 @@ type Workspace struct {
 	WorkspaceURL                        string                   `json:"workspace_url,omitempty" tf:"computed"`
 	WorkspaceStatus                     string                   `json:"workspace_status,omitempty" tf:"computed"`
 	WorkspaceStatusMessage              string                   `json:"workspace_status_message,omitempty" tf:"computed"`
+	ExpectedWorkspaceStatus             string                   `json:"expected_workspace_status,omitempty"`
```
mgyucht, line 100: We should also document this in docs/resources/mws_workspaces.md.

mws/resource_mws_workspaces.go
```diff
@@ -147,6 +148,10 @@ func (w *Workspace) MarshalJSON() ([]byte, error) {
 	if w.ComputeMode != "" {
 		workspaceCreationRequest["compute_mode"] = w.ComputeMode
 	}
+	// For now, only translate provisioning status
+	if w.ExpectedWorkspaceStatus == WorkspaceStatusProvisioning {
+		workspaceCreationRequest["workspace_state"] = "WORKSPACE_STATE_PROVISIONING"
+	}
```
mgyucht: I thought we were adding a new field to the API, but here it looks like we're reusing the `workspace_state` field in the API?

mws/resource_mws_workspaces.go
```diff
@@ -147,6 +148,10 @@ func (w *Workspace) MarshalJSON() ([]byte, error) {
 	if w.ComputeMode != "" {
 		workspaceCreationRequest["compute_mode"] = w.ComputeMode
 	}
+	// For now, only translate provisioning status
+	if w.ExpectedWorkspaceStatus == WorkspaceStatusProvisioning {
+		workspaceCreationRequest["workspace_state"] = "WORKSPACE_STATE_PROVISIONING"
```
mgyucht: Can we use the existing enum value rather than hard-coding the literal?

mws/resource_mws_workspaces.go
```diff
@@ -336,6 +359,16 @@ func (a WorkspacesAPI) Delete(mwsAcctID, workspaceID string) error {
...
 	if err != nil {
+		// Catch INVALID_STATE_TRANSITION error for workspaces in PROVISIONING status
+		var apiErr *apierr.APIError
+		if errors.As(err, &apiErr) && apiErr.ErrorCode == "INVALID_STATE_TRANSITION" {
+			// Check if workspace is in PROVISIONING status
+			workspace, readErr := a.Read(mwsAcctID, workspaceID)
+			if readErr == nil && workspace.WorkspaceStatus == WorkspaceStatusProvisioning {
+				log.Printf("[WARN] Workspace %s is in PROVISIONING status and cannot be deleted due to INVALID_STATE_TRANSITION. Treating as successfully deleted to avoid blocking Terraform operations.", workspaceID)
+				return nil
+			}
```
mgyucht: I don't think this is permissible. We need to allow users to delete workspaces in PROVISIONING state. Otherwise workspaces will be leaked. This violates the general expectations of Terraform by orphaning workspaces when you remove them from your config.

mws/resource_mws_workspaces.go
```diff
@@ -638,7 +680,16 @@ func ResourceMwsWorkspaces() common.Resource {
...
+			// The workspace returned from the API call does not have the expected_status field
+			// Therefore, we need to check the original Terraform configuration to not try to read workspace
+			// if the expected status is PROVISIONING
+			expectedStatus := d.Get("expected_workspace_status").(string)
+			if expectedStatus == WorkspaceStatusProvisioning {
+				log.Printf("[INFO] Expected status is PROVISIONING, skipping wait in read function")
+				err = nil
+			} else {
+				err = workspacesAPI.WaitForExpectedStatus(workspace, d.Timeout(schema.TimeoutRead))
+			}
```
mgyucht: We should have one place for logic around waiting for workspace status. I believe this is unnecessary because the implementation of WaitForExpectedStatus already handles this properly.

mws/resource_mws_workspaces.go
```diff
@@ -239,22 +244,40 @@ func (a WorkspacesAPI) explainWorkspaceFailure(ws Workspace) error {
...
+// If expected_workspace_status is specified, WaitForExpectedStatus will wait until workspace is in the expected status.
+// If not, it will wait until workspace is running, and otherwise will try to explain why it failed.
+func (a WorkspacesAPI) WaitForExpectedStatus(ws Workspace, timeout time.Duration) error {
+	// expected status should only ever be PROVISIONING or RUNNING
+	expected := ws.ExpectedWorkspaceStatus
+	if expected == "" {
+		expected = WorkspaceStatusRunning
+	}
+
+	// Skip waiting entirely if expected status is PROVISIONING since we cannot read the workspace while it is provisioning
```
mgyucht: Why can't we read the workspace while it is provisioning?

NEXT_CHANGELOG.md
```diff
@@ -17,3 +19,5 @@
 ### Exporter
 
 ### Internal Changes
+
+* Use `Jobs.Get` instead of `JobsGetByJobId` ([#5029](https://github.com/databricks/terraform-provider-databricks/pull/5029))
```
mgyucht:
```suggestion
```
This seems accidental.

mws/resource_mws_workspaces.go
```diff
@@ -147,6 +148,10 @@ func (w *Workspace) MarshalJSON() ([]byte, error) {
 	if w.ComputeMode != "" {
 		workspaceCreationRequest["compute_mode"] = w.ComputeMode
 	}
+	// For now, only translate provisioning status
+	if w.ExpectedWorkspaceStatus == WorkspaceStatusProvisioning {
+		workspaceCreationRequest["expected_workspace_status"] = WorkspaceStatusProvisioning
```
mgyucht:
```suggestion
	if w.ExpectedWorkspaceStatus != "" {
		workspaceCreationRequest["expected_workspace_status"] = w.ExpectedWorkspaceStatus
```

mws/resource_mws_workspaces.go
mgyucht: There is a new field in the create request.

mws/resource_mws_workspaces.go
```diff
@@ -310,7 +321,7 @@ func (a WorkspacesAPI) UpdateRunning(ws Workspace, timeout time.Duration) error
 	if err != nil {
 		return err
 	}
-	return a.WaitForRunning(ws, timeout)
+	return a.WaitForExpectedStatus(ws, WorkspaceStatusRunning, timeout)
```
mgyucht: Let's set the expected workspace status that is provided by the user in this request.

mws/mws_workspaces_test.go
```diff
@@ -258,6 +258,39 @@ func TestMwsAccGcpWorkspaces(t *testing.T) {
 	})
 }
 
+func TestMwsAccGcpWorkspacesWithExpectedProvisioning(t *testing.T) {
```
mgyucht: Separately, two more test cases:
1. New workspace in provisioning state -> set expected state to RUNNING
2. New workspace in provisioning state -> unset expected state, workspace should progress to RUNNING.

mgyucht (commented):
Aligned on the pathway forward, let's wrap this up.

mws/mws_workspaces_test.go
tinglin-db: Removed

mws/resource_mws_workspaces.go
tinglin-db: Updated

mws/resource_mws_workspaces.go
tinglin-db: Removed this code block

mws/resource_mws_workspaces.go
tinglin-db: Sorry this was a misleading comment, removed

mws/mws_workspaces_test.go
tinglin-db: Sounds good, added these test cases

mws/resource_mws_workspaces.go
tinglin-db: Updated

mws/resource_mws_workspaces.go
tinglin-db: I modified WaitForExpectedStatus to take in the expected_status. We need to pass it in from `d.Get("expected_workspace_status")` because the field is not returned by the GET during polling

mws/resource_mws_workspaces.go
tinglin-db: Is this true even if the feature is still in private preview?

mgyucht: Yes. Private preview services, methods and fields (and their documentation) are still part of the SDKs. Currently, private preview methods are not shown in the REST API documentation. For example: https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/feature_engineering_feature

I would like us to add badges on individual fields if their stability level lags behind the resource level, but we haven't gotten to that yet. In the meantime, we can add a note in the documentation that this feature is in private preview.

mws/resource_mws_workspaces.go
```diff
@@ -267,7 +284,7 @@ func (a WorkspacesAPI) WaitForRunning(ws Workspace, timeout time.Duration) error
 	})
 }
 
-var workspaceRunningUpdatesAllowed = []string{"credentials_id", "network_id", "storage_customer_managed_key_id", "private_access_settings_id", "managed_services_customer_managed_key_id", "custom_tags"}
+var workspaceRunningUpdatesAllowed = []string{"credentials_id", "network_id", "storage_customer_managed_key_id", "private_access_settings_id", "managed_services_customer_managed_key_id", "custom_tags", "expected_workspace_status"}
```
mgyucht: I know this isn't your fault, but would you mind formatting this list with one entry per line? Changes to this list are easier to review this way.

mws/resource_mws_workspaces.go
```diff
@@ -239,22 +243,35 @@ func (a WorkspacesAPI) explainWorkspaceFailure(ws Workspace) error {
 		ws.WorkspaceStatusMessage, strBuffer.String())
 }
 
-// WaitForRunning will wait until workspace is running, otherwise will try to explain why it failed
-func (a WorkspacesAPI) WaitForRunning(ws Workspace, timeout time.Duration) error {
+// If expected_workspace_status is specified, WaitForExpectedStatus will wait until workspace is in the expected status.
+// If not, it will wait until workspace is running, and otherwise will try to explain why it failed.
+func (a WorkspacesAPI) WaitForExpectedStatus(ws Workspace, expected_status string, timeout time.Duration) error {
```
mgyucht:
```suggestion
func (a WorkspacesAPI) WaitForExpectedStatus(ws Workspace, expectedStatus string, timeout time.Duration) error {
```

mws/resource_mws_workspaces.go
```diff
@@ -638,7 +667,10 @@ func ResourceMwsWorkspaces() common.Resource {
 			if err = common.StructToData(workspace, workspaceSchema, d); err != nil {
 				return err
 			}
-			err = workspacesAPI.WaitForRunning(workspace, d.Timeout(schema.TimeoutRead))
+			// The expected_workspace_status field is input only.
+			// Therefore, we need to read it from the original Terraform configuration.
+			expectedStatus := d.Get("expected_workspace_status").(string)
```
mgyucht, line 680:
```suggestion
			// PROVISIONING workspace import may fail because the "expected_workspace_status" is not included in the state during import, nor is it returned by the API.
			// As a result, the provider will (likely) wait for RUNNING state, which will never happen, and timeout.
			// TODO: fix thisl.
			expectedStatus := d.Get("expected_workspace_status").(string)
```

mws/resource_mws_workspaces.go
tinglin-db: Added documentation, let me know if the wording works

mws/resource_mws_workspaces.go
tinglin-db: Added comment

mgyucht (approved):
Thanks for working with us on this change!

PR #5029, opened 2025-09-16
[Internal] Use Jobs.Get instead of JobsGetByJobId

tanmay-db:
## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/jobs/job_test.go
+++ b/jobs/job_test.go
@@ -426,7 +426,9 @@ func TestAccPeriodicTrigger(t *testing.T) {
 			jobID, err := strconv.ParseInt(id, 10, 64)
 			assert.NoError(t, err)
 
-			res, err := w.Jobs.GetByJobId(ctx, jobID)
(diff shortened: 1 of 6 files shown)


alexott:
What is the difference, given that [GetByJobId](https://github.com/databricks/databricks-sdk-go/blob/main/service/jobs/api.go#L401) implementation is just the same call as in PR:

```golang
res, err := w.Jobs.Get(ctx, jobs.GetJobRequest{
	JobId: jobID,
})
```

And `Get` just [calls](https://github.com/databricks/databricks-sdk-go/blob/main/service/jobs/impl.go#L77) `Do` without pagination

tanmay-db:
@alexott the Get doesn't directly calls the underlying implementation, it handles pagination: https://github.com/databricks/databricks-sdk-go/blob/main/service/jobs/ext_api.go#L149

PR #5058, opened 2025-09-26
Surface table_id in databricks_sql_table

Oscarbachtiar759:
## Changes
Surfacing `table_id` in the `databricks_sql_table` resource. The [API already supports this](https://docs.databricks.com/api/workspace/tables/get#table_id), so this change is to surface it in Terraform so other resources can leverage it.

The change is to add TableID into [SqlTableInfo](https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_sql_table.go#L47). When reading the resource, the [getTable](https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_sql_table.go#L106) function is called to [fetch from the Get UC Table API](https://docs.databricks.com/api/workspace/tables/get#table_id), which should have the table_id field and will therefore be [copied to the tableInfo](https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_sql_table.go#L107).

## Tests
`go test -v ./catalog -run "TestResourceSqlTable"` passes.


catalog/resource_sql_table.go
```diff
@@ -64,6 +64,7 @@ type SqlTableInfo struct {
 	ClusterID           string            `json:"cluster_id,omitempty" tf:"computed"`
 	WarehouseID         string            `json:"warehouse_id,omitempty"`
 	Owner               string            `json:"owner,omitempty" tf:"computed"`
+	TableID             string            `json:"table_id,omitempty" tf:"computed"`
```
mgyucht: Table ID is an output-only field, so we can remove `omitempty` to ensure TF doesn't allow users to set it.
```suggestion
	TableID             string            `json:"table_id" tf:"computed"`
```

To verify that this won't break users, I created a table with the current version of the provider (so Table ID was not in state), then applied this change, and there is no diff. However, after refresh/apply, the table ID appears.

alexott:
@Oscarbachtiar759 please run `make fmt lint ws`

PR #5061, opened 2025-09-29
[Internal] Unified Terraform Provider: Plugin framework resources and data sources

tanmay-db:
## Changes

Add unified provider support for manually maintained plugin framework resources. This PR doesn't contain docs, will add them once core changes are reviewed and agreed on.

Note: This PR will be split into multiple PRs for different types of changes. Changelog will be updated in a separate PR which enables the feature for all resources SDKv2 and plugin framework. The PR is based on changes for client which would be generated when Go SDK is released and bumped (which was blocked due to issues with generation). 

Do Not Merge: This would be merged after bugbash

## Tests

Integration tests

NO_CHANGELOG=true


internal/providers/pluginfw/products/app/data_app.go
```diff
@@ -24,20 +24,22 @@ type dataSourceApp struct {
 }
 
 type dataApp struct {
-	Name types.String `tfsdk:"name"`
-	App  types.Object `tfsdk:"app"`
+	Name               types.String `tfsdk:"name"`
+	App                types.Object `tfsdk:"app"`
+	ProviderConfigData types.Object `tfsdk:"provider_config"`
```
mgyucht, line 29: Still best to keep these names matching with the tfsdk names where possible.

internal/providers/pluginfw/products/sharing/resource_acc_test.go
```diff
@@ -192,6 +192,9 @@ func TestUcAccUpdateShareAddObject(t *testing.T) {
 				data_object_type = "TABLE"
 				history_data_sharing_status = "ENABLED"
 			}
+			provider_config = {
+				workspace_id = "{env.THIS_WORKSPACE_ID}"
+			}
```
mgyucht, line 201: Add a PlanCheck on PreApply to verify that the share is not being deleted here.

mgyucht (commented):
1. make sure we are checking the plans for each step that resources aren't being recreated when that isn't expected
2. ditto for when we expect updates
3. fail planning with invalid configs (e.g. workspace_id = null or empty string)
4. use CustomizeDiff or RequiresReplaceIf to only replace the resource if the previous value is not null and is different from the current value.

PR #5062, opened 2025-09-29
[Internal] Unified Terraform Provider: Add support for client

tanmay-db:
## Changes

The changes have been merged, however pending bump in Go SDK followed by Go SDK release and then bump of Go SDK in TF. 

Go SDK bump contains breaking changes that should be prevented (https://github.com/databricks/databricks-sdk-go/pull/1296/files#diff-be6b23b09a52faa2956cd487dac672acbd641362243974fac28390d444c55fa9L2911). This would lead to delay. Adding the client changes post generation to unblock.

## Tests

Existing tests upstream + unit test

Note: we don't mock CurrentWorkspaceID, this code path would be better to be covered with integration tests for resources using unified provider. We integration test this in Go SDK as well: https://github.com/databricks/databricks-sdk-go/blob/main/internal/workspace_client_test.go#L9 

- [ ] `make test` run locally
- [ ] using Go SDK

NO_CHANGELOG=true


tanmay-db:
Fixing this upstream

PR #5068, opened 2025-09-30
[Feature] Perform workspace-level permission assignment by `user_name`, `group_name`, or `service_principal_name`

alexott:
## Changes

Use a previously not available API to perform permission assignment by user name, group name or SP application ID.  This solves a long-standing problem with the assignment of principals by workspace administrators who don't have access to account-level APIs.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/access/resource_permission_assignment.go
+++ b/access/resource_permission_assignment.go
@@ -2,7 +2,9 @@ package access
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strconv"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/terraform-provider-databricks/common"
@@ -22,9 +24,42 @@ type Permissions struct {
 	Permissions []string `json:"permissions"`
 }
 
-func (a PermissionAssignmentAPI) CreateOrUpdate(principalId int64, r Permissions) error {
-	path := fmt.Sprintf("/preview/permissionassignments/principals/%d", principalId)
-	return a.client.Put(a.context, path, r)
+func (a PermissionAssignmentAPI) CreateOrUpdate(assignment permissionAssignmentEntity) (principalInfo, error) {
+	if assignment.PrincipalId != 0 {
+		var resp permissionAssignmentResponseItem
+		path := fmt.Sprintf("/preview/permissionassignments/principals/%d", assignment.PrincipalId)
+		err := a.client.PutWithResponse(a.context, path, Permissions{Permissions: assignment.Permissions}, &resp)
+		if err != nil {
+			return principalInfo{}, err
+		}
+		return resp.Principal, nil
+	} else {
+		var principal permissionAssignmentResponse
+		path := "/preview/permissionassignments"
+		request := permissionAssignmentRequest{
+			PermissionAssignments: []permissionAssignmentRequestItem{
+				{
+					principalInfo: principalInfo{
+						ServicePrincipalName: assignment.ServicePrincipalName,
+						UserName:             assignment.UserName,
+						GroupName:            assignment.GroupName,
+					},
+					Permissions: assignment.Permissions,
+				},
+			},
+		}
+		err := a.client.Post(a.context, path, request, &principal)
+		if err != nil {
+			return principalInfo{}, err
+		}
+		if len(principal.PermissionAssignments) == 0 {
+			return principalInfo{}, fmt.Errorf("no permission assignment found")
+		}
+		if principal.PermissionAssignments[0].Error != "" {
+			return principalInfo{}, errors.New(principal.PermissionAssignments[0].Error)
+		}
+		return principal.PermissionAssignments[0].Principal, nil
+	}
 }
 
 func (a PermissionAssignmentAPI) Remove(principalId string) error {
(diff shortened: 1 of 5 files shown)


common/client.go
```diff
@@ -269,6 +269,11 @@ func (c *DatabricksClient) Put(ctx context.Context, path string, request any) er
 	return c.Do(ctx, http.MethodPut, path, nil, nil, request, nil, c.addApiPrefix)
 }
 
+// Put on path
+func (c *DatabricksClient) PutWithResponse(ctx context.Context, path string, request any, response any) error {
+	return c.Do(ctx, http.MethodPut, path, nil, nil, request, response, c.addApiPrefix)
+}
+
```
mgyucht: This file is now codegen'd. Can you just call the `c.Do` method directly in your code? If you want, I can add it to our internal copy.

access/resource_permission_assignment.go
```diff
@@ -22,39 +24,92 @@ type Permissions struct {
...
+		path := fmt.Sprintf("/preview/permissionassignments/principals/%d", assignment.PrincipalId)
+		err := a.client.PutWithResponse(a.context, path, Permissions{Permissions: assignment.Permissions}, &resp)
+		if err != nil {
+			return principalInfo{}, err
+		}
+		if resp.Error != "" {
+			return principalInfo{}, errors.New(resp.Error)
+		}
+		return resp.Principal, nil
+	} else {
```
mgyucht, line 38: We can inline the `else` case because the `if` case always returns.

access/resource_permission_assignment.go
```diff
@@ -22,39 +24,92 @@ type Permissions struct {
...
+		var principal permissionAssignmentResponse
+		path := "/preview/permissionassignments"
```
mgyucht: Can we either inline this constant or define it as a `const` outside of the method?

access/resource_permission_assignment.go
```diff
@@ -22,39 +24,92 @@ type Permissions struct {
...
+type permissionAssignmentResponseItem struct {
 	Permissions []string `json:"permissions"`
-	Principal   Principal
+	Principal   principalInfo
+	Error       string `json:"error,omitempty"`
 }
 
-type PermissionAssignmentList struct {
-	PermissionAssignments []PermissionAssignment `json:"permission_assignments"`
+type permissionAssignmentResponse struct {
```
mgyucht, line 94: thank you for unexporting these types!

access/resource_permission_assignment.go
```diff
@@ -63,48 +118,63 @@ func (l PermissionAssignmentList) ForPrincipal(principalId int64) (res Permissio
...
-		PrincipalId int64    `json:"principal_id"`
-		Permissions []string `json:"permissions" tf:"slice_as_set"`
-	}
-	s := common.StructToSchema(entity{}, common.NoCustomize)
+	s := common.StructToSchema(permissionAssignmentEntity{}, func(s map[string]*schema.Schema) map[string]*schema.Schema {
+		fields := []string{"principal_id", "service_principal_name", "user_name", "group_name"}
+		for _, field := range fields {
+			s[field].ExactlyOneOf = fields
+		}
+		common.CustomizeSchemaPath(s, "display_name").SetReadOnly().SetOptional()
```
mgyucht:
```suggestion
		common.CustomizeSchemaPath(s, "display_name").SetReadOnly()
```
Can users specify `display_name`? Otherwise it should probably just not be optional.

access/resource_permission_assignment.go
```diff
@@ -63,48 +118,63 @@ func (l PermissionAssignmentList) ForPrincipal(principalId int64) (res Permissio
...
 	return
 }
 
+type permissionAssignmentEntity struct {
+	PrincipalId          int64    `json:"principal_id,omitempty" tf:"computed,force_new"`
+	ServicePrincipalName string   `json:"service_principal_name,omitempty" tf:"computed,force_new"`
+	UserName             string   `json:"user_name,omitempty" tf:"computed,force_new"`
+	GroupName            string   `json:"group_name,omitempty" tf:"computed,force_new"`
+	Permissions          []string `json:"permissions" tf:"slice_as_set"`
+	DisplayName          string   `json:"display_name,omitempty" tf:"computed"`
```
mgyucht: If users can't specify display name, why have it as optional?
```suggestion
	DisplayName          string   `json:"display_name" tf:"computed"`
```

mgyucht: Actually, do we even need to include display name here? It should be available via the databricks_user/databricks_group/databricks_service_principal resource/data source.

common/client.go
alexott: I removed it for this implementation, but it would be nice to add it to the generated code as for `Do` I need to explicitly use `/api/2.0` prefix for path.

mgyucht (approved):
Thanks!

danielkwok72:
Hi @alexott 

I tried testing v1.92.0 or v1.93.0 with the databricks_permission_assignment resource to assign a group as ADMIN to a databricks workspace, and it's giving me the following error: 

module.assignment.databricks_permission_assignment.this: Creating...
**Error: cannot create permission assignment: Invalid permission: ADMIN is invalid or not supported for this workspace. with module.assignment.databricks_permission_assignment.this,**
on modules/permission-assignment/main.tf line 1, in resource
"databricks_permission_assignment" "this":
1: resource "databricks_permission_assignment" "this"

if I assign it to the "USER" role or first assign it to USER role and then update it to ADMIN role, then it works. But I think there is a bug there that stopping me to assign ADMIN role directly.

joaorpsoares:
Hi 👋 This is great, but unfortunately, it does not work if AIM is enabled and the identity (user/group/SP) is not yet provisioned at the account level, even though it exists in Entra ID. 
Opened this issue -> https://github.com/databricks/terraform-provider-databricks/issues/5414.

PR #5096, opened 2025-10-14
Do not update permissions for pipelines on delete

denik:
Logging requests, I see that it tries to set IS_OWNER to run_as service principal instead of actual owner.

Mitigating https://community.databricks.com/t5/data-engineering/dab-dlt-destroy-fails-due-to-ownership-permissions-mismatch/td-p/132101

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Breaking Changes
 
 * Remove stale resources/datasources/documentation related to Clean Room services.
-* databricks\_permissions resource no longer updates permissions on delete. This is to mitigate an issue with incorrect IS\_OWNER being set.
+* databricks\_permissions resource no longer updates permissions on delete. This is to mitigate an issue with incorrect IS\_OWNER being set ([#5096](https://github.com/databricks/terraform-provider-databricks/pull/5096))
 
 ### New Features and Improvements
 


permissions/resource_permissions.go
```diff
@@ -92,6 +92,13 @@ func (a PermissionsAPI) Update(objectID string, entity entity.PermissionsEntity,
 // by the current user and admin group. If the resource has IS_OWNER permissions, they are reset to the
 // object creator, if it can be determined.
 func (a PermissionsAPI) Delete(objectID string, mapping resourcePermissions) error {
+	if mapping.objectType == "pipelines" {
+		// There is a bug which causes the code below send IS_OWNER with run_as identity
+		// Which is of course wrong thing to do.
+		// For non-admin users this results in the error: https://community.databricks.com/t5/data-engineering/dab-dlt-destroy-fails-due-to-ownership-permissions-mismatch/td-p/132101
+		// For admin users situation is worse but there is no error, it silently changes owner to wrong identity.
+		return nil
+	}
```
alexott, line 101: Can we set access control list to an empty list? Admins should have inherited permissions...

denik: Not sure I understand, the goal is to delete the resource from the state but skip API calls. In which place do we "set access control list to an empty list"?

alexott: I meant - we can't make an API call with empty payload at all?

denik: I believe this is not done here because it risks locking out current IS_OWNER (or error that you cannot remove owner).

denik:
One integration test failed in one environment (azure-prod-ucws-is)

`--- FAIL: ***DataSourceShares (10.62s)`

It seems unrelated to this change.

Issue #1865, opened 2022-12-20
[FEATURE] add databricks_workspace data source to retrieve an existing workspace


riturr:
### Use-cases

Similar to #1651, I need to reference an existing workspace to attach it to a metastore.
Currently we can only hardcode the workspace ID.
### Proposal

```
data "databricks_workspace" "this" {
  name = "MyWorkspace"
}

resource "databricks_metastore_assignment" "this" {
  metastore_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  workspace_id = data.databricks_workspace.this.id
}
```
### References

#1651

natbaker:
I believe you can get this from [databricks_mws_workspaces](https://github.com/databricks/terraform-provider-databricks/blob/master/docs/data-sources/mws_workspaces.md)

`
provider "databricks" {
  alias = "mws"
  account_id = "abcd-0123"
}

data "databricks_mws_workspaces" "all" {
    provider = databricks.mws 
}

resource "databricks_metastore_assignment" "this" {
  metastore_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  workspace_id = data.databricks_mws_workspaces.all.ids["MyWorkspace"]
}`

alexott:
What about adding it to the `current_user` data source? We already have workspace URL there...

riturr:
> I believe you can get this from [databricks_mws_workspaces](https://github.com/databricks/terraform-provider-databricks/blob/master/docs/data-sources/mws_workspaces.md)
> 
> ```
> provider "databricks" {
>   alias = "mws"
>   account_id = "abcd-0123"
> }
> 
> data "databricks_mws_workspaces" "all" {
>     provider = databricks.mws 
> }
> 
> resource "databricks_metastore_assignment" "this" {
>   metastore_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
>   workspace_id = data.databricks_mws_workspaces.all.ids["MyWorkspace"]
> }
> ```

It doesn't seem to work for me, Am I missing something?. I'm using Azure Databricks.

When trying to resolve the `data` block Terraform throws an error not being able to parse the response: 

`Error: cannot read data mws workspaces: invalid character '<' looking for beginning of value
│ <!doctype html><html lang="en"><... (695 more bytes)`

[...]

PR #5100, opened 2025-10-16
Improve mws_workspaces data source

kishaningithub:
## Changes

Addresses https://github.com/databricks/terraform-provider-databricks/issues/1865#issuecomment-3014079353

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/mws/data_mws_workspaces.go
+++ b/mws/data_mws_workspaces.go
@@ -7,9 +7,70 @@ import (
 	"github.com/databricks/terraform-provider-databricks/common"
 )
 
+type WorkspaceData struct {
+	AccountID                           string `json:"account_id,omitempty" tf:"computed"`
+	WorkspaceName                       string `json:"workspace_name,omitempty" tf:"computed"`
+	DeploymentName                      string `json:"deployment_name,omitempty" tf:"computed"`
+	AwsRegion                           string `json:"aws_region,omitempty" tf:"computed"`
+	CredentialsID                       string `json:"credentials_id,omitempty" tf:"computed"`
+	CustomerManagedKeyID                string `json:"customer_managed_key_id,omitempty" tf:"computed"`
+	StorageConfigurationID              string `json:"storage_configuration_id,omitempty" tf:"computed"`
+	ManagedServicesCustomerManagedKeyID string `json:"managed_services_customer_managed_key_id,omitempty" tf:"computed"`
+	StorageCustomerManagedKeyID         string `json:"storage_customer_managed_key_id,omitempty" tf:"computed"`
+	PricingTier                         string `json:"pricing_tier,omitempty" tf:"computed"`
+	PrivateAccessSettingsID             string `json:"private_access_settings_id,omitempty" tf:"computed"`
+	NetworkID                           string `json:"network_id,omitempty" tf:"computed"`
+	IsNoPublicIPEnabled                 bool   `json:"is_no_public_ip_enabled,omitempty" tf:"computed"`
+	WorkspaceID                         int64  `json:"workspace_id,omitempty" tf:"computed"`
+	WorkspaceURL                        string `json:"workspace_url,omitempty" tf:"computed"`
+	WorkspaceStatus                     string `json:"workspace_status,omitempty" tf:"computed"`
+	WorkspaceStatusMessage              string `json:"workspace_status_message,omitempty" tf:"computed"`
+	ExpectedWorkspaceStatus             string `json:"expected_workspace_status,omitempty" tf:"computed"`
+	CreationTime                        int64  `json:"creation_time,omitempty" tf:"computed"`
+	//ExternalCustomerInfo                *externalCustomerInfo    `json:"external_customer_info,omitempty" tf:"computed"`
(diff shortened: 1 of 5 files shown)


kishaningithub:
@tanmay-db Have completed addressing all the items in the PR check list. Kindly review the PR and let me know if further changes are required.

tanmay-db (commented):
Hi @kishaningithub, thanks for the contribution. I have triggered integration tests.

mws/data_mws_workspaces.go
```diff
@@ -9,7 +9,8 @@ import (
 
 func DataSourceMwsWorkspaces() common.Resource {
 	type mwsWorkspacesData struct {
-		Ids map[string]int64 `json:"ids" tf:"computed"`
+		Ids             map[string]int64  `json:"ids" tf:"computed"`
+		DeploymentNames map[string]string `json:"deployment_names" tf:"computed"`
```
tanmay-db: If we expand the data source to contain more information. Then I think this should become a map of workspace ids to workspaceInfo which contains information about a workspace including id, deployment name etc...

what do you think @alexott ?

alexott: Yes, it makes sense to include the complete workspace info struct, similar to https://github.com/databricks/terraform-provider-databricks/pull/5164

tanmay-db: Hi @kishaningithub, could you update the PR similar to what @alexott has for the `databricks_service_principals` data source linked above?

kishaningithub: It definitely makes sense to include all the fields that is coming out of the API. I was mainly concerned about breaking changes to the terraform state. Will take a look at your PR @alexott and work on it

tanmay-db: @alexott for changes to terraform state (from ids to workspaceInfo) this would just show up in the plan right?

alexott: yes, there will be no breaking changes

kishaningithub: @alexott / @tanmay-db Do you have any suggestions on handling the following nested fields in the `Workspace` struct?

Link: https://github.com/databricks/terraform-provider-databricks/blob/5c9ecc9b114c53ab2e9eb3f61a73e9bb3407ba2a/mws/resource_mws_workspaces.go#L102

Screenshot:
image

kishaningithub: @tanmay-db / @alexott Have updated the code based on my understanding. Kindly review on the field mapping and the test part

tanmay-db: Hi @kishaningithub , thanks for the update. I was on holidays, will take a look.

mws/data_mws_workspaces.go
```diff
@@ -7,9 +7,70 @@ import (
 	"github.com/databricks/terraform-provider-databricks/common"
 )
 
+type WorkspaceData struct {
```
tanmay-db: @kishaningithub Instead of creating a new type, I think we should use the existing one: https://github.com/databricks/terraform-provider-databricks/blob/main/mws/resource_mws_workspaces.go#L82 to reduce redundancy.

mws/data_mws_workspaces.go
```diff
@@ -7,9 +7,70 @@ import (
...
+	WorkspaceID                         int64  `json:"workspace_id,omitempty" tf:"computed"`
+	WorkspaceURL                        string `json:"workspace_url,omitempty" tf:"computed"`
+	WorkspaceStatus                     string `json:"workspace_status,omitempty" tf:"computed"`
+	WorkspaceStatusMessage              string `json:"workspace_status_message,omitempty" tf:"computed"`
+	ExpectedWorkspaceStatus             string `json:"expected_workspace_status,omitempty" tf:"computed"`
+	CreationTime                        int64  `json:"creation_time,omitempty" tf:"computed"`
+	//ExternalCustomerInfo                *externalCustomerInfo    `json:"external_customer_info,omitempty" tf:"computed"`
+	//CloudResourceBucket                 *CloudResourceContainer  `json:"cloud_resource_container,omitempty" tf:"computed"`
+	//GCPManagedNetworkConfig             *GCPManagedNetworkConfig `json:"gcp_managed_network_config,omitempty" tf:"computed"`
+	//GkeConfig                           *GkeConfig               `json:"gke_config,omitempty" tf:"computed"`
```
tanmay-db: Any reason these are commented out?

mws/data_mws_workspaces.go
kishaningithub: Have addressed the above feedback. Can you take a look?

mws/data_mws_workspaces.go
kishaningithub: Have addressed the above feedback. Can you take a look?

kishaningithub:
@alexott / @tanmay-db I am mainly concerned about this structure as we cannot change it in the future without breaking backwards compatibility. Can you let me know your thoughts? (screenshot below)

image

tanmay-db:
Hi @kishaningithub , thanks for the update, we will take a look

kishaningithub:
Not stale

PR #5109, opened 2025-10-17
Fix `ADMIN` assignment in `databricks_permission_assignment` resource

### LINKED ISSUE #5106: [ISSUE] Issue with `databricks_permission_assignment` resource
### Configuration

```hcl
# Create a local module with following resource block
resource "databricks_permission_assignment" "this" {
  group_name = var.group_name
  permissions  = ["ADMIN"]
}

# Calling the module in the TF workspace
module "assignment" {
  source = "./modules/permission-assignment"

  providers = {
    databricks = databricks.workspace
  } 

  group_name = "dan-guo-grp"
}

# databricks provider
provider "databricks" {
  alias = "workspace"
  host                        = "https://adb-2432776786089467.7.azuredatabricks.net"
  azure_workspace_resource_id = "xxxxxxxxxx"
  azure_client_id             = data.vault_generic_secret.azurerm_auth.data["xxxx"]
  azure_client_secret         = data.vault_generic_secret.azurerm_auth.data["xxxx"]
  azure_tenant_id             = data.vault_generic_secret.azurerm_auth.data["xxxx"]
}
```

### Expected Behavior
Terraform apply should be successful

### Actual Behavior
It's giving me the following error:

module.assignment.databricks_permission_assignment.this: Creating...
╷
│ Error: cannot create permission assignment: Invalid permission: ADMIN is invalid or not supported for this workspace.
│ 
│   with module.assignment.databricks_permission_assignment.this,
│   on modules/permission-assignment/main.tf line 1, in resource "databricks_permission_assignment" "this":
│    1: resource "databricks_permission_assignment" "this" {

### Steps to Reproduce

terraform apply

1.92.0 or 1.93.0

### Is it a regression?

This is a new feature released in 1.92.0

alexott:
## Changes

It was found that assignment by name works only for the `USER` role, so the direct assignment as `ADMIN` fails.  This PR fixes this issue by doing assignment as `USER` first, and then changing it to `ADMIN`.

Resolves #5106

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/access/resource_permission_assignment.go
+++ b/access/resource_permission_assignment.go
@@ -150,6 +150,8 @@ func ResourcePermissionAssignment() common.Resource {
 			var assignment permissionAssignmentEntity
 			common.DataToStructPointer(d, s, &assignment)
 			api := NewPermissionAssignmentAPI(ctx, c)
+			// We need this because assignment by name doesn't work for admins, so we need to
+			// first assign them as users.  And then reassign them as admins.
 			shouldReassignAdmin := false
 			if assignment.PrincipalId == 0 && slices.Contains(assignment.Permissions, "ADMIN") {
 				shouldReassignAdmin = true


access/resource_permission_assignment.go
```diff
@@ -148,11 +150,26 @@ func ResourcePermissionAssignment() common.Resource {
 			var assignment permissionAssignmentEntity
 			common.DataToStructPointer(d, s, &assignment)
 			api := NewPermissionAssignmentAPI(ctx, c)
+			shouldReassignAdmin := false
+			if assignment.PrincipalId == 0 && slices.Contains(assignment.Permissions, "ADMIN") {
+				shouldReassignAdmin = true
+				assignment.Permissions = []string{"USER"}
+			}
```
tanmay-db, line 159: Would be good to add why we are doing this in comments.

access/resource_permission_assignment_test.go
```diff
@@ -101,6 +101,128 @@ func TestPermissionAssignmentCreateWithUserName(t *testing.T) {
 	}.ApplyNoError(t)
 }
 
+func TestPermissionAssignmentCreateWithUserNameAndAdminPermissions(t *testing.T) {
```
tanmay-db, line 104: Would be good to also have integration test for this

access/resource_permission_assignment.go
alexott: Added

access/resource_permission_assignment_test.go
alexott: Do we support integration tests with two providers? Because we need an account-level provider to create a group, and then a workspace-level provider to perform the assignment...

PR #5110, opened 2025-10-17
Fix crash when error happens during reading `databricks_job`

alexott:
## Changes

The crash occurred when reading a job and backend returned an error, so the `job` became `nil`, but we still tried to access fields inside the struct, instead of returning the error immediately.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


jobs/jobs_api_go_sdk.go
```diff
@@ -313,6 +313,9 @@ func Read(jobID int64, w *databricks.WorkspaceClient, ctx context.Context) (job
 		JobId: jobID,
 	})
 	err = wrapMissingJobError(err, fmt.Sprintf("%d", jobID))
+	if err != nil {
+		return
+	}
```
tanmay-db, line 318: Should we add a test for this? i.e before it panics and now it works?

alexott: Hmmmm, how can we produce panic when it happens in the code that is skipped by this condition?

tanmay-db: My understanding is that the panic happens when the job is removed on backend and .Get return nil?
> backend changed the response from zeroed struct to nil, or sdk decodes it differently

If so we should be able to return nil from the test fixtures?

tanmay-db: something like TestResourceJobRead but with Nil response.

tanmay-db: I am able to panic with it
````
	d, err := qa.ResourceFixture{
		Fixtures: []qa.HTTPFixture{
			{
				Method:   "GET",
				Resource: "/api/2.0/jobs/get?job_id=789",
				Response: nil,
			},
		},
		Resource: ResourceJob(),
		Read:     true,
		New:      true,
		ID:       "789",
	}.Apply(t)
````

tanmay-db: Error:      	Received unexpected error:
        	            	panic: runtime error: invalid memory address or nil pointer dereference
        	Test:       	TestResourceJobRead
    resource_job_test.go:2309:

tanmay-db: we would also need to return an error so we check it before doing job.Settings

alexott: No, API doesn't return `nil` - it returns normal HTTP code, and then SDK converts it into `nil, err`...

PR #5111, opened 2025-10-17
Ensure `workload_size` is marked as non-computed for `databricks_model_serving`

mgyucht:
## Changes
`workload_size` is currently marked as `optional` and `computed`. As a result, if a user sets a model serving endpoint using `workload_size` initially, then removes it in favor of `min_provisioned_concurrency`/`max_provisioned_concurrency`, `workload_size` remains in the state. As a result, it is also included in the request sent to the REST API, which is invalid.

By marking this field as non-computed, it should be removed from the plan when a user removes it from their config. This means it won't be sent to the REST API.

## Tests
- [ ] Manual test: created a model serving endpoint with `workload_size` set, then removed it and set min/max provisioned concurrency fields, and the apply succeeded.


alexott:
@mgyucht this change introduces configuration drift:

```
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # databricks_model_serving.endpoint will be updated in-place
          ~ resource "databricks_model_serving" "endpoint" {
                id                  = "terraform-test-model-serving-pt-l0dri"
                name                = "terraform-test-model-serving-pt-l0dri"
                # (3 unchanged attributes hidden)
        
              ~ config {
                  ~ served_entities {
                      + environment_vars            = {}
                        name                        = "pt_model"
                      - workload_size               = "Small" -> null
                        # (10 unchanged attributes hidden)
                    }
        
                    # (1 unchanged block hidden)
                }
            }
```

mgyucht:
> @mgyucht this change introduces configuration drift

Yup, I need to address this before merging, but it is expected unfortunately because this field is always set in the response

PR #5146, opened 2025-10-24
[Internal] Unified provider support for resources and all data sources on SDKv2 using Go SDK

tanmay-db:
## Changes

- Add unified provider support for resources and datasources using Go SDK
- Add support for jobs and clusters resources 
- Support for all data sources using Go SDK

Keeping no change for these and will announce all together as part of documentation of unified TF. Keeping changelog now will cause lots of conflict during merging. 

## Tests

Integration tests
Unit tests

NO_CHANGELOG=true


common/unified_provider.go
```diff
@@ -0,0 +1,82 @@
...
+// NamespaceCustomizeSchema is used to customize the schema for the provider configuration
+// for a single schema.
+func NamespaceCustomizeSchema(s *CustomizableSchema) {
+	s.SchemaPath("provider_config", "workspace_id").SetValidateFunc(workspaceIDValidateFunc())
+}
+
+// NamespaceCustomizeSchemaMap is used to customize the schema for the provider configuration
+// in a map of schemas.
+func NamespaceCustomizeSchemaMap(m map[string]*schema.Schema) map[string]*schema.Schema {
+	if providerConfig, ok := m["provider_config"]; ok {
```
mgyucht: let's panic here if these fields are not found

common/unified_provider.go
```diff
@@ -0,0 +1,82 @@
...
+// across resources onboarded to SDKv2.
+type ProviderConfig struct {
+	WorkspaceID string `json:"workspace_id"`
+}
+
+// workspaceIDValidateFunc is used to validate the workspace ID for the provider configuration
+func workspaceIDValidateFunc() func(interface{}, string) ([]string, []error) {
+	return validation.All(
+		validation.StringIsNotEmpty,
+		validation.StringMatch(regexp.MustCompile(`^\d+$`), "workspace_id must be a valid integer"),
```
mgyucht:
```suggestion
		validation.StringMatch(regexp.MustCompile(`^[1-9]\d*$`), "workspace_id must be a valid integer"),
```

common/unified_provider_test.go
```diff
@@ -0,0 +1,524 @@
...
+		}
+
+		result := NamespaceCustomizeSchemaMap(testSchema)
+		assert.Equal(t, testSchema, result, "Schema should be returned unchanged when provider_config is not present")
+	})
+}
+
+func TestNamespaceCustomizeDiff(t *testing.T) {
+	testSchema := map[string]*schema.Schema{
+		"provider_config": {
```
mgyucht: What is this testing?

common/unified_provider_test.go
```diff
@@ -0,0 +1,524 @@
...
+			} else {
+				assert.NoError(t, err, tc.description)
+				assert.NotNil(t, result)
+				assert.Equal(t, mockWorkspaceClient, result)
+			}
+		})
+	}
+}
+
+func TestWorkspaceIDExtractionLogic(t *testing.T) {
```
mgyucht: This also is effectively not guarding against incorrect non-test code.

common/unified_provider_test.go
```diff
@@ -0,0 +1,524 @@
...
+					DatabricksClient: &client.DatabricksClient{
+						Config: &config.Config{
+							Host:  "https://test.cloud.databricks.com",
+							Token: "test-token",
+						},
+					},
+					cachedWorkspaceClient: mockWorkspaceClient,
+					cachedWorkspaceID:     tc.cachedWorkspaceID,
+				}
+			}
```
mgyucht: Let's remove the layer of indirection here and just include this in the test case. This makes the test cases slightly larger but makes the test logic much simpler and thus easier for reviewers to understand what each test case actually covers
```suggestion
			dc = &DatabricksClient{
				DatabricksClient: &client.DatabricksClient{
					Config: tc.Config,
					cachedWorkspaceClient: tc.mockWorkspaceClient,
					cachedWorkspaceID:     tc.cachedWorkspaceID,
				},
			}
```

common/unified_provider.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// workspaceIDSchemaKey is the key for the workspace ID in schema
+const workspaceIDSchemaKey = "provider_config.0.workspace_id"
+
+type Namespace struct {
```
mgyucht, line 20: Can you leave a comment here about where this will be used and what it needs to be kept in sync with?

common/unified_provider_test.go
```diff
@@ -0,0 +1,337 @@
...
+						Host:  "https://test.cloud.databricks.com",
+						Token: "test-token",
+					},
+				},
+				cachedWorkspaceClient: mockWorkspaceClient,
+				cachedWorkspaceID:     789012,
+			},
+			expectError: false,
+			description: "Should handle different workspace IDs correctly",
+		},
```
mgyucht, line 296: This should test when the cached workspace ID doesn't match the ID provided in the resource data.

PR #5149, opened 2025-10-24
[Internal] Unified Provider support for UC resources

tanmay-db:
## Changes

Unified Provider Support for the following resources: 
- Catalog

Note: UC Resources will not have `NamespaceCustomizeDiff` which does `d.ForceNew`

## Tests

Added in main branch

NO_CHANGELOG=true


catalog/resource_catalog.go
```diff
@@ -188,7 +194,7 @@ func ResourceCatalog() common.Resource {
 			return bindings.AddCurrentWorkspaceBindings(ctx, d, w, ci.Name, bindings.BindingsSecurableTypeCatalog)
```
mgyucht: This probably needs to be moved up to the top of Update(), otherwise the WorkspaceClient used when changing provider_config.workspace_id may not have access to the isolated catalog (unless that workspace also happens to have a workspace binding separately).

catalog/resource_catalog.go
```diff
@@ -142,6 +148,12 @@ func ResourceCatalog() common.Resource {
 			common.DataToStructPointer(d, catalogSchema, &updateCatalogRequest)
 			updateCatalogRequest.Name = d.Id()
 
+			// Bind the current workspace if the catalog is isolated, otherwise the read will fail
+			err = bindings.AddCurrentWorkspaceBindings(ctx, d, w, updateCatalogRequest.Name, bindings.BindingsSecurableTypeCatalog)
```
alexott, line 152: Hmmm, typically the binding is done on creation. Should we add a check if the isolation mode is changing?

catalog/resource_catalog.go
```diff
@@ -142,6 +148,12 @@ func ResourceCatalog() common.Resource {
...
+			if err != nil {
+				return err
```
alexott, line 154: We also need to handle error like this: https://github.com/databricks/terraform-provider-databricks/pull/5149/files#diff-2dee1865e5fcab955309d55246c5d88e1692d4ea8008c9d5c133505fa6877130R187

alexott (commented):
In general should be ok, but we may need to discuss how to handle bindings

PR #5150, opened 2025-10-24
[Internal] Refactor data source methods for SDKv2 resources

tanmay-db:
## Changes

These changes are done upstream since these shouldn't be part of autogeneration as they are used by legacy SDKv2 resources

## Tests

Existing tests
NO_CHANGELOG=true

### DIFF
--- a/aws/user_role_test.go
+++ b/aws/user_role_test.go
@@ -7,6 +7,9 @@ import (
 )
 
 func TestAccUserRole(t *testing.T) {
+	if !acceptance.IsAws(t) {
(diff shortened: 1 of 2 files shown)


common/datasource.go
```diff
@@ -0,0 +1,251 @@
+package common
```
mgyucht, line 1: nit: call this file `datasource.go` (in spite of it being used in these functions, "data" is too broad)

mgyucht (approved):
One suggestion, otherwise LGTM!

PR #5161, opened 2025-10-29
Added `databricks_permission` resource for managing permissions for individual principals.

alexott:
## Changes

This resource provides fine-grained control over permissions by managing a single principal's access to a single object, unlike `databricks_permissions`, which manages all principals' access to an object at once. This is particularly useful for:

- Managing permissions for different teams independently
- Token and password authorization permissions that previously required all principals in one resource
- Avoiding conflicts when multiple configurations manage different principals on the same object

Caveat: Since we cannot remove an individual permission, the `Delete` operation is performed as `Read/Put`, so we need to use a lock around each object.

## Tests

- [x] `make test` run locally
- [x] tested manually
- [x] using Go SDK


docs/resources/permission.md
```diff
@@ -0,0 +1,282 @@
...
+subcategory: "Security"
+---
+
+# databricks_permission Resource
+
+This resource allows you to manage permissions for a single principal on a Databricks workspace object. Unlike `databricks_permissions`, which manages all principals' permissions for an object at once, this resource is authoritative for a specific object-principal pair only.
+
+-> This resource can only be used with a workspace-level provider!
+
+~> This resource is _authoritative_ for the specified object-principal pair. Configuring this resource will manage the permission for the specified principal only, without affecting permissions for other principals.
```
mgyucht, line 11: Maybe good to note that you can't use both `databricks_permission` and `databricks_permissions` for the same object.

docs/resources/permission.md
```diff
@@ -0,0 +1,282 @@
...
+
+* `permission_level` - (Required) The permission level to grant. The available permission levels depend on the object type. Common values include `CAN_MANAGE`, `CAN_USE`, `CAN_VIEW`, `CAN_RUN`, `CAN_EDIT`, `CAN_READ`, `CAN_RESTART`, `CAN_ATTACH_TO`.
+
+Exactly one of the following principal identifiers must be specified:
+
+* `user_name` - (Optional) User email address to grant permissions to. Conflicts with `group_name` and `service_principal_name`.
+* `group_name` - (Optional) Group name to grant permissions to. Conflicts with `user_name` and `service_principal_name`.
+* `service_principal_name` - (Optional) Application ID of the service principal. Conflicts with `user_name` and `group_name`.
+
+Exactly one of the following object identifiers must be specified:
```
mgyucht, line 168: Just curious: do you like the pattern of having a separate field for each permission type? I was thinking that it might be more scalable to split a bit from how `databricks_permissions` works and instead have `object_type` and `object_id` fields. That way, as new object types are added, this resource doesn't need to be updated. I think the IDs are all the same type IIUC. We should then point people to the REST API docs for the set of allowed types.

internal/providers/pluginfw/products/permissions/object_mutex.go
```diff
@@ -0,0 +1,62 @@
...
+
+// objectMutexManager manages mutexes per object ID to prevent concurrent
+// operations on the same Databricks object that could lead to race conditions.
+//
+// This is particularly important for Delete operations where multiple
+// databricks_permission resources for the same object might be deleted
+// concurrently, each doing GET -> filter -> SET, which could result in
+// lost permission updates.
+type objectMutexManager struct {
+	mutexes map[string]*sync.Mutex
```
mgyucht, line 15: Suggestion: you can use `sync.Map` and then don't need a separate `mapLock`.

internal/providers/pluginfw/products/permissions/object_mutex.go
```diff
@@ -0,0 +1,62 @@
...
+func lockObject(objectID string) {
+	globalObjectMutexManager.Lock(objectID)
+}
+
+// unlockObject releases the lock for the given object ID.
+// This should be called at the end of any operation that modifies permissions.
+// Use defer to ensure it's always called even if the operation panics.
+func unlockObject(objectID string) {
+	globalObjectMutexManager.Unlock(objectID)
+}
```
mgyucht, line 62: Are these used outside of tests? Do we need these?

internal/providers/pluginfw/products/permissions/object_mutex.go
```diff
@@ -0,0 +1,62 @@
...
+
+// globalObjectMutexManager is the singleton instance used by all permission resources
+var globalObjectMutexManager = &objectMutexManager{
+	mutexes: make(map[string]*sync.Mutex),
+}
+
+// Lock acquires a mutex for the given object ID.
+// Each object ID gets its own mutex to allow concurrent operations on different objects
+// while serializing operations on the same object.
+func (m *objectMutexManager) Lock(objectID string) {
```
mgyucht, line 27: Since this is just used for permissions, maybe we accept object type and object ID as parameters and synthesize an appropriate key.

internal/providers/pluginfw/products/permissions/object_mutex_test.go
```diff
@@ -0,0 +1,167 @@
...
+		mutexes: make(map[string]*sync.Mutex),
+	}
+
+	objectID := "/clusters/test-cluster"
+
+	// First lock should succeed immediately
+	manager.Lock(objectID)
+
+	// Verify mutex was created
+	manager.mapLock.Lock()
```
mgyucht, line 23: You don't need to lock this here; in the context of your test, this is only accessed by one go-routine.

internal/providers/pluginfw/products/permissions/object_mutex_test.go
```diff
@@ -0,0 +1,167 @@
...
+		resultsMutex.Unlock()
+	}()
+
+	wg.Wait()
+
+	// Both operations should complete (order doesn't matter)
+	assert.Len(t, results, 2, "Both operations should complete")
+}
+
+func TestObjectMutexManager_SerializeSameObject(t *testing.T) {
```
mgyucht, line 107: What is the difference between this test and TestObjectMutexManager_ConcurrentAccess?

internal/providers/pluginfw/products/permissions/object_mutex_test.go
```diff
@@ -0,0 +1,167 @@
...
+func TestObjectMutexManager_UnlockNonexistent(t *testing.T) {
+	manager := &objectMutexManager{
+		mutexes: make(map[string]*sync.Mutex),
+	}
+
+	// Unlocking a non-existent object should not panic
+	assert.NotPanics(t, func() {
+		manager.Unlock("/clusters/does-not-exist")
+	}, "Unlocking non-existent object should not panic")
+}
```
mgyucht, line 167: Do we need these?

internal/providers/pluginfw/products/permissions/permission_level_validator.go
```diff
@@ -0,0 +1,84 @@
...
+func (v permissionLevelValidator) Description(ctx context.Context) string {
+	return "validates that the permission level is valid for the configured object type"
+}
+
+func (v permissionLevelValidator) MarkdownDescription(ctx context.Context) string {
+	return "validates that the permission level is valid for the configured object type"
+}
+
+func (v permissionLevelValidator) ValidateString(ctx context.Context, req validator.StringRequest, resp *validator.StringResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
```
mgyucht, line 25: Should we fail if the permission level is null?

internal/providers/pluginfw/products/permissions/permission_level_validator.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"fmt"
+
+	"github.com/databricks/terraform-provider-databricks/permissions"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// permissionLevelValidator validates that the permission_level is valid for the configured object type
```
mgyucht, line 13: Question: why not let the API handle validation? We can add this but then we're on the hook every time new permission levels are added.

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+			Description: "Permission level for the principal on the object (e.g., CAN_MANAGE, CAN_USE, CAN_VIEW).",
+			Validators: []validator.String{
+				ValidatePermissionLevel(),
+			},
+		},
+	}
+
+	// Dynamically add object identifier attributes from permission definitions
+	// Each object identifier is mutually exclusive with all others
+	for i, mapping := range allPermissions {
```
mgyucht, line 133: Related question from above: why not just use`object_type` and `object_id` as the parameters? I know it doesn't match `databricks_permissions` but it requires schema changes and upgrading the provider every time a new resource is released.

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+
+	// Get the object mapping and ID (reads object identifiers dynamically from plan)
+	mapping, objectID, objectFieldName, objectFieldValue, err := r.getObjectMappingAndID(ctx, w, req.Plan)
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to get object mapping", err.Error())
+		return
+	}
+
+	// Lock the object to prevent concurrent modifications
+	lockObject(objectID)
```
mgyucht, line 206: Don't you need to consider the object type when locking as well?

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("id"), &id)...)
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("user_name"), &userName)...)
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("group_name"), &groupName)...)
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("service_principal_name"), &servicePrincipalName)...)
+	resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("permission_level"), &permissionLevel)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
mgyucht, line 348: You probably want to read the entire plan to an instance of PermissionResourceModel

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+
+	w, err := r.Client.WorkspaceClient()
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to get workspace client", err.Error())
+		return
+	}
+
+	// Parse the ID to get object ID and principal
+	objectID, _, err := r.parseID(id.ValueString())
```
mgyucht, line 357: This should not be necessary in update, since the object ID should be in the PermissionResourceModel

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Update state using SetAttribute
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(mapping.GetField()), objectFieldValue)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("permission_level"), currentPermissionLevel)...)
+}
+
+func (r *PermissionResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
mgyucht, line 336: Are create and update basically the same? (which makes sense because permissions are essentially a singleton) Should we deduplicate by putting the create and update implementations in a single method?

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+		resp.Diagnostics.AddError("Failed to get workspace client", err.Error())
+		return
+	}
+
+	// Parse the ID to get object ID and principal
+	objectID, principal, err := r.parseID(id.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to parse resource ID", err.Error())
+		return
+	}
```
mgyucht, line 427: Ditto, we shouldn't need to parse the ID to get details about the permission, it should be in state. The only place it may not be in state is in Read (post-import).

internal/providers/pluginfw/products/permissions/resource_permission.go
```diff
@@ -0,0 +1,652 @@
...
+
+	if servicePrincipalName != "" {
+		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_principal_name"), types.StringValue(servicePrincipalName))...)
+	} else {
+		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_principal_name"), types.StringNull())...)
+	}
+
+	// Set permission level
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("permission_level"), types.StringValue(permissionLevel))...)
```
mgyucht, line 588: Should we populate an instance of PermissionResourceModel and then write the entire resource to the plan?

permissions/resource_permission_acc_test.go
```diff
@@ -0,0 +1,518 @@
+package permissions_test
```
mgyucht, line 1: Can you move this test to the internal/providers/pluginfw/products/permissions package?

mgyucht (commented):
Got through a pass. LMK what you think!

docs/resources/permission.md
alexott: It could be easier to read in some cases, but let me try to play with it

alexott: Also, we use specific field name to validate what are supported permissions... and additional customization. So it makes sense to support only explicit fields. Plus, compatibilit with `permissions` resouce

tanmay-db (commented):
Hi @alexott, before reviewing this PR, are the comments from Miles addressed?

alexott:
@tanmay-db We need to discuss if we go this way, or #5186 is preferable

PR #5164, opened 2025-10-30
Improve `databricks_service_principals` data source

### LINKED ISSUE #4997: [FEATURE] Add display names to output of databricks_service_principals data source
### Use-cases
We've been using the `databricks_service_principals` data source with a filter on the name to assign things like permissions for multiple service principals. This is great and working fine.

The data source only returns a list of application_ids which we then iterate over. This issue with this is that for validating plan output the application_ids are pretty much meaningless since no one will know which service principal they belong to, making review of changes to infrastructure either kind of pointless since you don't know which service principal things are being assigned to or a lot of work because you have to lookup which service principal actually belongs to which application_id.

### Attempted Solutions
I don't think there's a possible solution or workaround for this at the moment.

### Proposal
Include the display names in the output of the `databricks_service_principals` data source.
I think adding an additional attribute with a map probably makes the most sense, then one can use the display name as the key and the application id as the value, allowing easy use of it and the existing `application_ids` would keep working as is.

That way one would go from
```
databricks_grant.service_principals["7e424903-eea1-473c-a4f9-a1f8403d3961"]
```
to
```
databricks_grant.service_principals["service-principal-foo"]
```
in the resources that are getting instantiated based on the output of the `databricks_service_principals` data source which is much more usable feedback.

One thing to note is that as far as I know service principal names are not required to be unique (might be a good improvement?), so this won't be usable in all cases. We don't have service principals with the same name and tbh I expect any sane and well managed Databricks account to not have duplicate service principals names, but a note or warning might be useful to flag this.

### LINKED ISSUE #5162: [ISSUE] databricks_service_principals fails when there are no matching service principals
### Configuration
```hcl
data "databricks_service_principals" "doesnt_exist" {
  provider = databricks.mws

  display_name_contains = "name-that-doesnt-exist"
}
```

### Expected Behavior
If there are no matching service principals return an empty list

### Actual Behavior
Errors with
```
Error: cannot read service principals: cannot read data service principals: cannot find SPs with display name containing name-that-doesnt-exist
```

### Steps to Reproduce
Run `terraform plan`

```
OpenTofu v1.10.6
on linux_amd64
+ provider registry.opentofu.org/databricks/databricks v1.95.0
```

### Important Factoids
We'd like to dynamically define some resources that should automatically be added based on this lookup. In some cases the service principal(s) don't exit and then we don't want the resources to be created. This use-case breaks with the current implementation.

Also, since the name of the service principal isn't included in the output of `databricks_service_principals` (https://github.com/databricks/terraform-provider-databricks/issues/4997) using no filter at all and then filtering all service principals based on their name afterwards is clunky because it needs an additional intermediate `databricks_service_principal` data source for each (this also means a lot more API calls).

Potentially, if this would be OK and merged soonish.

alexott:
## Changes

Changes include:

- don't throw an error when nothing is found/got an empty result
- Added `service_principals` attribute to contain more information about service principals - SCIM ID, display name, etc.

Resolves #5162, Resolves #4997

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


simonvanderveldt:
Very happy to see this! :)
I believe this would also fix https://github.com/databricks/terraform-provider-databricks/issues/4997

simonvanderveldt:
@alexott Just wanted to check when you expect a release of the provider with these fixes to be available? Releases seem to generally be every week on Thursdays, but last week it didn't happen.

alexott:
Technically, it should be this week...

simonvanderveldt:
OK, that would be great, fingers crossed :crossed_fingers: 
We've already validated the changes from this PR locally, all working as expected, thanks for that! :+1:

PR #5180, opened 2025-11-03
Relaxed `force_new` constraint on `catalog` attribute in `databricks_pipeline` resource to allow changing the default catalog for existing pipelines

### LINKED ISSUE #4567: [FEATURE] Better support of direct publishing mode in `databricks_pipeline`
Right now, the `catalog` field is marked as `ForceNew` in `databricks_pipeline` ([src](https://github.com/databricks/terraform-provider-databricks/blob/main/pipelines/resource_pipeline.go#L208). With the release of direct publishing mode it's now possible to change `catalog` without the need to recreate a pipeline.

But we still need to recreate a pipeline when switching from `storage` to `catalog` or vice versa.

### LINKED ISSUE #4692: [FEATURE] Relax `force_new` on `catalog`/`schema` in `databricks_pipeline` resource
With the latest releases, it's now possible to change `catalog` and `schema` for existing pipelines, so we need to remove `force_new` from these fields

But we need to investigate how to handle change from `storage` to `catalog`/`schema` - this may require an implementation of custom diff function to do the `force_new` for such combination.

alexott:
## Changes

Pipelines will still be recreated when switching between `storage` and `catalog` modes, but changing the catalog value in an existing catalog-based pipeline no longer requires recreation.

Resolves #4692, Resolves #4567

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


pipelines/resource_pipeline.go
```diff
@@ -335,5 +336,41 @@ func ResourcePipeline() common.Resource {
 		Timeouts: &schema.ResourceTimeout{
 			Default: schema.DefaultTimeout(DefaultTimeout),
 		},
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff) error {
+			// Allow changing catalog value in existing pipelines, but force recreation
+			// when switching between storage and catalog (or vice versa).
+			// This should only run on update, thus we skip this check if the ID is not known.
+			if d.Id() != "" {
```
pietern: You can return early if it is empty to achieve the same with less indentation.

pipelines/resource_pipeline.go
```diff
@@ -335,5 +336,41 @@ func ResourcePipeline() common.Resource {
...
+
+					// Force new if switching from catalog to storage
+					if oldCatalog != "" && oldCatalog != nil && newStorage != "" && newStorage != nil {
+						if err := d.ForceNew("catalog"); err != nil {
+							return err
+						}
+						if err := d.ForceNew("storage"); err != nil {
+							return err
+						}
+					}
```
pietern: Shouldn't a change in either always result in recreation, unconditionally?

If not, it would be nice to include a comment stating why the logic is necessary.

pietern (approved):

PR #5185, opened 2025-11-04
[Internal] Sort entities/models in model serving endpoint resources

alexott:
## Changes

Backend has started to return entities in non-deterministic order, so let's try to workaround it.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


alexott:
Didn't help

PR #5205, opened 2025-11-10
Fix catalog name update using NewName field

pierskarsenbarg:
## Changes

The catalog Update operation was using the Name field to identify the catalog to be updated, but when the catalog name should be changed, it looks like the API requires using the NewName field instead. This fixes the bug where attempting to rename a catalog would silently fail because the old name was always used.

Also adds a unit test (TestUpdateCatalogName) that reproduces and verifies the fix for this issue.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Bug Fixes
 
+* Fix `databricks_catalog` name update using NewName field ([#5205](https://github.com/databricks/terraform-provider-databricks/pull/5205))
+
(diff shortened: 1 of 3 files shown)


renaudhartert-db (changes requested):
Thanks for the PR @pierskarsenbarg! While the API allows for renaming, the operation is often problematic in TF as it mutates the identity of the resource and makes the apply non-idempotent (i.e. applying twice in a row will fail). We currently do not intend to support renaming in TF for UC resources.

PR #5212, opened 2025-11-13
[Exporter] Added support for Database Instance resource (aka Lakebase)

alexott:
## Changes

Also adding more generic functions, i.e. handle of effective fields, etc.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/exporter/abstractions_test.go
+++ b/exporter/abstractions_test.go
@@ -1,0 +1,1 @@
+package exporter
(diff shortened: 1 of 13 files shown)


exporter/abstractions_test.go
```diff
@@ -0,0 +1 @@
+package exporter
```
rauchy, line 1: Is this placeholder file intentional? If there are no tests yet for abstractions.go, might be worth either adding tests for `copyEffectiveFieldsToInputFieldsWithConverters` or removing the empty file.

rauchy (approved):

exporter/abstractions_test.go
alexott: Ok, I'll check with the next PRs - I think that it's used there. If not, will remove it

PR #5217, opened 2025-11-14
[Exporter] Added support for `databricks_budget_policy` resource

alexott:
## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -20,14 +20,14 @@ jobs:
     outputs:
       has_token: ${{ steps.set-token-status.outputs.has_token }}
     steps:
-      - name: Check if DECO_WORKFLOW_TRIGGER_APP_ID is set
+      - name: Check if required secrets are set
         id: set-token-status
         run: |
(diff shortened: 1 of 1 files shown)


exporter/impl_billing.go
```diff
@@ -0,0 +1,114 @@
...
+		ID:       "accounts/" + accountID + "/budgetPolicies/" + r.ID + "/ruleSets/default",
+	})
+
+	bindingWorkspaceIdsRaw := r.DataWrapper.Get("binding_workspace_ids")
+	if bindingWorkspaceIdsRaw != nil {
+		// Convert to slice of int64
+		var bindingWorkspaceIds []int64
+		if workspaceIdsList, ok := bindingWorkspaceIdsRaw.([]int64); ok {
+			bindingWorkspaceIds = workspaceIdsList
+		}
```
rauchy, line 54: Maybe worth erroring if you encounter something which is not `[]int64`?

alexott: Our API returns []int64

PR #5237, opened 2025-11-28
[Exporter] Adding more services: DQMv2 + OAuth

alexott:
## Changes

Added support for:
- `databricks_account_federation_policy`
- `databricks_custom_app_integration`
- `databricks_quality_monitor_v2`
- `databricks_service_principal_federation_policy`

## Tests

- [x] `make test` run locally
- [x] using Go SDK


exporter/impl_oauth.go
```diff
@@ -0,0 +1,114 @@
...
+			ID:       policy.PolicyId,
+			Name:     fmt.Sprintf("acc_fed_policy_%s", name),
+		})
+	}
+	return nil
+}
+
+func listServicePrincipalFederationPolicies(ic *importContext) error {
+	// First, list all service principals
+	sps, err := ic.accountClient.ServicePrincipals.ListAll(ic.Context, iam.ListAccountServicePrincipalsRequest{
```
rauchy, line 64: Should you use `ServicePrincipalFederationPolicy` instead and just go over the service principals that _have_ a federation policy?

alexott: The List operation there requires a specific SP SCIM ID :-( 

https://docs.databricks.com/api/account/serviceprincipalfederationpolicy/list

PR #5238, opened 2025-11-30
[Exporter] Add support for serverless egress controls

alexott:
## Changes

This includes support for:

- `databricks_account_network_policy`
- `databricks_workspace_network_option`

Also introduced `NameUnified` for correct generation of TF resource names

## Tests

- [x] `make test` run locally
- [x] using Go SDK


exporter/impl_serverless.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Get the workspace network option to emit the associated network policy
+	option, err := ic.accountClient.WorkspaceNetworkConfiguration.GetWorkspaceNetworkOptionRpc(ic.Context,
+		settings.GetWorkspaceNetworkOptionRequest{
+			WorkspaceId: workspaceId,
+		})
+	if err != nil {
+		log.Printf("[WARN] Failed to get workspace network option for workspace %d: %s", workspaceId, err.Error())
+		return nil
+	}
```
rauchy, line 78: Should we fail here?

rauchy (changes requested):
Can you add some test coverage?

exporter/impl_serverless.go
alexott: Not really - we may need to be a bit permissive for the exporter.

alexott:
Regarding tests, the current test framework doesn't work for account-level resources. It requires a completely new implementation

alexott:
@rauchy I've added tests, and then https://github.com/databricks/terraform-provider-databricks/pull/5262 adds even more

840 (approved):

PR #5241, opened 2025-12-01
[Internal] Add integration test for share resource with Volume type

tanmay-db:
## Changes

There was a regression with share resource using VOLUME type for shared data object. With the change in Go SDK, this test verifies that the VOLUME type is supported.

## Tests

Added integration test

NO_CHANGELOG=true

### DIFF
--- a/internal/providers/pluginfw/products/sharing/resource_share_acc_test.go
+++ b/internal/providers/pluginfw/products/sharing/resource_share_acc_test.go
@@ -783,7 +783,7 @@ func TestAccShare_ProviderConfig_Remove(t *testing.T) {
 
 const shareVolumeTemplate = preTestTemplate + `
 resource "databricks_volume" "this" {
-	name         = "share_volume_test{var.RANDOM}"
+	name         = "share_volume_test{var.STICKY_RANDOM}"
 	catalog_name = databricks_catalog.sandbox.name
 	schema_name  = databricks_schema.things.name
 	volume_type  = "MANAGED"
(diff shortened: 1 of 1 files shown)


alexott:
@tanmay-db The problem is that in old implementation it was a string, but in the new one - it's ENUM, and REST API missed it. I already filed the ES ticket last week, and Delta Sharing team assigned it to oncall

alexott (approved):

internal/providers/pluginfw/products/sharing/resource_share_acc_test.go
```diff
@@ -780,3 +780,35 @@ func TestAccShare_ProviderConfig_Remove(t *testing.T) {
 		},
 	})
 }
+
+const shareVolumeTemplate = preTestTemplate + `
+resource "databricks_volume" "this" {
+	name         = "share_volume_test{var.RANDOM}"
```
alexott: Should it be sticky random?

alexott:
Fails with 

```
When expanding the plan for databricks_share.myshare to include new values
        learned so far during apply, provider
        "registry.terraform.io/hashicorp/databricks" produced an invalid new value
        for .object[0].name: was
        cty.StringVal("sandboxsfbfcdadfhiel.thingssfbfcdadfhiel.share_volume_testtdcbdeebcehaj"),
        but now
        cty.StringVal("sandboxsfbfcdadfhiel.thingssfbfcdadfhiel.share_volume_testtkdjibgfbaadj").
```

PR #5244, opened 2025-12-02
[Documentation] Update documentation for data object type for share resource

tanmay-db:
## Changes

The documentation for data object type is out of date as it's manually maintained. Updating it to include types that are supported eg: `VOLUME` . Also added `content` and `partition` as part of documentation.

Note: List is based on the changes merged internally.

## Tests

N/A


docs/resources/share.md
```diff
@@ -91,7 +91,7 @@ The following arguments are required:
 ### object Configuration Block
 
 * `name` - (Required) Full name of the object, e.g. `catalog.schema.name` for a tables, views, volumes and models, or `catalog.schema` for schemas.
-* `data_object_type` - (Required) Type of the data object, currently `TABLE`, `VIEW`, `SCHEMA`, `VOLUME`, and `MODEL` are supported.
+* `data_object_type` - (Required) Type of the data object. Supported types: `TABLE`, `FOREIGN_TABLE`, `SCHEMA`, `VIEW`, `MATERIALIZED_VIEW`, `STREAMING_TABLE`, `MODEL`, `NOTEBOOK_FILE`, `FUNCTION`, `FEATURE_SPEC`, and `VOLUME`.
```
alexott, line 94: `NOTEBOOK_FILE` could be problematic - we don't have API to put notebooks into the share

tanmay-db: I see, this is based on list internally. Let me add you to the thread.

alexott: I was wrong - we have it now: https://docs.databricks.com/api/workspace/shares/update#updates-data_object-content, but it's not documented in our TF docs (plus some other fields like `partitions`)

tanmay-db: I will also add the new fields as part of this PR.

tanmay-db: added

PR #5248, opened 2025-12-03
[Fix] Force new on `databricks_credential` name change

ashenm:
## Changes
Update `databricks_credential` attribute `name` to recreate resource on change;

Currently while changing credential name it marks the resource as "update in-place" instead of "recreate" which attempts and to update credential name and succeeds silently without changing.

## Tests

- [X] `make test` run locally
- [ ] using Go SDK


NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### New Features and Improvements
 
+* Add `force_new` constraints on `name` attribute in `databricks_credential` to match specification on documentation.
```
alexott: Please add a link to PR to changelog

alexott:
Also, please add a unit test like this: https://github.com/databricks/terraform-provider-databricks/blob/main/catalog/resource_schema_test.go#L342

PR #5263, opened 2025-12-06
Wait for app in DELETING state before create

varundeepsaini:
## Changes

Fixes: [#4006](https://github.com/databricks/cli/issues/4006)

Wait for app in DELETING state before creating a new app. When a DAB app is destroyed and immediately redeployed, the create fails with "An app with the same name already exists" because the app is still in DELETING state. This change adds a wait before create to handle this race condition.

I couldn't replicate the issue locally but have added unit tests to verify the behavior.

## Tests

- [x] `make test` run locally
- [x] using Go SDK


internal/providers/pluginfw/products/app/resource_app.go
```diff
@@ -211,6 +216,24 @@ func (a *resourceApp) waitForApp(ctx context.Context, w *databricks.WorkspaceCli
 	})
 }
 
+func waitForAppDeleted(ctx context.Context, w *databricks.WorkspaceClient, name string) error {
+	retrier := retries.New[struct{}](retries.WithTimeout(-1), retries.WithRetryFunc(shouldRetry))
```
alexott: I would still recommend to define a max time to wait for deletion...

internal/providers/pluginfw/products/app/resource_app.go
```diff
@@ -127,6 +127,11 @@ func (a *resourceApp) Create(ctx context.Context, req resource.CreateRequest, re
 		return
 	}
 
+	if err := waitForAppDeleted(ctx, w, appGoSdk.Name); err != nil {
```
alexott: Maybe it's better to wait in the Delete operation?

varundeepsaini:
@alexott have made the changes. Please review again 
@andrewnester can you also take a look

varundeepsaini:
@andrewnester bump ^^

hectorcast-db (changes requested):
.

Issue #5227, opened 2025-11-24
[FEATURE] Tag support for managing ABAC tag policy in databricks_policy_info


excavator-matt:
Update: There is a [resource for polices](https://github.com/databricks/terraform-provider-databricks/issues/5227#issuecomment-3590952832), but it doesn't support tags

It is a bit surprising, but I can't find any resource to manage an [ABAC tag policy](https://docs.databricks.com/aws/en/data-governance/unity-catalog/abac/policies) that to my understanding can be used for column masking. While it is a beta, this feature seems very useful and we're redoing data privacy in Databricks right now.

There is a [databricks_tag_policy](https://registry.terraform.io/providers/databricks/databricks/1.97.0/docs/resources/tag_policy), but is seems this is used to "manage governed tags". The name is a bit misleading.

### Use-cases

I want my data privacy policy completely version controlled.

### Attempted Solutions

I can create tags using databricks_tag_policy and assign tags on relevant columns in my transformation tool. The final step of applying a ABAC policy can only be clicked at the moment.

### Proposal

1. Rename databricks_tag_policy to databricks_governed_tag
2. [Create udf resource](https://github.com/databricks/terraform-provider-databricks/issues/4074)
3. Create a ABAC policy resource that combines the two as a policy

### References
https://docs.databricks.com/aws/en/data-governance/unity-catalog/filters-and-masks/

excavator-matt:
Regarding the function, it says in [this issue](https://github.com/databricks/terraform-provider-databricks/issues/4550) there needs to be an API update for creating a function.

Okay then. Can we still have the ABAC policy and hardcode the function? 

Also, it seems the Databricks CLI did the same mistake for namespace. This makes it a bit harder for the Terraform team to correct the course Databricks is currently following.

`databricks tag-policies list-tag-policies`

ctivanovich:
@excavator-matt have you explored this resource? 

https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/policy_info

Perhaps the name is too generic, making it difficult for people to recognize that it is for ABAC policies. I would suggest a name like data_object_policy or something a bit more specific, but broad enough to cover the fact that UC policies for masking/filtering don't necessarily require use of attributes, that is just one option for identifying targets for masking/filtering (the other being column name). 

I also dislike the whole "ABAC" marketing buzzword, as one tends to think of attributes as being more than just tags on objects, or DSL queries to check membership in a group or something, and it can get confusing for people. 


[...]

PR #5264, opened 2025-12-07
Document how the different tag resources relate

excavator-matt:
## Changes

Document how the different ([confusingly named](https://github.com/databricks/terraform-provider-databricks/issues/5227)) tag resources relate.

## Tests

N/A

- [ ] `make test` run locally
- [ ] using Go SDK


alexott:
Thank you for the PR, Matt! But this file is auto-generated as well, so it should be fixed in the API spec from which it's generated. I passed this PR to the dev team.

You can check [.gitattributes file](https://github.com/databricks/terraform-provider-databricks/blob/main/.gitattributes#L80) to see if a specific file is auto-generated or not.

PR #5284, opened 2025-12-17
Add support for default_workspace_id

Divyansh-db:
## Changes
Added support for default workspace ID in plugin framework and sdkv2 configure methods. A new defaultWorkspaceID field is a part of DatabricksClient.

## Tests
Unit test.

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/common/unified_provider_test.go
+++ b/common/unified_provider_test.go
@@ -360,7 +360,7 @@ func TestWorkspaceClientUnifiedProvider(t *testing.T) {
 				},
 			},
 			expectError:   true,
-			errorContains: "workspace_id is not set, please set the workspace_id in the provider_config",
+			errorContains: "workspace_id is not set in provider_config and default_workspace_id is not configured at provider level",
 			description:   "Account-level provider requires workspace_id to be set",
 		},
 	}


common/client.go
```diff
@@ -78,6 +78,11 @@ type DatabricksClient struct {
 	// configured for the provider
 	cachedAccountClient *databricks.AccountClient
 
+	// defaultWorkspaceID is the default workspace ID to use when workspace_id is not
+	// specified in provider_config at the resource level. This is set from the provider
+	// configuration and serves as a fallback for unified provider resources.
+	defaultWorkspaceID string
+
```
tanmay-db, line 85: Note: These changes would need to be done upstream as this file is generated.

internal/providers/sdkv2/sdkv2.go
```diff
@@ -321,6 +321,22 @@ func providerSchema() map[string]*schema.Schema {
...
+			if v == "" {
+				return
+			}
+			if !regexp.MustCompile(`^[1-9]\d*$`).MatchString(v) {
+				errs = append(errs, fmt.Errorf("default_workspace_id must be a positive integer without leading zeros"))
+			}
+			return
+		},
+		Description: "Default workspace ID to use when workspace_id is not specified in provider_config at the resource level",
+	}
```
tanmay-db, line 339: Do we need to add this separately or can this be part of ConfigAttributes like it's done for all the other attributes. See line above 316

Divyansh-db: ConfigAttributes come from config in Go SDK. default_workspace_id only makes sense in the context of terraform provider so in my opinion, it should not be the part of Go SDK config.

common/client.go
Divyansh-db: I will add it to the generated file template

chrisst: Follow up PR - can you add a header to all autogenerated files and then add claude.md that will parse that and know what to do? Make sure the claude.md doesn't leak internal details, draft it from the perspective of somebody using claude from outside databricks

NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### New Features and Improvements
 
+* Add support for `default_workspace_id` provider-level configuration ([#5284](https://github.com/databricks/terraform-provider-databricks/pull/5284)).
```
chrisst, line 9: This is not a great changelog comment as is. There are so many caveats about how and where this works (how your provider is configured, what the host is, which resource you're using, etc) that as a stand along single sentence this is confusing at best. TBH I would leave this out of the changelog entirely since it's still pretty 1/2 baked at this point or add more details like "experimental support for .... " or something like that.

PR #5309, opened 2026-01-12
[Internal] Add support for unified host - Migrate to HostType

tanmay-db:
## Changes

- Add support for unified host using experimental flag.
- IsAccountClient is deprecated and is migrated to HostType

## Tests

- Unit tests
- Manual E2E tests, terraform apply works when using unified host with experimental_is_unified_host. Integration test would be added when support for testing against unified hosts is added in test infra


common/client.go
```diff
@@ -111,7 +111,7 @@ func (c *DatabricksClient) GetWorkspaceClientForUnifiedProvider(
 	ctx context.Context, workspaceID string,
 ) (*databricks.WorkspaceClient, error) {
 	// The provider can be configured at account level or workspace level.
-	if c.Config.IsAccountClient() {
```
tanmay-db: Will create PR upstream once approved here.

hectorcast-db (commented):
Can you explain a bit more how it works? What happens when the flag is set? Does the host type change? 
I know the end goal, but I fail to understand the impact of this specific PR on users.

tanmay-db:
> Can you explain a bit more how it works? What happens when the flag is set? Does the host type change? I know the end goal, but I fail to understand the impact of this specific PR on users.

Yes, the impact of this specific PR is that users can now specify the spog (or unified) host in their provider or profile along with the flag. The flag is necessary because without flag users cannot specify the unified host.

If the flag is not set, the GO SDK panics with unified host. Ref: https://github.com/databricks/databricks-sdk-go/blob/main/config/config.go#L351. To support this we migrate to HostType methods which checks for the flag: https://github.com/databricks/databricks-sdk-go/blob/main/config/config.go#L373 which returns a UnifiedHost

The HostType logic after checking for the flag is same as IsAccountClient

hectorcast-db (approved):

PR #5312, opened 2026-01-13
Return empty string for `data_source_id` if data source API call fails

raymondguo-db:
## Changes

The [Databricks SQL data source API](https://docs.databricks.com/api/workspace/datasources) is deprecated and will be removed in the future. This API is used to populate the `data_source_id` field in `databricks_sql_warehouse` and `databricks_sql_endpoint`.

In order to prevent this from breaking the TF provider, adding fallback behavior to return empty string for `data_source_id` if the call to the data source API fails.

## Tests

Tests for `databricks_sql_warehouse` and `databricks_sql_endpoint`.

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/sql/data_sql_warehouse.go
+++ b/sql/data_sql_warehouse.go
@@ -27,7 +27,31 @@ func DataSourceWarehouse() common.Resource {
 		var dataSourceId string
 
 		dataSources, err := w.DataSources.List(ctx)
-		if err == nil {
+		if err != nil {
+			// DataSources.List failed; fall back to finding warehouse without DataSourceId
+			if data.Name != "" {
+				// Find warehouse by name using Warehouses.ListAll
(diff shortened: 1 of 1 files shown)


sql/data_sql_warehouse.go
```diff
@@ -22,38 +22,68 @@ func DataSourceWarehouse() common.Resource {
...
-		if err != nil {
-			return nil, err
-		}
-		for _, source := range dataSources {
-			if data.Name != "" && source.Name == data.Name {
-				selected = append(selected, source)
-			} else if data.Id != "" && source.WarehouseId == data.Id {
-				selected = append(selected, source)
-				break
+		if err == nil {
```
tanmay-db: Could you refactor to how it was before i.e. 
if err != nil { // logic when API fails} 
// happy path

raymondguo-db: Updated!

PR #5320, opened 2026-01-16
Support custom scopes

tejaskochar-db:
## Changes
Added support for slices to enable users to pass custom scopes in their configurations.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

---
NO_CHANGELOG=true

### DIFF
--- a/internal/providers/pluginfw/pluginfw.go
+++ b/internal/providers/pluginfw/pluginfw.go
@@ -167,7 +167,12 @@ func (p *DatabricksProviderPluginFramework) setAttribute(
 		elements := attrValue.Elements()
 		strSlice := make([]string, len(elements))
 		for i, elem := range elements {
-			strSlice[i] = elem.(types.String).ValueString()
+			strElem, ok := elem.(types.String)
+			if !ok {
+				diags.Append(diag.NewErrorDiagnostic(fmt.Sprintf("unexpected type for attribute %s", attr.Name), fmt.Sprintf("expected types.String, got %T", elem)))
+				return false, diags
+			}
+			strSlice[i] = strElem.ValueString()
 		}
 		err := attr.Set(cfg, strSlice)
 		if err != nil {


internal/providers/pluginfw/pluginfw.go
```diff
@@ -69,8 +69,6 @@ func (p *DatabricksProviderPluginFramework) Configure(ctx context.Context, req p
 	resp.ResourceData = client
 }
 
-// Function returns a schema.Schema based on config attributes where each attribute is mapped to the appropriate
-// schema type (BoolAttribute, StringAttribute, Int64Attribute).
```
tanmay-db, line 73: Why remove this comment instead of appending slice?

internal/providers/pluginfw/pluginfw.go
```diff
@@ -90,15 +88,20 @@ func providerSchemaPluginFramework() schema.Schema {
...
 		}
 	}
 	return schema.Schema{
 		Attributes: ps,
 	}
 }
 
-// setAttribute sets the attribute value in the SDK config corresponding to the attribute name in the provider configuration.
-// It returns true if the attribute was set, false if it was not set (because it was unknown or null), and a diag.Diagnostics object in case of error.
+// setAttribute returns true if the attribute was set, false if null/unknown.
```
tanmay-db: same here

internal/providers/pluginfw/pluginfw.go
```diff
@@ -149,6 +152,29 @@ func (p *DatabricksProviderPluginFramework) setAttribute(
...
+		var attrValue types.List
+		diags.Append(providerConfig.GetAttribute(ctx, path.Root(attr.Name), &attrValue)...)
+		if diags.HasError() {
+			return false, diags
+		}
+		if attrValue.IsNull() || attrValue.IsUnknown() {
+			return false, diags
+		}
+		elements := attrValue.Elements()
+		// Treat empty lists as unset for consistency with SDKv2's GetOk behavior.
```
tanmay-db: Why do we return false with 0 elements?

tejaskochar-db: Based on offline discussion, fixed it to distinguish between the two.

internal/providers/pluginfw/pluginfw.go
```diff
@@ -98,7 +104,6 @@ func providerSchemaPluginFramework() schema.Schema {
 }
 
 // setAttribute sets the attribute value in the SDK config corresponding to the attribute name in the provider configuration.
-// It returns true if the attribute was set, false if it was not set (because it was unknown or null), and a diag.Diagnostics object in case of error.
```
tanmay-db, line 101: why do we remove this?

internal/providers/providers_test.go
```diff
@@ -387,6 +387,57 @@ func testOAuthFetchesToken(t *testing.T, c *common.DatabricksClient) {
...
+			"DATABRICKS_HOST":  "x",
+			"DATABRICKS_TOKEN": "x",
+		},
+		assertAuth:   "pat",
+		assertHost:   "https://x",
+		assertScopes: nil,
+	}.apply(t)
+}
+
+func TestConfig_EmptyScopes(t *testing.T) {
```
tanmay-db, line 428: what happens if we use empty string as scope? scopes: []string{""}

internal/providers/sdkv2/sdkv2.go
```diff
@@ -334,6 +344,16 @@ func ConfigureDatabricksClient(ctx context.Context, d *schema.ResourceData, conf
 	attrsUsed := []string{}
 	for _, attr := range config.ConfigAttributes {
 		if value, ok := d.GetOk(attr.Name); ok {
+			// SDKv2's GetOk returns []interface{} for lists, but the SDK expects []string.
+			if attr.Kind == reflect.Slice {
+				if rawList, ok := value.([]interface{}); ok {
```
tanmay-db: if value.([]interface{}) fails, we directly will do attr.Set(cfg, value) below right? That would also fail?

internal/providers/providers_test.go
tejaskochar-db: That will be treated as a scope and result in an auth error down the line when it is used to mint a token

internal/providers/sdkv2/sdkv2.go
tejaskochar-db: Yes. Fixed it

PR #5326, opened 2026-01-16
[Fix] Only set 'enabled' when the field is updated for ncc endpoint rules

### LINKED ISSUE #5306: [ISSUE] with mws/resource_mws_ncc_private_endpoint_rule.go  enable resource update for Azure
### Problem description : 

Update an existing endpoint rule is not supported for Azure. 
When trying to update the domain list of the rule, it throws an error : 

> Error: cannot update mws ncc private endpoint rule: Updating fields enabled for Azure private endpoint rule 4e569a53-xxx-xxx-xxxx-39907e327bf2 is not supported. No updates were attempted. 

According the to code mws/resource_mws_ncc_private_endpoint_rule.go 
it seems that update is only supported on AWS. Currently on Azure, databricks starts to support App Gateway as [target](https://learn.microsoft.com/en-us/azure/databricks/security/network/serverless-network-security/serverless-private-link#-configure-private-link-to-azure-app-gateway-v2), the provider should allow update on the endpoint rules on Azure   

### Steps to Reproduce
1. changing the domain list in the databricks_mws_ncc_private_endpoint_rule resource  
2. run terraform apply 
 

1.100.0

bgrams:
## Changes
Only set the `enabled` request field when updating an NCC endpoint rule if it has been changed in the configuration. Including this for non-S3 endpoint rules results in a client error during apply.

Fixes #5306 

## Tests
Unit test added, existing unit tests updated, tested locally

- [x] `make test` run locally
- [x] using Go SDK


tanmay-db:
Hi @bgrams, thanks for the fix, by local testing you meant you are able to update the resource on non-aws cloud with local TF build?

bgrams:
@tanmay-db yes a local build, using standard AWS VPCE rules (i.e. `domain_names` in the resource). I don't have access to an environment to test more comprehensively than this. 

Here's the local config I'm testing with for reference:
```terraform
provider "databricks" {
  profile = "account"
}

resource "databricks_mws_network_connectivity_config" "config" {
  name   = "test"
  region = "us-west-2"
}

resource "databricks_mws_ncc_private_endpoint_rule" "rule" {
  endpoint_service = "com.amazonaws.vpce.us-west-2.vpce-svc-xxxxxxxxxxxxxxxx"
  domain_names     = ["foo.bar"]

  network_connectivity_config_id = databricks_mws_network_connectivity_config.config.network_connectivity_config_id
}

```

bgrams:
hi @tanmay-db - when might you have a moment to review?

tanmay-db:
Hi @bgrams, thanks for the reminder, @840 will be taking a look.

mws/resource_mws_ncc_private_endpoint_rule.go
```diff
@@ -24,7 +24,9 @@ func ResourceMwsNccPrivateEndpointRule() common.Resource {
 		}
 
 		common.CustomizeSchemaPath(m, "network_connectivity_config_id").SetRequired().SetForceNew()
-		common.CustomizeSchemaPath(m, "enabled").SetOptional().SetComputed()
+
+		// `enabled` can only be set for AWS S3 private endpoint services, identified by `resource_names`
+		common.CustomizeSchemaPath(m, "enabled").SetOptional().SetComputed().SetConflictsWith([]string{"domain_names", "group_id"})
```
tanmay-db: Hi @bgrams, this could be a breaking change for existing users if they have enabled set as part of their config even though they are using azure fields like domain_names, group_id.

Shall we remove this? What do you think?

bgrams: Yeah, that's up to you I think. You're right that it's not required for this fix. Lmk if you'd like me to remove it.

bgrams: Though I will say I think the use cases you're describing will all have failed at apply time, so this change was intended to trigger the failure earlier in the process.

Tl;dr yes it's a breaking change to the interface but I don't believe any production scenarios will be impacted.

Either way, I can remove this if you think that's most appropriate, and we can punt this discussion to a later date if it becomes necessary.

tanmay-db: I see, yeah, let's just remove it for now

bgrams: done in 732c5f95740d8ef4aadadfc996d836c174fa758c

tanmay-db:
Triggered Integration tests

tanmay-db:
Thanks @bgrams for the contribution.

tanmay-db (approved):

PR #5345, opened 2026-01-29
fix: ensure rfa_access_request_destinations are cleared on delete

micheledaddetta-databricks:
## Changes
This PR fixes a bug in the `databricks_rfa_access_request_destinations` resource: deleting the resource from Terraform configuration would leave the destinations intact in the Databricks Workspace UI.

## Tests

- [X] `make test` run locally
- [X] using Go SDK
- [X] using TF Plugin Framework


alexott:
@micheledaddetta-databricks thank you for the PR, but this resource is automatically generated from the API spec, and any change in the code will be overwritten. Please open ES ticket against the team that is responsible for this functionality

PR #5349, opened 2026-02-01
[Fix] databricks apps resource fail when deleted outside terraform

### LINKED ISSUE #5308: [ISSUE] Issue with `databricks_users` resource - `extra_attributes` not working
### Configuration

```hcl
data "databricks_users" "all" {
  extra_attributes = "active"
}
```

### Expected Behavior
The `active` field should be included for all users in `data.databricks_users.users`, and it should be a boolean.

### Actual Behavior
The attribute `active` is always `null`.

 TF 1.13.3, provider 1.102.0

### Is it a regression?
?

### Important Factoids
It's working when calling the rest api directly like
```
curl --location 'https://accounts.azuredatabricks.net/api/2.0/accounts/xxx/scim/v2/Users?attributes=id%2CuserName%2Cactive%2CdisplayName%2CexternalId' 
```

jlieow:
## Changes

Fixed two bugs in the databricks_users data source where the extra_attributes parameter was not working correctly:

1. Account-level: extra_attributes used .String() instead of .ValueString(), which wrapped single-attribute values in extra quotes (e.g., "active" became "\"active\""), causing the SCIM API to silently ignore them.
2. Workspace-level: extra_attributes was not being forwarded to the SCIM API at all. The Attributes field was missing from the ListUsersRequest.

## Tests

New acceptance tests:                                                                                                                                        
- TestAccDataSourceUsers_SingleExtraAttribute (account level)                                                                                                
- TestWorkspaceDataSourceUsers_SingleExtraAttribute (workspace level)  

- [x] `make test` run locally                                                                                                                                        
- [x] using Go SDK                                                                                                                                                   

Docs don't need updating. The extra_attributes argument is already documented and the bug fix doesn't change any behaviour from the user's perspective. It just makes the existing documented behaviour work correctly.     

Fixes #5308


jlieow:
Realised I mixed up the branches. Closing this and will link the correct the PR.

PR #5366, opened 2026-02-09
[Fix] databricks_sql_table times out without adding resource to state

### LINKED ISSUE #5340: [FEATURE] Add timeouts block support for databricks_sql_table resource
### Use-cases                                                                                                                                                                              
                                                                                                                                                                                             
  Creating external tables pointing to S3 locations with many Parquet files fails with `context deadline exceeded` after exactly 50 seconds. The resource has `MaxSqlExecWaitTimeout = 50`   
  hardcoded and doesn't support the `timeouts` block.                                                                                                                                        
                                                                                                                                                                                             
  ### Attempted Solutions                                                                                                                                                                    
                                                                                                                                                                                             
  ```hcl                                                                                                                                                                                     
  resource "databricks_sql_table" "results" {                                                                                                                                                
    warehouse_id       = databricks_sql_endpoint.this.id                                                                                                                                     
    storage_location   = "s3://bucket/path"                                                                                                                                                  
    data_source_format = "PARQUET"                                                                                                                                                           
                                                                                                                                                                                             
    timeouts {                                                                                                                                                                               
      create = "20m"                                                                                                                                                                         
    }                                                                                                                                                                                        
  }                                                                                                                                                                                          
  ```                                                                                                                                                                                        
                                                                                                                                                                                             
  Result: `Blocks of type "timeouts" are not expected here.`                                                                                                                                 
                                                                                                                                                                                             
  Debug log:                                                                                                                                                                                 
  ```                                                                                                                                                                                        
  module.databricks_tf_module.databricks_sql_table.results-table: Still creating... [40s elapsed]                                                                                            
  module.databricks_tf_module.databricks_sql_table.results-table: Still creating... [50s elapsed]                                                                                            
  ╷                                                                                                                                                                                          
  │ Error: cannot create sql table: Post "https://dbc-xxxxx.cloud.databricks.com/api/2.0/sql/statements": context deadline exceeded                                                          
  │                                                                                                                                                                                          
  │   with module.databricks_tf_module.databricks_sql_table.results-table,                                                                                                                   
  │    1: resource "databricks_sql_table" "results-table" {                                                                                                                                  
  ```                                                                                                                                                                                        
                                                                                                                                                                                             
  ### Proposal                                                                                                                                                                               
                                                                                                                                                                                             
  Add `timeouts` block support for `databricks_sql_table` resource.

jlieow:
## Changes
The `databricks_sql_table` resource times out with custom `warehouse_id` when SQL tables take longer than 50 seconds to create and Terraform does not record it in state. On the next `terraform apply`, Terraform tries to create the table again but fails because it already exists, leaving the user stuck and having to import the resource into the state.                                                

This issue is not present when the `warehouse_id` is excluded. When you provide a `warehouse_id`, it routes to the SQL Statement Execution API which uses a SQL Warehouse and has a timeout limit of 50s. When you exclude it, it is sent to a completely different Command Execution API which uses a Cluster and does not have any timeout upper limit, resulting in a different user experience.

Ideally the experience should be the same regardless of the chosen compute. However, the SQL Statement Execution API has a `wait_timeout` parameter that controls how long the API waits before responding, with a maximum of 50 seconds. It also has an `on_wait_timeout` parameter with two options:
- **Cancel**: The statement is aborted after the timeout. However, the table may still end up being created on the Databricks side. Since Terraform saw an error, it never records the table in its state file.
- **Continue**: The statement keeps running after the timeout, but the API returns without a final result. The table is eventually created on the Databricks side, but Terraform still has no success response to record in state.

Users requested adding a Terraform `timeouts` block. However, this does not fully solve the issue either. If a user sets a timeout that is too short for their table to be created, the same problem occurs. The provider gives up waiting, never records the table in state, but the table may still be created on the Databricks side. This puts the burden on the user to guess the right timeout value, and getting it wrong leads to the same stuck state.

The most reliable fix is to use `Continue` instead of `Cancel` (which keeps the statement running after the timeout) and then poll a separate API (`GetStatementByStatementId`) internally to check the statement's progress until it completes. By always waiting for a definitive result (either success or failure), the provider ensures the table is recorded in state when it is created, and future `terraform apply` runs work seamlessly without manual intervention. This also aligns the user experience for both compute types and means no changes to the resource block structure are required.

The `Continue` behavior is applied for all users rather than making it optional because providing a choice would risk users leaving the default as `Cancel` or choosing `Cancel` explicitly, which would still result in tables not being added to state when they take longer than 50 seconds to create. 
  
## Tests
Added `TestResourceSqlTableCreateTable_ExistingSQLWarehouse_Polling` to verify polling behavior when statements take longer than the initial 50-second wait period.

- [x] `make test` run locally
- [x] covered with integration tests in `catalog/resource_sql_table_test.go`
- [x] using Go SDK

No changes to the terraform block is made, just an alignment of user experience, so no update to docs is required. 

The changes are made to SDKv2 as this update is too big to also include a migration to the Plugin Framework.

Fixes #5340


venkad-intc:
Hi,

We see this issue even when the warehouse id is excluded.

Do we have any ETA on this fix?

ShaneMazur:
Hello - also waiting on this fix. Seems that the `ALTER TABLE ... CLUSTER BY AUTO` statements can regularly take more than 50 seconds for any table of realistic size. Anything still blocking this?

jlieow:
> Hi,
> 
> We see this issue even when the warehouse id is excluded.
> 
> Do we have any ETA on this fix?

Hi, could you share the full error message you're seeing? Specifically, does it mention context deadline exceeded?

This PR targets the 50s timeout in the SQL Statement Execution API (the warehouse_id path). Without warehouse_id, the resource should use the Command Execution API instead which does not have that 50s limit, but Terraform has a default 20 minute context timeout limit. If your cluster needs a cold start, that could eat into the budget significantly.

ShaneMazur:
@venkad-intc in case your notifications are only on mentions ⬆️

Divyansh-db:
@alexott do you have any thoughts on this PR. It introduces a client side polling to overcome the 50s timeout from the server.

aderkach-mirai:
Any updates?

PR #5373, opened 2026-02-10
[Fix] Do not delete scope manage ACL when principal is deployer

shreyas-goenka:
## Summary
Prevents users from accidentally locking themselves out of managing a secret scope by skipping ACL deletion when the current user is the principal with MANAGE permission.

Note: The ACLs are cleaned up when the associated scope is deleted.

## Test Plan
- Added unit tests for all delete scenarios
- Manually verified that the changes work as expected, and that ACLs are always cleaned up when the scope is deleted.


secrets/resource_secret_acl.go
```diff
@@ -85,6 +85,18 @@ func ResourceSecretACL() common.Resource {
 			if err != nil {
 				return err
 			}
+			// Skip deleting the ACL if the current user is the principal with MANAGE permission.
+			// This prevents users from accidentally locking themselves out of managing the secret scope.
+			permission := d.Get("permission").(string)
+			if permission == string(workspace.AclPermissionManage) {
+				me, err := w.CurrentUser.Me(ctx)
```
alexott, line 92: We have cached CurrentUser in the client, no need to call API explicitly: https://github.com/databricks/terraform-provider-databricks/blob/main/common/client.go#L25

shreyas-goenka: It should already be using the cached value. The client from `c.WorkspaceClientUnifiedProvider(ctx, d)` seems to be using the c.WorkspaceClient() underneath the hood. 

Am I missing something?

alexott: ah, sorry - missed that we override `CurrentUser`

alexott (approved):

PR #5376, opened 2026-02-10
[Internal] Validate unified provider at plan phase for `workspace_id`

tanmay-db:
## Changes

Add support for `workspace_id` mismatch during plan phase (`CustomizeDiff`) so that mismatches are caught earlier without needing to apply.

## Tests

- Integration test
- Acceptance tests updated with `PlanOnly: true` to verify plan-phase validation

NO_CHANGELOG=true

### DIFF
--- a/common/unified_provider.go
+++ b/common/unified_provider.go
@@ -86,23 +86,25 @@ func NamespaceCustomizeSchemaMap(m map[string]*schema.Schema) map[string]*schema
 
 // namespaceForceNew marks the workspace_id field as ForceNew if it changed.
 func namespaceForceNew(d *schema.ResourceDiff) error {
-	workspaceIDKey := workspaceIDSchemaKey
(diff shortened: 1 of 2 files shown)


common/unified_provider.go
```diff
@@ -95,6 +96,26 @@ func NamespaceCustomizeDiff(ctx context.Context, d *schema.ResourceDiff, c *Data
...
+	}
+	workspaceIDInt, err := parseWorkspaceID(newWSID)
+	if err != nil {
+		return err
+	}
+	w, err := c.WorkspaceClient()
+	if err != nil {
+		return err
+	}
+	err = c.validateWorkspaceIDFromProvider(ctx, workspaceIDInt, w)
```
hectorcast-db: It makes total sense for TF, but I just noticed that this means that people are not able to change the default workspace in their Profile.

PR #5377, opened 2026-02-10
[Feature] Support `git_repository` configuration on `databricks_app` resource

jlieow:
## Changes

This change enables the `databricks_app` resource block to support `git_repository` configuration. This involved reworking the `Update` method to use the async `CreateUpdate` API instead of the synchronous `Update` API, because the synchronous API does not support `git_repository` changes or fields that require async processing.

The async `CreateUpdate` API has a few specific design requirements:

1. **UpdateMask**: The `CreateUpdate` API uses an `update_mask` parameter to specify which fields are being changed. Based on the [Databricks Apps API docs](https://databricks-sdk-py.readthedocs.io/en/latest/workspace/apps/apps.html), it is recommended to always explicitly list the fields being updated and avoid full field replacement, as it can lead to unintended results if the API changes in the future. The implementation compares plan and state to build a comma-separated list of only the changed field names (e.g. `"description,git_repository"`).

2. **Re-fetching app state**: Since the API is async, the immediate response from the waiter is an update status object (`apps.AppUpdate`), not the full app. After the update completes, we must call `GetByName` to retrieve the actual server-side app state so Terraform's state accurately reflects reality.

3. **Timeout**: As noted in (2), this is an async API that requires polling until the update completes, so a timeout is required. A 20-minute timeout is used when waiting for the async update to complete. This aligns with the established convention throughout the codebase:
   - Clusters (`clusters/clusters_api_sdk.go`): `20*time.Minute` for `WaitGetClusterRunning` and `WaitGetClusterTerminated`
   - Pipelines (`pipelines/resource_pipeline.go`): `const DefaultTimeout = 20 * time.Minute`
   - MWS Workspaces (`mws/resource_mws_workspaces.go`): `const DefaultProvisionTimeout = 20 * time.Minute`
   - Storage Mounts (`storage/resource_mount.go`): `schema.DefaultTimeout(20 * time.Minute)`

---

### New imports

```go
"strings"
"time"
```

- `strings` is needed for `strings.Join` when building the comma-separated update mask.
- `time` is needed to specify the 20-minute timeout when waiting for the async update to complete.

---

### Read both plan and state (lines 263-273)

**Before:**
```go
var app AppResource
resp.Diagnostics.Append(req.Plan.Get(ctx, &app)...)
```

**After:**
```go
var plan AppResource
resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)

var state AppResource
resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
```

**Why:** The previous code only read the desired plan. Now we also read the current state so we can compare the two and determine exactly which fields changed. The variable was renamed from `app` to `plan` for clarity alongside the new `state` variable.

---

### Build an update mask from changed fields (lines 294-300)

```go
updateMask := a.buildUpdateMask(ctx, plan, state)
if updateMask == "" {
    resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
    return
}
```

**Why:** Instead of sending the entire object to the API, we compute a mask of only the fields that differ between plan and state (e.g. `"description,git_repository"`). If nothing changed, we skip the API call entirely.

---

### Switch from `Update` to `CreateUpdate` (lines 302-311)

**Before:**
```go
response, err := w.Apps.Update(ctx, apps.UpdateAppRequest{
    App:  appGoSdk,
    Name: app.Name.ValueString(),
})
```

**After:**
```go
waiter, err := w.Apps.CreateUpdate(ctx, apps.AsyncUpdateAppRequest{
    AppName:    appGoSdk.Name,
    App:        &appGoSdk,
    UpdateMask: updateMask,
})
```

**Why:** The synchronous `Update` API does not support `git_repository` changes. The async `CreateUpdate` API accepts an `UpdateMask` to specify which fields to update and returns a waiter for polling completion.

---

### Wait for async completion and check status (lines 313-324)

```go
update, err := waiter.GetWithTimeout(20 * time.Minute)
if err != nil {
    resp.Diagnostics.AddError("error waiting for app update to complete", err.Error())
    return
}

if update.Status.State == apps.AppUpdateUpdateStatusUpdateStateFailed {
    resp.Diagnostics.AddError("app update failed", update.Status.Message)
    return
}
```

**Why:** Since `CreateUpdate` is async, we must poll until the update finishes. After completion, we check whether the update status is failed and surface the error message to the user.

---

### Re-fetch app state after update (lines 326-331)

**Before:**
```go
var newApp AppResource
resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, response, &newApp)...)
```

**After:**
```go
finalApp, err := w.Apps.GetByName(ctx, appGoSdk.Name)
if err != nil {
    resp.Diagnostics.AddError("failed to read updated app", err.Error())
    return
}

var newApp AppResource
resp.Diagnostics.Append(converters.GoSdkToTfSdkStruct(ctx, finalApp, &newApp)...)
```

**Why:** The `CreateUpdate` waiter returns an update status object, not the full app. We call `GetByName` to fetch the actual server-side app state so that Terraform state reflects the real values.

---

### Variable rename in state assignment (lines 340-341)

**Before:**
```go
newApp.NoCompute = app.NoCompute
newApp.ProviderConfig = app.ProviderConfig
```

**After:**
```go
newApp.NoCompute = plan.NoCompute
newApp.ProviderConfig = plan.ProviderConfig
```

**Why:** Pure rename from `app` to `plan` to match the new variable naming. No behavioral change.

---

### New `buildUpdateMask` method (lines 348-384)

```go
func (a *resourceApp) buildUpdateMask(ctx context.Context, plan, state AppResource) string {
    var changedFields []string

    if !plan.Description.Equal(state.Description) {
        changedFields = append(changedFields, "description")
    }
    if !plan.GitRepository.Equal(state.GitRepository) {
        changedFields = append(changedFields, "git_repository")
    }
    if !plan.Resources.Equal(state.Resources) {
        changedFields = append(changedFields, "resources")
    }
    if !plan.BudgetPolicyId.Equal(state.BudgetPolicyId) {
        changedFields = append(changedFields, "budget_policy_id")
    }
    if !plan.ComputeSize.Equal(state.ComputeSize) {
        changedFields = append(changedFields, "compute_size")
    }
    if !plan.UsagePolicyId.Equal(state.UsagePolicyId) {
        changedFields = append(changedFields, "usage_policy_id")
    }
    if !plan.UserApiScopes.Equal(state.UserApiScopes) {
        changedFields = append(changedFields, "user_api_scopes")
    }

    return strings.Join(changedFields, ",")
}
```

**Why:** Compares each updatable field between plan and state using Terraform's `.Equal()` method. Returns a comma-separated string of only the changed field names, which is passed as the `UpdateMask` to the `CreateUpdate` API.

### Integration Tests (`resource_app_acc_test.go`)

**TestAccAppResource_GitRepository_Update** — Tests adding and modifying `git_repository` configuration:
1. Create app without `git_repository`
2. Add `git_repository` pointing to `https://github.com/databricks/databricks-sdk-py`
3. Update `git_repository` URL to `https://github.com/databricks/databricks-sdk-go`
- Verifies `update_time` changes after each modification
- Verifies `git_repository.0.url` attribute is correctly set

### Unit Tests (`resource_app_test.go`)

These tests verify that `buildUpdateMask` correctly compares plan vs state and produces a comma-separated mask of only the fields that differ:

- **TestBuildUpdateMask_GitRepository**: State has no `git_repository` and description `"Original description"`. Plan adds a `git_repository` object (`gitHub`, `https://github.com/user/repo`) and changes description to `"Updated description"`. Asserts the mask contains both `"description"` and `"git_repository"`.
- **TestBuildUpdateMask_MultipleFields**: State has `description="Original description"` and `compute_size="SMALL"`. Plan changes description, compute_size to `"MEDIUM"`, and adds `budget_policy_id` and `usage_policy_id`. Asserts all four fields appear in the mask.
- **TestBuildUpdateMask_NoChanges**: State and plan are identical (all nullable fields explicitly set to null, same description). Asserts the mask is an empty string, meaning no API call would be made.
- **TestBuildUpdateMask_DescriptionOnly**: State and plan are identical except description changes from `"Old description"` to `"New description"`. Asserts the mask is exactly `"description"` — confirming unchanged fields are excluded.
- **TestBuildUpdateMask_Resources**: State has `resources` as null. Plan has `resources` as an empty list (non-null). Asserts the mask contains `"resources"` — verifying that null-to-empty-list is detected as a change.
- **TestBuildUpdateMask_UserApiScopes**: State has `user_api_scopes` as null. Plan sets it to `["read", "write"]`. Asserts the mask contains `"user_api_scopes"` — verifying list-type field changes are detected.

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,7 @@
 * Add `role_arn` field to `databricks_mws_storage_configurations` resource to support sharing S3 buckets between root storage and Unity Catalog ([#5222](https://github.com/databricks/terraform-provider-databricks/issues/5222))
 
 ### Bug Fixes
+* Fix the `databricks_mws_ncc_private_endpoint_rule` resource to support updates for non-S3 endpoint rules ([#5326](https://github.com/databricks/terraform-provider-databricks/pull/5326))
 
 * Fixed `databricks_users` data source `extra_attributes` parameter issues ([#5308](https://github.com/databricks/terraform-provider-databricks/issues/5308)): (1) Single-attribute inputs (e.g., `extra_attributes = "active"`) were silently ignored at account level due to incorrect value quoting. (2) Complex multi-valued attributes like `emails` and `roles` returned null at account level even when explicitly requested in `extra_attributes`. (3) `extra_attributes` were not forwarded to the SCIM API at workspace level.
 * Fix `databricks_app` resource fail to read app when deleted outside terraform ([#5365](https://github.com/databricks/terraform-provider-databricks/pull/5365))
--- a/mws/resource_mws_ncc_private_endpoint_rule.go
+++ b/mws/resource_mws_ncc_private_endpoint_rule.go
@@ -87,11 +87,13 @@ func ResourceMwsNccPrivateEndpointRule() common.Resource {
 			// only enabled, domain names & resource names are updatable
 			// they do require update_mask to be set
 			// resource_names are not applicable to Azure, so we exclude them from the update
-			updateMask := []string{"enabled"}
-			updatePrivateEndpointRule := settings.UpdatePrivateEndpointRule{
-				Enabled: d.Get("enabled").(bool),
-			}
+			updateMask := []string{}
+			updatePrivateEndpointRule := settings.UpdatePrivateEndpointRule{}
 
+			if d.HasChange("enabled") {
+				updateMask = append(updateMask, "enabled")
+				updatePrivateEndpointRule.Enabled = d.Get("enabled").(bool)
+			}
 			if d.HasChange("domain_names") {
 				updateMask = append(updateMask, "domain_names")
 				newDomainNames := []string{}
--- a/mws/resource_mws_ncc_private_endpoint_rule_test.go
+++ b/mws/resource_mws_ncc_private_endpoint_rule_test.go
@@ -120,7 +120,7 @@ func TestResourceNccPrivateEndpointRulePrivateEndpointRuleUpdateDomainName(t *te
 				PrivateEndpointRule: settings.UpdatePrivateEndpointRule{
 					DomainNames: []string{"my-new-example.exampledomain.com", "my-new-example2.exampledomain.com"},
 				},
-				UpdateMask: "enabled,domain_names",
+				UpdateMask: "domain_names",
 			}).Return(getTestNccRule(), nil)
 			e.GetPrivateEndpointRuleByNetworkConnectivityConfigIdAndPrivateEndpointRuleId(mock.Anything, "ncc_id", "rule_id").Return(
 				&settings.NccPrivateEndpointRule{
@@ -167,7 +167,7 @@ func TestResourceNccPrivateEndpointRulePrivateEndpointRuleUpdateResourceName(t *
 				PrivateEndpointRule: settings.UpdatePrivateEndpointRule{
 					ResourceNames: []string{"bucket1", "bucket2"},
 				},
-				UpdateMask: "enabled,resource_names",
+				UpdateMask: "resource_names",
 			}).Return(getTestNccRule(), nil)
 			e.GetPrivateEndpointRuleByNetworkConnectivityConfigIdAndPrivateEndpointRuleId(mock.Anything, "ncc_id", "rule_id").Return(
 				&settings.NccPrivateEndpointRule{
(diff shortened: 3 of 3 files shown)


jlieow:
PR was incorrectly opened from main. Will reopen from the correct feature branch.

PR #5383, opened 2026-02-12
Fix databricks_cluster to send gcp_attributes.local_ssd_count=0 durin…

stav-cohen12:
…g updates

When updating a cluster with gcp_attributes.local_ssd_count = 0, the value was not being sent to the Databricks API, causing the cluster to fall back to "Default" behavior (at least 1 local SSD). This fix ensures that zero values are explicitly sent during updates, matching the behavior already present for cluster creation (added in PR #3631).

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


tanmay-db (commented):

PR #5391, opened 2026-02-13
[Fix] Permanent permissions drift when `user_name` casing in `databricks_permissions` `access_control` blocks differs from the API response

### LINKED ISSUE #5183: [ISSUE] Issue with `databricks_permissions` resource when `user_name` is used for access_control. There is a permanent drift with each terraform plan and apply
### Configuration

Terraform 1.9.0.

```hcl
    databricks = {
      version = "= 1.96.0"
    }
```

### Expected Behavior

With consecutive terraform plan runs, I would not expect a configuration drift. Terraform apply will now remove and apply the same databricks permissios.

### Actual Behavior

The following resources were working fine. No configuration drift. Terraform would say "no infrastructure change'.

```hcl
resource "databricks_permissions" "compass_sql_endpoint_permissions" {
  sql_endpoint_id = databricks_sql_endpoint.compass_sql_endpoint_x_small.id

  dynamic "access_control" {
    for_each = var.databricks_groups_assign_to_sql_endpoint
    content {
      group_name       = access_control.key
      permission_level = access_control.value
    }
  }
  dynamic "access_control" {
    for_each = var.databricks_service_principals_assign_to_sql_endpoint
    content {
      service_principal_name = access_control.key
      permission_level       = access_control.value
    }
  }
  provider = databricks.databricks_ws
}
```
When I added the user_name, the resource block became as the following

```hcl
resource "databricks_permissions" "compass_sql_endpoint_permissions" {
  sql_endpoint_id = databricks_sql_endpoint.compass_sql_endpoint_x_small.id

  dynamic "access_control" {
    for_each = var.databricks_groups_assign_to_sql_endpoint
    content {
      group_name       = access_control.key
      permission_level = access_control.value
    }
  }

  dynamic "access_control" {
    for_each = var.databricks_users_assign_to_sql_endpoint
    content {
      user_name        = access_control.key
      permission_level = access_control.value
    }
  }

  dynamic "access_control" {
    for_each = var.databricks_service_principals_assign_to_sql_endpoint
    content {
      service_principal_name = access_control.key
      permission_level       = access_control.value
    }
  }
  provider = databricks.databricks_ws
}
```

Now with each terraform plan, I see

### Steps to Reproduce

Use user_name in a databricks_permission.

```hcl
  dynamic "access_control" {
    for_each = var.databricks_users_assign_to_sql_endpoint
    content {
      user_name        = access_control.key
      permission_level = access_control.value
    }
  }
```

The variables are shown below:

```hcl
databricks_users_assign_to_sql_endpoint = {
  "svcDatabricksD2PowerbiP@sadflksadfkas.com" : "CAN_MANAGE"
}
databricks_groups_assign_to_sql_endpoint = {
  "AadDptITDataPlatformEngineers" : "CAN_MANAGE",
  "AadDptITAnalyticsEngineers" : "CAN_MONITOR",
  "AadDptITReportingEngineers" : "CAN_MONITOR"
}
```

Terraform version 1.9.0.

### Debug Output

If this is enough to reproduce the issue

Image
Image

, I would not like to share it because of sensitive information with subscription_id etc. etc.

### LINKED ISSUE #5712: [ISSUE] databricks_permissions with instance_pool_id shows permanent drift when deployer is a workspace admin
### Configuration

```hcl
resource "databricks_instance_pool" "example" {
  instance_pool_name                    = "example-pool"
  min_idle_instances                    = 0
  node_type_id                          = "Standard_D4ds_v5"
  idle_instance_autotermination_minutes = 10
}

resource "databricks_permissions" "pool_usage" {
  instance_pool_id = databricks_instance_pool.example.id

  access_control {
    group_name       = "DataEngineering"
    permission_level = "CAN_ATTACH_TO"
  }
}
```

The deploying identity (user or service principal) is a **workspace admin**.

### Expected Behavior

After a successful `terraform apply`, a subsequent `terraform plan` should show no changes.

### Actual Behavior

Every `terraform plan` shows the `access_control` blocks being destroyed and recreated. The Permissions API (`GET /api/2.0/permissions/instance-pools/<id>`) returns the deploying admin's inherited `CAN_MANAGE` in the `access_control_list` response. Since this entry is not in the Terraform config, the provider sees a diff on every plan and produces a destroy/recreate cycle that never converges.

The plan output looks like:

```
# databricks_permissions.pool_usage will be updated in-place
~ resource "databricks_permissions" "pool_usage" {
    id = "/instance-pools/xxx"

  - access_control {
      - permission_level = "CAN_MANAGE" -> null
      - user_name        = "deployer@domain.com" -> null
    }
  - access_control {
      - permission_level = "CAN_ATTACH_TO" -> null
      - group_name       = "DataEngineering" -> null
    }
  + access_control {
      + permission_level = "CAN_ATTACH_TO"
      + group_name       = "DataEngineering"
    }
}
```

### Steps to Reproduce

1. Authenticate Terraform as a workspace admin (user or service principal)
2. `terraform apply` — creates the pool and permissions successfully
3. `terraform plan` — shows drift; `access_control` blocks will be destroyed and recreated
4. `terraform apply` — re-applies the same permissions
5. `terraform plan` — drift appears again. Repeat forever.

```
Terraform v1.x
databricks/databricks (latest)
Cloud: Azure
```

### Is it a regression?

Unknown. This may have existed since `databricks_permissions` was introduced for instance pools.

### Debug Output

The relevant diff is visible in a standard `terraform plan`. The root cause is in the Read function: the API response includes `access_control_list` entries where `inherited == true` and `inherited_from_object` contains `/admins`. The provider does not filter these out before comparing to state.

### Important Factoids

**All known workarounds fail for this specific scenario:**

| Workaround | Why it fails |
|---|---|
| Explicitly declare deployer's `CAN_MANAGE` in config | Provider auto-strips admin `access_control` blocks: *"It is not possible to lower permissions for admins, so Databricks Terraform Provider removes those access_control blocks automatically"* ([docs](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/permissions)) |
| Use a non-admin SP as deployer | `allow_instance_pool_create` entitlement can only be granted to **workspace-local groups** (legacy), not account groups. Databricks recommends migrating away from workspace-local groups, making this a non-starter for orgs using account-level identity federation. |
| Transfer pool ownership via API | No `change-owner` API exists for instance pools. Only clusters have `/api/2.1/clusters/change-owner`. |
| Use `databricks_access_control_rule_set` instead | Docs state: *"cannot be used to manage access rules for resources supported by databricks_permissions"* |
| `lifecycle { ignore_changes = [access_control] }` | Suppresses all permission drift including real changes — not acceptable for governed environments |

**Related issues:**

- #4734 — Same drift pattern for `cluster_id` (open, no assignee)
- #3730 — Same drift pattern for `sql_endpoint_id` (closed/fixed). The fix added automatic `IS_OWNER` handling, but instance pools do not have an `IS_OWNER` permission level, so the same fix cannot be directly applied.
- #2543 — Feature request to change cluster owner via provider (open). Does not cover instance pools.

**Proposed fix:** During the Read/refresh of `databricks_permissions` for `instance_pool_id`, filter out `access_control_list` entries where `inherited == true` and `inherited_from_object` contains `/admins` before writing to state. This prevents inherited admin permissions from appearing in the diff. This is analogous to the #3730 fix, adapted for the fact that pools lack `IS_OWNER`.

Not at this time, but the fix path is well-defined: modify the Read function in the `databricks_permissions` resource to filter inherited admin entries from the API response before comparing to state. The pattern already exists in the #3730 fix for SQL warehouses.

jlieow:
Closes #5183, Fixes #5712

## Changes

Fixes permanent permissions drift when `user_name` casing in `databricks_permissions` `access_control` blocks differs from the API response.

**Problem:** The Databricks API normalizes user emails to lowercase (e.g. config has `User@Example.com`, API returns `user@example.com`). Because emails are case-insensitive, both refer to the same user, but Terraform treats them as different entries, causing a remove/re-add diff on every plan that never converges.

**Fix:** Normalize `user_name` and `service_principal_name` to lowercase throughout the permissions resource so that casing differences between config and API are ignored. The fix touches three areas:

1. **Hash function**: Lowercase `user_name` and `service_principal_name` before hashing so config and API values produce the same hash.
2. **API response handling**: Lowercase `user_name` and `service_principal_name` when reading the API response into state.
3. **Current-user comparison**: Use case-insensitive comparison (`strings.EqualFold`) when checking if the current user is in the config. Without this, the provider silently dropped the current user's permissions from state when their email casing differed, a second source of drift.

---

### Hash function case normalization (`permissions/resource_permissions.go`, lines 191-217)

**Before:**
```go
if strVal, ok := val.(string); ok {
    if strVal != "" {
        normalized[key] = strVal
    }
}
```

**After:**
```go
if strVal, ok := val.(string); ok {
    if strVal != "" {
        if key == "user_name" || key == "service_principal_name" {
            strVal = strings.ToLower(strVal)
        }
        normalized[key] = strVal
    }
}
```

**Why:** The `access_control` block uses `schema.TypeSet`, which relies on a hash function to determine element identity. If the user's config has `User@Example.com` and the API returns `user@example.com`, they produced different hashes, so Terraform saw them as two different set elements: one to remove and one to add. Lowercasing before hashing ensures both values produce the same hash.

---

### API response lowercasing (`permissions/permission_definitions.go`, lines 247-248)

**Before:**
```go
entity.AccessControlList = append(entity.AccessControlList, iam.AccessControlRequest{
    GroupName:            accessControl.GroupName,
    UserName:             accessControl.UserName,
    ServicePrincipalName: accessControl.ServicePrincipalName,
    PermissionLevel:      permission.PermissionLevel,
})
```

**After:**
```go
entity.AccessControlList = append(entity.AccessControlList, iam.AccessControlRequest{
    GroupName:            accessControl.GroupName,
    UserName:             strings.ToLower(accessControl.UserName),
    ServicePrincipalName: strings.ToLower(accessControl.ServicePrincipalName),
    PermissionLevel:      permission.PermissionLevel,
})
```

**Why:** Even with matching hashes, if the stored string value in state differs from the config value, Terraform detects a diff. Lowercasing the API response ensures the state value matches a lowercase config value.

---

### Case-insensitive current-user comparison (`permissions/permission_definitions.go`, line 231)

**Before:**
```go
if me == accessControl.UserName || me == accessControl.ServicePrincipalName {
```

**After:**
```go
if strings.EqualFold(me, accessControl.UserName) || strings.EqualFold(me, accessControl.ServicePrincipalName) {
```

**Why:** `==` is case-sensitive, so `"Admin@Example.com" == "admin@example.com"` returns `false`. `strings.EqualFold` compares strings case-insensitively, correctly matching the same user regardless of casing.

---

### Case-insensitive lookup in `ContainsUserOrServicePrincipal` (`permissions/entity/permissions_entity.go`, line 17)

**Before:**
```go
if ac.UserName == name || ac.ServicePrincipalName == name {
```

**After:**
```go
if strings.EqualFold(ac.UserName, name) || strings.EqualFold(ac.ServicePrincipalName, name) {
```

**Why:** Same as above. `==` is case-sensitive, so `"Admin@Example.com" == "admin@example.com"` returns `false`. `strings.EqualFold` compares strings case-insensitively, correctly matching the same user regardless of casing.

---

## Tests

- **TestPermissionsDrift_UserNameNoDriftWithSameCasing**: Simulates an existing resource in state with `user_name="user@example.com"`, then performs a Read where the API returns the same casing. Constructs a full `InstanceState` with computed set hashes and verifies that the Read produces exactly 2 `access_control` entries (no drift).

- **TestPermissionsDrift_UserNameNoDriftWithDifferentCasing**: Simulates the exact drift from issue #5183: `InstanceState` has `user_name="user@example.com"` but the API returns `User@Example.com`. Verifies that after the fix, the Read normalizes the casing so that:
  1. The `access_control` set still has exactly 2 entries (not 3, which would indicate drift)
  2. The `user_name` value in state is `"user@example.com"` (lowercase)

Both tests use `MockWorkspaceClientFunc` with mocked `CurrentUser.Me()` and `Permissions.Get()` responses, and construct realistic `InstanceState` maps with pre-computed set hashes to simulate the post-apply state that Terraform would produce.

- [x] `make test` run locally
- [x] using Go SDK

## Notes

- No documentation changes needed. This is a bug fix with no new fields or behavioral changes to document.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,7 @@
 * Add support for managing permissions of Agent Bricks resources ([#5708](https://github.com/databricks/terraform-provider-databricks/issues/5669)). Reverts [#5582](https://github.com/databricks/terraform-provider-databricks/pull/5708).
 
 ### Bug Fixes
+* Fix permanent permissions drift when `user_name` casing in `databricks_permissions` `access_control` blocks differs from the API response ([#5183](https://github.com/databricks/terraform-provider-databricks/issues/5183))
 
 * Fix `databricks_metastore` so that updating `external_access_enabled` from `true` to `false` is sent in the PATCH request. Previously the field was silently dropped from the request body, so the change never reached the API.
 * Fixed destroying of UC objects when workspace binding removed before actual destroy ([#5581](https://github.com/databricks/terraform-provider-databricks/pull/5581)).
--- a/permissions/entity/permissions_entity.go
+++ b/permissions/entity/permissions_entity.go
@@ -1,6 +1,10 @@
 package entity
 
-import "github.com/databricks/databricks-sdk-go/service/iam"
+import (
+	"strings"
+
+	"github.com/databricks/databricks-sdk-go/service/iam"
+)
 
 // PermissionsEntity is the one used for resource metadata
 type PermissionsEntity struct {
@@ -10,7 +14,7 @@ type PermissionsEntity struct {
 
 func (p PermissionsEntity) ContainsUserOrServicePrincipal(name string) bool {
 	for _, ac := range p.AccessControlList {
-		if ac.UserName == name || ac.ServicePrincipalName == name {
+		if strings.EqualFold(ac.UserName, name) || strings.EqualFold(ac.ServicePrincipalName, name) {
 			return true
 		}
 	}
--- a/permissions/permission_definitions.go
+++ b/permissions/permission_definitions.go
@@ -228,7 +228,7 @@ func (p resourcePermissions) prepareResponse(objectID string, objectACL *iam.Obj
 		// If the user doesn't include an access_control block for themselves, do not include it in the state.
 		// On create/update, the provider will automatically include the current user in the access_control block
 		// for appropriate resources. Otherwise, it must be included in state to prevent configuration drift.
-		if me == accessControl.UserName || me == accessControl.ServicePrincipalName {
+		if strings.EqualFold(me, accessControl.UserName) || strings.EqualFold(me, accessControl.ServicePrincipalName) {
 			if !existing.ContainsUserOrServicePrincipal(me) {
 				continue
 			}
@@ -244,8 +244,8 @@ func (p resourcePermissions) prepareResponse(objectID string, objectACL *iam.Obj
 			}
 			entity.AccessControlList = append(entity.AccessControlList, iam.AccessControlRequest{
 				GroupName:            accessControl.GroupName,
-				UserName:             accessControl.UserName,
-				ServicePrincipalName: accessControl.ServicePrincipalName,
+				UserName:             strings.ToLower(accessControl.UserName),
+				ServicePrincipalName: strings.ToLower(accessControl.ServicePrincipalName),
 				PermissionLevel:      permission.PermissionLevel,
 			})
 		}
--- a/permissions/resource_permissions.go
+++ b/permissions/resource_permissions.go
@@ -188,28 +188,30 @@ func ResourcePermissions() common.Resource {
 		}
 		s["access_control"].MinItems = 1
 
-		// Use a custom hash function that only considers non-empty fields.
-		// This prevents spurious diffs when comparing {group_name: "X", permission_level: "Y"}
-		// with {group_name: "X", permission_level: "Y", service_principal_name: "", user_name: ""}
+		// Use a custom hash function that normalizes case-insensitive identity
+		// fields before hashing. Emails (user_name) and service principal names
+		// are case-insensitive, but the API may return them with different casing
+		// than the user's config, causing permanent drift (#5183).
 		acSchema := s["access_control"].Elem.(*schema.Resource).Schema
 		s["access_control"].Set = func(v interface{}) int {
 			m, ok := v.(map[string]interface{})
 			if !ok {
 				return 0
 			}
-			// Build a normalized map with only non-empty string fields for hashing
 			normalized := make(map[string]interface{})
 			for key, val := range m {
 				if _, exists := acSchema[key]; !exists {
-					continue // Skip fields not in schema
+					continue
 				}
 				if strVal, ok := val.(string); ok {
 					if strVal != "" {
+						if key == "user_name" || key == "service_principal_name" {
+							strVal = strings.ToLower(strVal)
+						}
 						normalized[key] = strVal
 					}
 				}
 			}
-			// Use HashResource with a schema that only includes the fields we care about
 			hashSchema := &schema.Resource{Schema: acSchema}
 			return schema.HashResource(hashSchema)(normalized)
 		}
--- a/permissions/resource_permissions_test.go
+++ b/permissions/resource_permissions_test.go
@@ -2062,3 +2062,188 @@ func TestResourcePermissionsCreate_SupervisorAgent_InvalidLevel(t *testing.T) {
 		`,
 	}.ExpectError(t, "permission_level CAN_VIEW is not supported with supervisor_agent_id objects; allowed levels: CAN_MANAGE, CAN_QUERY")
 }
+
+// TestPermissionsDrift_UserNameNoDriftWithSameCasing simulates an existing resource
+// in state with user_name, then performs a Read where the API returns the same
+// casing. Verifies no drift occurs.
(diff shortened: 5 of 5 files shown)


alexott:
@jlieow can you please resolve conflicts here

jlieow:
> @jlieow can you please resolve conflicts here

Hi @alexott resolved.

PR #5393, opened 2026-02-13
[Fix] databricks_sql_endpoint resource doesn't set min_num_clusters to default value resulting in unmanaged drift if min_num_clusters is updated outside terraform.

### LINKED ISSUE #5294: [ISSUE] `databricks_sql_endpoint` resource doesn't set `min_num_clusters` to default value
### Configuration

```hcl
resource "databricks_sql_endpoint" "custom" {
  name           = "custom_endpoint"
  cluster_size   = "2X-Small"
  auto_stop_mins = 20
  warehouse_type = "PRO"
}
```

### Expected Behavior

One of our team members manually changed max and min number of clusters to 4 through the UI (previously they were set to the default of 1 during creation by TF):

Image

On the next run of Terraform, I would expect those to be set back to their default value.

### Actual Behavior

On the next run of Terraform, `max_num_clusters` is set back to the default value of 1, but `min_num_clusters` is not:

```hcl
  # databricks_sql_endpoint.custom will be updated in-place
  ~ resource "databricks_sql_endpoint" "custom" {
        id                        = "{endpoint_id}"
      ~ max_num_clusters          = 4 -> 1
        name                      = "custom_endpoint"
        # (15 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

Therefore, during TF apply we get an error, probably because we're not allowed to set max to be less than the min:

> Error: cannot update sql endpoint: 1 is not a valid value for max_num_clusters. The value must be greater than or equal to 4, and less than or equal to 40.

When I add `min_num_clusters` into the config, everything works fine:

```hcl
resource "databricks_sql_endpoint" "custom" {
  name           = "custom_endpoint"
  cluster_size   = "2X-Small"
  min_num_clusters = 1
  auto_stop_mins = 20
  warehouse_type = "PRO"
}
```

But I think it's unexpected that `max_num_clusters` and `min_num_clusters` work differently here.

```bash
$ terraform version
Terraform v1.14.3
on linux_amd64
```
Provider version is v1.100.0

jlieow:
Closes #5294

## Changes

The `databricks_sql_endpoint` resource sets defaults differently for `max_num_clusters` and `min_num_clusters`. `max_num_clusters` had `Default: 1` while `min_num_clusters` used `SuppressDiff()`.

`SuppressDiff()` meant Terraform did not manage `min_num_clusters` when unset by the user. Any changes outside of Terraform (e.g. via the Databricks UI) would go undetected, allowing drift to persist silently in Terraform state.

For example, given this config where neither `min_num_clusters` nor `max_num_clusters` is set:

```hcl
resource "databricks_sql_endpoint" "custom" {
  name           = "custom_endpoint"
  cluster_size   = "2X-Small"
  auto_stop_mins = 20
  warehouse_type = "PRO"
}
```

If someone changes both `min_num_clusters` and `max_num_clusters` to 4 via the UI, the next `terraform plan` would only detect drift on `max_num_clusters` (resetting it to its default of `1`) while `min_num_clusters` stays at `4`:

```
~ max_num_clusters = 4 -> 1
```

The subsequent `terraform apply` then fails because `max_num_clusters` (1) < `min_num_clusters` (4):

> Error: cannot update sql endpoint: 1 is not a valid value for max_num_clusters. The value must be greater than or equal to 4, and less than or equal to 40.

This PR replaces `SuppressDiff()` on `min_num_clusters` with `SetDefault(1)` and `IntBetween(1, MaxNumClusters)` validation, matching the existing `max_num_clusters` schema exactly. The Databricks Warehouses API (`Warehouses.Create`, `Warehouses.Edit`) default for `min_num_clusters` is also `1`, so this is a safe, non-breaking change.

---

### `sql/resource_sql_endpoint.go`

```go
// Before
common.CustomizeSchemaPath(m, "min_num_clusters").SetSuppressDiff()

// After
common.SetDefault(m["min_num_clusters"], 1)
m["min_num_clusters"].ValidateDiagFunc = validation.ToDiagFunc(
    validation.IntBetween(1, MaxNumClusters))
```

**Why:** Removed `common.CustomizeSchemaPath(m, "min_num_clusters").SetSuppressDiff()` because it meant Terraform did not manage `min_num_clusters` when unset by the user, allowing changes outside of Terraform to persist silently. Replaced with `common.SetDefault(m["min_num_clusters"], 1)` so Terraform actively manages the field with a known default of `1`, detecting and correcting drift. Added `validation.IntBetween(1, MaxNumClusters)` to validate user input is within the allowed range. Both are identical to how `max_num_clusters` already works.

---

### `sql/resource_sql_endpoint_test.go`

- Added `MinNumClusters: 1` to the shared `createRequest` fixture and all inline `CreateWarehouseRequest` / `EditWarehouseRequest` mock expectations (6 locations).
- Added `"min_num_clusters"` entry to `ExpectedDiff` maps in `TestResourceSQLEndpointUpdateHealthNoDiff` and `TestResourceSQLEndpointUpdateNoAutoTermination`.

**Why:** With the new default, Terraform now includes `min_num_clusters: 1` in all API requests. Test fixtures must match the expected request payloads.

---

### `sql/sql_endpoint_test.go`

- Added `TestAccSQLEndpointMinNumClustersDefault` acceptance test.

**Why:** Validates end-to-end that when `min_num_clusters` is not set in the config, both `MinNumClusters` and `MaxNumClusters` default to `1` on the API side after apply.

---

### Unit Tests (`sql/resource_sql_endpoint_test.go`)

All 12 existing `TestResourceSQLEndpoint*` tests updated to include `MinNumClusters: 1` in API request expectations and passing.

### Acceptance Test (`sql/sql_endpoint_test.go`)

- **TestAccSQLEndpointMinNumClustersDefault** - Creates a warehouse without setting `min_num_clusters` or `max_num_clusters`, then verifies via the API that both default to `1`.

- [x] `make test` run locally
- [x] using Go SDK

## Notes

- No `docs/` change is needed. `docs/resources/sql_endpoint.md` already documents `min_num_clusters` as optional. The default value of `1` now matches `max_num_clusters` behavior.


tanmay-db (commented):

PR #5397, opened 2026-02-16
Update Go SDK to v0.108.0

rauchy:
## Changes
Update Go SDK to v0.108.0 and API Spec Version to match the SDK.

### DIFF
--- a/jobs/resource_job_test.go
+++ b/jobs/resource_job_test.go
@@ -2287,7 +2287,7 @@ func TestResourceJobCreateFromGitSourceWithoutProviderFail(t *testing.T) {
 			}
 		}
 	`,
-	}.ExpectError(t, "git source is not empty but Git Provider is not specified and cannot be guessed by url &{GitBranch: GitCommit: GitProvider: GitSnapshot:<nil> GitTag:0.4.8 GitUrl:https://custom.git.hosting.com/databricks/terraform-provider-databricks JobSource:<nil> ForceSendFields:[]}")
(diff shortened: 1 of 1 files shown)


docs/resources/endpoint.md
```diff
@@ -0,0 +1,71 @@
...
+### Azure Private Endpoints
+This resource manages Azure Private Endpoints that connect to Databricks workspaces through Private Link Service.
+
+
+## Example Usage
+### Example for Azure cloud
+This is an example for creating an endpoint in Azure cloud:
+```hcl
+resource "databricks_endpoint" "this" {
+  account_id    = "eae3abf6-1496-494e-9983-4660a5ad5aab"
```
alexott, line 20: it's an attribute, it shouldn't be in the docs

docs/resources/endpoint.md
```diff
@@ -0,0 +1,71 @@
...
+  endpoint_name = "my-private-endpoint"
```
alexott, line 21: is it correct? we don't have it in the doc below, but have `display_name`

docs/resources/endpoint.md
```diff
@@ -0,0 +1,71 @@
...
+}
+```
+
+
+## Arguments
+The following arguments are supported:
+* `display_name` (string, required) - The human-readable display name of this endpoint.
+  The input should conform to RFC-1034, which restricts to letters, numbers, and hyphens,
+  with the first character a letter, the last a letter or a number, and a 63 character maximum
+* `parent` (string, required)
```
alexott, line 36: it's marked as `required`, but it isn't in the example

PR #5410, opened 2026-02-19
[Internal] Add support for host agnostic SQL global config resource

tanmay-db:
## Changes

Remove cloud-specific runtime validation for instance_profile_arn (AWS) and google_service_account (GCP) 

## Tests

Integration tests

### DIFF
--- a/sql/resource_sql_global_config_test.go
+++ b/sql/resource_sql_global_config_test.go
@@ -185,32 +185,6 @@ func TestResourceSQLGlobalConfigCreateWithData(t *testing.T) {
 	assert.Equal(t, "PASSTHROUGH", d.Get("security_policy"))
 }
 
-func TestResourceSQLGlobalConfigCreateError(t *testing.T) {
-	_, err := qa.ResourceFixture{
-		Resource: ResourceSqlGlobalConfig(),
-		Fixtures: []qa.HTTPFixture{
-			{
-				Method:       "GET",
-				Resource:     "/api/2.0/sql/config/warehouses",
-				ReuseRequest: true,
-				Response: GlobalConfigForRead{
-					SecurityPolicy: "DATA_ACCESS_CONTROL",
-				},
-			},
-		},
-		Create: true,
-		Azure:  true,
-		State: map[string]any{
-			"security_policy":      "PASSTHROUGH",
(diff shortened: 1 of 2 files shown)


sql/resource_sql_global_config.go
```diff
@@ -73,18 +72,10 @@ func (a globalConfigAPI) Set(gc GlobalConfig, d *schema.ResourceData) error {
 		}
 	}
 	if gc.InstanceProfileARN != "" {
-		if a.client.IsAws() {
```
hectorcast-db, line 76: Did we check with teams that it's ok to send this?

sql/resource_sql_global_config.go
```diff
@@ -127,6 +118,8 @@ func ResourceSqlGlobalConfig() common.Resource {
 		m map[string]*schema.Schema) map[string]*schema.Schema {
 		m["enable_serverless_compute"].Deprecated = "This field is intended as an internal API " +
 			"and may be removed from the Databricks Terraform provider in the future"
+		m["instance_profile_arn"].ConflictsWith = []string{"google_service_account"}
```
hectorcast-db: So this only applies to user provided data? Because we are are filling in both in the change above.

sql/resource_sql_global_config.go
tanmay-db: The API errors if these are set for the wrong cloud (will send details internally).

sql/resource_sql_global_config.go
tanmay-db: Yes, since the API errors out for the wrong cloud, we can prevent this at plan phase but it might be the case that they have these values set and we ignored the issues so far because we only pass these values to the API depending on cloud, removing schema conflicts to back to original behavior

sql/resource_sql_global_config.go
tanmay-db: and we also document this in our docs: https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/sql_global_config

google_service_account :Please note that this parameter is only for GCP, and will generate an error if used on other clouds.
instance_profile_arn : Please note that this parameter is only for AWS, and will generate an error if used on other clouds.

PR #5437, opened 2026-02-27
[Internal] Update unified host pathway for incoming support for default workspace ID

tanmay-db:
## Changes

- We are going to add support for default workspace id. With this support, the databricks profiles created for unified host that have workspace_id in them can be properly used even when the workspace ID is not specified at the resource level. Route unified host through account-level code paths instead of falling through to workspace-level.
- Bump Go SDK to v0.118.0

## Tests

- Integration tests

NO_CHANGELOG=true

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -272,6 +272,7 @@ func (c *DatabricksClient) WorkspaceClientForWorkspace(ctx context.Context, work
 	if err != nil {
 		// Fallback: create workspace client on the same host with workspace_id set.
 		// This works for unified hosts that can route by workspace_id through X-Databricks-Org-Id header.
+		// Note: GetWorkspaceClient(*workspace) supports all host, this is the case when users don't have access to the account
 		w, err = c.tryWorkspaceClientDirect(workspaceId)
 		if err != nil {
 			return nil, err


common/client.go
```diff
@@ -267,21 +267,46 @@ func (c *DatabricksClient) WorkspaceClientForWorkspace(ctx context.Context, work
...
 	}
-	a, err := c.accountClient()
-	if err != nil {
-		return nil, err
-	}
-	workspace, err := a.Workspaces.Get(ctx, provisioning.GetWorkspaceRequest{WorkspaceId: workspaceId})
-	if err != nil {
-		return nil, err
+	var w *databricks.WorkspaceClient
+	if c.Config.HostType() == config.UnifiedHost {
```
hectorcast-db: If everything goes to plan, by the next week, we won't have config.UnifiedHost. 
If we are adding dependencies on it, we need a plan to move away ASAP.

internal/acceptance/unified_host_acc_test.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	`
+
+	run(t, []Step{
+		{
+			Template: workspaceTemplate,
+		},
+		{
+			PreConfig: func() {
```
hectorcast-db: As discussed, let's not do this. Too slow and the same can be tested with an existing workspace.

Issue #1238, opened 2022-04-11
[ISSUE] Resource databricks_cluster removes spark_env_vars without warning


mstreuhofer:
Terraform removes manually configured `spark_env_vars` every time any change to a `databricks_cluster` needs to be done even though `spark_env_vars` is not configured/maintained by terraform. Even while trying to use `lifecycle` `ignore_changes` `spark_env_vars` terraform always removes any manually configured environment variables.

### Configuration
```hcl
resource "databricks_cluster" "this" {
  cluster_name     = "Cluster Display Name"
  single_user_name = lower("example@example.com")

  node_type_id            = "Standard_DS3_v2"
  spark_version           = "9.1.x-cpu-ml-scala2.12"
  autotermination_minutes = 15
  is_pinned               = true

  autoscale {
    min_workers = 1
    max_workers = 2
  }

  spark_conf = {
    "spark.databricks.passthrough.enabled" : true
    "spark.hadoop.fs.permissions.umask-mode" : "000"
  }
}
```

### Expected Behavior
Cluster should be created (works).
Changes to e.g. `autotermination_minutes` should be done if changed in terraform config (works).
Environment variables which are set by the user should not be touched (does not work).

### Actual Behavior
Every time terraform decides that a change is necessary it will do so and warn about what will change. It never mentiones `spark_env_vars`. Yet the list of configured environment variables is always reset to an empty list.

### Steps to Reproduce
1. Let a cluster be created.
2. Manually configure Spark environment variables.
3. Let the cluster be changed by e.g. increasing the `autotermination_minutes`.
4. Manually configured Spark environment variables are gone.

### Terraform and provider versions

Terraform: **v1.1.7 on darwin_amd64**
Databricks: **0.5.4**

nfx:
@mstreuhofer works as designed. all manual configuration changes would _always_ be overridden by tf.

mstreuhofer:
@nfx thanks for your response. but i do have a follow-up question if i may.

shouldn't a manual change be shown as such before being reverted?

shouldn't i be able to ignore certain changes by using constructs like the following?

```
  lifecycle {
    ignore_changes = [
      spark_env_vars,
    ]
  }
```

mstreuhofer:
@nfx does your reopening of this issue mean the "invalid" tag assigned to this issue is infact invalid? ;)

nfx:
Following up - is this issue still relevant?

mstreuhofer:
[...]

PR #5438, opened 2026-02-27
[Fix] Preserve externally set `spark_env_vars` during `databricks_cluster` updates when `ignore_changes` is used

jlieow:
Related to #1238

## Changes

When `spark_env_vars` are set externally (e.g. by cluster policies or manual edits) but not declared in the Terraform config, updating any other cluster field (like `cluster_name`) causes the Edit API to wipe those env vars. This happens because the Clusters Edit API (`POST /api/2.1/clusters/edit`) does a full replacement — omitted fields are cleared. The provider was sending nil for `spark_env_vars` since `StructToData` skips writing API-returned values to state when the field is not in the user's config.

This fix carries over existing `spark_env_vars` from the current cluster state into the Edit request when the user has not configured any. This also fixes `lifecycle { ignore_changes = [spark_env_vars] }` which previously had nothing to preserve.

No schema changes. No changes to Read behavior.

---

### Preserve spark_env_vars in update path (`clusters/resource_cluster.go`, lines 622-636)

**Before:**
```go
		} else {
			SetForceSendFieldsForCluster(&cluster, d)
```

**After:**
```go
		} else {
			// Preserve externally-set spark_env_vars (e.g. from cluster policies)
			// when not configured by the user. The Edit API does a full replacement,
			// so omitting spark_env_vars would clear them.
			//
			// cluster.SparkEnvVars = from Terraform state (HCL config)
			// clusterInfo.SparkEnvVars = from GET API (actual cluster values)
			//
			// HCL has env vars | Cluster has env vars | Action
			// false            | false                | nothing to preserve
			// false            | true                 | carry over values set outside of Terraform
			// true             | false                | values in Terraform takes precedence
			// true             | true                 | values in Terraform takes precedence
			if len(cluster.SparkEnvVars) == 0 && len(clusterInfo.SparkEnvVars) > 0 {
				cluster.SparkEnvVars = clusterInfo.SparkEnvVars
			}
			SetForceSendFieldsForCluster(&cluster, d)
```

**Why:** The `EditCluster` struct is built from Terraform state via `DataToStructPointer` (line 546). When `spark_env_vars` is not in the user's config, state has no value for it, so `cluster.SparkEnvVars` is nil. The Edit API interprets this as "clear spark_env_vars". The current cluster details are already fetched at line 578 (`clusterInfo`), so we carry over the existing env vars when the user hasn't configured any.

---

## Tests

- **TestResourceClusterUpdate_PreservesExternalSparkEnvVars**: Simulates an update where the cluster GET returns `spark_env_vars` set externally (`PYSPARK_PYTHON`, `ENV_FROM_POLICY`) but the user's Terraform state has no `spark_env_vars`. Asserts via `ExpectedRequest` that the Edit API call includes the externally-set env vars rather than omitting them.

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,9 +8,11 @@
 
 ### Bug Fixes
 
-* Fixed `databricks_cluster` to preserve externally-set `spark_env_vars` (e.g. from cluster policies) during updates when not configured in Terraform. This also fixes `lifecycle { ignore_changes = [spark_env_vars] }` which previously failed to prevent deletion of externally-set values.
+* Fixed `databricks_cluster` to preserve externally-set `spark_env_vars` (e.g. from cluster policies) during updates when not configured in Terraform. This also fixes `lifecycle { ignore_changes = [spark_env_vars] }` which previously failed to prevent deletion of externally-set values ([#1238](https://github.com/databricks/terraform-provider-databricks/issues/1238)).
 
 ### Documentation
+
+* Added documentation note about whitespace handling in `MAP` column types for `databricks_sql_table`.
 
 ### Exporter
 


clusters/resource_cluster.go
```diff
@@ -619,6 +619,21 @@ func resourceClusterUpdate(ctx context.Context, d *schema.ResourceData, c *commo
...
+			// clusterInfo.SparkEnvVars = from GET API (actual cluster values)
+			//
+			// HCL has env vars | Cluster has env vars | Action
+			// false            | false                | nothing to preserve
+			// false            | true                 | carry over values set outside of Terraform
+			// true             | false                | values in Terraform takes precedence
+			// true             | true                 | values in Terraform takes precedence
+			if len(cluster.SparkEnvVars) == 0 && len(clusterInfo.SparkEnvVars) > 0 {
+				cluster.SparkEnvVars = clusterInfo.SparkEnvVars
+			}
```
tanmay-db, line 636: @jlieow if Edit does full replacement, isn't the expected behaviour for `terraform apply` that whatever is in the config is applied? Should these variables be part of config then?

NEXT_CHANGELOG.md
```diff
@@ -8,9 +8,9 @@
 
 ### Bug Fixes
 
-### Documentation
+* Fixed `databricks_cluster` to preserve externally-set `spark_env_vars` (e.g. from cluster policies) during updates when not configured in Terraform. This also fixes `lifecycle { ignore_changes = [spark_env_vars] }` which previously failed to prevent deletion of externally-set values ([#1238](https://github.com/databricks/terraform-provider-databricks/issues/1238)).
 
-* Added documentation note about whitespace handling in `MAP` column types for `databricks_sql_table`.
```
tanmay-db, line 13: This shouldn't be removed

clusters/resource_cluster.go
jlieow: This is related to #1238. In that issue, we can see that users are trying to manage parts of the resource in Terraform and others in the UI.

The user is attempting to use `ignore_changes` to let Terraform and an external process manage different parts of the same object, which is what `ignore_changes` is supposed to do as per the lifecycle reference [doc](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes).

However, due to the full replacement behaviour of the Edit API, this breaks the Terraform lifecycle behaviour. `ignore_changes` only suppresses drift detection in the plan, but the provider still sends an empty map to the API, which clears the externally-set values. This impacts users who manage this resource via multiple processes.

This fix assumes that native Terraform features like `ignore_changes` should take precedence since these are core features that users expect to work across all resources and not be tied to API limitations.

NEXT_CHANGELOG.md
jlieow: I have amended the commit and restored the note.

jlieow:
Superseded by #5510

PR #5443, opened 2026-03-01
[Fix] Fix empty `databricks_share` creation failing with 'no fields to update'

### LINKED ISSUE #5442: [ISSUE] Creating empty `databricks_share` fails with "UpdateShare no fields to update"
## Description

Creating a `databricks_share` resource with no `object` blocks fails during `terraform apply` with:

```
Error: failed to update share

UpdateShare no fields to update.
```

## Configuration to Reproduce

```hcl
resource "databricks_share" "example" {
  name = "my_empty_share"
}
```

## Expected Behavior

An empty share (with no objects) should be created successfully. The Databricks API supports creating shares without objects — the `CreateShare` call succeeds, but the subsequent unconditional `UpdateShare` call fails because there is nothing to update.

## Actual Behavior

The Plugin Framework implementation of `databricks_share` in `internal/providers/pluginfw/products/sharing/resource_share.go` calls `w.Shares.Update()` unconditionally in the `Create` method, even when the plan contains no objects, no owner, and no comment. The API rejects this with "no fields to update".

The relevant code path:

```go
shareInfo, err := w.Shares.Create(ctx, createShare)
// ...
shareChanges := shareChanges(planGoSDK, string(sharing.SharedDataObjectUpdateActionAdd))
updatedShareInfo, err := w.Shares.Update(ctx, shareChanges)  // <-- fails when no objects
```

## Root Cause

The `Create` method calls `w.Shares.Update()` unconditionally after `w.Shares.Create()`. When the plan has no `object` blocks, `shareChanges()` produces an `UpdateShare` request with an empty `Updates` slice and no `Owner`/`Comment`, which the API rejects.

Note that the `Update` method already has a guard for this:
```go
if len(changes) > 0 || !plan.Comment.IsNull() {
```

But the `Create` method is missing an equivalent guard.

## Proposed Fix

Wrap the `Update` call in the `Create` method with a guard that checks whether there is actually anything to update, mirroring the pattern in the `Update` method:

```go
if len(shareChanges.Updates) > 0 || shareChanges.Owner != "" || !plan.Comment.IsNull() {
    updatedShareInfo, err := w.Shares.Update(ctx, shareChanges)
    // ...
}
```

## Provider Version

v1.110.0

## Terraform Version

v1.12.1

slimbnsalah:
## Changes

The Plugin Framework implementation of `databricks_share` calls `w.Shares.Update()` unconditionally in the `Create` method after `w.Shares.Create()`. When the resource is created with no `object` blocks, the `UpdateShare` request has an empty `Updates` slice, no `Owner`, and no `Comment`, causing the API to reject it with:

```
Error: failed to update share
UpdateShare no fields to update.
```
### Fix

Wrapped the `Update` call in the `Create` method with a guard that skips it when there is nothing to update:

```go
if len(shareChanges.Updates) > 0 || shareChanges.Owner != "" || !plan.Comment.IsNull() {
    // ... call Update ...
}
```

This mirrors the guard already present in the `Update` method at line 386:

```go
if len(changes) > 0 || !plan.Comment.IsNull() {
```

When there's nothing to update, the share returned by `Create` is used directly for state.

Fixes #5442


alexott (approved):
lgtm, thank you

alexott:
@slimbnsalah please resolve conflicts

slimbnsalah:
@alexott done

hectorcast-db:
@rauchy @tanmay-db Shares is a delicate resource. You have more knowledge on it.

tanmay-db (commented):
Hi @slimbnsalah, thanks for creating the PR, is this something which used to work before and now not working? Trying to understand if this was always a bug vs something changed on the API side?

slimbnsalah:
hi @tanmay-db, this was always a bug in terraform, afaik nothing changed on the API because it always allowed creating empty shares

PR #5449, opened 2026-03-03
[Internal] Replace error string matching with workspace GetStatus API for notebook parent folder check

tanmay-db:
## Changes

  - When creating a notebook, if the import fails, use `GET /api/2.0/workspace/get-status` to check
    if the parent folder exists instead of string-matching the error message for "parent folder does not exist"
  - Uses structured `apierr.IsMissing` check instead of fragile `isParentDoesntExistError` string matching

## Tests

- Unit tests
- Integration tests


alexott:
This introduces much bigger load on the backend and we'll hit rate limits much earlier. The current implementation was a result of a performance analysis that we did for big plans.

tanmay-db:
> This introduces much bigger load on the backend and we'll hit rate limits much earlier. The current implementation was a result of a performance analysis that we did for big plans.

Moved the creation post creation error

hectorcast-db (approved):

PR #5451, opened 2026-03-03
[Fix] Cannot provision `databricks_app` when `description` is omitted despite it being `Optional`

jlieow:
## Changes

When users create a `databricks_app` resource without specifying `description`, Terraform fails with:
```
provider produced inconsistent result after apply: .description: was null, but now cty.StringVal("")
```

The `description` attribute was marked as `Optional` only. When omitted, Terraform planned it as `null`. The Databricks API always returns `"description": ""` in its response, and the SDK unmarshaller + converter pipeline produced `types.StringValue("")` instead of `types.StringNull()`, causing a plan-vs-state mismatch.

This fix marks `description` as `Optional + Computed`, so Terraform treats an omitted description as "unknown" during planning and accepts the API's empty-string response without error.

---

### Mark `description` as Computed (`resource_app.go:43`)

**Before:**
```go
s = apps_tf.App{}.ApplySchemaCustomizations(s)
return s
```

**After:**
```go
s = apps_tf.App{}.ApplySchemaCustomizations(s)
s["description"] = s["description"].SetComputed()
return s
```

**Why:** The auto-generated `ApplySchemaCustomizations` in `apps_tf.App` sets `description` as `Optional` only. Adding `SetComputed()` after that call makes it `Optional + Computed`. When the user omits `description`, Terraform plans it as "unknown" rather than "null", so the API returning `""` is accepted without error.

---

## Tests

- **TestAppResourceSchema_DescriptionIsOptionalAndComputed**: Builds the `databricks_app` resource schema and asserts that the `description` attribute is `Optional` and `Computed`. This test would have caught the original bug by detecting that `description` lacked the `Computed` flag.

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* Fixed `databricks_app` producing "inconsistent result after apply" error when `description` is omitted.
+
 ### Documentation
 
 * Added documentation note about whitespace handling in `MAP` column types for `databricks_sql_table`.
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bug Fixes
 
-* Fixed `databricks_app` producing "inconsistent result after apply" error when `description` is omitted.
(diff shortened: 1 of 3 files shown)


jlieow:
@fjakobs

tanmay-db (commented):

PR #5454, opened 2026-03-05
Fix `databricks_instance_pool` ignoring `enable_elastic_disk = false`

### LINKED ISSUE #5462: [Bug] `databricks_instance_pool` ignores `enable_elastic_disk = false` due to `omitempty` JSON tag
### Description

Setting `enable_elastic_disk = false` on `databricks_instance_pool` is silently ignored, causing an infinite plan/replace cycle.

### Root Cause

The `EnableElasticDisk` field in both `InstancePool` and `InstancePoolAndStats` structs uses `json:"enable_elastic_disk,omitempty"`. Since `false` is the zero value for `bool` in Go, `omitempty` omits it from the serialized JSON. The API never receives `enable_elastic_disk: false` and defaults to `true`.

Combined with `tf:"force_new"`, this creates an infinite cycle:
1. User sets `enable_elastic_disk = false`
2. Terraform creates pool — but `omitempty` drops the field from the request
3. API defaults to `true`
4. Next `terraform plan` sees drift (`true → false`), wants to recreate
5. Repeat forever

### Verification

Direct REST API call confirms Databricks **does** support `enable_elastic_disk: false`:

```bash
# Create with false
curl -X POST ".../api/2.0/instance-pools/create" \
  -d '{"instance_pool_name":"test","node_type_id":"Standard_F4","enable_elastic_disk":false,...}'

# Read back — confirmed false
curl ".../api/2.0/instance-pools/get?instance_pool_id=..."
# Returns: "enable_elastic_disk": false
```

### Impact

Users with NVMe-equipped VMs (e.g. `Standard_E4d_v4`, `Standard_D4ds_v5`) are forced to pay for both the included NVMe local SSD **and** an additional managed disk (elastic disk), since there is no way to disable elastic disk via Terraform.

### Fix

Remove `omitempty` from the JSON tag on `EnableElasticDisk` in both structs so that `false` is explicitly sent to the API.

PR: https://github.com/databricks/terraform-provider-databricks/pull/5454

### Environment

- Terraform Provider Version: latest
- Cloud: Azure

Bafff:
## Summary

`enable_elastic_disk = false` on `databricks_instance_pool` is silently ignored, causing an infinite plan/replace cycle.

Fixes https://github.com/databricks/terraform-provider-databricks/issues/5462

## Root Cause

The `EnableElasticDisk` field in both `InstancePool` and `InstancePoolAndStats` structs uses `json:"enable_elastic_disk,omitempty"`. Since `false` is the zero value for `bool` in Go, `omitempty` omits it from the serialized JSON. The API never receives `enable_elastic_disk: false` and defaults to `true`.

Combined with `tf:"force_new"`, this creates an infinite cycle:
1. User sets `enable_elastic_disk = false`
2. Terraform creates pool — but `omitempty` drops the field from the request
3. API defaults to `true`
4. Next `terraform plan` sees drift (`true → false`), wants to recreate
5. Repeat forever

## Verification

Direct REST API call confirms Databricks **does** support `enable_elastic_disk: false`:

```bash
# Create with false
curl -X POST ".../api/2.0/instance-pools/create" \
  -d '{"instance_pool_name":"test","node_type_id":"Standard_F4","enable_elastic_disk":false,...}'

# Read back — confirmed false
curl ".../api/2.0/instance-pools/get?instance_pool_id=..."
# Returns: "enable_elastic_disk": false
```

## Fix

Remove `omitempty` from the JSON tag on `EnableElasticDisk` in both structs so that `false` is explicitly sent to the API.

## Test plan

- [ ] Existing unit tests pass (`go test ./pools/...`)
- [ ] Create instance pool with `enable_elastic_disk = false` — verify API receives the value
- [ ] Subsequent `terraform plan` shows no changes

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bug Fixes
 
 * Mark `effective_file_event_queue` as read-only in `databricks_external_location` to prevent Terraform drift.
+* Fixed `databricks_instance_pool` ignoring `enable_elastic_disk = false` due to `omitempty` JSON tag ([#5454](https://github.com/databricks/terraform-provider-databricks/pull/5454)).
 ### Documentation
 
 ### Exporter
--- a/pools/resource_instance_pool.go
+++ b/pools/resource_instance_pool.go
@@ -89,7 +89,7 @@ type InstancePool struct {
 	GcpAttributes                      *InstancePoolGcpAttributes      `json:"gcp_attributes,omitempty" tf:"force_new,suppress_diff"`
 	NodeTypeID                         string                          `json:"node_type_id,omitempty" tf:"suppress_diff,force_new,conflicts:instance_pool_fleet_attributes"`
 	CustomTags                         map[string]string               `json:"custom_tags" tf:"optional"`
-	EnableElasticDisk                  bool                            `json:"enable_elastic_disk,omitempty" tf:"force_new,suppress_diff"`
+	EnableElasticDisk                  bool                            `json:"enable_elastic_disk" tf:"optional,force_new,suppress_diff"`
 	DiskSpec                           *InstancePoolDiskSpec           `json:"disk_spec,omitempty" tf:"force_new"`
 	PreloadedSparkVersions             []string                        `json:"preloaded_spark_versions,omitempty" tf:"force_new"`
 	PreloadedDockerImages              []clusters.DockerImage          `json:"preloaded_docker_images,omitempty" tf:"force_new,slice_set,alias:preloaded_docker_image"`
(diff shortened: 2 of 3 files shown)


Bafff:
Hey!

What should I do to help with this?
@Divyansh-db

Divyansh-db:
Hey @Bafff — thanks so much for catching this bug and writing the fix! Our CI test suite can't run on PRs from forks, which is why this one couldn't get past the checks. I've opened #5802 with the same change from an upstream branch so the tests can run. The diagnosis, code change, and unit test are all credited to you in that PR.

Closing this in favor of #5802. Thanks again for the contribution!

PR #5476, opened 2026-03-11
[Internal] Use TEST_ENVIRONMENT_TYPE for acceptance test environment detection

tanmay-db:
## Changes

  - Replaces ad-hoc environment variable checks (`DATABRICKS_ACCOUNT_ID`, `TEST_METASTORE_ID`) with a single `TEST_ENVIRONMENT_TYPE` variable for determining test execution context
  - Introduces `skipIfNotEnvironmentType()` helper that centralizes skip logic for all four test levels: `ACCOUNT`, `WORKSPACE`, `UC_ACCOUNT`, `UC_WORKSPACE`
  - Simplifies `LoadWorkspaceEnv`, `LoadAccountEnv`, `LoadUcwsEnv`, and `LoadUcacctEnv` to one-liners

## Tests

NO_CHANGELOG=true


internal/acceptance/init.go
```diff
@@ -436,35 +436,33 @@ func setDebugLogger() {
 
 func LoadWorkspaceEnv(t *testing.T) {
 	initTest(t, "workspace")
-	if os.Getenv("DATABRICKS_ACCOUNT_ID") != "" {
```
hectorcast-db, line 439: I think tthat here we want to run it in both WORKSPCE and UC_WORKSPACE. That is the current behavior.

tanmay-db: Added

PR #5480, opened 2026-03-15
[Feature] Add `databricks_secret` write-only attributes

ashenm:
## Changes
Add `string_value_wo` and `string_value_wo_version` ephemeral attributes to `databricks_secret` resource allowing write-only secret value population avoiding secret value being written to state

## Tests
Tested via unit-tests and also deployed locally using local provider build

- [X] `make test` run locally
- [ ] using Go SDK


rauchy (changes requested):
@ashenm Thanks for the PR!

The write-only approach makes sense, but I'm not a fan of changing provider infrastructure (`common/resource.go`, `qa/`) for this - that affects every resource in the provider and I'd rather keep the blast radius small.

I think you may avoid all of that by adding an Update handler to the resource. `PutSecret` is already an upsert so it'd be basically the same as Create. That makes the resource CRUD instead of CRD, which means the ForceNew loop in `common/resource.go` won't kick in and you don't need to patch it. As a bonus, `string_value_wo_version` wouldn't need `ForceNew` either - bumping it would just trigger an update instead of destroy+recreate, which is nicer for users doing secret rotation.

ashenm:
@rauchy yep make sense; updated to CRUD as suggested removing `common/` changes. Leaving the `qa/` changes as it's required to mimic the actual behaviour of `RawConfig` for tests

rauchy (approved):

hectorcast-db (commented):
@rauchy 
I see you approved it, but the approval was removed. Can you take a look?

ashenm:
@rauchy @hectorcast-db forced push was due to conflicts in NEXT_CHANGELOG.md; could you guys help to review and reapprove?

rauchy (approved):

ashenm:
https://github.com/databricks/terraform-provider-databricks/pull/5782

PR #5486, opened 2026-03-17
Add disabled tasks support, tests, and doc

ricardoc-db:
## Changes

Add disabled tasks support, tests, and doc

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK


jobs/job_test.go
```diff
@@ -318,6 +318,87 @@ func TestAccJobTasks(t *testing.T) {
 	})
 }
 
+func TestAccJobDisabledTask(t *testing.T) {
```
tanmay-db, line 310: I ran the integration test locally and it failed with 

> expected task 'b' to have Disabled=true

From the logs, I don't see disabled as true in the API response, is this expected?

ricardoc-db: Removed the integration test from this PR as the field is not yet working in every single environment.
Integration test was moved to another PR: https://github.com/databricks/terraform-provider-databricks/pull/5506

tanmay-db (approved):

jobs/job_test.go
ricardoc-db: @tanmay-db reintroduced the integration test in this PR as the feature is GA now. Could you run on your end again, please?

PR #5492, opened 2026-03-18
workspace_id support of SDKv2 resources using Go SDK

Divyansh-db:
## Summary

  Adds workspace_id support for all SDKv2 resources in the unified provider. Users can now set workspace_id on an account-level provider and all SDKv2 resources will be managed against that workspace — the same behavior already available for Plugin Framework resources.

  What this enables

  An account-level provider with workspace_id can now manage SDKv2 workspace-level resources:

  provider "databricks" {
    host         = "https://accounts.cloud.databricks.com"
    account_id   = "..."
    workspace_id = "123456789"  # ← SDKv2 resources now respect this
  }

  resource "databricks_directory" "example" {
    path = "/Shared/my-dir"
  }

  Resources can also override the provider default per-resource:

  resource "databricks_notebook" "cross_workspace" {
    path           = "/Shared/notebook"
    content_base64 = "..."
    language       = "PYTHON"
    provider_config {
      workspace_id = "987654321"  # ← targets a different workspace
    }
  }

  Changing the effective workspace (via provider workspace_id or per-resource provider_config) triggers ForceNew — the resource is destroyed in the old workspace and recreated in the new one.

###  Key changes

  - Workspace ID resolution: The effective workspace ID is resolved from (in priority order): resource-level provider_config.workspace_id, provider-level workspace_id, or lazily via the SCIM /Me API for workspace hosts. The resolved ID is persisted in state so that workspace changes are detectable across plans.
  - Plan-time validation (NamespaceCustomizeDiff): Validates that the target workspace is reachable and triggers ForceNew when the effective workspace ID changes. Dual resources (e.g., metastore, storage_credential) operating at account level are skipped via api field guards.
  - State population (populateProviderConfigInState): Post-Read hook writes the effective workspace ID into state after Create and Import. On subsequent reads, preserves the existing state value so CustomizeDiff can detect workspace changes.
  - Client routing: WorkspaceClientUnifiedProvider (Go SDK resources) and DatabricksClientForUnifiedProvider (HTTP-path resources) resolve a workspace-scoped client based on the effective workspace ID, with caching to avoid redundant client creation.
  - Schema change: provider_config.workspace_id changed from Required to Optional+Computed — when omitted, the provider's workspace_id is used as the default.
  - Removed HostType dependency: All config.HostType() checks removed from validation and state population, aligning with the PF approach of host-agnostic behavior.
  - Test infrastructure (qa/testing.go): Pre-populates cachedWorkspaceClients in test fixtures when ProviderWorkspaceID is set, enabling account-level provider simulation in unit tests.

 ## Test coverage

  - 3114 unit tests pass (including 700+ new/updated lines in unified_provider_test.go)
  - 44 acceptance tests across 4 suites (workspace GoSDK, workspace HTTP, account GoSDK, account HTTP) covering 22 CUJs each: implicit defaults, explicit overrides, migrations between workspace and account providers, ForceNew on workspace change, provider upgrades, and error cases

 ### Tests

  - make test run locally
  - relevant change in docs/ folder
  - covered with integration tests in internal/acceptance
  - using Go SDK
  - using TF Plugin Framework
  - has entry in NEXT_CHANGELOG.md file

  NO_CHANGELOG=true


internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+
+// ==========================================
+// Validation Tests
+// ==========================================
+
+// TestAccWorkspaceID_InvalidWorkspaceID tests that
+// invalid workspace_id values in the provider block are rejected.
+// The SDK's workspace_id field does not have schema-level validation,
+// so invalid values are caught during workspace client creation.
+func TestAccWorkspaceID_InvalidWorkspaceID(t *testing.T) {
```
tanmay-db: The account tests are skipped since the name needs to match the target test environment filter (eg for aws prod account, it's `TestMwsAcc`), each test environment has it's own test filter.

That's why this and other account tests doesn't appear in the account test logs.

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+}
+
+// ==========================================
+// Account-Level: New Setup
+// ==========================================
+//
+// Account provider + workspace_id → create workspace-level resource.
+// Verifies: resource created, state has correct provider_config.0.workspace_id.
+
+func TestAccWorkspaceID_AccountNewSetup(t *testing.T) {
```
tanmay-db: same

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+			`workspace_id = "{env.TEST_WORKSPACE_ID}"`,
+			"",
+		),
+		Check: checkProviderConfigWSIDFromEnv(directoryResource, "TEST_WORKSPACE_ID"),
+	})
+}
+
+// TestAccWorkspaceID_AccountNewSetupWithOverride tests that provider_config.workspace_id
+// takes precedence over the provider-level workspace_id.
+func TestAccWorkspaceID_AccountNewSetupWithOverride(t *testing.T) {
```
tanmay-db: same

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+// Existing resource on workspace-level provider. Switch to account provider
+// with workspace_id pointing to the SAME workspace.
+// Expected: no changes (noop).
+//
+// Step 1: provider { host = TEST_WORKSPACE_URL } → workspace-level create.
+// Step 2: provider { workspace_id = TEST_WORKSPACE_ID } → account-level, same ws.
+//
+// Requires: account-level OAuth credentials that also work against the workspace host.
+
+func TestAccWorkspaceID_MigrationSameWorkspace(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+// Migration: Different Workspace
+// ==========================================
+//
+// Same as MigrationSameWorkspace, but workspace_id points to a DIFFERENT workspace.
+// Expected: ForceNew (destroy in old workspace, recreate in new).
+//
+// Step 1: provider { host = TEST_WORKSPACE_URL } → create in workspace 1.
+// Step 2: provider { workspace_id = TEST_WORKSPACE_ID_2 } → ForceNew.
+
+func TestAccWorkspaceID_MigrationDiffWorkspace(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+
+// ==========================================
+// Add provider_config Override (Same Workspace)
+// ==========================================
+//
+// Resource was created using workspace_id. User adds explicit
+// provider_config with the SAME workspace ID.
+// Expected: Noop. State already has the same value from the default; no diff.
+
+func TestAccWorkspaceID_AddOverrideSame(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+
+// ==========================================
+// Add provider_config Override (Different Workspace)
+// ==========================================
+//
+// Resource was created using workspace_id = TEST_WORKSPACE_ID.
+// User adds provider_config { workspace_id = TEST_WORKSPACE_ID_2 }.
+// Expected: ForceNew (effective workspace changes).
+
+func TestAccWorkspaceID_AddOverrideDiff(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+// ==========================================
+// Remove provider_config Override (Falls Back to Same Default)
+// ==========================================
+//
+// Resource has provider_config { workspace_id = X }. User removes it.
+// workspace_id is also X.
+// Expected: Noop. Optional+Computed preserves state value, and effective
+// workspace is unchanged (default fallback is the same value).
+
+func TestAccWorkspaceID_RemoveOverrideSame(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+// ==========================================
+//
+// Resource has provider_config { workspace_id = TEST_WORKSPACE_ID_2 }.
+// workspace_id = TEST_WORKSPACE_ID (different).
+// User removes provider_config → effective workspace changes → ForceNew.
+//
+// Step 1: Create in TEST_WORKSPACE_ID_2 (override wins over default).
+// Step 2: Remove override → falls back to TEST_WORKSPACE_ID → ForceNew.
+
+func TestAccWorkspaceID_RemoveOverrideDiff(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+
+// ==========================================
+// Change workspace_id (No provider_config)
+// ==========================================
+//
+// User changes workspace_id from TEST_WORKSPACE_ID to TEST_WORKSPACE_ID_2.
+// Resources have no explicit provider_config.
+// Expected: ForceNew for all affected resources.
+
+func TestAccWorkspaceID_ChangeDefault(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+// ==========================================
+// Change workspace_id (Resource Has Explicit Override)
+// ==========================================
+//
+// User changes workspace_id from TEST_WORKSPACE_ID to TEST_WORKSPACE_ID_2.
+// Resource has explicit provider_config { workspace_id = TEST_WORKSPACE_ID }.
+// Expected: Noop. The explicit override is the same value in both steps;
+// the default changed but the override shields the resource from any diff.
+
+func TestAccWorkspaceID_ChangeDefaultWithOverride(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+}
+
+// ==========================================
+// Account Provider Without workspace_id or provider_config
+// ==========================================
+//
+// Account-level provider with no workspace_id. Resource has no provider_config.
+// Expected: error during CRUD — no workspace_id available for routing.
+
+func TestAccWorkspaceID_NoDefaultNoOverride(t *testing.T) {
```
tanmay-db: ditto

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,572 @@
...
+//
+// User had workspace_id and resources relying on it (no explicit provider_config).
+// User removes workspace_id.
+// Expected: error — provider_config.workspace_id is in state but no source for it
+// in config and no workspace_id set.
+//
+// Step 1: Create with workspace_id.
+// Step 2: Remove workspace_id → plan error.
+
+func TestAccWorkspaceID_RemoveDefault(t *testing.T) {
```
tanmay-db: ditto

common/resource.go
```diff
@@ -6,13 +6,57 @@ import (
...
+// host) on the first time — when no workspace ID exists in state yet (after Create).
+// On subsequent reads, it preserves whatever is already in state.
+func populateProviderConfigInState(ctx context.Context, d *schema.ResourceData, c *DatabricksClient) error {
+	// If provider_config.workspace_id already exists in state, preserve it.
+	// During refresh reads the state value reflects the workspace the Read
+	// actually targeted. Overwriting it would prevent CustomizeDiff from
+	// detecting workspace changes.
+	if existing := d.Get(workspaceIDSchemaKey); existing != nil {
+		if existingStr, ok := existing.(string); ok && existingStr != "" {
+			d.Set("provider_config", []map[string]any{{"workspace_id": existingStr}})
```
tanmay-db: should we check for err returned from d.Set()?

common/resource.go
```diff
@@ -6,13 +6,57 @@ import (
...
+	wsID, _ := workspaceIDFromRawConfig(d)
+	if wsID == "" && c.DatabricksClient != nil && c.Config != nil {
+		wsID = c.Config.WorkspaceID
+	}
+	if wsID == "" && c.cachedWorkspaceID != 0 {
+		wsID = strconv.FormatInt(c.cachedWorkspaceID, 10)
+	}
+
+	if wsID != "" {
+		d.Set("provider_config", []map[string]any{{"workspace_id": wsID}})
```
tanmay-db: same

common/unified_provider.go
```diff
@@ -48,6 +50,7 @@ func AddNamespaceInSchema(m map[string]*schema.Schema) map[string]*schema.Schema
 	m["provider_config"] = &schema.Schema{
 		Type:     schema.TypeList,
 		Optional: true,
+		Computed: true,
```
tanmay-db, line 52: Would current users not using provider_config see any diff when they upgrade their terraform version during terraform plan?

tanmay-db: Also should check `make diff-schema` since we are making schema changes

common/unified_provider.go
```diff
@@ -168,9 +244,15 @@ func (c *DatabricksClient) DatabricksClientForUnifiedProvider(ctx context.Contex
...
-	// and can return the current client.
+	// If the workspace_id is not passed in the resource configuration,
+	// fall back to workspace_id for account-level providers.
+	// We don't error here when no workspace_id is found because the caller may
+	// use AccountOrWorkspaceRequest to route to account APIs, which doesn't need
+	// a workspace-scoped client.
 	if workspaceID == "" {
+		if c.DatabricksClient != nil && c.Config.HostType() == config.AccountHost && c.Config.WorkspaceID != "" {
+			return c.getDatabricksClientForUnifiedProvider(ctx, c.Config.WorkspaceID)
+		}
```
tanmay-db: Why do we need this? If workspaceID is not passed through provider_config or in the c.Config then we can just return the client, is this something that should be moved after this check?

internal/providers/sdkv2/sdkv2.go
```diff
@@ -373,6 +373,19 @@ func ConfigureDatabricksClient(ctx context.Context, d *schema.ResourceData, conf
...
+	// Validate workspace_id is not used with workspace-level providers
+	if databricksClient.Config.WorkspaceID != "" && databricksClient.Config.HostType() == config.WorkspaceHost {
+		return nil, diag.FromErr(fmt.Errorf("workspace_id cannot be used with a workspace-level provider; " +
+			"it is only supported when the provider is configured at the account level"))
+	}
+	// For workspace-level providers, eagerly resolve and cache the workspace ID
+	// so it's available for populateProviderConfigInState in the post-Read hook.
+	// This ensures the workspace ID from the host is stored in state.
+	if databricksClient.Config.HostType() != config.AccountHost {
+		if _, err := databricksClient.CurrentWorkspaceID(ctx); err != nil {
```
tanmay-db: What would happen if CurrentWorkspaceID fails as we don't throw the error and warn instead?

common/client.go
```diff
@@ -124,11 +124,14 @@ func (c *DatabricksClient) GetWorkspaceClientForUnifiedProvider(
...
-	// TODO: Link to the documentation once migration guide is published
+	// If workspace_id is not provided in provider_config, use the provider-level
+	// workspace_id from SDK config as fallback
 	if workspaceID == "" {
-		return nil, fmt.Errorf("workspace_id is not set, please set the workspace_id in the provider_config")
+		workspaceID = c.Config.WorkspaceID
+	}
+	if workspaceID == "" {
+		return nil, fmt.Errorf("managing workspace-level resources requires a workspace_id, " +
+			"but none was found in provider_config or the provider configuration")
```
tanmay-db: provider_config and provider configuration could be confusing

would be good to clarify provider_config in the resource and provider block for example

internal/providers/sdkv2/sdkv2.go
```diff
@@ -373,6 +373,19 @@ func ConfigureDatabricksClient(ctx context.Context, d *schema.ResourceData, conf
 	if err != nil {
 		return nil, diag.FromErr(err)
 	}
+	// Validate workspace_id is not used with workspace-level providers
+	if databricksClient.Config.WorkspaceID != "" && databricksClient.Config.HostType() == config.WorkspaceHost {
+		return nil, diag.FromErr(fmt.Errorf("workspace_id cannot be used with a workspace-level provider; " +
+			"it is only supported when the provider is configured at the account level"))
+	}
```
tanmay-db: Shouldn't the validation be part of getting workspace client for unified provider later? is this needed as part of configuring client?

common/unified_provider.go
Divyansh-db: > Would current users not using provider_config see any diff when they upgrade their terraform version during terraform plan?

It does not show in plan as it is suppressed by our customised diff for this field.

make diff-schema return "no changes to schema"

common/client.go
Divyansh-db: done

internal/providers/sdkv2/sdkv2.go
Divyansh-db: why so? If we inforce this at schema level itself, then we don't even allow the provider to be made when workspace_id is defined at provider level for workspace host?

common/unified_provider.go
Divyansh-db: This check is necessary to see if workspace_id is defined at the provider level. Suppose a user upgrades TF, shifts to Account/Unified host with workspace_id at provider level, for the first plan/apply, they should get a databricks client resolved from the workspace_id. But if this check does not exist, then it will simply return c.

Divyansh-db: The current migration (Suppose a user upgrades TF, shifts to Account/Unified host with workspace_id at resource level - provider_config, for the first plan/apply) using provider_config might have issues as it will aslo return 'c' instead of resolving from workspace_id defined in provider_config

internal/providers/sdkv2/sdkv2.go
Divyansh-db: I have moved this function to lazily cache the workspace id from a workspace host and have a hard error if it fails. Have modified the testing framework to set a default cached workspace_id so that the existing tests do not break due to an extra API call.

common/unified_provider.go
```diff
@@ -74,21 +78,85 @@ func NamespaceCustomizeSchemaMap(m map[string]*schema.Schema) map[string]*schema
...
+// workspace_id fallback) and triggers ForceNew when the effective
+// workspace changes.
+//
+// With provider_config marked as Optional+Computed, Terraform preserves the state
+// value when the config doesn't specify provider_config. This means d.GetChange()
+// shows no change in that case. We use GetRawConfigAt to inspect the actual user
+// config and determine the true effective new workspace ID.
+func namespaceForceNew(ctx context.Context, d *schema.ResourceDiff, c *DatabricksClient) error {
+	// Dual resources operating at account level have no workspace to track.
+	if d.Get("api") != nil && IsAccountLevelFromDiff(d, c) {
```
tanmay-db, line 107: Are we using this to check if a resource is a dual resource?

common/unified_provider.go
```diff
@@ -74,21 +78,85 @@ func NamespaceCustomizeSchemaMap(m map[string]*schema.Schema) map[string]*schema
...
+	if configHasProviderConfig {
+		// Config explicitly sets provider_config.workspace_id
+		newEffective = configWsID
+	} else {
+		// Config does not have provider_config; use workspace_id
+		newEffective = c.Config.WorkspaceID
+	}
+	// Lazy resolution from workspace host: if newEffective is still empty and
+	// there's an old value in state to compare against, resolve the workspace ID
+	// via the cached workspace ID or SCIM /Me API call. Errors are swallowed so
```
tanmay-db: Should we swallow the error? what will happens if we don't?

access/sql_permissions_provider_config_test.go
```diff
@@ -29,14 +29,16 @@ func TestAccSqlPermissions_ProviderConfig_Invalid(t *testing.T) {
 	})
 }
 
-func TestAccSqlPermissions_ProviderConfig_Required(t *testing.T) {
+// TestAccSqlPermissions_ProviderConfig_EmptyBlock verifies that an empty
+// provider_config {} block is valid (workspace_id is Optional+Computed).
+func TestAccSqlPermissions_ProviderConfig_EmptyBlock(t *testing.T) {
```
tanmay-db: These tests won't serve a purpose given we are making workspace_id as optional + computed, and we already have other tests above regarding provider_config. So it would be better to remove these tests with empty block

catalog/catalog_test.go
```diff
@@ -324,7 +313,7 @@ func TestAccCatalog_ProviderConfig_Match(t *testing.T) {
 		`, workspaceIDStr)),
 		ConfigPlanChecks: resource.ConfigPlanChecks{
 			PreApply: []plancheck.PlanCheck{
-				plancheck.ExpectResourceAction("databricks_catalog.this", plancheck.ResourceActionUpdate),
+				plancheck.ExpectResourceAction("databricks_catalog.this", plancheck.ResourceActionNoop),
```
tanmay-db, line 316: What will the terraform plan show in this step after above?

catalog/catalog_test.go
```diff
@@ -354,7 +343,7 @@ func TestAccCatalog_ProviderConfig_Recreate(t *testing.T) {
 		`),
 		ConfigPlanChecks: resource.ConfigPlanChecks{
 			PostApplyPreRefresh: []plancheck.PlanCheck{
-				plancheck.ExpectResourceAction("databricks_catalog.this", plancheck.ResourceActionUpdate),
+				plancheck.ExpectResourceAction("databricks_catalog.this", plancheck.ResourceActionDestroyBeforeCreate),
```
tanmay-db, line 346: Why this instead of update?

catalog/credential_provider_config_test.go
```diff
@@ -33,14 +33,15 @@ func TestAccCredential_ProviderConfig_Invalid(t *testing.T) {
 	})
 }
 
-func TestAccCredential_ProviderConfig_Required(t *testing.T) {
+// Empty provider_config {} is valid since workspace_id is Optional+Computed.
```
tanmay-db: ditto

catalog/storage_credential_provider_config_test.go
```diff
@@ -32,14 +32,15 @@ func TestAccStorageCredential_ProviderConfig_Invalid(t *testing.T) {
 	})
 }
 
-func TestAccStorageCredential_ProviderConfig_Required(t *testing.T) {
+// Empty provider_config {} is valid since workspace_id is Optional+Computed.
```
tanmay-db: same + for other tests

common/client.go
```diff
@@ -522,11 +531,28 @@ func GetApiLevel(d *schema.ResourceData) string {
...
+// or empty string if not set. This mirrors GetApiLevel but works with ResourceDiff.
+func GetApiLevelFromDiff(d *schema.ResourceDiff) string {
+	if v, ok := d.GetOk("api"); ok {
+		level := v.(string)
+		if level == ApiLevelAccount || level == ApiLevelWorkspace {
+			return level
+		}
+	}
+	return ""
+}
```
tanmay-db, line 545: Which value in the diff are we using here? If it's new let's add a comment for that + also let's add unit tests for this.

common/resource.go
```diff
@@ -6,13 +6,90 @@ import (
...
+// new one and trigger ForceNew when they differ. If this hook resolved the "new
+// effective" workspace ID and wrote it during refresh, it would overwrite the old
+// value before CustomizeDiff runs, making workspace-change detection impossible.
+//
+// Therefore this hook only resolves from provider-level sources (workspace_id,
+// host) on the first time — when no workspace ID exists in state yet (after Create).
+// On subsequent reads, it preserves whatever is already in state.
+func populateProviderConfigInState(ctx context.Context, d *schema.ResourceData, c *DatabricksClient) error {
+	// Dual resources operating at account level have no workspace to track.
+	if d.Get("api") != nil && IsAccountLevel(d, c) {
```
tanmay-db, line 49: we are using this to check if a resource is a dual resource?

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,686 @@
+package acceptance
```
tanmay-db, line 1: workspace_id_sdkv2_gosdk_acc_test and workspace_id_sdkv2_http_acc_test are nearly identical, is it possible to reduce redundancy by abstracting common logic?

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
```diff
@@ -0,0 +1,686 @@
...
+
+import (
```
tanmay-db, line 3: Can we also add a test which makes sure the scim/me API call is not made again to store workspace_id once it's cached?

common/unified_provider.go
```diff
@@ -185,13 +277,24 @@ func (c *DatabricksClient) getDatabricksClientForUnifiedProvider(ctx context.Con
 		return nil, err
 	}
 
+	// newDatabricksClient wraps a cached client.DatabricksClient and propagates
+	// any already-resolved workspace client for this workspace ID so that
+	// WorkspaceClient() returns it without recreating.
+	newDatabricksClient := func(dc *client.DatabricksClient) *DatabricksClient {
```
tanmay-db, line 295: why do we need to create a new client here?

common/unified_provider.go
Divyansh-db: yes - dual resources and data sources

catalog/catalog_test.go
Divyansh-db: no changes detected

catalog/catalog_test.go
Divyansh-db: because workspace id for the resource changes in this step, so the customised schema will flag this as recreate operation

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
Divyansh-db: done

internal/acceptance/workspace_id_sdkv2_gosdk_acc_test.go
Divyansh-db: leaving it as it for readibility and consistency with other acceptance files of pf and migrated resources

common/client.go
Divyansh-db: the new one

common/unified_provider.go
Divyansh-db: We were doing it before as well. Just abstracted the logic in a function and added caching.

common/unified_provider.go
Divyansh-db: made the comment more explicit

PR #5408, opened 2026-02-18
Mark `effective_enable_file_events` as read-only in `databricks_external_location`

elenathfgs:
## Changes

The UC API now returns a new `effective_enable_file_events` output-only field on external locations (following [AIP-129](https://google.aip.dev/129) for optional fields with server defaults). This field reflects the actual file events provisioning state, including when server-side defaults are applied.

Without marking it as `SetReadOnly()` (i.e., `Computed: true`), Terraform detects drift on every `plan` because:
- The API response includes `effective_enable_file_events` (e.g., `true`)
- The user's HCL config does not specify it
- Terraform sees a diff between state and config

This PR adds `"effective_enable_file_events"` to the existing read-only fields list alongside `created_at`, `created_by`, `credential_id`, etc.

## Tests

- [x] `go test ./catalog/ -run TestExternalLocation` run locally (all 19 tests pass)
- [x] using Go SDK
- [x] has entry in `NEXT_CHANGELOG.md` file

PR #5507, opened 2026-03-23
Mark `effective_file_event_queue` as read-only in `databricks_external_location`

elenathfgs:
## Changes

Mark the new `effective_file_event_queue` field as read-only in the `databricks_external_location` resource to prevent Terraform drift.

### Background

The Unity Catalog backend now returns an `effective_file_event_queue` field on `ExternalLocationInfo` (an `OUTPUT_ONLY` field). This is the queue-level counterpart to the existing `effective_enable_file_events` field (handled in #5408). Without marking it read-only, Terraform would detect a diff between the API response (which includes the field) and the user's HCL config (which never specifies it).

### Changes

1. **`catalog/resource_external_location.go`** — Added `SetReadOnly()` for `"effective_file_event_queue"`
2. **`catalog/resource_external_location_test.go`** — Added `TestCreateExternalLocationWithEffectiveFileEventQueue` that verifies the API returning a server-populated `effective_file_event_queue` with a managed SQS queue doesn't cause drift
3. **`NEXT_CHANGELOG.md`** — Added bug fix entry

### Note

This PR depends on a Go SDK update that picks up the new `EffectiveFileEventQueue` field from the Unity Catalog API spec (universe PR already merged). Once the SDK is bumped, the test will compile and pass.

### Tests

- [x] `TestCreateExternalLocationWithEffectiveFileEventQueue` — mirrors the existing `TestCreateExternalLocationWithEffectiveEnableFileEvents` pattern

### DIFF
--- a/catalog/resource_external_location.go
+++ b/catalog/resource_external_location.go
@@ -36,9 +36,10 @@ func ResourceExternalLocation() common.Resource {
 			common.CustomizeSchemaPath(m, "isolation_mode").SetComputed()
 			common.CustomizeSchemaPath(m, "owner").SetComputed()
 			common.CustomizeSchemaPath(m, "metastore_id").SetComputed()
-			for _, key := range []string{"created_at", "created_by", "credential_id", "updated_at", "updated_by", "browse_only", "effective_enable_file_events", "effective_file_event_queue"} {
+			for _, key := range []string{"created_at", "created_by", "credential_id", "updated_at", "updated_by", "browse_only", "effective_enable_file_events"} {
 				common.CustomizeSchemaPath(m, key).SetReadOnly()
 			}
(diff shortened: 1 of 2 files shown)


tanmay-db:
Hi @elenathfgs, is this ready for review? if I understand correctly this is still blocked on Go SDK bump right?

elenathfgs:
> Hi @elenathfgs, is this ready for review? if I understand correctly this is still blocked on Go SDK bump right?

@tanmay-db Yes exactly. I guess the release of our Go SDK is also delayed at this moment?

tanmay-db:
> > Hi @elenathfgs, is this ready for review? if I understand correctly this is still blocked on Go SDK bump right?
> 
> @tanmay-db Yes exactly. I guess the release of our Go SDK is also delayed at this moment?

Yes

PR #5510, opened 2026-03-24
[Fix] `lifecycle { ignore_changes }` for computed fields on `databricks_cluster`

### LINKED ISSUE #1238: [ISSUE] Resource databricks_cluster removes spark_env_vars without warning

jlieow:
Closes #1238
Supersedes #5438

## Changes

`lifecycle { ignore_changes = [...] }` does not work for several `databricks_cluster` fields (`spark_conf`, `spark_env_vars`, `custom_tags`, `use_ml_runtime`, `is_single_node`, `kind`, `single_user_name`, `driver_node_type_flexibility`, `worker_node_type_flexibility`). Values set externally -- via cluster policies, the UI, or API defaults -- are silently wiped on every `terraform apply` because the provider never writes them to state.

Root cause:
1. These fields are `Optional` but not `Computed`, so `StructToData` filters out API-returned values before they reach Terraform state.
2. A shared filter gate in `common/reflect_resource.go` skips all `Optional` fields with empty/nil values, even `Computed` ones -- causing a perpetual `(known after apply)` diff when the API returns empty maps.

Fix: mark the 9 affected fields as `Optional + Computed` and add a `!fieldSchema.Computed` guard to the shared filter gate so empty values for `Computed` fields are written to state.

---

### `common/reflect_resource.go` -- `StructToData` empty/nil filter

**Before:**
```go
if fieldSchema.Optional && isValueNilOrEmpty(valueField, fieldPath) {
    return nil
}
```

**After:**
```go
if fieldSchema.Optional && !fieldSchema.Computed && isValueNilOrEmpty(valueField, fieldPath) {
    return nil
}
```

Without this, `StructToData` skips `Optional + Computed` fields when the API returns an empty map. Adding `!fieldSchema.Computed` lets empty values through for `Computed` fields while preserving existing behavior for `Optional`-only fields.

### `clusters/resource_cluster.go` -- `ClusterSpec.CustomizeSchema`

**Before:**
```go
s.SchemaPath("spark_conf").SetCustomSuppressDiff(SparkConfDiffSuppressFunc)
```

**After:**
```go
s.SchemaPath("spark_conf").SetComputed().SetCustomSuppressDiff(SparkConfDiffSuppressFunc)
s.SchemaPath("spark_env_vars").SetComputed()
s.SchemaPath("custom_tags").SetComputed()
s.SchemaPath("use_ml_runtime").SetComputed()
s.SchemaPath("is_single_node").SetComputed()
s.SchemaPath("kind").SetComputed()
s.SchemaPath("single_user_name").SetComputed()
s.SchemaPath("driver_node_type_flexibility").SetComputed()
s.SchemaPath("worker_node_type_flexibility").SetComputed()
```

All 9 fields can be set server-side by cluster policies, the UI, or API defaults. Adding `Computed` to an `Optional` field is a [non-breaking schema change](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#optional-and-computed).

---

## Tests

- **`TestStructToData_OptionalComputedEmptyMapWrittenToState`** (`common/reflect_resource_test.go`): Verifies that an empty map for an `Optional + Computed` field is written to state (regression test for the perpetual diff). Creates a resource with `Computed: true` on a map field, calls `StructToData` with an empty map, and asserts the field is present in state.

- **`TestStructToData_OptionalOnlyEmptyMapNotWrittenToState`** (`common/reflect_resource_test.go`): Verifies that `Optional`-only (non-`Computed`) fields still filter out empty maps -- confirming the Gate 1 fix does not change behavior for existing resources.

- **`TestResourceClusterRead_ExternallySetComputedFields`** (`clusters/resource_cluster_test.go`): Simulates a Read where the API returns externally-set values for `spark_env_vars`, `custom_tags`, `spark_conf`, and `use_ml_runtime` that were not in the user's config. Asserts all values are preserved in state.

- **`TestResourceClusterUpdate_PreservesExternallySetComputedFields`** (`clusters/resource_cluster_test.go`): Simulates an Update where state contains externally-set values. Asserts the Edit API request includes those values (they are not wiped) and state is correct after update.

- [x] `make test` run locally
- [ ] using Go SDK


parthban-db (changes requested):
Hi @jlieow, thanks for the PR, but it can't be merged because it makes a behavioral change for the clusters resource.

PR #5518, opened 2026-03-24
[Internal] Pin remaining GitHub Actions

pietern:
## Summary

- Follow-up to #5508 — pins remaining action references that were missed

## Test plan

- [x] Verified all action references pass the pin scan

This pull request was AI-assisted by Isaac.


hectorcast-db:
Force merge to bypass the tests since GitHub actions don't run

PR #5523, opened 2026-03-25
[Enhancement] Implement caching for workspace permission assignments

tagirb:
In the current implementation workspace permission assignments are fetched for every [databricks_mws_permission_assignment](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/mws_permission_assignment) resource, whereas they could be fetched only once on the first resource and cached locally for subsequent use by other resources.

## Changes
This PR adds the optimization described above by adding a cache.

## Tests
- [x] `make test` run locally
- [x] using Go SDK


tagirb:
Obsoleted by https://github.com/databricks/terraform-provider-databricks/pull/5535

PR #5535, opened 2026-03-27
[Performance] Cache SCIM and permission-assignment list calls to reduce API requests during plan

tagirb:
## [Performance] Cache SCIM and permission-assignment list calls to reduce API requests during plan

For large IAM deployments, `terraform plan` was issuing one individual API call per resource Read —
e.g. 620 `GET /Users/{id}` calls for 620 `databricks_user` resources, 9,001 per-group member
lookups for 9,001 `databricks_group_member` resources, and so on. Under high concurrency this also
caused 429 rate-limit errors.

**Root cause:** Each resource's `Read` handler called the API independently with no sharing between
parallel goroutines.

**Fix:** Each affected resource type now has a shared in-memory cache populated by a single list
call. All goroutines for resources of the same type share the one result. Mutations
(Create/Update/Delete) invalidate the cache so the next Read re-fetches fresh data.

**Concurrency safety:** Caches use `sync.RWMutex` (concurrent reads on a warm cache) combined with
`golang.org/x/sync/singleflight` (deduplicating in-flight cold-cache fetches so exactly one
goroutine calls the API while all others wait and share the result).

**Per-resource changes:**

| Resource | Before | After |
|---|---|---|
| `databricks_mws_permission_assignment` (~4,400) | 1 `ListByWorkspaceId` per resource | 1 per `workspace_id` |
| `databricks_permission_assignment` | 1 `List` per resource | 1 per host |
| `databricks_group` (~1,466) | 1 `GET /Groups/{id}` per resource | 1 `GET /Groups?attributes=id,...` total |
| `databricks_user` (~620) | 1 `GET /Users/{id}` per resource | 1 `GET /Users?attributes=id,...` total |
| `databricks_group_member` (~9,001) | up to 1,466 concurrent `GET /Groups/{id}?attributes=members` → 429s | 1 `GET /Groups?attributes=id,members` total |

All caches fall back to individual reads for resources absent from the list response (e.g. created
concurrently after the cache was populated).

## Changes

Shared in-memory list caches for five IAM resource types, reducing total SCIM/IAM API calls during
a plan cycle from O(N) to O(1) per resource type. No schema, provider interface, or user-facing
behaviour changes.

## Tests

- [x] `make test` run locally
- [ ] relevant change in `docs/` folder — no user-facing schema or behaviour changes; existing resource documentation remains accurate
- [x] covered with integration tests in `internal/acceptance` — existing acceptance tests in `scim/group_test.go`, `scim/user_test.go`, `scim/group_member_test.go` cover full CRUD paths for all changed resources
- [ ] using Go SDK — not applicable; all changed resources use the SDKv2 `client.Scim()` helper, not the Go SDK IAM client
- [ ] using TF Plugin Framework — not applicable; all changed resources (`databricks_group`, `databricks_user`, `databricks_group_member`, `databricks_permission_assignment`, `databricks_mws_permission_assignment`) are SDKv2 resources

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -13,3 +13,7 @@
 ### Exporter
 
 ### Internal Changes
+
+* Use account host check instead of account ID check in `databricks_access_control_rule_set` to determine client type ([#5484](https://github.com/databricks/terraform-provider-databricks/pull/5484)).
+
+* Significantly reduced the number of SCIM and IAM API calls during `terraform plan`/`apply` for large deployments by introducing shared in-memory caches with `sync.RWMutex` and `singleflight` deduplication. Resources `databricks_group`, `databricks_user`, `databricks_group_member`, `databricks_permission_assignment`, and `databricks_mws_permission_assignment` now each issue a single list API call per plan cycle instead of one call per resource instance, eliminating redundant requests and rate-limit (429) errors.
--- a/access/resource_permission_assignment.go
+++ b/access/resource_permission_assignment.go
@@ -8,6 +8,9 @@ import (
 	"net/http"
 	"slices"
 	"strconv"
+	"sync"
+
+	"golang.org/x/sync/singleflight"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/terraform-provider-databricks/common"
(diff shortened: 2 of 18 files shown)


martinzuern:
@hectorcast-db @nfx @pietern any chance to get a look at this? It would be tremendously helpful to improve the performance of our larger permission provisioning

alexott:
## Code Review — SCIM Caching

### Critical Issues

**1. `groupCache` (used by `databricks_group_member`) has no host/context qualifier in cache keys**

`scim/resource_group_member.go:32-33` — The cache maps raw group IDs with no workspace host or API level qualifier. In contrast, `groupsListCache` and `usersListCache` correctly use `host + "|" + apiLevel` as the key. If a single Terraform config uses both account-level and workspace-level providers, overlapping group IDs would return wrong member data from cache.

The `bulkInitialized` flag (line 35) is also a single global bool with no per-context discrimination — once set for one workspace, a second workspace's groups are never bulk-fetched.

**Fix:** Key by `host|apiLevel|groupID` and move `bulkInitialized`/`bulkSg` into a per-context map, matching the pattern in the other caches.

**2. Inconsistent cache invalidation pattern in `mws/resource_mws_permission_assignment.go`**

Lines 149-155 and 187-188 use eager invalidation instead of `defer`, unlike all other resource handlers in this PR. In particular, the Delete handler invalidates the cache *before* checking the error return, so a failed delete still clears the cache. Not incorrect per se, but inconsistent and fragile.

**Fix:** Use `defer globalWorkspaceAssignmentsCache.invalidate(...)` to match the pattern elsewhere.

### Important Issues

**3. Pagination uses truncating `int32` cast**

`scim/groups.go:77` and `scim/users.go:76` — `int32(len(result)) >= page.TotalResults` casts a 64-bit int to int32. While the ceiling is unreachable in practice, the cast is unnecessary and confusing. Use `len(result) >= int(page.TotalResults)` instead.

**4. No test for cache-miss fallback path**

`resource_group.go:61-62` and `resource_user.go:62-63` have a fallback that calls `api.Read()` directly when the resource is absent from a warm cache. No test exercises the successful fallback scenario (ListAll returns empty, direct Read succeeds).

**5. Significant code duplication**

Five nearly identical cache implementations repeat the same `sync.RWMutex` + `singleflight` + double-checked locking pattern. A generic `SingleValueCache[K, V]` utility would reduce divergence risk and maintenance burden. The inconsistency in issue #2 above is a direct consequence of this duplication.

tagirb:
@alexott Thank you for the review, I updated the PR addressing your feedback.

tagirb:
@alexott any updates?

tagirb:
Here are some numbers to highlight the impact of this performance improvement.

Our Databricks Terraform Project currently contains ~16k resources.

```bash
$ terraform state list | wc -l
   16328
```

I compared terraform plan execution times using the official provider and my patched version on my work laptop, a MacBook Pro M4:

* Official provider, state refresh disabled

```bash
$ terraform plan -lock=false -refresh=false  1120.81s user 21.02s system 731% cpu 2:36.02 total
```

* Patched provider, state refresh disabled

```bash
$ terraform plan -lock=false -refresh=false  1121.46s user 21.41s system 724% cpu 2:37.76 total
```

* Official provider, state refresh enabled

```bash
$ terraform plan -lock=false -refresh=true  1321.81s user 44.83s system 116% cpu 19:28.62 total
```

* Patched provider, state refresh enabled

```bash
$ terraform plan -lock=false -refresh=true  1118.54s user 22.98s system 646% cpu 2:56.64 total
```

| Provider            | State Refresh | Wall-clock Time |
|---------------------|----------------|-----------------|
| Official provider   | disabled       | 2:36            |
| Patched provider    | disabled       | 2:38            |
| Official provider   | enabled        | 19:29           |
| Patched provider    | enabled        | 2:57            |

Using the execution time without state refresh as a baseline, the SCIM-related refresh overhead went from roughly 16–17 minutes with the official provider to around 15–20 seconds with the patched provider.
That is almost a **60x improvement**.
The improvement comes from replacing individual SCIM requests with batched requests, caching the fetched data, and safely sharing it between parallel goroutines using mutex protection. This removes a large amount of duplicated SCIM traffic during refresh operations and produces significant benefits both for Terraform users and for Databricks backend infrastructure.

This PR has been hanging about for 2 months, I hope someone finally finds time to review it again, because it brings a lot of value to both the user community and Databricks.

PR #5542, opened 2026-03-30
Workspace ID support for Apps

Divyansh-db:
## Summary

  Adds unified provider workspace_id support to the databricks_app Plugin Framework resource. This is the first PF resource to support
  provider-level workspace_id for workspace routing, enabling account-level providers to target specific workspaces.

###  Key changes:

  - Resource-level workspace_id lifecycle management (resource_app.go): Added ModifyPlan with workspaceDriftDetection (detects workspace changes
  → ForceNew) and validateWorkspaceID (validates reachability during plan). Create/Read call PopulateProviderConfigInState to resolve and persist
   the effective workspace ID.
  - Lazy workspace ID resolution: Removed eager workspace ID resolution from PF provider Configure. Validation is now handled lazily at the
  resource level during ModifyPlan, matching the SDKv2 pattern.
  - ProviderConfig schema: workspace_id defaults to Required at the shared ProviderConfig level (for future SDKv2-migrated resources). The app
  resource overrides this to Optional+Computed in its own ApplySchemaCustomizations, allowing the provider to populate it during apply when the
  user omits provider_config.
  - ProviderConfigPlanModifier (unified_provider.go): Handles provider_config lifecycle — sets plan to unknown on create (so provider can
  populate), copies prior state on update (prevents perpetual diffs).
  - PopulateProviderConfigInState (unified_provider.go): Resolves effective workspace ID at the end of Create/Read using fallback chain: state →
  Config.WorkspaceID → CurrentWorkspaceID().
  - CurrentWorkspaceID (unified_provider.go): Lazily caches the workspace ID from the X-Databricks-Org-Id header for workspace-level providers.
  - Improved error messages (client.go): getWorkspaceClientForAccountUnifiedHost now falls back to provider-level Config.WorkspaceID before
  erroring, with a clearer message directing users to set workspace_id in either provider config or resource provider_config.

##  Test plan

  - make test — 3150 unit tests pass, 0 failures
  - 22 acceptance tests (6 workspace-level, 16 account-level) covering:
    - Implicit default from provider workspace_id (no perpetual diff)
    - Explicit provider_config override (precedence over provider default)
    - Add/remove/change provider_config and provider workspace_id
    - ForceNew on workspace change, noop on same workspace
    - Migration from host-based to workspace_id-based provider config
    - Provider upgrade backward compatibility (no unexpected diffs)
    - Plan-time validation errors (PlanOnly: true confirms all errors surface during plan, not apply):
        - Invalid workspace_id, missing workspace_id, workspace_id mismatch
      - Removing workspace_id with no fallback (both provider-level and resource-level variants)
  - Exporter tests updated for workspace ID resolution in mock environments

- [x] `make test` run locally
- [x] using Go SDK

NO_CHANGELOG=true

### DIFF
--- a/internal/providers/pluginfw/tfschema/unified_provider.go
+++ b/internal/providers/pluginfw/tfschema/unified_provider.go
@@ -223,7 +223,7 @@ func (m ProviderConfigPlanModifier) MarkdownDescription(ctx context.Context) str
 
 func (m ProviderConfigPlanModifier) PlanModifyObject(ctx context.Context, req planmodifier.ObjectRequest, resp *planmodifier.ObjectResponse) {
 	// If config has a value, let it through (user explicitly set provider_config).
-	if !req.ConfigValue.IsNull() {
+	if !req.ConfigValue.IsNull() && !req.ConfigValue.IsUnknown() {
 		return
 	}
 	// Config is null (user didn't set provider_config).


internal/providers/pluginfw/products/app/resource_app.go
```diff
@@ -101,23 +113,99 @@ func (a *resourceApp) Configure(ctx context.Context, req resource.ConfigureReque
...
+	var newWsID string
+	if !cfg.ProviderConfig.IsNull() && !cfg.ProviderConfig.IsUnknown() {
+		newWsID, _ = tfschema.GetWorkspaceIDResource(ctx, cfg.ProviderConfig)
+	}
+	if newWsID == "" {
+		// Config does not have provider_config; use provider-level workspace_id.
+		newWsID = a.client.Config.WorkspaceID
+	}
+	// Fallback to cached workspace ID for workspace-level providers.
+	if newWsID == "" && a.client.Config.HostType() != config.AccountHost {
```
Divyansh-db: Removing this will lead to an extra API call.

common/unified_provider.go
```diff
@@ -61,6 +61,23 @@ func AddNamespaceInSchema(m map[string]*schema.Schema) map[string]*schema.Schema
 	return m
 }
 
+// CurrentWorkspaceID returns the workspace ID for a workspace-level provider.
+// It uses the cached value if available, otherwise makes an API call to resolve it.
+func (c *DatabricksClient) CurrentWorkspaceID(ctx context.Context) (int64, error) {
+	if c.cachedWorkspaceID != 0 {
```
tanmay-db: nit: I think it might be better to have getCachedWorkspaceID, we already have setCachedWorkspaceID

internal/providers/pluginfw/tfschema/unified_provider.go
```diff
@@ -196,3 +202,154 @@ func GetWorkspaceIDDataSource(ctx context.Context, providerConfig types.Object)
...
+	return "Handles provider_config for create and update lifecycle."
+}
+
+func (m ProviderConfigPlanModifier) MarkdownDescription(ctx context.Context) string {
+	return m.Description(ctx)
+}
+
+func (m ProviderConfigPlanModifier) PlanModifyObject(ctx context.Context, req planmodifier.ObjectRequest, resp *planmodifier.ObjectResponse) {
+	// If config has a value, let it through (user explicitly set provider_config).
+	if !req.ConfigValue.IsNull() {
```
tanmay-db: we don't need to check for providerConfig.IsUnknown()?

internal/providers/pluginfw/tfschema/unified_provider.go
```diff
@@ -196,3 +202,154 @@ func GetWorkspaceIDDataSource(ctx context.Context, providerConfig types.Object)
...
+// first time (after Create/Import) when no workspace ID is in state yet.
+// Preserving the prior state value is critical: CustomizeDiff compares the old
+// effective workspace ID against the new one to trigger ForceNew when they
+// differ. Overwriting it here would make that detection impossible.
+func PopulateProviderConfigInState(ctx context.Context, client *common.DatabricksClient,
+	providerConfig types.Object, state *tfsdk.State) diag.Diagnostics {
+	// GetWorkspaceIDResource reads from providerConfig, which during Read
+	// comes from the prior state. If the state already has a workspace ID,
+	// it is returned here and the fallbacks below are skipped — preserving
+	// the old value for CustomizeDiff.
```
tanmay-db: `CustomizeDiff` should this be ModifyPlan as it's a plugin framework resource

internal/providers/pluginfw/tfschema/unified_provider.go
```diff
@@ -196,3 +202,154 @@ func GetWorkspaceIDDataSource(ctx context.Context, providerConfig types.Object)
...
+
+// PopulateProviderConfigInState resolves the effective workspace ID and sets it
+// in the response state. Call this at the end of every Read method for
+// unified-provider Plugin Framework resources.
+//
+// During refresh reads (terraform plan), providerConfig comes from the prior
+// state (req.State). If it already contains a workspace ID, that value is
+// preserved — the fallback to workspace_id / host is only used on the
+// first time (after Create/Import) when no workspace ID is in state yet.
+// Preserving the prior state value is critical: CustomizeDiff compares the old
```
tanmay-db: Same, CustomizeDiff to ModifyPlan?

internal/providers/pluginfw/tfschema/unified_provider.go
```diff
@@ -196,3 +202,154 @@ func GetWorkspaceIDDataSource(ctx context.Context, providerConfig types.Object)
...
+		id, err := client.CurrentWorkspaceID(ctx)
+		if err != nil {
+			return diag.Diagnostics{diag.NewErrorDiagnostic(
+				"failed to resolve workspace ID for provider_config state population", err.Error())}
+		}
+		wsID = strconv.FormatInt(id, 10)
+	}
+	if wsID != "" {
+		diags.Append(state.SetAttribute(ctx, path.Root("provider_config"),
+			ProviderConfig{WorkspaceID: types.StringValue(wsID)}.ToObjectValue(ctx))...)
```
tanmay-db, line 355: This would need to be updated when new fields in ProviderConfig are added, we should add a comment here

common/unified_provider.go
Divyansh-db: This name will come from the code generator.

internal/providers/pluginfw/tfschema/unified_provider.go
Divyansh-db: done

internal/providers/pluginfw/tfschema/unified_provider.go
Divyansh-db: added a comment above the namespace definition

PR #5573, opened 2026-04-09
[BLOCKED][Internal] Add HostTypeForTerraform() to decouple host type detection from SDK

tanmay-db:
## Summary
  - Adds `DatabricksClient.HostTypeForTerraform()` method that
  replicates the Go SDK's `Config.HostType()` logic, allowing the
  Terraform provider to customize host type detection independently
  - Uses `ResolvedHostType` from SDK host metadata resolution as the
  primary signal, falling back to URL-prefix matching (`accounts.` /
  `accounts-dod.`) for account host detection
  - Removes dependency on SDK's `Experimental_IsUnifiedHost` and
  `isTesting` checks, which are marked as TODOs for removal in the SDK

  ## Tests
  - 10 unit tests covering:
    - Resolved host type precedence (AccountHost, UnifiedHost,
  WorkspaceHost)
    - URL-prefix fallback for `accounts.` and `accounts-dod.` prefixes
    - Edge cases: no scheme, empty host, localhost
    - ResolvedHostType takes precedence over URL-based detection

### DIFF
--- a/common/client.go
+++ b/common/client.go
@@ -574,6 +574,36 @@ func (c *DatabricksClient) Scim(ctx context.Context, method, path string, reques
 	}, nil, request, response, c.addApiPrefix, c.scimVisitorForLevel(apiLevel))
 }
 
+// HostTypeForTerraform returns the type of host the provider is configured for.
+// This replicates the SDK's Config.HostType() logic so that the Terraform provider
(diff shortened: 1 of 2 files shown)


hectorcast-db (commented):
@tanmay-db Do we close this?

PR #5582, opened 2026-04-13
Add provider level workspace id support for migrated resources

Divyansh-db:
## Summary

  - Migrate databricks_share, databricks_library, and databricks_quality_monitor from Namespace_SdkV2 (types.List) to Namespace (types.Object) for provider_config, unifying them with the pattern already used by databricks_app
  - Make ProviderConfig.workspace_id Optional+Computed centrally, eliminating per-resource overrides (removes the manual override from resource_app.go)
  - Skip SingleNestedAttributeBuilder in convertAttributesToBlocks() so that ConfigureAsSdkV2Compatible() no longer panics when the schema contains types.Object fields like provider_config
  - Add WorkspaceDriftDetection and ValidateWorkspaceID to all three resources' ModifyPlan, and PopulateProviderConfigInState to their Create/Read methods
  - Remove common.Namespace from the SDKv2 implementations of migrated resources (share, library, quality_monitor) to enable seamless bidirectional SDKv2↔PF migration at schema version 0 without a state upgrader
  - Add 22 acceptance tests for databricks_share covering all workspace ID routing scenarios (override precedence, drift detection, ForceNew on workspace change, error cases)
  - Added note that SDKv2 fallback doesn't support provider_config/workspace_id 

###  Details

  **Why types.Object instead of types.List?**

  provider_config with types.List + Namespace_SdkV2 was the original SDKv2-compatible approach, but it caused a perpetual diff: Terraform Core's proposedNewBlock logic for ListNestedBlock does not preserve prior state when config is absent, forcing a diff on every plan. types.Object + SingleNestedAttribute with Optional+Computed avoids this because TC's optionalValueNotComputable preserves the attribute from prior state when all children are Computed.

  **Why skip SingleNestedAttributeBuilder in convertAttributesToBlocks?**

  SingleNestedAttributeBuilder implements Blockable but its ToBlock() panics — types.Object has no SDKv2 block equivalent. The skip ensures SDKv2-migrated resources that embed Namespace (types.Object) can call ConfigureAsSdkV2Compatible() without crashing. The ConfigureProviderConfig() helper then applies the required Optional+Computed flags and ProviderConfigPlanModifier after the SdkV2-compat conversion.

  **Why no state upgrader for bidirectional migration?**

  The migrated resources never had provider_config in their original SDKv2 implementations — common.Namespace support is added per-resource, and these resources were migrated to PF before it existed. By keeping both SDKv2 and PF at schema version 0 (the Go default), the framework's passthrough path (IgnoreUndefinedAttributes: true) handles both directions transparently:
  - SDKv2→PF: Missing provider_config in state → null (Optional+Computed, valid)
  - PF→SDKv2: Extra provider_config in state → silently ignored
  - SDKv2 timeouts (e.g. quality_monitor): Also silently ignored by the passthrough

  This allows users to freely switch between SDKv2 (USE_SDK_V2_RESOURCES) and PF without state manipulation.

 ## Test plan

  - Unit tests pass
  - 22 share workspace_id acceptance tests pass (6 workspace-level ucws + 16 account-level ucacct), covering: default routing, override precedence, drift detection, ForceNew on workspace change, remove/add override, error cases
  - 6 bidirectional migration tests pass: TestUcAccShareMigrationFromSDKv2, TestUcAccShareMigrationFromPluginFramework, TestUcAccUpdateQualityMonitorTransitionFromSdkV2, TestUcAccUpdateQualityMonitorTransitionFromPluginFw, TestAccLibraryUpdateTransitionFromSdkV2, TestUcAccShareVolume
  - 14 ProviderConfig validation tests pass (Invalid, Mismatched, EmptyID, NotProvided, Match, Recreate, Remove for share; Mismatched for quality_monitor and library; data source tests)
  - 1 ProviderConfig test skipped (TestAccShare_ProviderConfig_Multiple) — HCL-level "Attribute redefined" error breaks the test framework's post-match state retrieval; this is a Terraform testing framework limitation, not a provider bug
  - Existing app acceptance tests unaffected
  - make diff-schema confirms no breaking schema changes for existing users (provider_config was already Optional for app; share/library/quality_monitor move from ListNestedBlock to SingleNestedAttribute — validated via acceptance tests)

  NO_CHANGELOG=true


catalog/resource_quality_monitor.go
```diff
@@ -35,7 +35,6 @@ func WaitForMonitor(w *databricks.WorkspaceClient, ctx context.Context, monitorN
 
 type QualityMonitorSchemaStruct struct {
```
tanmay-db, line 36: For each of the resource, could you add the path forward.

Example, for quality monitor we have v2 in plugin framework https://github.com/databricks/terraform-provider-databricks/tree/main/internal/providers/pluginfw/products/quality_monitor_v2

Also which one is used in production (SDKv2 vs pluginfw) by default.

clusters/resource_library.go
```diff
@@ -15,7 +15,6 @@ import (
 
 type LibraryResource struct {
```
tanmay-db, line 16: Just to confirm, we use the pluginfw implementation by default right? https://github.com/databricks/terraform-provider-databricks/blob/main/internal/providers/pluginfw/products/library/resource_library.go#L95

Could you mention since when we started using the pluginfw implementation and if users face the issues because we are removing the namespace from older SDKv2 implementation, what would be their path forward? for example, they shouldn't use the SDKv2 override flag.

internal/providers/pluginfw/products/sharing/resource_share_acc_test.go
```diff
@@ -630,13 +630,12 @@ func shareTemplate(provider_config string) string {
 func TestAccShare_ProviderConfig_Invalid(t *testing.T) {
 	acceptance.UnityWorkspaceLevel(t, acceptance.Step{
 		Template: preTestTemplateSchema + shareTemplate(`
-			provider_config {
+			provider_config = {
```
tanmay-db, line 633: This would be a breaking change for users right? if they are already using provider_config {}

internal/providers/pluginfw/tfschema/attribute_converter.go
```diff
@@ -11,11 +11,17 @@ type Blockable interface {
 
 // convertAttributesToBlocks converts all attributes implementing the Blockable interface to blocks, returning
 // a new NestedBlockObject with the converted attributes and the original blocks.
+// SingleNestedAttributeBuilder (types.Object) is excluded from conversion because it is used by
+// fields like provider_config that are new to the Plugin Framework and have no SDKv2 block equivalent.
```
tanmay-db, line 15: So the breaking change would happen for other such fields as well? i.e. earlier they would be block vs now attribute?

Divyansh-db: No other field is affect as only `provider_config` uses SingleNestedAttribute for migrated resources. Before this PR, if any other resource tried using SingleNestedAttribute, they would get paniced by ToBlock() implementation of SingleNestedAttributeBuilder

internal/providers/pluginfw/products/sharing/resource_share_acc_test.go
Divyansh-db: Only if they are using provider_config already

clusters/resource_library.go
Divyansh-db: The PF implementation is the default since September, 2025. The users who are overriding to the non-default SDKv2 implementation will not get the full support for unified provider. The path forward for them, if they want to use unified provider is they need to start using the default PF implementation.

catalog/resource_quality_monitor.go
Divyansh-db: PF is used as default

Divyansh-db: Added a note for all of these migrated resources

tanmay-db (approved):

PR #5598, opened 2026-04-16
Vector Search timeout fixes

janniklasrose:
## Changes

Fixes two timeout bugs in `databricks_vector_search_index`:

1. **Hardcoded 15-minute creation timeout** — increased the default to 75 minutes, matching `databricks_vector_search_endpoint`. The timeout is now also user-overridable via the standard `timeouts` block (e.g. `timeouts { create = "90m" }`).

2. **Deletion reusing the creation timeout** — `waitForVectorSearchIndexDeletion` was using `defaultIndexProvisionTimeout`. Introduced a separate `defaultIndexDeletionTimeout` (15 minutes, old value) so deletion keeps its original budget.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


vectorsearch/resource_vector_search_index.go
```diff
@@ -16,10 +16,11 @@ import (
 	"github.com/databricks/databricks-sdk-go/service/vectorsearch"
 )
 
-const defaultIndexProvisionTimeout = 15 * time.Minute
+const defaultIndexProvisionTimeout = 75 * time.Minute
+const defaultIndexDeletionTimeout = 15 * time.Minute
```
tanmay-db, line 20: @janniklasrose what is the expected timeout for deletion (API behaviour)? 

From PR description earlier it was 75, now it's 15 min, if deletion takes time, could this lead to earlier timeout?

janniklasrose: Sorry, unclear PR description - it was 15 min before (delete was using ProvisionTimeout), so now it's still 15min but has its own timeout value since Provision was bumped to 75min

tanmay-db (approved):

PR #5600, opened 2026-04-18
[Feature] Allow to preserve SDP tables on pipeline delete

### LINKED ISSUE #5577: [FEATURE] Preserve tables owned by SDP pipeline when removing a pipeline
API now allows to keep tables created by SDP pipeline when deleting a pipeline: https://learn.microsoft.com/en-us/azure/databricks/ldp/unity-catalog#delete-a-pipeline

P.S. We need to wait until this is available in the Go SDK

alexott:
## Changes

Pipelines API now allows to keep tables created by SDP pipeline when deleting a pipeline ([doc](https://learn.microsoft.com/en-us/azure/databricks/ldp/unity-catalog#delete-a-pipeline)). This PR adds a new attribute that allows that: `preserve_tables_on_delete`.

Resolves #5577

## Tests

- [x] `make test` run locally
- [x] using Go SDK


alexott:
Ok, we need to handle HMS pipelines separately, otherwise we get:

> cannot delete pipeline: Cannot set the cascade parameter to true for HMS pipelines. SDP service cannot cascade delete HMS tables on pipeline deletion. Please set the cascade parameter to false or omit to delete pipeline

PR #5602, opened 2026-04-20
[Internal] Mark effective_file_event_queue as read-only in databricks_external_location

tanmay-db:
## Changes

Based on contributions from @elenathfgs, PR: https://github.com/databricks/terraform-provider-databricks/pull/5507

Mark the new effective_file_event_queue field as read-only in the databricks_external_location resource to prevent Terraform drift.

Background
The Unity Catalog backend now returns an effective_file_event_queue field on ExternalLocationInfo (an OUTPUT_ONLY field). This is the queue-level counterpart to the existing effective_enable_file_events field (handled in https://github.com/databricks/terraform-provider-databricks/pull/5408). Without marking it read-only, Terraform would detect a diff between the API response (which includes the field) and the user's HCL config (which never specifies it).
## Tests

Unit tests

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug Fixes
 
+* Mark `effective_file_event_queue` as read-only in `databricks_external_location` to prevent Terraform drift.
 ### Documentation
 
 ### Exporter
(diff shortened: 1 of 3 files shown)


elenathfgs (approved):
LGTM! Thx!

hectorcast-db (approved):

PR #5613, opened 2026-04-21
[Internal] Improve dual resource identification

Divyansh-db:
## Summary

  Three improvements to unified provider internals: explicit dual-resource detection, dedicated client and CustomizeDiff routing for dual resources, and simplified error flow.

 1. **Explicit dual-resource detection via IsDual field**

 Previously, workspace-tracking logic identified dual resources by probing d.Get("api") != nil at runtime — this could false-match if a non-dual resource ever added a coincidental api field. Now, each dual resource explicitly declares IsDual: true on its common.Resource struct, and dual-resource handling is split into dedicated wrappers rather than guarded centrally.

  2. **DatabricksClientForDualResource for CRUD client routing**

  Introduced DatabricksClientForDualResource() — a thin wrapper that checks IsAccountLevel(d, c) and returns the
  current client for account-level operations, or delegates to DatabricksClientForUnifiedProvider() for workspace
  routing. All 18 dual resources (55 callsites) now call this instead of DatabricksClientForUnifiedProvider directly.

  This keeps DatabricksClientForUnifiedProvider clean — it no longer needs any dual-resource awareness, since only
  non-dual resources call it.

3. **CustomizeDiffDualResources short-circuits at account level**

CustomizeDiffDualResources (the CustomizeDiff entry point for dual resources) now checks IsAccountLevelFromDiff(d, c) after running ValidateApiLevelForUnifiedHost, and returns nil for account-level operations — they have no workspace to track, so NamespaceCustomizeDiff (ForceNew detection + workspace validation) is skipped.

This mirrors the DatabricksClientForDualResource pattern: the dual-resource entry point handles account-level routing internally, and NamespaceCustomizeDiff stays focused on workspace-level concerns.

  4. **Early return in namespaceForceNew error path**

  Simplified the lazy workspace ID resolution block: instead of swallowing the CurrentWorkspaceID error and falling through to a separate newEffective == "" check 10 lines later, the error case now returns the actionable error message immediately. Same behavior, fewer code paths.

##  Tests

  - make test run locally (3162 tests pass)

  NO_CHANGELOG=true


common/unified_provider.go
```diff
@@ -275,25 +261,29 @@ func (c *DatabricksClient) WorkspaceClientUnifiedProvider(ctx context.Context, d
...
-//     a. api = "account" → return current client (account-level operation).
-//     b. api = "workspace"s → resolve workspace_id, return workspace-scoped client.
-//  3. api field not in schema (pure workspace resource) → resolve workspace_id, return workspace-scoped client.
+// DatabricksClientForDualResource returns the appropriate client for a dual resource
+// (one that can operate at both account and workspace level via the "api" field).
+// When api="account", returns the current client unchanged (no workspace routing).
+// When api="workspace" or unset, delegates to DatabricksClientForUnifiedProvider
+// for workspace-scoped client resolution.
+// Dual resources should call this instead of DatabricksClientForUnifiedProvider.
+func (c *DatabricksClient) DatabricksClientForDualResource(ctx context.Context, d *schema.ResourceData) (*DatabricksClient, error) {
```
tanmay-db, line 270: In future all resources would be autogenerated + not dual so this won't be an issue for them but for maintenance of existing dual resources we would need to think how to couple dual methods like this one / CustomizeDiffDualResources / IsDual fields so we error if a non dual resource tries to use them or error if a dual resource tries to use other methods.

tanmay-db (approved):

common/unified_provider.go
Divyansh-db: As we have a fixed set of legacy handwritten resources, this should not be a major concern

PR #5630, opened 2026-04-23
Remove legacy tag-triggered release workflow

hectorcast-db:
NO_CHANGELOG=true

## Summary
The legacy tag-triggered release workflow is superseded by the secure release pipeline in
`databricks/secure-public-registry-releases-eng`, which dispatches `tagging.yml` and
`package.yml` here and handles the GitHub release (with zips + SHA256SUMS) externally.
This PR removes the obsolete tag-trigger path while preserving the `merge_group` snapshot build that runs as a build-breakage check in the merge queue.

## What changed
- `.github/workflows/release-legacy.yml`: dropped `push.tags` trigger, removed the strictly tag-gated steps (release-notes extraction and the non-snapshot `goreleaser release` run), and removed the now-redundant `if: ! startsWith(github.ref, 'refs/tags/v')` guard on the snapshot step. The `merge_group` trigger and the `goreleaser release --clean --snapshot` build-check step are preserved.

## What's preserved
- `tagging.yml` (auto-tag entry point dispatched by the secure release pipeline)
- `package.yml` (GPG-signed GoReleaser build dispatched by the secure release workflow)
- `release-legacy.yml`'s `merge_group` GoReleaser snapshot run (catches build breakage in the merge queue)

## Blast radius
Terraform Registry polls the GH Release assets. The secure release pipeline now produces
the exact asset set (`*.zip`, `*_SHA256SUMS`, `*_SHA256SUMS.sig`) that it produced before.

### DIFF
--- a/.github/workflows/release-build-check.yml
+++ b/.github/workflows/release-build-check.yml
@@ -1,4 +1,4 @@
-name: release-legacy
+name: Release Build Check
 
 on:
   merge_group:
--- a/.github/workflows/release-legacy.yml
+++ b/.github/workflows/release-legacy.yml
@@ -1,9 +1,6 @@
 name: release-legacy
 
 on:
-  push:
-    tags:
-      - "*"
   merge_group:
     types: [checks_requested]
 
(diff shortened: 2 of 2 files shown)


hectorcast-db:
Added rename commit bb4b52fe: `release-legacy.yml` -> `release-build-check.yml` with top-level `name:` updated to `Release Build Check`. Since the workflow now only runs a merge_group snapshot build (no release path), the filename and display name reflect that.

Heads-up: any branch-protection rules or required-status-check configs referencing the old check name (`release-legacy`) will need to be updated to `Release Build Check` after this merges.

mihaimitrea-db (approved):

PR #5639, opened 2026-04-23
Doc: `grant_rule` optional for `access_control_rule_set`

### LINKED ISSUE #5638: [DOC] databricks_access_control_rule_set: grant_rules is documented as required but appears optional
### Affected Resource(s)
- `databricks_access_control_rule_set`

### Summary
[The documentation currently marks `grant_rules` as required](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/access_control_rule_set#grant_rules-1:~:text=grant_rules%20%2D%20(Required)%20The%20access%20control%20rules%20to%20be%20granted%20by%20this%20rule%20set%2C%20consisting%20of%20a%20set%20of%20principals%20and%20roles%20to%20be%20granted%20to%20them.), but in practice the resource can be created/applied without any `grant_rules` blocks.

When applied this way, the provider successfully creates/updates the rule set and the target entity ends up with **no grants**.

We were using provider version 1.112.0

### Expected Behavior
The documentation should match the actual provider behavior.

Either:
1. `grant_rules` should be documented as optional, with the note that omitting it results in an empty rule set / removal of all grants, or
2. the provider should reject configurations that omit `grant_rules` if that behavior is not intended.

### Actual Behavior
A `databricks_access_control_rule_set` resource can be instantiated without `grant_rules`, and this removes all grants from the affected entity.

Because the docs say `grant_rules` is required, it is unclear whether this is a supported feature or an implementation bug.

dario6695:
## Changes
Clarified the `access_control_rule_set` documentation for `grant_rules`.

The docs previously marked `grant_rules` as required, but the provider accepts a configuration without `grant_rules`, which results in removing all grants from the affected entity. This PR updates the resource documentation to reflect the actual behavior and make the effect of omitting `grant_rules` explicit.

Resolves https://github.com/databricks/terraform-provider-databricks/issues/5638 if this is the intended behaviour.

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

NO_CHANGELOG=true

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -11,7 +11,7 @@
 * Mark `effective_file_event_queue` as read-only in `databricks_external_location` to prevent Terraform drift.
 ### Documentation
 
-In resource `access_control_rule_set` the field `grant_rule` is marked as optional to match the current behaviour.
+* In resource `access_control_rule_set` the field `grant_rule` is marked as optional to match the current behaviour.
 
 ### Exporter
 
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -11,6 +11,8 @@
 * Mark `effective_file_event_queue` as read-only in `databricks_external_location` to prevent Terraform drift.
 ### Documentation
 
+In resource `access_control_rule_set` the field `grant_rule` is marked as optional to match the current behaviour.
+
 ### Exporter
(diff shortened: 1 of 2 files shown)


dario6695:
This is relative to this issue https://github.com/databricks/terraform-provider-databricks/issues/5638 if the current one is the intended behavior

PR #5643, opened 2026-04-24
[Internal] Mark `effective_file_event_queue` as computed with diff suppression in `databricks_external_location`

tanmay-db:
## Changes
 Correctly annotate `effective_file_event_queue` as computed with diff suppression on `databricks_external_location` so the resource doesn't drift when the Unity Catalog backend populates the field.

 Uses `SetComputed().SetSuppressDiff()`. Matches the existing pattern used for `effective_predictive_optimization_flag` and `provisioning_info` on `databricks_catalog`.

## Tests

- Unit tests
- TestUcAccExternalLocation passing on the CI


elenathfgs (approved):
Thanks for taking care of this

alexott (approved):

rauchy (approved):

PR #5646, opened 2026-04-24
Add API documentation links to all resource docs

chrisst:
## Summary

- Adds an API documentation link to every resource doc page (129 files), following the pattern used by Google Cloud's Terraform provider
- Each resource now includes a callout like: `-> **API Documentation** https://docs.databricks.com/api/workspace/clusters`
- Links point to the correct workspace or account-level API docs page
- 5 resources have TODO markers where the API mapping could not be confidently determined (feature engineering, online store, RFA)

NO_CHANGELOG=true

## Test plan

- [x] ran a script to verify that all of these URLs are valid.


docs/resources/account_federation_policy.md
```diff
@@ -4,6 +4,8 @@ subcategory: "OAuth"
 # databricks_account_federation_policy Resource
 [![GA](https://img.shields.io/badge/Release_Stage-GA-green)](https://docs.databricks.com/aws/en/release-notes/release-types)
 
+-> **API Documentation** https://docs.databricks.com/api/account/accountfederationpolicy
```
tanmay-db: These docs (for autogenerated resources) would get overwritten so the changes for them would need to be done upstream.

docs/resources/access_control_rule_set.md
```diff
@@ -4,6 +4,8 @@ subcategory: "Security"
 
 # databricks_access_control_rule_set Resource
 
+-> **API Documentation** https://docs.databricks.com/api/workspace/accountaccesscontrolproxy
```
tanmay-db: This would appear with a Note banner like below for "This resource can be used..." [link](https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/access_control_rule_set) and would be the first things users would see, should we move this below introduction?

Image

tanmay-db (approved):
LGTM for manual resources

docs/resources/access_control_rule_set.md
chrisst: hmm. The Note isn't really the right style, I'll move it to just vanilla text.

docs/resources/account_federation_policy.md
chrisst: 👍🏽 Sent you the upstream PR as well.

PR #5648, opened 2026-04-24
Fix NCC transient failures during terraform apply and destroy (#5292)

### LINKED ISSUE #5292: [ISSUE] Issue with `mws/resource_mws_ncc_private_endpoint_rule.go` resource
### Context

When provisioning and tearing down AWS Network Connectivity Configurations (NCCs) via Terraform, two recurrent problems have been observed:

- **NCC private endpoint rule creation timeouts**
- **NCC deletion failures due to “still has private endpoint rules / workspaces”**

These issues occur even when Terraform dependencies and destroy ordering are correct. The root causes are how the provider currently calls the NCC APIs and handles long‑running / eventually consistent operations.

Relevant Databricks APIs:

- **Create private endpoint rule**  
  `POST /api/2.0/accounts/{account_id}/network-connectivity-configs/{ncc_id}/private-endpoint-rules`
- **Delete NCC**  
  `DELETE /api/2.0/accounts/{account_id}/network-connectivity-configs/{ncc_id}`

---

### 1.1 Customer-facing symptom

Example error (from `terraform apply` logs, module `module.cloud_resources`):

```text
Error: cannot create mws ncc private endpoint rule: Post "https://accounts.cloud.databricks.com/api/2.0/accounts/{account_id}/network-connectivity-configs/{ncc_id}/private-endpoint-rules": request timed out after 1m5s of inactivity
```

This is an **HTTP client inactivity timeout** message, not a Databricks business‑logic error.

### 1.2 Current provider implementation

`mws/resource_mws_ncc_private_endpoint_rule.go`

```go
func ResourceMwsNccPrivateEndpointRule() common.Resource {
    ...
    return common.Resource{
        Schema: s,
            var create settings.CreatePrivateEndpointRuleRequest
            common.DataToStructPointer(d, s, &create.PrivateEndpointRule)
            create.NetworkConnectivityConfigId = d.Get("network_connectivity_config_id").(string)
            acc, err := c.AccountClient()
            if err != nil {
                return err
            }
            rule, err := acc.NetworkConnectivity.CreatePrivateEndpointRule(ctx, create)
            if err != nil {
                return err
            }
            common.StructToData(rule, s, d)
            p.Pack(d)
            return nil
        },
        ...
    }
}
```

Key points:

- The provider makes **one** call to `CreatePrivateEndpointRule`.
- If the HTTP client sees no activity for ~60s, it surfaces a timeout:  
  `"request timed out after 1mXs of inactivity"`.
- The resource **does not retry**, even though the provider has an existing helper specifically for this pattern: `common.RetryOnTimeout` (see `common/retry.go`).

### 1.3 Root cause

- NCC private endpoint rule creation may legitimately take **longer than the HTTP inactivity timeout**, due to backend processing.
- The Go SDK returns a plain error string; the provider forwards it directly to Terraform.
- Without retry, any single slow call causes the entire `terraform apply` to fail, even though retrying the same call is safe and often succeeds.

### 1.4 Proposed provider fix

**Pattern:** Reuse `common.RetryOnTimeout` so that *only* the known inactivity timeout case is retried, matching patterns already used for workspace paths, notebooks, etc.

Updated `Create` implementation (local changes):

```go
    var create settings.CreatePrivateEndpointRuleRequest
    common.DataToStructPointer(d, s, &create.PrivateEndpointRule)
    create.NetworkConnectivityConfigId = d.Get("network_connectivity_config_id").(string)
    acc, err := c.AccountClient()
    if err != nil {
        return err
    }
    // Retry on “request timed out after ... of inactivity”
    rule, err := common.RetryOnTimeout(ctx, func(ctx context.Context) (*settings.NccPrivateEndpointRule, error) {
        return acc.NetworkConnectivity.CreatePrivateEndpointRule(ctx, create)
    })
    if err != nil {
        return err
    }
    common.StructToData(rule, s, d)
    p.Pack(d)
    return nil
}
```

**Why this is aligned with existing code:**

- `RetryOnTimeout` is already used in:
  - `workspace/resource_directory.go`
  - `workspace/resource_notebook.go`
  - `workspace/resource_workspace_file.go`
- It detects the same timeout text with `timeoutRegex = regexp.MustCompile("request timed out after .* of inactivity")`.
- The NCC rule create API is idempotent for the same payload, making retries safe.

---

### 2.1 Customer-facing symptom

Example destroy error (from `terraform destroy` on `databricks_mws_network_connectivity_config`):

```text
Error: cannot delete mws network connectivity config: Network Connectivity Config 4410739a-a37e-42ae-9bdd-e7ed9f87906a is unable to be deleted because it has one or more private endpoint rules.
```

A related variant also appears:

```text
... is unable to be deleted because it is attached to one or more workspaces: <workspace_id>
```

### 2.2 Terraform graph / ordering is correct

In the customer HCL (module `cloud_resources`):

- `databricks_mws_ncc_private_endpoint_rule.*` reference `databricks_mws_network_connectivity_config.ncc_new` → rules depend on NCC.
- `databricks_mws_ncc_binding` depends on a `time_sleep` which depends on the rules, enforcing the order:

**Destroy order:**

1. `databricks_mws_ncc_binding`  
2. `databricks_mws_ncc_private_endpoint_rule.*`  
3. `databricks_mws_network_connectivity_config.ncc_new`

So Terraform is deleting resources in the correct dependency order; the failing delete is due to how the backend enforces “NCC still in use,” not due to a graph bug.

### 2.3 Current provider implementation

`mws/resource_mws_network_connectivity_config.go`

```go
func ResourceMwsNetworkConnectivityConfig() common.Resource {
    ...
    p := common.NewPairSeparatedID("account_id", "network_connectivity_config_id", "/")
    return common.Resource{
        Schema: s,
        ...
            _, nccId, err := p.Unpack(d)
            if err != nil {
                return err
            }
            acc, err := c.AccountClient()
            if err != nil {
                return err
            }
            // Single delete, no retry:
            return acc.NetworkConnectivity.DeleteNetworkConnectivityConfigurationByNetworkConnectivityConfigId(ctx, nccId)
        },
    }
}
```

- A single call is made to delete the NCC.
- If Databricks still sees private endpoint rules or workspace bindings attached, the API returns an `apierr.APIError` with a message like:
  - `"has one or more private endpoint rules"` or
  - `"attached to one or more workspaces"`.
- The provider treats that as a **hard failure**, even though in practice it is often a **transient state** while previous deletes propagate.

### 2.4 Root cause

- NCC delete is correctly blocked when:
  - private endpoint rules still exist, or
  - workspace bindings still exist.
- Rule/binding deletes can succeed before the NCC’s internal “still in use” check observes their removal, due to **eventual consistency**.
- The provider does not attempt to **retry NCC delete while it’s “still in use”**, unlike other resources that already implement poll‑until‑gone behavior.

### 2.5 Proposed provider fix

**New helper** (classify retryable “NCC still in use” errors):

```go
func isNccStillInUseError(err error) bool {
    var apiErr *apierr.APIError
    if !errors.As(err, &apiErr) {
        return false
    }
    msg := strings.ToLower(apiErr.Message)
    if strings.Contains(msg, "has one or more private endpoint rules") {
        return true
    }
    if strings.Contains(msg, "attached to one or more workspaces") {
        return true
    }
    return false
}
```

**Updated Delete with `RetryContext`:**

```go
    _, nccId, err := p.Unpack(d)
    if err != nil {
        return err
    }
    acc, err := c.AccountClient()
    if err != nil {
        return err
    }
    // Retry deletion while the backend reports that the NCC is still in use
    // (attached to workspaces or has private endpoint rules). This handles
    // eventual consistency between rule/binding deletes and NCC deletion.
    const timeout = 15 * time.Minute
    return resource.RetryContext(ctx, timeout, func() *resource.RetryError {
        err := acc.NetworkConnectivity.DeleteNetworkConnectivityConfigurationByNetworkConnectivityConfigId(ctx, nccId)
        if err == nil {
            return nil
        }
        if isNccStillInUseError(err) {
            return resource.RetryableError(err)
        }
        return resource.NonRetryableError(err)
    })
}
```

**Why this matches existing provider patterns:**

- `mws/resource_mws_networks.go`:

```go
return resource.RetryContext(a.context, 60*time.Second, func() *resource.RetryError {
    network, err := a.Read(mwsAcctID, networksID)
    if apierr.IsMissing(err) {
        log.Printf("[INFO] Network %s/%s is removed.", mwsAcctID, networksID)
        return nil
    }
    if err != nil {
        return resource.NonRetryableError(err)
    }
    msg := fmt.Errorf("Network %s is not removed yet. VPC Status: %s", network.NetworkName, network.VPCStatus)
    log.Printf("[INFO] %s", msg)
    return resource.RetryableError(msg)
})
```

- `mws/resource_mws_workspaces.go` uses `RetryContext` and classifies workspace statuses as retryable vs fatal.
- The provider already uses **message pattern matching** in other places where structured error codes are missing (e.g. `RetryOnTimeout`, “Cluster X does not exist” in library status handling).

---

## 3. Net Effect for Users

With these changes:

- **Create NCC private endpoint rules**:
  - Transient HTTP inactivity timeouts (e.g. 1m5s) are automatically retried.
  - Users no longer see spurious `cannot create mws ncc private endpoint rule: ... request timed out` failures when the backend is just slow.

- **Delete NCC**:
  - NCC delete is retried while the Databricks API returns:
    - `"has one or more private endpoint rules"` or  
    - `"attached to one or more workspaces"`.
  - This absorbs normal propagation delays between rule/binding deletion and NCC deletion, without requiring users to sprinkle extra `time_sleep` or custom scripts in their Terraform.

The Terraform graph in user configuration remains simple and logically correct; the provider takes responsibility for handling timeouts and eventual consistency in a way that matches the rest of the codebase.

steven-muschler:
## Changes

Fixes #5292 — two transient failures in NCC resources that break `terraform apply` / `terraform destroy` even though the backend eventually reaches the desired state.

- **`databricks_mws_ncc_private_endpoint_rule` Create** — wrap `CreatePrivateEndpointRule` in `common.RetryOnTimeout` so the SDK's `"request timed out after X of inactivity"` error retries instead of failing the apply. The backend often completes the create successfully even when the HTTP client gives up. Same pattern already used in `workspace/resource_directory.go`, `workspace/resource_notebook.go`, `workspace/resource_workspace_file.go`.

- **`databricks_mws_network_connectivity_config` Delete** — add `isNccStillInUseError` classifier and wrap Delete in `resource.RetryContext` with a 15-minute timeout. Retries while the backend reports `has one or more private endpoint rules` or `attached to one or more workspaces` — these are propagation races between dependent deletes and the NCC's own deletion check. Non-retryable errors (e.g. permission denied) still fail fast. Same pattern already used in `mws/resource_mws_networks.go`.

No public API or schema change — existing TF configs behave identically, retries happen transparently inside the provider.

Credit: bug + fix proposed by @michael-saltzman-databricks in the issue body.

## Tests

Four new unit tests in `mws/`:

- `TestResourceNccPrivateEndpointRuleCreate_RetriesOnTimeout` — first call returns the inactivity-timeout error, second succeeds.
- `TestResourceNccDelete_RetriesOnStillInUseRules` — first delete returns "has one or more private endpoint rules", second succeeds.
- `TestResourceNccDelete_RetriesOnStillInUseWorkspaces` — first delete returns "attached to one or more workspaces", second succeeds.
- `TestResourceNccDelete_NonRetryableErrorFailsFast` — single "permission denied" response surfaces immediately without retrying.

All pre-existing `TestResourceNcc*` unit tests still pass.

**On integration tests:** intentionally omitted. The retry paths only trigger on transient backend timing (HTTP inactivity cutoffs, eventual-consistency windows on delete) that can't be reliably reproduced in a live acceptance harness. The existing `TestAccDataSourceMwsNetworkConnectivityConfigTest` still exercises the happy path end-to-end, so CI continues to verify no regression in the non-retry code path. Happy to add a happy-path `TestMwsAccNccPrivateEndpointRule…` as a follow-up if reviewers would prefer.

- [x] `make test` run locally
- [ ] relevant change in `docs/` folder — n/a, no public behavior change
- [ ] covered with integration tests in `internal/acceptance` — see note above
- [x] using Go SDK
- [ ] using TF Plugin Framework — n/a, this resource is still SDKv2

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,8 @@
 ### Bug Fixes
 
 * Mark `effective_file_event_queue` as read-only in `databricks_external_location` to prevent Terraform drift.
+* Retry `databricks_mws_ncc_private_endpoint_rule` creation on transient HTTP inactivity timeouts ([#<PR>](https://github.com/databricks/terraform-provider-databricks/pull/<PR>)). Fixes spurious "request timed out after 1m5s of inactivity" failures during `terraform apply`.
+* Retry `databricks_mws_network_connectivity_config` deletion while the backend reports that private endpoint rules or workspaces are still attached ([#<PR>](https://github.com/databricks/terraform-provider-databricks/pull/<PR>)). Absorbs eventual-consistency delays during `terraform destroy`.
 ### Documentation
 
 ### Exporter
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,8 +9,8 @@
 ### Bug Fixes
 
 * Mark `effective_file_event_queue` as read-only in `databricks_external_location` to prevent Terraform drift.
-* Retry `databricks_mws_ncc_private_endpoint_rule` creation on transient HTTP inactivity timeouts ([#<PR>](https://github.com/databricks/terraform-provider-databricks/pull/<PR>)). Fixes spurious "request timed out after 1m5s of inactivity" failures during `terraform apply`.
-* Retry `databricks_mws_network_connectivity_config` deletion while the backend reports that private endpoint rules or workspaces are still attached ([#<PR>](https://github.com/databricks/terraform-provider-databricks/pull/<PR>)). Absorbs eventual-consistency delays during `terraform destroy`.
+* Retry `databricks_mws_ncc_private_endpoint_rule` creation on transient HTTP inactivity timeouts ([#5648](https://github.com/databricks/terraform-provider-databricks/pull/5648)). Fixes spurious "request timed out after 1m5s of inactivity" failures during `terraform apply`.
+* Retry `databricks_mws_network_connectivity_config` deletion while the backend reports that private endpoint rules or workspaces are still attached ([#5648](https://github.com/databricks/terraform-provider-databricks/pull/5648)). Absorbs eventual-consistency delays during `terraform destroy`.
 ### Documentation
 
 ### Exporter
--- a/mws/resource_mws_ncc_private_endpoint_rule.go
+++ b/mws/resource_mws_ncc_private_endpoint_rule.go
@@ -51,7 +51,13 @@ func ResourceMwsNccPrivateEndpointRule() common.Resource {
 			if err != nil {
 				return err
 			}
-			rule, err := acc.NetworkConnectivity.CreatePrivateEndpointRule(ctx, create)
+			// Retry on "request timed out after ... of inactivity" — the backend
+			// can exceed the SDK's HTTP inactivity timeout on NCC private endpoint
+			// creation. Retries are safe because (resource_id, group_id, ncc_id)
+			// is the natural key and the API returns a conflict on duplicates.
+			rule, err := common.RetryOnTimeout(ctx, func(ctx context.Context) (*settings.NccPrivateEndpointRule, error) {
+				return acc.NetworkConnectivity.CreatePrivateEndpointRule(ctx, create)
+			})
 			if err != nil {
 				return err
 			}
--- a/mws/resource_mws_ncc_private_endpoint_rule_test.go
+++ b/mws/resource_mws_ncc_private_endpoint_rule_test.go
@@ -1,6 +1,7 @@
 package mws
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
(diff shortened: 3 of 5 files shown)


steven-muschler:
Closing — based on discussion with the NCC backend team, the right fix is server-side: making `CreatePrivateEndpointRule` async (returning immediately with `connection_state=CREATING`) eliminates the HTTP inactivity timeouts and retry races at the root, rather than papering over them client-side. The async-API behavior is being rolled out at the account level, which addresses the flakiness in #5292 without a provider change.

PR #5662, opened 2026-04-28
[Internal] Use correct customised diff for UC resources

Divyansh-db:
## Summary

Wire workspace_id-aware CustomizeDiff into the 17 SDKv2 UC resources in the `catalog/` package, so that changing the provider's `workspace_id` correctly routes API calls to the new workspace instead of silently pinning to the workspace where the resource was originally created.

UC resources are metastore-scoped, not workspace-scoped — the underlying object lives in a metastore that may be reachable from multiple workspaces. Before this PR, the unified-provider plumbing was wired only on a few resources (cluster, notebook, job, table, credential), so for the rest, `WorkspaceClientUnifiedProvider` would resolve `provider_config.0.workspace_id` from the prior state and route every Read/Update/Delete through the original workspace, even after the user changed the provider's `workspace_id`. The user's intent was silently ignored.

## How this fixes it

Two new helpers in `common/unified_provider.go`:

- **`NamespaceCustomizeDiffNoForceNew`** — for non-dual UC resources. Detects the workspace_id change (via raw config inspection), publishes the new effective workspace_id to the planned state via `SetNew`, and validates the new workspace is reachable. Does NOT call `ForceNew` (since the underlying metastore object isn't tied to a single workspace).
- **`CustomizeDiffDualResourcesNoForceNew`** — same for dual resources (those with an `api` field). Skips workspace tracking at account level, delegates to the non-dual variant at workspace level.

Internally, `namespaceForceNew` now takes a `forceNewOnChange bool` parameter to control whether ForceNew is triggered. The existing `NamespaceCustomizeDiff` (used by workspace-scoped resources like cluster, job, notebook) keeps its ForceNew behavior; UC resources opt into the no-ForceNew variant.

## Resources updated by category

**Group A** (already had a CustomizeDiff): `metastore_assignment`, `storage_credential`, `metastore` (all dual; swapped `CustomizeDiffDualResources` → `…NoForceNew`), and `sql_table` (non-dual; prepended a `NamespaceCustomizeDiffNoForceNew` call).

**Group B** (no CustomizeDiff): added `CustomizeDiff: common.NamespaceCustomizeDiffNoForceNew` to `catalog`, `artifact_allowlist`, `connection`, `registered_model`, `external_location`, `grants`, `schema`, `grant`, `volume`, `system_schema`.

**Group C/D** — UC resources without an `Update` function (`metastore_data_access`, `online_table`, `workspace_binding`): need a special treatment because the auto-ForceNew loop in `common/resource.go:198-216` would otherwise fail to handle a non-ForceNew change to `provider_config` (terraform errors `"doesn't support update"`). Solution: mark the nested `provider_config.workspace_id` attribute as `ForceNew = true` at schema construction time. When the user changes workspaces, the resource is destroyed via the old workspace's client and recreated via the new — consistent with the resource's existing immutable semantics (any field change already triggered Replace via the auto-ForceNew loop).

## Acceptance tests

New file `internal/acceptance/uc_resources_workspace_id_acc_test.go` with one test per category. All four pass against `aws-prod-ucacct`:

- `TestUcAccCatalog_ChangeProviderWorkspaceID_TriggersUpdateNotForceNew` — Group A representative (catalog, has Update). Expects `ResourceActionUpdate` (not Replace) on workspace switch.
- `TestUcAccSchema_ChangeProviderWorkspaceID_TriggersUpdateNotForceNew` — Group B representative. Same expectation.
- `TestUcAccStorageCredential_ChangeProviderWorkspaceID_TriggersUpdateNotForceNew` — Group A dual representative (uses `api = "workspace"`).
- `TestUcAccMetastoreDataAccess_ChangeProviderWorkspaceID_TriggersReplace` — Group C dual representative (no Update). Expects `ResourceActionDestroyBeforeCreate`.
- `TestUcAccWorkspaceBinding_ChangeProviderWorkspaceID_TriggersReplace` — Group D non-dual representative (no Update). Same expectation.

All tests use `PlanOnly: true` + `PostApplyPreRefresh` plan checks, so they verify the plan-time action without depending on whether the test's `TEST_WORKSPACE_ID_2` shares a metastore with `TEST_WORKSPACE_ID`.

## Tests

- [x] `make test` run locally (no regressions in `common/`, `catalog/`, `scim/`)
- [x] using Go SDK

NO_CHANGELOG=true


catalog/resource_schema.go
```diff
@@ -43,7 +43,8 @@ func ResourceSchema() common.Resource {
 			return s
 		})
 	return common.Resource{
-		Schema: s,
+		Schema:        s,
```
tanmay-db, line 46: we should also do this for system_schema

catalog/resource_storage_credential.go
```diff
@@ -80,7 +80,7 @@ func ResourceStorageCredential() common.Resource {
 		IsDual: true,
 		Schema: storageCredentialSchema,
 		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, c *common.DatabricksClient) error {
-			return common.CustomizeDiffDualResources(ctx, d, c)
+			return common.CustomizeDiffDualResourcesNoForceNew(ctx, d, c)
```
tanmay-db, line 83: also for resource_credential, resource_table

common/unified_provider.go
```diff
@@ -136,16 +141,19 @@ func namespaceForceNew(ctx context.Context, d *schema.ResourceDiff, c *Databrick
...
 		// When Optional+Computed preserves the old state value (config removes
-		// provider_config), the SDK sees no diff for workspace_id. ForceNew
-		// requires HasChange, so we SetNew on the top-level provider_config key
-		// to create the diff entry before calling ForceNew.
+		// provider_config), the SDK sees no diff for workspace_id. Both ForceNew
+		// (which requires HasChange) and the metastore-scoped flow (which needs
+		// the new value committed to state via Update) require us to SetNew on
+		// the top-level provider_config key to create the diff entry.
 		if !d.HasChange(workspaceIDKey) {
 			if err := d.SetNew("provider_config", []map[string]interface{}{{"workspace_id": newEffective}}); err != nil {
```
tanmay-db, line 168: This would overwrite the whole provider_config block, would be good to add a comment mentioning this in case a field is planned to be added later in the block

catalog/resource_metastore_data_access.go
```diff
@@ -75,6 +75,13 @@ var dacSchema = common.StructToSchema(StorageCredentialInfo{},
 		common.AddApiField(m)
 		common.AddNamespaceInSchema(m)
 		common.NamespaceCustomizeSchemaMap(m)
+		// metastore_data_access has no real Update API (immutable after Create).
+		// Mark provider_config.workspace_id as ForceNew so a workspace_id switch
+		// destroys and recreates the resource via the new workspace, instead of
+		// erroring at apply with "doesn't support update". ForceNew must be on
+		// the nested attribute (not the list block) so attribute changes inside
+		// the block trigger Replace.
+		m["provider_config"].Elem.(*schema.Resource).Schema["workspace_id"].ForceNew = true
```
tanmay-db: since this is used in other places, might be good to abstract this out to maybe ?

```go
func NamespaceCustomizeSchemaMapImmutable(m map[string]*schema.Schema) map[string]*schema.Schema {
        m = NamespaceCustomizeSchemaMap(m)
        elem := m["provider_config"].Elem.(*schema.Resource)
        elem.Schema["workspace_id"].ForceNew = true
        return m
  }
```

tanmay-db (approved):
Would be good to run terraform apply in the integration tests instead of plan only, we could create a workspace as part of the test

Also should do this for other UC resources, databricks_recipient, databricks_provider

PR #5663, opened 2026-04-29
Fix `databricks_cluster` failing with "unexpected state Restarting" when cluster attributes and `library` blocks are modified together

alexott:
## Changes

The Edit triggers an asynchronous restart; the resource now waits for the cluster to reach `RUNNING` before applying library changes instead of attempting to start a restarting cluster.

The issue was reported on Slack

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


Saugat168:
Do we have to update our databricks provider version to get this fix? We reported it originally via MS and have now been waiting for 2 months now and we still have the same error and always need to manually terminate the cluster each time, before a terraform apply. We are on databricks provider version 1.121.0

PR #5689, opened 2026-05-07
[Fix] Add `SkipProviderConfigStatePopulation` to opt-out account data sources from state population and deprecate provider_config

tanmay-db:
## Changes

Fix #5672: introduce a `SkipProviderConfigStatePopulation` opt-out on `common.Resource` and set it on `databricks_mws_workspaces` and `databricks_mws_credentials` so the post-Read `populateProviderConfigInState` hook is skipped for these account-only data sources, which have no workspace ID to resolve. Also marks `provider_config` and its nested `workspace_id` as deprecated (schema + docs) on both data sources and adds a NEXT_CHANGELOG entry.

## Tests

- Regression tests (fail on base branch, passes on these changes)
- Manually verified end-to-end against `accounts.cloud.databricks.com` (AWS prod): without the flag, plan fails for both data sources with the issue's exact wrapper; with the flag, both reads complete in <1s and the deprecation warning fires when users still set `provider_config`.

PR #5694, opened 2026-05-08
[Fix] Opt account-only settings out of post-Read provider_config hook

Divyansh-db:
## Summary

- Account-only settings built via `accountSetting` (currently only `databricks_disable_legacy_features_setting`) fail on plan/apply with `cannot populate provider_config for disable legacy features setting: failed to resolve workspace_id: ... Unable to load OAuth Config`. The post-Read `populateProviderConfigInState` hook tries to resolve a workspace_id against an account host that has no workspace context.
- #5689 introduced the `SkipProviderConfigStatePopulation` opt-out and applied it to `databricks_mws_workspaces` and `databricks_mws_credentials`, but missed account-level settings.
- This PR applies the opt-out at the `makeSettingResource` builder in `settings/generic_setting.go`, so any future `accountSetting`-based resource inherits it automatically. It detects `accountSettingDefinition[T]` at builder time and sets `SkipProviderConfigStatePopulation: true` on the returned `common.Resource{}`.
- Mirroring #5689, the auto-injected `provider_config` block and its nested `workspace_id` are marked deprecated for account-only settings.
- `workspaceSettingDefinition` and `accountWorkspaceSettingDefinition` paths are untouched — they still need the post-Read hook.

## Test plan

- [x] Unit tests pass: `go test ./settings/`
- [x] New regression test `TestDisableLegacyFeaturesSetting_SkipProviderConfigStatePopulation` asserts `SkipProviderConfigStatePopulation == true`, `provider_config` block still present in schema, and `provider_config.Deprecated` is set.
- [ ] Manual verification against `accounts.cloud.databricks.com`: with the fix, `terraform plan` for `databricks_disable_legacy_features_setting` no longer fails on the post-Read hook.

NO_CHANGELOG=false (entry added in `NEXT_CHANGELOG.md` under Bug Fixes).


settings/generic_setting.go
```diff
@@ -357,11 +357,37 @@ func (aw accountWorkspaceSetting[T]) GetCustomizeSchemaFunc() func(map[string]*s
 
 var _ accountWorkspaceSettingDefinition[struct{}] = accountWorkspaceSetting[struct{}]{}
 
+// deprecateProviderConfigInSchema marks the auto-injected provider_config
+// block (added by AddNamespaceInSchema) and its nested workspace_id as
+// deprecated for account-only settings. These resources have no workspace
+// context, so the field has never had a meaningful effect.
+func deprecateProviderConfigInSchema(s map[string]*schema.Schema) {
```
tanmay-db: We have this for mws, it would be better to abstract this out in a common method to reduce redundancy

settings/generic_setting.go
```diff
@@ -419,8 +445,18 @@ func makeSettingResource[T, U any](defn genericSettingDefinition[T, U]) common.R
...
+		Schema:                            resourceSchema,
+		SkipProviderConfigStatePopulation: isAccountOnly,
 		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, c *common.DatabricksClient) error {
+			// Account-only settings have no workspace context. NamespaceCustomizeDiff
+			// would call NamespaceValidateWorkspaceID, which falls back to
+			// c.Config.WorkspaceID (empty for account-level providers) and then
+			// errors out of GetWorkspaceClientForUnifiedProvider with
+			// "managing workspace-level resources requires a workspace_id".
+			// Skip workspace-tracking entirely for these resources.
+			if isAccountOnly {
```
tanmay-db, line 442: Should we keep this consistent with how it's done with CRUD? 

eg: case accountSettingDefinition[T]

settings/generic_setting.go
```diff
@@ -357,11 +357,37 @@ func (aw accountWorkspaceSetting[T]) GetCustomizeSchemaFunc() func(map[string]*s
...
 	resourceSchema := common.StructToSchema(defn.SettingStruct(),
 		defn.GetCustomizeSchemaFunc())
 	common.AddNamespaceInSchema(resourceSchema)
 	common.NamespaceCustomizeSchemaMap(resourceSchema)
+	// Account-only settings have no workspace context; the post-Read
+	// populateProviderConfigInState hook would try to resolve a workspace_id
+	// that does not exist (account host) and fail. Mark these resources to
+	// skip the hook and deprecate the auto-injected provider_config block.
+	// See https://github.com/databricks/terraform-provider-databricks/issues/5672.
+	_, isAccountOnly := defn.(accountSettingDefinition[T])
```
tanmay-db: same, do we need this here if we use switch like it's done in other places?

settings/generic_setting.go
Divyansh-db: I think its easier to just extract the account level flag and use it rather then to write switch case everywhere.

settings/generic_setting.go
Divyansh-db: done

settings/generic_setting.go
Divyansh-db: done

tanmay-db (approved):
Let's also add an unit / integration test which would have failed earlier and now would pass

I think TestDisableLegacyFeaturesSetting_SkipProviderConfigStatePopulation doesn't check the full apply / plan succeeds

Divyansh-db:
The integration tests that are failing started failiing on main as well since today (but they passed on 10th and the last commit on main is from 9th, so not related to TF changes). Seems to be a server-side issue. retrying the tests multiple times make them pass. Having internal discussion on fixing this.

Issue #5680, opened 2026-05-05
[BUG] databricks_entitlements: provider_config ignored when using account-level provider — AccountHost guard fires before DatabricksClientForUnifiedProvider


dvanwart-par:
## Summary

`databricks_entitlements` silently ignores `provider_config { workspace_id }` when `provider` is an account-level provider. The resource's `Create` function has a premature `AccountHost` guard that returns an error before `DatabricksClientForUnifiedProvider` is ever called, making `provider_config` completely non-functional for this resource.

## Error

```
Error: cannot create entitlements: entitlements can only be managed with a provider configured at the workspace-level

  with module.my_sp.databricks_entitlements.workspace_entitlements["datalake_dev"],
  on .terraform/modules/my_sp/main.tf line 77, in resource "databricks_entitlements" "workspace_entitlements":
  77: resource "databricks_entitlements" "workspace_entitlements" {
```

## Configuration to Reproduce

```hcl
provider "databricks" {
  alias      = "account"
  host       = "https://accounts.cloud.databricks.com"
  account_id = var.account_id
  client_id     = var.client_id
  client_secret = var.client_secret
}

resource "databricks_mws_permission_assignment" "workspace_assignment" {
  provider     = databricks.account
  workspace_id = 3459791227301047
  principal_id = databricks_service_principal.service_principal.id
  permissions  = ["USER"]
}

# ❌ Fails — provider_config is ignored, AccountHost guard fires first
resource "databricks_entitlements" "workspace_entitlements" {
  provider             = databricks.account
  service_principal_id = databricks_service_principal.service_principal.id

  provider_config {
    workspace_id = 3459791227301047
  }

  databricks_sql_access = true

  depends_on = [databricks_mws_permission_assignment.workspace_assignment]
}
```

## Root Cause

In `scim/resource_entitlement.go`, the `Create` function checks the host type **before** calling `DatabricksClientForUnifiedProvider`:

```go
Create: func(ctx context.Context, d *schema.ResourceData, c *common.DatabricksClient) error {
    if c.HostTypeForTerraform() == config.AccountHost {
        return fmt.Errorf("entitlements can only be managed with a provider configured at the workspace-level")
    }
    newClient, err := c.DatabricksClientForUnifiedProvider(ctx, d)  // never reached
    ...
```

[...]

Issue #5672, opened 2026-04-29
[ISSUE] Issue with `databricks_mws_workspaces` datasource "failed to resolve workspace_id"


red8888:
### Configuration

```hcl
provider "databricks" {
  alias                  = "accounts"
  host                   = "https://accounts.gcp.databricks.com"
  google_service_account = var.terraform_service_account_email
  account_id             = var.databricks_account_id
}

data "databricks_mws_workspaces" "all" {
  provider = databricks.accounts
}
```

### Expected Behavior
No error

### Actual Behavior
`│ Error: cannot populate provider_config for mws workspaces: failed to resolve workspace_id: failed to get the workspace_id: strconv.ParseInt: parsing "": invalid syntax`

### Steps to Reproduce
1. `terraform plan`

### Terraform and provider versions
Terraform v1.9.5
on darwin_arm64
+ provider registry.terraform.io/databricks/databricks v1.114.0
+ provider registry.terraform.io/hashicorp/google v7.30.0
+ provider registry.terraform.io/hashicorp/google-beta v7.30.0

### Is it a regression?
I dont know

### Debug Output
```
2026-04-29T17:20:22.206-0400 [ERROR] provider.terraform-provider-databricks_v1.114.0: Response contains error diagnostic: @module=sdk.proto diagnostic_detail="" diagnostic_summary="cannot populate provider_config for mws workspaces: failed to resolve workspace_id: failed to get the workspace_id: strconv.ParseInt: parsing \"\": invalid syntax" tf_proto_version=6.11 tf_req_id=7579c203-0c2b-b286-c62f-0b6bb9e5d533 tf_rpc=ReadDataSource @caller=/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.30.0/tfprotov6/internal/diag/diagnostics.go:58 diagnostic_severity=ERROR tf_data_source_type=databricks_mws_workspaces tf_provider_addr=registry.terraform.io/databricks/databricks timestamp=2026-04-29T17:20:22.206-0400
2026-04-29T17:20:22.206-0400 [ERROR] vertex "data.databricks_mws_workspaces.all" error: cannot populate provider_config for mws workspaces: failed to resolve workspace_id: failed to get the workspace_id: strconv.ParseInt: parsing "": invalid syntax
[...]

Issue #5668, opened 2026-04-29
[ISSUE] Issue with `databricks_git_credential` and `databricks_token` resource since 1.114.0


hanson-ma-8451:
### Configuration

```hcl
resource "databricks_mws_permission_assignment" "add_sp_to_workspace" {
  workspace_id = var.databricks_workspace_id
  principal_id = var.service_principal_databricks_id
  permissions  = ["USER"]

  provider = databricks.account_console
}

resource "databricks_token" "pat" {
  comment = "Terraform Provisioning"

  provider = databricks.azure_sp

  depends_on = [databricks_mws_permission_assignment.add_sp_to_workspace]
}

resource "databricks_git_credential" "ado" {
  depends_on = [databricks_mws_permission_assignment.add_sp_to_workspace]

  git_username          = "[REDACTED]"
  git_provider          = "[REDACTED]"
  personal_access_token = var.git_pat
  force                 = true
  provider              = databricks.azure_sp
}
```

### Expected Behavior
The plan should have gone through and not errored out (can't finish terraform plan like it used to)

### Actual Behavior
Terraform plan errors out with message: "Error: failed to validate workspace_id: failed to get the workspace_id: User not authorized." for both resources

### Terraform and provider versions
Terraform Version: `1.7.5`
Databricks Provider: `1.114.0`

### Is it a regression?
Yes, databricks provider version `1.113.0`

hanson-ma-8451:
Here is a minimal set of TF that we used to reproduce the provider issue that is less dependent on our overall terraform config. We received the same error in 1.114.0 but it works and succeeds in version 1.113.0

```hcl
locals {
  key_vault_resource_id        = "[REDACTED]"
  sp_client_id_secret_name     = "[REDACTED]"
  sp_client_secret_secret_name = "[REDACTED]"

  azure_tenant_id                  = "[REDACTED]"
  databricks_workspace_url         = "[REDACTED]"
  databricks_workspace_resource_id = "[REDACTED]"
}

terraform {
  required_version = "~> 1.7"
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~>4.0"
    }
    databricks = {
      source  = "databricks/databricks"
      version = "= 1.114"
    }
  }
}

data "azurerm_key_vault_secret" "service_principal_client_id" {
  name         = local.sp_client_id_secret_name
  key_vault_id = local.key_vault_resource_id
}

[...]

Issue #5664, opened 2026-04-29
[ISSUE] Issue with `databricks_service_principal ` resource since 1.114.0


BoEFerrariJ:
### Configuration
```hcl
data "databricks_service_principal" "spns" {
  for_each     = local.spns[var.env_short]
  display_name = each.value.entra_name
  provider     = databricks.accounts
}
```

### Expected Behavior
Data resource reads Service Principal.

### Actual Behavior
On `1.114.0` we now get this error:

```
Error: cannot populate provider_config for service principal: failed to resolve workspace_id: failed to get the workspace_id: strconv.ParseInt: parsing "": invalid syntax
```

We have not changed any code, this worked fine on `1.113.0`

### Steps to Reproduce
`terraform plan`

### Terraform and provider versions
Terraform Version `1.12.2`
Databricks Provider `1.114.0`

### Is it a regression?
Yes, tested and still working on `1.113.0`

maxnsdia:
We experience the same.
Adding a provider_config block (within the data block) with workspace ID solves it for us
```
provider_config {
    workspace_id = <workspaceid>
  }
```
This was not necessary for 1.113.0 but is now for 1.114.0.

Maybe related to
https://github.com/databricks/terraform-provider-databricks/releases
_Fail at plan time with "please set api to account or workspace" for dual workspace/account resources when the provider is configured against a unified host and the resource's api field is not set._

pvieito:
+1 Same issue here!

tanmay-db:
Hi, thanks for reporting, we are looking at the issue.

bhourigan-ei:
Same issue for our org. we are looking a workaround.

vw-rm:
The same issue with `databricks_mws_workspaces` Databricks account level `data resource` started from v1.114.0. 
This is working fine with v1.113.0.

tanmay-db:
Hi @bhourigan-ei @vw-rm @pvieito, fix for the issue is in progress, please use the version 1.113.0 to mitigate in meantime.

charlesfu4:
Our team happened to encounter it as well, specifically for oidc sp policy

tanmay-db:
Hi @charlesfu4 

> Our team happened to encounter it as well, specifically for oidc sp policy

Could you mention the sample HCL as well? Which data source is this?

charlesfu4:
> Hi [@charlesfu4](https://github.com/charlesfu4)
> 
[...]

PR #5699, opened 2026-05-10
[Fix] Defer plan-time workspace_id validation when not user-typed

Divyansh-db:
## Summary

Plan-time `workspace_id` validators (`NamespaceValidateWorkspaceID` for SDKv2, `ValidateWorkspaceID` for PF) previously fell back to the auto-resolved `Config.WorkspaceID` and made API calls against it. That caused the v1.114.0 cluster of regressions:

- **#5664** — account-host data sources fail with `strconv.ParseInt: parsing "": invalid syntax`.
- **#5668** — workspace-host providers using SPs fail with `failed to validate workspace_id: failed to get the workspace_id: User not authorized` at plan time.
- **#5672** — same root cause as #5664, surfaced in `databricks_mws_workspaces` data source.
- **Bootstrap pattern** — `provider_config.workspace_id = databricks_mws_workspaces.this.workspace_id` (cross-resource ref) breaks at plan because SDKv2 shims unknown→`""` and the validator falls back to an empty `c.Config.WorkspaceID`.

This PR makes the plan-time validators a fail-fast UX hint only when the user explicitly typed `provider_config.workspace_id` — null/unknown/absent values defer to apply. The `Config.WorkspaceID` fallback for apply-time client lookup is centralized in `GetWorkspaceClientForUnifiedProvider` so both SDKv2 and Plugin Framework callers benefit without per-resource changes.

### Changes

- `common/client.go` — `GetWorkspaceClientForUnifiedProvider` does the `Config.WorkspaceID` fallback at the top; the duplicate fallback inside `getWorkspaceClientForAccountUnifiedHost` is removed. Single source of truth for SDKv2 + PF + the 38 PF autogen resources.
- `common/unified_provider.go` — `NamespaceValidateWorkspaceID` uses `workspaceIDFromRawDiffConfig` to detect explicit user input; defers when null / unknown / absent / empty.
- `internal/providers/pluginfw/tfschema/unified_provider.go` — `ValidateWorkspaceID` reads from `req.Config` (not `resp.Plan`), walks the inner `workspace_id` attribute directly to preserve the unknown signal that `UnhandledUnknownAsEmpty: true` would otherwise collapse.
- Tests: `diffCustomizeDiff` upgraded to set cty values so `GetRawConfigAt` works; new defer-on-not-explicit unit tests on both SDKv2 and PF; 4 acceptance tests dropped `PlanOnly: true` (errors now surface at apply via the dispatcher); `TestValidateWorkspaceID_ReadsFromRespPlan` replaced with three focused PF tests; `scim/resource_entitlement_test.go` updated to expect the entitlements `Create`-method guard error (which becomes the first failure point now that the validator no longer catches it first — see follow-up #5680).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./common/... ./internal/...`
- [x] `make ws` (1110 checked, 0 failed)
- [x] `gofmt` clean
- [x] Unit tests across `common/`, `internal/providers/...`, `catalog/`, `tokens/`, `scim/`, `settings/`, `jobs/`, `workspace/`, `mws/`, `pools/`, `clusters/`, `repos/`, `sql/`, `pipelines/`, `permissions/`, `access/`, `aws/`, `storage/`, `secrets/`, `sharing/`, `serving/`, `mlflow/`, `vectorsearch/`, `dashboards/`, `policies/` — all pass
- [x] Acceptance tests against `aws-prod-acct` (deco):
  - `TestMwsAccWorkspaceID_NoDefaultNoOverride` PASS — `cannot create directory: managing workspace-level resources requires a workspace_id, ...`
  - `TestMwsAccWorkspaceIDTagPolicy_NoDefaultNoOverride` PASS — `failed to get workspace client: managing workspace-level resources requires a workspace_id, ...`
  - `TestMwsAccWorkspaceIDApp_NoDefaultNoOverride` PASS — `failed to get workspace client: managing workspace-level resources requires a workspace_id, ...`
  - `TestMwsAccWorkspaceIDHttp_NoDefaultNoOverride` PASS — relaxed regex (`cannot create notebook`); HTTP path takes a different code path because `databricks_notebook.Create` calls `DatabricksClientForUnifiedProvider` first and discards the workspace-client error
  - `TestMwsAccWorkspaceID_ImplicitFromProviderDefault` PASS — positive sanity check for the v1.114-victim case (account-level provider with provider-level `workspace_id`, resource without `provider_config`)
- [x] No schema changes (validators only — `make diff-schema` should be empty)

## Risks

- **Apply-time error UX shift for `*NoDefaultNoOverride`** — users typing nothing for `workspace_id` previously errored at plan; now they error at apply with the same message. Acceptable trade-off: the existing failure surface on plan was the cost of plan-time aggressiveness.
- **Cross-resource ref bootstrap** — `workspace_id = databricks_mws_workspaces.this.workspace_id` plans cleanly even when the workspace doesn't yet exist; apply-time client lookup is the source of truth (matches v1.113.0).
- **Workspace-host equality check on every call** — Edit 1 means workspace-host providers with `c.Config.WorkspaceID` set always pass a non-empty `workspaceID` into `getWorkspaceClientForWorkspaceConfiguredProvider`, triggering the cached equality check. Cheap (cached), and on workspace-host the value matches by construction.

## Out of scope (follow-ups)

- The redundant `Config.WorkspaceID` fallback at `common/unified_provider.go:353` inside `DatabricksClientForUnifiedProvider` — kept for now; removal requires careful tracing through `setCachedDatabricksClient`'s cache path.
- Issue #5680 — `databricks_entitlements` premature `AccountHost` guard in `Create`. The test update in this PR documents that issue as a follow-up.
- PF analog of `SkipProviderConfigStatePopulation` for any future account-only PF data source.


common/client.go
```diff
@@ -108,9 +108,23 @@ func (c *DatabricksClient) GetWorkspaceClientForUnifiedProviderWithDiagnostics(
...
+//
+// On workspace-level providers this also engages validateWorkspaceIDFromProvider
+// and catches the case where the user typed a workspace_id at provider level
+// that does not match the workspace the host actually points at — preventing
+// the resource from being created in workspace A while state records it as
+// workspace B.
 func (c *DatabricksClient) GetWorkspaceClientForUnifiedProvider(
 	ctx context.Context, workspaceID string,
 ) (*databricks.WorkspaceClient, error) {
+	if workspaceID == "" && c.DatabricksClient != nil && c.Config != nil {
```
tanmay-db, line 125: why are we checking for c.DatabricksClient != nil here?

common/unified_provider_test.go
```diff
@@ -678,14 +684,73 @@ func diffCustomizeDiff(
 	t.Helper()
 	ctx := context.Background()
 	ctx = context.WithValue(ctx, ResourceName, "test_resource")
-	var is *terraform.InstanceState
+
+	ctyVal := testResourceConfigToCtyValue(resource, rawConfig)
+	block := schema.InternalMap(resource.Schema).CoreConfigSchema()
+	rc := terraform.NewResourceConfigShimmed(ctyVal, block)
```
tanmay-db, line 690: what is this doing?

common/unified_provider_test.go
```diff
@@ -678,14 +684,73 @@ func diffCustomizeDiff(
...
+		if !ok {
+			vals[name] = cty.NullVal(attrType)
+			continue
+		}
+		vals[name] = goValueToCty(raw, attrType)
+	}
+	return cty.ObjectVal(vals)
+}
+
+func goValueToCty(v any, t cty.Type) cty.Value {
```
tanmay-db, line 721: This seems like a generic utility function, is this not present somewhere else / do we have to define it?

internal/providers/pluginfw/tfschema/unified_provider.go
```diff
@@ -317,25 +317,57 @@ func WorkspaceDriftDetection(ctx context.Context, client UnifiedProviderClient,
...
+// signal into "", which would then route into validation as if the user had
+// typed nothing.
+//
 // Call this from any PF resource's ModifyPlan that has a provider_config block.
 func ValidateWorkspaceID(ctx context.Context, client UnifiedProviderClient, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
-	var planPC types.Object
-	// Read from resp.Plan (not req.Plan) because WorkspaceDriftDetection may
-	// have updated provider_config in the plan to reflect a new workspace_id.
-	resp.Diagnostics.Append(resp.Plan.GetAttribute(ctx, path.Root("provider_config"), &planPC)...)
+	var cfgPC types.Object
```
tanmay-db, line 349: nit: better name for cfgPC

internal/providers/pluginfw/tfschema/unified_provider.go
```diff
@@ -317,25 +317,57 @@ func WorkspaceDriftDetection(ctx context.Context, client UnifiedProviderClient,
...
+	resp.Diagnostics.Append(req.Config.GetAttribute(ctx, path.Root("provider_config"), &cfgPC)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
-	workspaceID, diags := GetWorkspaceIDResource(ctx, planPC)
+	if cfgPC.IsNull() || cfgPC.IsUnknown() {
+		return
+	}
+	var ns ProviderConfig
```
tanmay-db, line 357: what does ns mean? nit: let's use a better name

exporter/exporter_test.go
```diff
@@ -3400,6 +3403,7 @@ func TestAlertsV2Export(t *testing.T) {
 		},
 	}, func(ctx context.Context, client *common.DatabricksClient) {
 		client.Config.WorkspaceID = testProviderWorkspaceID
+		client.SetCachedWorkspaceID(testProviderWorkspaceIDInt)
```
tanmay-db, line 3406: why do we need this now in exporter tests?

tanmay-db (commented):
Let's add an integration test that should have failed earlier and now will pass over the changes.

common/client.go
Divyansh-db: its for testing. Some legacy tests did not have it.

PR #5703, opened 2026-05-10
[Fix] Defer ForceNew and drift detection when workspace_id is an unknown ref

Divyansh-db:
## Summary

Stacked on top of #5699. Fixes a latent defect in `workspaceIDFromRawDiffConfig` that conflates "user did not write `provider_config`" with "user wrote `provider_config.workspace_id` as an unknown reference (not yet resolved at plan time)."

The unknown case shows up in the canonical bootstrap pattern:

```hcl
resource "databricks_mws_workspaces" "foo" { ... }

resource "databricks_token" "x" {
  provider_config {
    workspace_id = databricks_mws_workspaces.foo.workspace_id   # unknown until apply
  }
  comment = "..."
}
```

Both halves of `NamespaceCustomizeDiff` (SDKv2) and `WorkspaceDriftDetection` (PF) misinterpreted unknown as "absent" and silently substituted `c.Config.WorkspaceID`:

- **Account host** with no provider-level `workspace_id`: the fallback yields `""`, lazy resolution via `CurrentWorkspaceID` fails (no workspace context on account host), and the plan errors with `resource has provider_config.workspace_id = "..." in state, but managing workspace-level resources requires a workspace_id` — even though the user wrote a valid HCL reference.
- **Workspace host** with provider-level `workspace_id` set: silently uses the provider's value as the new effective and may trigger speculative `ForceNew` (when the unknown ref will eventually resolve to a different value) or no diff (when it would resolve to the same value).

The validator (`NamespaceValidateWorkspaceID`) accidentally did the right thing — it conflated both deferral cases into one — but the drift/force-new path didn't.

## Changes

- `common/unified_provider.go::workspaceIDFromRawDiffConfig` — now returns `(value string, explicit, known bool)`. `explicit=false` = user didn't write `provider_config`; `explicit=true, known=false` = user wrote it but value is unknown (cross-resource ref); `explicit=true, known=true` = user wrote a literal.
- `common/unified_provider.go::namespaceForceNew` — defers (returns nil) when `explicit && !known`. Terraform Core will show `(known after apply)` for `provider_config.workspace_id` in the diff; apply-time client construction handles the resolved value.
- `common/unified_provider.go::NamespaceValidateWorkspaceID` — updated to the tri-state signature; semantics unchanged (both unknown and absent already defer here).
- `internal/providers/pluginfw/tfschema/unified_provider.go::WorkspaceDriftDetection` — adds two guards mirroring the SDKv2 fix: skip when the outer `provider_config` is unknown, and walk the inner `workspace_id` attribute (without `UnhandledUnknownAsEmpty`) so an inner unknown defers too.

## Test plan

- [x] `go build ./...`
- [x] `make test` — 3199 tests pass, 0 failures.
- [x] New unit test `TestNamespaceCustomizeDiff_DefersWhenInnerUnknown` in `common/unified_provider_test.go` reproduces failure mode A (account host + unknown ref + state with `"12345"`): asserts no error and no `RequiresNew`. Without the fix this test fails with `resource has provider_config.workspace_id = "12345" in state, but managing workspace-level resources requires a workspace_id`.
- [x] New unit test `TestWorkspaceDriftDetection_DefersWhenInnerUnknown` in `internal/providers/pluginfw/tfschema/unified_provider_test.go` exercises the PF path with a fake account-host client; asserts no diagnostics, no `RequiresReplace`, no `ValidateWorkspaceAccess` call.

## Risks

- **Plan-time drift detection is weakened for the unknown-ref case.** This is correct behavior: when a value is unknown at plan, the provider has no basis for deciding whether to ForceNew. Terraform Core's standard `(known after apply)` UX handles this; the resolved value gets validated and routed at apply.
- **No state-corruption risk.** Apply-time client construction in `GetWorkspaceClientForUnifiedProvider` still runs `validateWorkspaceIDFromProvider` on workspace-host providers and `parseWorkspaceID` on account-host providers. Any mismatch between the resolved unknown and the workspace the provider can reach is caught before `d.SetId` is called.

## Out of scope

- Validator-side semantics of explicit-unknown are unchanged — `NamespaceValidateWorkspaceID` and PF `ValidateWorkspaceID` already defer for both null and unknown. The tri-state helper makes the deferral explicit instead of accidental, but the user-visible behavior of the validator is the same as #5699.

NO_CHANGELOG=true

### DIFF
--- a/common/unified_provider.go
+++ b/common/unified_provider.go
@@ -128,11 +128,23 @@ func namespaceForceNew(ctx context.Context, d *schema.ResourceDiff, c *Databrick
 	// Determine the new effective workspace ID by inspecting the raw config.
 	// With Optional+Computed, d.GetChange may return the preserved state value
 	// rather than reflecting the actual config, so we check the raw config directly.
-	configWsID, configHasProviderConfig := workspaceIDFromRawDiffConfig(d)
+	configWsID, configHasProviderConfig, configIsKnown := workspaceIDFromRawDiffConfig(d)
+
+	// Explicit-but-unknown: user wrote provider_config.workspace_id as a
+	// reference that isn't resolved yet (e.g. `workspace_id =
+	// databricks_mws_workspaces.foo.workspace_id` for a workspace being created
+	// in the same plan). Defer all force-new and drift decisions; falling back
+	// to c.Config.WorkspaceID here would silently substitute a different value
+	// than the user wrote, causing speculative ForceNew or spurious
+	// "managing workspace-level resources requires a workspace_id" errors on
+	// account-host providers. Apply will see the resolved value and act on it.
+	if configHasProviderConfig && !configIsKnown {
+		return nil
+	}
 
 	var newEffective string
 	if configHasProviderConfig {
-		// Config explicitly sets provider_config.workspace_id
+		// Config explicitly sets provider_config.workspace_id to a known value.
 		newEffective = configWsID
 	} else {
 		// Config does not have provider_config; use workspace_id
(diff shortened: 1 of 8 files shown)


tanmay-db (commented):
We should add an integration test that would have failed earlier and passes now

PR #5705, opened 2026-05-11
[Fix] Treat CLI `workspace_id = none` sentinel as no workspace_id

tanmay-db:
## Changes
The Databricks CLI writes `workspace_id = none` into ~/.databrickscfg for account-level profiles with no workspace bound. The Go SDK passes that literal through to `Config.WorkspaceID`, which downstream `parseWorkspaceID` calls then fail to parse as int64, breaking `terraform plan` for any workspace resource declared under that profile with:

>   failed to parse workspace_id, please check if the workspace_id in
>   provider_config is a valid integer: strconv.ParseInt: parsing "none":
>   invalid syntax

Normalize the sentinel once in PrepareDatabricksClient (the single configure chokepoint for both SDKv2 and Plugin Framework providers) by rewriting "none" to "" right after the SDK resolves the config. Downstream code already handles empty correctly (resource-level provider_config.workspace_id takes over, or a clear "managing workspace-level resources requires a workspace_id" error is surfaced).

## Tests

- Unit test
- Manual E2E test: Verified end-to-end with a locally-built binary against the account profile -- creating a workspace and a job in it.  

Before:
```sh
strconv.ParseInt: parsing "none"
```

After: 
```sh

managing workspace-level resources requires a workspace_id
```

Note: The after issue is because of plan time validation of workspace_id which doesn't exists as it's also being created. This is planned to be fixed separately after release of v1.115.0

NO_CHANGELOG=true


common/unified_provider_test.go
```diff
@@ -1591,6 +1591,33 @@ func TestValidateWorkspaceID(t *testing.T) {
...
 }
 
+// TestGetWorkspaceClientForUnifiedProvider_NoWorkspaceID_AccountHost is the
+// regression test for the bug where the Databricks CLI wrote
+// `workspace_id = none` into ~/.databrickscfg for account-level profiles. After
+// the normalization fix in client.PrepareDatabricksClient, "none" is rewritten
+// to "" before it reaches this code path, so the user-facing error must be the
+// "managing workspace-level resources requires a workspace_id" message — not
+// the lower-level strconv.ParseInt error from parseWorkspaceID.
+func TestGetWorkspaceClientForUnifiedProvider_NoWorkspaceID_AccountHost(t *testing.T) {
```
rauchy: I'm all for extra tests, but it feels unrelated to the PR and I'm not sure if it even adds value.

rauchy (approved):

internal/providers/client/databricks_client_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := &config.Config{
+				Host:        "https://accounts.cloud.databricks.com",
+				AccountID:   "00000000-0000-0000-0000-000000000000",
+				Token:       "test-token",
+				WorkspaceID: tc.workspaceID,
+			}
+			pc, err := PrepareDatabricksClient(context.Background(), cfg, nil)
```
simonfaltum, line 66: Could we make this test hermetic? `PrepareDatabricksClient` calls `cfg.EnsureResolved()`, and with the default SDK loaders this can read ambient `DATABRICKS_*` env vars / `~/.databrickscfg`; with the real accounts host it can also attempt `/.well-known/databricks-config`. That means the empty-workspace case can be overwritten by local config or discovery. Setting `Loaders` to a no-op and stubbing `HostMetadataResolver` (or using a local mock transport) would keep the test focused on the `none` normalization.

simonfaltum (commented):
One test isolation comment.

internal/providers/client/databricks_client_test.go
tanmay-db: Updated

simonfaltum (approved):

Divyansh-db (approved):

PR #5714, opened 2026-05-12
[Doc] Fix `azure_private_endpoint_info` syntax in `databricks_endpoint` example

chrisst:
## Changes

The HCL example for `databricks_endpoint` used block syntax for `azure_private_endpoint_info`:

```hcl
azure_private_endpoint_info {
  ...
}
```

`azure_private_endpoint_info` is declared as a `SingleNestedAttribute` in the Plugin Framework schema (`internal/service/networking_tf/model.go`), so the correct HCL is the attribute-assignment form:

```hcl
azure_private_endpoint_info = {
  ...
}
```

Reported by a customer.

## Tests

- [ ] Doc-only change; no code paths affected.

### DIFF
--- a/docs/resources/endpoint.md
+++ b/docs/resources/endpoint.md
@@ -20,7 +20,7 @@ resource "databricks_endpoint" "this" {
   parent       = "accounts/123e4567-e89b-12d3-a456-426614174000"
   display_name = "my-private-endpoint"
   region       = "westus"
-  azure_private_endpoint_info {
+  azure_private_endpoint_info = {
     private_endpoint_name          = "my-pe"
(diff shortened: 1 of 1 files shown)


docs/resources/endpoint.md
alexott, line 1: This file is [autogenerated](https://github.com/databricks/terraform-provider-databricks/blob/main/.gitattributes#L112), so it makes no sense to fix it here - it will be overwritten next time

alexott (changes requested):

chrisst:
Closing this in favor of an upstream fix. As @alexott noted, `docs/resources/endpoint.md` is autogenerated (marked `linguist-generated=true` in `.gitattributes`), so patching it here would be overwritten on the next regeneration.

PR #5725, opened 2026-05-15
Exclude entitlements from CurrentUser.Me requests

rauchy:
This PR makes the provider pass `excludedAttributes=entitlements` on every SCIM `Me` request it issues. The provider never reads `User.Entitlements` from any of those responses, but the SCIM server still computes the field by default - and computing it scales with the total number of entitlement grants in the workspace, not with what the caller actually needs. On large workspaces that's a meaningful amount of work for nothing.

## What changed

Every call to `w.CurrentUser.Me(ctx)` now passes a populated `iam.MeRequest`. Touched:

- `permissions/resource_permissions.go`
- `secrets/resource_secret_acl.go`
- `catalog/resource_sql_table.go`
- `scim/data_current_user.go`
- `exporter/context.go`
- `internal/acceptance/init.go` (integration-test bootstrap)
- `secrets/secret_acl_test.go`, `secrets/secret_scope_test.go` (integration-test bootstrap that builds raw SDK clients)
- `common/client.go` - the existing `cachedMe` wrapper now matches the new `CurrentUserInterface.Me` signature.
- `mws/resource_mws_workspaces.go` - the workspace-reachability probe doesn't go through the iam interface (uses the legacy `wsClient.Get`), so the query string is appended inline.

## Why filter, not whitelist

The provider reads various fields across call sites - `UserName`, `Id`, `ExternalId`, `Emails`, `Groups[].Display`. Using `excludedAttributes=entitlements` (blacklist the one heavy field) is more robust than enumerating an `attributes=...` whitelist: when a future caller starts reading another field, no one has to remember to update the whitelist.

## Review Guide

- **Core**: the call-site updates listed above, all mechanical.
- **Wrapper**: `common/client.go::cachedMe` swaps `iam.CurrentUserService` (deprecated) for `iam.CurrentUserInterface` and forwards the new `MeRequest` argument. No behavior change beyond signature.
- **Tests**: `permissions/resource_permissions_test.go` updates ~30 mock expectations from `Me(mock.Anything)` to `Me(mock.Anything, mock.Anything)` to match the new SDK interface. `common/client_test.go` adds a small test asserting the wrapper forwards `ExcludedAttributes` correctly.
- **MWS probe**: one-line URL change in `verifyWorkspaceReachable`.

### DIFF
--- a/scim/data_current_user.go
+++ b/scim/data_current_user.go
@@ -54,7 +54,7 @@ func DataSourceCurrentUser() common.Resource {
 			if err != nil {
 				return err
 			}
-			me, err := w.CurrentUser.Me(ctx, iam.MeRequest{ExcludedAttributes: "entitlements"})
+			me, err := w.CurrentUser.Me(ctx, iam.MeRequest{})
 			if err != nil {
 				return err
 			}
--- a/scim/data_current_user.go
+++ b/scim/data_current_user.go
@@ -54,7 +54,7 @@ func DataSourceCurrentUser() common.Resource {
 			if err != nil {
 				return err
 			}
-			me, err := w.CurrentUser.Me(ctx, iam.MeRequest{})
+			me, err := w.CurrentUser.Me(ctx, iam.MeRequest{ExcludedAttributes: "entitlements"})
 			if err != nil {
 				return err
 			}
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -13,3 +13,7 @@
 ### Exporter
 
 ### Internal Changes
+
+* Pass `excludedAttributes=entitlements` on SCIM `/Me` requests ([#5725](https://github.com/databricks/terraform-provider-databricks/pull/5725)).
+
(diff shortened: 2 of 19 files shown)


common/client.go
```diff
@@ -22,24 +22,24 @@ import (
...
 
-func newCachedMe(inner iam.CurrentUserService) *cachedMe {
+func newCachedMe(inner iam.CurrentUserInterface) *cachedMe {
 	return &cachedMe{
 		internalImpl: inner,
 	}
 }
 
-func (a *cachedMe) Me(ctx context.Context) (*iam.User, error) {
+func (a *cachedMe) Me(ctx context.Context, request iam.MeRequest) (*iam.User, error) {
```
tanmay-db, line 36: Build is failing with "undefined: iam.MeRequest"

scim/data_current_user.go
```diff
@@ -54,7 +54,7 @@ func DataSourceCurrentUser() common.Resource {
 			if err != nil {
 				return err
 			}
-			me, err := w.CurrentUser.Me(ctx, iam.MeRequest{})
```
tanmay-db, line 57: Do we want to exclude entitlements as part of this data source?

tanmay-db (approved):
Flagging 2 use cases, for others it looks like entitlements isn't required.

scim/data_current_user.go
rauchy: Good question. Should be safe because we don't even return it in the response, but I removed it to be on the safe side.

PR #5732, opened 2026-05-20
[Fix] `databricks_metastore`: send `external_access_enabled = false` on update

parthban-db:
## 🥞 Stacked PR
Use this [link](https://github.com/databricks/terraform-provider-databricks/pull/5732/files) to review incremental changes.
- [**stack/fix-access-enabled**](https://github.com/databricks/terraform-provider-databricks/pull/5732) [[Files changed](https://github.com/databricks/terraform-provider-databricks/pull/5732/files)]

---------

## Summary

Currently, setting `external_access_enabled = false` on a `databricks_metastore` was a no-op: Terraform's plan showed `true -> false`, but the PATCH request body to `/api/2.1/unity-catalog/metastores/{id}` never contained the field, so the API never applied the change.

Root cause: the Databricks Go SDK declares `ExternalAccessEnabled bool` with `omitempty`, and its custom JSON marshaller treats `false` as the empty value for booleans — so `false` is silently dropped from the request payload unless the field name is also present in `ForceSendFields`.

The fix extends `updateForceSendFieldsWorkspaceLevel` / `updateForceSendFieldsAccountLevel` in `catalog/resource_metastore.go` to also append `"ExternalAccessEnabled"` to `ForceSendFields` whenever `d.HasChange("external_access_enabled")` is true. This mirrors the existing pattern
## Test plan

Unit tests.


rauchy (approved):
LGTM

PR #5739, opened 2026-05-21
[Internal] Update comments to reflect new workspace routing header name

Divyansh-db:
## Changes

On unified Databricks hosts that serve multiple workspaces, the SDK sends a routing header so the gateway can dispatch each request to the correct workspace. The platform is consolidating workspace addressing onto a single header name, `X-Databricks-Workspace-Id`, which the gateway can populate from a broader range of workspace identifier formats. The previous header, `X-Databricks-Org-Id`, continues to be accepted by the platform for rollback safety.

The request-side rename ships independently of the response-side rename — the server still echoes the legacy `X-Databricks-Org-Id` header on `/api/2.0/preview/scim/v2/Me` responses, and `CurrentWorkspaceID` continues to read from that header.

This PR updates two test-file comments to reflect the new request-side header name and to document the asymmetric request/response migration:

- `common/unified_provider_test.go` (1 line) — comment near `tryWorkspaceClientDirect` updated to reference the new request header name.
- `internal/providers/pluginfw/tfschema/unified_provider_test.go` (4-line comment block added; mock unchanged) — explains why the mock continues to set the legacy `X-Databricks-Org-Id` response header on `/Me`.

No production code, schema, validator, or changelog changes; this PR is comments-only. The eventual switch to send `X-Databricks-Workspace-Id` on outgoing requests lands when the corresponding SDK release is consumed.

## Tests

- [x] `make test` run locally
- [x] using Go SDK

NO_CHANGELOG=true


hectorcast-db (commented):
Only comments are updated in the diff, no the code itself.

Divyansh-db:
> Only comments are updated in the diff, no the code itself.

This is mostly because all of the header handling happens in Go SDK. Commit 1 tried to replace the response header but it was not required, so now, its just comments changes and Go SDK bump required in TF for connection ID support

hectorcast-db (approved):

PR #5746, opened 2026-05-24
[Internal] Add internal/retrier package

renaudhartert-db:
## Summary

The repository has several ways to implement simple retry loops. This PR adds a new `internal/retrier` package that captures the common denominator across all of them. 

This PR already opportunistically migrates `common/retry.go` to use this new package internally. Subsequent PRs will migrate existing call sites to it, removing unnecessary middleware.

Design choices:

- **Value-aware predicate.** `Retrier[V].IsRetriable(V, error)` sees both the polled value and any error, so state-driven polling (the dominant pattern across the provider's waiters) is supported natively without sentinel-error wrapping.
- **Factory-per-Run.** `Run` takes a `func() Retrier[V]` and invokes it once per call. Each invocation gets its own retrier instance with independent backoff state, making `Run` safe to use from multiple goroutines with the same factory without locking.
- **Deterministic exponential backoff, no jitter.** The Databricks SDK already absorbs transient errors, and state-polling waiters here do not need cross-client decorrelation. The package is intended to sit on top of the SDK for longer polls (e.g. resource state transitions); jitter adds complexity without a corresponding benefit in that scenario. This is not a one-way door; it can be revisited if a use case calls for it.

## Test plan

- [x] `go test ./internal/retrier/...` passes (12 cases across BackoffPolicy initialization and exponential growth, `Run` retry/value-aware/ctx-cancellation paths, `RunErr`, `RetryIf` wiring and factory independence, `RetryIfErr` forwarding, `sleep` cancellation).
- [x] `go tool staticcheck ./internal/retrier/...` clean.
- [x] `make fmt lint ws` clean.
- [x] CI green.

### DIFF
--- a/common/retry.go
+++ b/common/retry.go
@@ -4,40 +4,58 @@ import (
 	"context"
 	"errors"
 	"regexp"
+	"time"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/logger"
-	"github.com/databricks/databricks-sdk-go/retries"
+	"github.com/databricks/terraform-provider-databricks/internal/retrier"
 )
 
 var timeoutRegex = regexp.MustCompile(`request timed out after .* of inactivity`)
 
-func RetryOnTimeout[T any](ctx context.Context, f func(context.Context) (*T, error)) (*T, error) {
-	r := retries.New[T](retries.WithRetryFunc(func(err error) bool {
-		msg := err.Error()
-		isTimeout := timeoutRegex.MatchString(msg)
-		if isTimeout {
-			logger.Debugf(ctx, "Retrying due to timeout: %s", msg)
-		}
-		return isTimeout
-	}))
-	return r.Run(ctx, func(ctx context.Context) (*T, error) {
-		return f(ctx)
-	})
+// transientErrorBackoff is the backoff used by both helpers below. Transient
+// errors typically resolve in a few seconds, so we start fast and cap quickly
+// rather than using the retrier package defaults of 10s/5m.
+var transientErrorBackoff = retrier.BackoffPolicy{
+	Initial: 1 * time.Second,
+	Maximum: 30 * time.Second,
+	Factor:  2,
 }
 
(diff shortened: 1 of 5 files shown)


common/retry.go
```diff
@@ -4,40 +4,47 @@ import (
...
-		isTimeout := timeoutRegex.MatchString(msg)
-		if isTimeout {
-			logger.Debugf(ctx, "Retrying due to timeout: %s", msg)
-		}
-		return isTimeout
-	}))
-	return r.Run(ctx, func(ctx context.Context) (*T, error) {
-		return f(ctx)
-	})
+	return retrier.Run(ctx, retryOnErr[*T](isTimeoutError), f)
```
mihaimitrea-db, line 17: [minor] The SDK implementation of the `WithRetryFunc()` creates a `RetryConfig` object whose timeout value defaults to 20 minutes. We're loosing that here, so if `ctx` does not already set a deadline we don't bound the retry.

Would using `WithTimeout(ctx, 20*time.Minute)` keep it bounded?

common/retry.go
```diff
@@ -4,40 +4,47 @@ import (
...
-		return true
-	}))
-	return r.Run(ctx, func(ctx context.Context) (*T, error) {
-		return f(ctx)
+	return retrier.Run(ctx, retryOnErr[*T](is504Error), f)
+}
+
+// retryOnErr adapts an error-only predicate to a value-aware retrier factory
+// using the transient backoff policy.
+func retryOnErr[V any](isRetriable func(error) bool) func() retrier.Retrier[V] {
```
mihaimitrea-db, line 29: [tiny nit] I find it a big confusing that in this file we have the two public `RetryOn_` functions call `RetryOnErr`. Doing a quick read I would've thought that all of them were doing a call to the retrier.

Also this function is a workaround the `RetryIfErr` func from `retrier.go` to be able to call `Run` and not `RunErr` for it's `*T` output. I'm not sure how prevalent this behaviour is, but maybe it's better for future users of `RunErr` to discard the `*T` output rather than have the ones needing it require workarounds?

mihaimitrea-db (approved):

PR #5748, opened 2026-05-26
[Fix] `databricks_external_location`: send `enable_file_events = false` on create when user asks for it

alexmos-db:
## Changes

Setting `enable_file_events = false` on a new `databricks_external_location` was a no-op: Terraform's plan showed the field being applied, but the POST body to `/api/2.1/unity-catalog/external-locations` never contained it, so the API silently fell back to its `true` default and `effective_enable_file_events` came back `true`.

Root cause: the Databricks Go SDK declares `EnableFileEvents bool` with `omitempty`, and its custom JSON marshaller treats `false` as the empty value for booleans — so `false` is dropped from the request payload unless the field name is also present in `ForceSendFields`. The Update path on this resource already handles `read_only`, `fallback`, and `enable_file_events` correctly (`catalog/resource_external_location.go:145-150`); the Create path did not.

This PR mirrors the Update path's three-field list onto Create using the existing `common.SetForceSendFields` helper. The helper uses `d.GetOkExists` + zero-value detection, so we only force-send the field when the user explicitly set it in HCL (`false` typed in config → forced; field omitted → still omitted from the wire). Same shape as the recent `databricks_metastore` fix in #5732.

## Tests

- [X] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/catalog/resource_external_location.go
+++ b/catalog/resource_external_location.go
@@ -82,6 +82,10 @@ func ResourceExternalLocation() common.Resource {
 			}
 			var createExternalLocationRequest catalog.CreateExternalLocation
 			common.DataToStructPointer(d, s, &createExternalLocationRequest)
+			// SDK marshals these bools with omitempty; force-send so an explicit
+			// false from HCL reaches the server instead of being elided.
+			common.SetForceSendFields(&createExternalLocationRequest, d,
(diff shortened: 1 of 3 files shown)


catalog/resource_external_location.go
alexmos-db, line 149: This is the update path which handles it correctly.

PR #5749, opened 2026-05-27
[Feature] Add `principal_id` support to `databricks_git_credential` for service principal credential management

rauchy:
## Changes

The Databricks Go SDK already supports `principal_id` in Git Credentials API requests, but the Terraform resource did not expose it. This meant there was no way to manage Git credentials on behalf of service principals.

This PR adds the `principal_id` argument to `databricks_git_credential` and passes it through all CRUD operations (Create, Read, Delete, and force-List). Without `principal_id` in Read/Delete/List, the API searches the calling user's credentials instead of the target service principal's, resulting in errors like:

```
databricks_git_credential.this: Creating...
╷
│ Error: cannot read git credential: Git credential with id <credential_id> could not be found.
│
│   with databricks_git_credential.this,
│   on main.tf line 20, in resource "databricks_git_credential" "this":
│   20: resource "databricks_git_credential" "this" {
```

Documentation for the `principal_id` argument is also added.

---

### Read: pass `principal_id` to Get (`repos/resource_git_credential.go:92-99`)

**Before:**
```go
resp, err := w.GitCredentials.Get(ctx, workspace.GetCredentialsRequest{CredentialId: cred_id})
```

**After:**
```go
principalId := int64(d.Get("principal_id").(int))
resp, err := w.GitCredentials.Get(ctx, workspace.GetCredentialsRequest{
    CredentialId: cred_id,
    PrincipalId:  principalId,
})
```

**Why:** The Get endpoint requires `principal_id` as a query parameter to retrieve credentials belonging to a service principal. Without it, the API searches the calling user's credentials and returns "not found".

### Delete: pass `principal_id` to Delete (`repos/resource_git_credential.go:134-141`)

**Before:**
```go
return w.GitCredentials.DeleteByCredentialId(ctx, cred_id)
```

**After:**
```go
principalId := int64(d.Get("principal_id").(int))
return w.GitCredentials.Delete(ctx, workspace.DeleteCredentialsRequest{
    CredentialId: cred_id,
    PrincipalId:  principalId,
})
```

**Why:** `DeleteByCredentialId` is a convenience wrapper that omits `principal_id`. Switching to the full `Delete` method allows passing it. Without this, `terraform destroy` fails for service principal credentials.

### Create (force flow): pass `principal_id` to List (`repos/resource_git_credential.go:63-66`)

**Before:**
```go
var listRequest workspace.ListCredentialsRequest
common.DataToStructPointer(d, s, &listRequest)
creds, err := w.GitCredentials.ListAll(ctx, listRequest)
```

**After:**
```go
principalId := int64(d.Get("principal_id").(int))
creds, err := w.GitCredentials.ListAll(ctx, workspace.ListCredentialsRequest{
    PrincipalId: principalId,
})
```

**Why:** `ListCredentialsRequest.PrincipalId` is tagged `json:"-"`, so `DataToStructPointer` (which uses JSON reflection) never populates it. When `force = true` triggers the list-then-update fallback, it lists the calling user's credentials (0 results) instead of the service principal's, causing "list of credentials is either empty or have more than one entry (0)".

### Documentation (`docs/resources/git_credential.md`)

Added `principal_id` to the Argument Reference and a new example section showing service principal credential management.

---

## Tests

**TestResourceGitCredentialCreateWithPrincipalId**: Verifies `principal_id` is sent in Create request and passed through to the subsequent Get call. Asserts `principal_id` persists in state.
**TestResourceGitCredentialDeleteWithPrincipalId**: Verifies Delete passes `principal_id` to the `DeleteCredentialsRequest`.
**TestResourceGitCredentialDelete** (updated): Updated to match new `Delete` call signature (was `DeleteByCredentialId`).

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/repos/resource_git_credential.go
+++ b/repos/resource_git_credential.go
@@ -60,9 +60,10 @@ func ResourceGitCredential() common.Resource {
 				if !d.Get("force").(bool) || !isOnlyOneGitCredentialForProviderError(err) {
 					return err
 				}
-				var listRequest workspace.ListCredentialsRequest
-				common.DataToStructPointer(d, s, &listRequest)
-				creds, err := w.GitCredentials.ListAll(ctx, listRequest)
+				principalId := int64(d.Get("principal_id").(int))
+				creds, err := w.GitCredentials.ListAll(ctx, workspace.ListCredentialsRequest{
+					PrincipalId: principalId,
+				})
 				if err != nil {
 					return err
 				}
@@ -91,7 +92,11 @@ func ResourceGitCredential() common.Resource {
 			if err != nil {
 				return err
 			}
-			resp, err := w.GitCredentials.Get(ctx, workspace.GetCredentialsRequest{CredentialId: cred_id})
+			principalId := int64(d.Get("principal_id").(int))
+			resp, err := w.GitCredentials.Get(ctx, workspace.GetCredentialsRequest{
+				CredentialId: cred_id,
+				PrincipalId:  principalId,
+			})
 			if err != nil {
 				return err
 			}
@@ -129,7 +134,11 @@ func ResourceGitCredential() common.Resource {
 			if err != nil {
 				return err
 			}
-			return w.GitCredentials.DeleteByCredentialId(ctx, cred_id)
+			principalId := int64(d.Get("principal_id").(int))
+			return w.GitCredentials.Delete(ctx, workspace.DeleteCredentialsRequest{
+				CredentialId: cred_id,
+				PrincipalId:  principalId,
+			})
 		},
 	}
 }
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Documentation
 
+* Documented `principal_id` argument for `databricks_git_credential` resource, allowing management of Git credentials on behalf of service principals.
+
 ### Exporter
 
 ### Internal Changes
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,13 +6,21 @@
 
 ### New Features and Improvements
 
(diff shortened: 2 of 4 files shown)


repos/resource_git_credential.go
```diff
@@ -60,9 +60,10 @@ func ResourceGitCredential() common.Resource {
 				if !d.Get("force").(bool) || !isOnlyOneGitCredentialForProviderError(err) {
 					return err
 				}
-				var listRequest workspace.ListCredentialsRequest
-				common.DataToStructPointer(d, s, &listRequest)
-				creds, err := w.GitCredentials.ListAll(ctx, listRequest)
+				principalId := int64(d.Get("principal_id").(int))
```
tanmay-db, line 63: do we need both int64() and .(int) ?

repos/resource_git_credential.go
```diff
@@ -129,7 +134,11 @@ func ResourceGitCredential() common.Resource {
 			if err != nil {
 				return err
 			}
-			return w.GitCredentials.DeleteByCredentialId(ctx, cred_id)
+			principalId := int64(d.Get("principal_id").(int))
+			return w.GitCredentials.Delete(ctx, workspace.DeleteCredentialsRequest{
+				CredentialId: cred_id,
+				PrincipalId:  principalId,
```
tanmay-db, line 140: Now that we are always passing principal_id, any chance this could be a behaviour change for existing users?

tanmay-db (approved):

repos/resource_git_credential.go
rauchy: When principal_id isn't set, d.Get(...) returns int(0), which the SDK's url:"principal_id,omitempty" tag drops at query-encoding time (go-querystring/query). Existing users' requests are byte-identical to before.

Issue #4082, opened 2024-10-09
[FEATURE] AI-BI Genie Terraform Support


pcsrijith:
Hi Community,

Can we have Terraform support for AI-BI Genie features for Databricks.

### Use-cases
1. Enabling Genie workspace feature at account/workspace levels.
2. Creation of Genie Workspace
3. Tagging of tables in genie workspaces
4. Giving ACLs to users/groups without removing existing roles in genie workspace.
5. Creation of sql queries inside genie workspace.
6. Creation of sql functions inside genie workspace.
7. Add/remove benchmarks in genie workspace.
8. Attaching SQL Warehouse to any new/existing genie workspace.

and any other features which I might have missed logging here.

alexott:
there is no public API for that (yet)

alexott:
Connect with your account team - they can talk with the product team. Terraform team doesn't have insights into the timelines

VOVELEE:
hey @alexott ,

I assume some of the requirements above can be met via https://docs.databricks.com/api/azure/workspace/genie.
Is my understanding correct?

P.S. I also require this in order to properly govern genie spaces on top of workspaces :)

alexott:
There is still no Create/Update API for that.

VOVELEE:
yes, indeed :)
@alexott but what about the following use case - list all genie spaces (list is possible) and assigned mandatory permissions (potentially using databricks_permissions)

alexott:
New resources will be autogenerated as soon as the API is complete, so it's better to wait. Also, I don't see permissions API for genie...

VOVELEE:
:( 
Indeed it looks like inmature implementation in its current state.
Thanks for checking

shadzic:
I am interested in this feature as well.
Doesn't this Set Object Permission in databricks REST API include genie?
https://docs.databricks.com/api/azure/workspace/permissions/set

pcsrijith:
@alexott : I see that create genie space is in Beta phase. Do you think, this can be picked up for curation? [Reference API Link](https://docs.databricks.com/api/azure/workspace/genie/createspace)

alexott:
New resources are automatically generated when the corresponding engineering team decides that the API is stable.

nijanthanvijayakumar:
Hi @alexott,

[...]

PR #5770, opened 2026-05-30
Add databricks_genie_space resource and data source

dinbab1984:
## Summary

Adds the `databricks_genie_space` Plugin Framework resource and the
`databricks_genie_spaces` list data source, plus a `genie_space_id` attribute
on `databricks_permissions` so Genie Space ACLs can be managed end-to-end.

The resource mirrors `databricks_dashboard`:
- `parent_path` is required and auto-created on first apply
- `serialized_space` is JSON-normalized to suppress whitespace and key-order diffs
- Delete is trash-aware (already-trashed space is treated as success)

The data source supports a `title_contains` filter and paginates over `ListSpaces`.

The permissions entry registers `CAN_READ` / `CAN_RUN` / `CAN_EDIT` / `CAN_MANAGE`
with `AddCurrentUserAsManage` customizers on update and delete to prevent lockout.

Tags can be attached via the existing `databricks_workspace_entity_tag_assignment`
resource with `entity_type = "geniespaces"` (no code change needed; the SDK already
accepts that value).

## Test plan

- [x] Unit tests: schema, plan modifier (JSON normalization), pagination, filter, import parsing, regression test for post-update state composition (`TestPostUpdateState_TrustsPlanForUserControlledFieldsWhenGetIsStale`)
- [x] Integration test: `TestAccGenieSpacesDataSource` (data source)
- [ ] Resource acceptance test deferred to a follow-up PR. The Genie API rejects hand-written `serialized_space` blobs (it validates against an internal schema); a known-minimal seed needs to be confirmed against the live API before being committed as a test fixture.
- [x] Live verification against a real workspace:
  - Create + permissions + tags (`env=smoke`, `owner=terraform`)
  - In-place update (title change)
  - Idempotency after update (JSON normalization works)
  - `parent_path` change triggers replacement with auto-mkdir on the new folder
  - Trash-aware delete (manually trashing the space, then `terraform destroy` succeeds)
  - Data source list + case-insensitive substring filter + pagination
- [x] `make fmt lint ws test` clean
- [x] `make diff-schema` reports purely additive changes (new resource, new data source, new `genie_space_id` permission attribute)

## Notes for reviewers

- Bug found during live test: Genie's `GET` cache lags behind `UpdateSpace` and returned pre-update `title`, triggering "Provider produced inconsistent result after apply". Fixed by trusting plan values for user-controlled fields in `Update` (matches the SDKv2 `databricks_dashboard` pattern). See `postUpdateState` + the regression test.
- The `databricks_workspace_entity_tag_assignment` resource has a separate pre-existing race condition ("Concurrent modification detected") when applying multiple tags to the same entity in parallel. Not blocking; affects all entity types. Should be tracked as a separate issue.


alexott:
Thank you for your contribution, but it would be better to discuss it before implementation. The current requirement is that all new data sources and resources must be automatically generated from the API spec, and that this be driven by the corresponding product team, not the Terraform dev team.

dinbab1984:
@alexott Thank you for your feedback! Apologies for not aligning first — I followed the §"Adding a new resource" steps in CONTRIBUTING.md (write CRUD + schema based on resource_app.go, add tests, open PR) and didn't see the autogen/product-team requirement called out there. Background: I built a custom solution using APIs for a customer because DAB doesn't yet support Genie Spaces and it added a hard Terraform dependency, so contributing the same upstream seemed useful.

Happy to share the requirements + live-test notes with the Genie product team — who should I reach out to?

Close the PR, or leave it open as a reference until the autogen version lands?

pcsrijith:
Tagging issue. #4082

schwarztrinker:
I don't care how this is called Genie Space -> Genie Agent or whatever renaming Databricks is already thinking about :D Please add this to the provider!

achistef:
Hello!
Is there a plan to merge this PR?

pcsrijith:
I believe `genie_spaces` needs to be renamed with `genie_agents` which in line with current nomenclature.

PR #5780, opened 2026-06-03
[Internal] Group manual SDKv2 resources and data sources into Account, Workspace and Dual categories

tanmay-db:
## Changes

Based on https://github.com/databricks/terraform-provider-databricks/pull/5752

Move SDKv2 resource and data source registration out of sdkv2.go into a new resources.go, split into Workspace*, Account*, and Dual* maps. sdkv2.go now calls Resources() / DataSources(), which flatten the three maps via merge. This is an organizational change with no runtime effect today.

## Tests

  - **`TestApiField_RegistrationConsistency`** — every `Dual*` entry must carry the `api` schema attribute (the single signal that a resource is dual, set by `common.AddApiField`), and every `Workspace*` / `Account*` entry must not.
  - **`TestProviderConfig_AccountRegistrationConsistency`** — account resources and data sources must not carry a `provider_config` block, because they have no workspace context. The two account-only data sources that retain a *deprecated* `provider_config` block for backward compatibility (`databricks_mws_credentials`, `databricks_mws_workspaces`) are pinned as exceptions, so no new account resource can silently gain the block. Scope note: account-only *settings* (e.g. `databricks_disable_legacy_features_setting`) are registered
  through the separate `settings.AllSettingsResources()` path and are covered by the `settings` package's own tests, so they are intentionally out of scope here.
  - **`TestNoDuplicateKeys`** — no Terraform name is registered in more than one map within a namespace (resources and data sources checked separately). A single map literal made duplicate keys a compile error; splitting into three maps removed that guarantee, and this test restores it.
- Manually drift tested the map of resources to make there is sure no new additional / deletion. Might be better to have it as a CI check but that's out of scope for this PR. https://github.com/databricks/terraform-provider-databricks/pull/5784/changes#diff-1492dfe82344b8da8362a77e4041fd29c20b9595756d71cb88797a2d8acd355a

Note: The failing integration tests are unrelated, tests on main are failing and needs to be fixed first. 

NO_CHANGELOG=true


Divyansh-db (changes requested):
Flagging a subtle but real issue with this refactor — I don't think it's a no-op at runtime; it can make the outgoing `User-Agent` grow without bound, so requesting changes.

### Problem

Moving the maps to package-level `var`s means each `*schema.Resource` is created **once** (at package init) and then **shared across every `DatabricksProvider()` call** — `Resources()` / `merge` (`maps.Copy`) copy the pointers, not the resources.

`DatabricksProvider()` then calls `common.AddContextToAllResources(p, "databricks")`, which **mutates each resource in place** by wrapping its CRUD funcs:

```go
r.ReadContext = addName(op(r.ReadContext)) // wrapper calls useragent.InContext(ctx, "resource", name) + "sdk"
```

Since the pointers are shared, every `DatabricksProvider()` call re-wraps the **same** objects, nesting the wrapper one layer deeper each time:

```
build 1:  ReadContext = W(original)
build 2:  ReadContext = W(W(original))
build N:  ReadContext = W(W(…W(original)…))
```

At request time each layer runs `InContext(ctx, "resource", <name>)`, so the `User-Agent` ends up with `resource/<name> sdk/sdkv2` repeated **once per build**.

This is invisible in normal CLI usage (the provider is built once per process), but anything that builds the provider repeatedly in a single process — notably the acceptance test suite, which constructs the provider once per test — accumulates these segments across the run (hundreds of repeats over a full suite). An oversized `User-Agent` can then get dropped/rejected downstream.

Quick repro — wrapping the same resource N times vs. a fresh one per build:

```
shared pointer (this PR):  … sdk/sdkv2 resource/directory   (×N)
fresh per build (today):   … sdk/sdkv2 resource/directory   (×1)
```

### Suggested fix

Keep the grouping, but make the groups **functions that return fresh maps per call** instead of package-level `var`s, so each build gets fresh `.ToResource()` objects (wrapped exactly once):

```go
// instead of:
var WorkspaceResources = map[string]*schema.Resource{
	"databricks_directory": workspace.ResourceDirectory().ToResource(),
	// …
}

// do:
func WorkspaceResources() map[string]*schema.Resource {
	return map[string]*schema.Resource{
		"databricks_directory": workspace.ResourceDirectory().ToResource(),
		// …
	}
}
```

Then `Resources()` / `DataSources()` call them (`merge(WorkspaceResources(), …)`), and the consistency tests just add `()`.

A small regression test would lock it in:

```go
r1 := DatabricksProvider().ResourcesMap["databricks_directory"]
r2 := DatabricksProvider().ResourcesMap["databricks_directory"]
require.NotSame(t, r1, r2) // shared pointer => AddContextToAllResources re-wraps it => UA grows per build
```

(A defensive de-dup is also landing in the Go SDK's user-agent builder — databricks/databricks-sdk-go#1718 — but the root cause is the shared pointers here, so worth fixing in this PR too.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

renaudhartert-db (approved):

Divyansh-db (approved):

tanmay-db (commented):
found a bug — metastores data source was in workspace datasource list instead of account, created a fix + regression test: https://github.com/databricks/terraform-provider-databricks/pull/5780/changes/0d00d2c17516f14fead90df5db90ea45fd21b9d9.

renaudhartert-db (approved):

PR #5781, opened 2026-06-04
[Doc] Note in AGENTS.md that generated docs must be fixed upstream

chrisst:
## Changes

Adds a short guideline to `AGENTS.md` (and `CLAUDE.md`, which is a symlink to it) noting that auto-generated documentation must not be edited or PR'd directly.

Docs marked `linguist-generated=true` in `.gitattributes` (most files under `docs/resources/` and `docs/data-sources/`) are auto-generated. Direct edits to them are overwritten on the next regeneration, so they must be updated at their upstream source instead.

This is the same issue that led to #5714 being closed — the goal is to steer contributors (and LLM agents) away from patching the generated files directly.

NO_CHANGELOG=true

## Tests

- [ ] Doc-only change to agent guidance; no code paths affected.


AGENTS.md
```diff
@@ -143,6 +143,7 @@ Within each section, maintain alphabetical order.
 
 ### Documentation
 - All resources and data sources require Terraform Registry compatible documentation in `docs/`
+- Docs marked `linguist-generated=true` in `.gitattributes` (most files under `docs/resources/` and `docs/data-sources/`) are auto-generated. Do not edit them or open PRs against them — direct changes are overwritten on the next regeneration. They must be updated at their upstream source instead.
```
alexott, line 146: I think that we need to add a similar comment for the source code - I see that people sometimes open PRs that fixes auto-generated code

alexott (approved):

PR #5783, opened 2026-06-04
pipelines: send serverless explicitly so ingestion pipelines can use classic compute

shreyas-goenka:
## Problem

LakeFlow Connect ingestion pipelines fail to deploy on classic compute:

> Error: cannot create pipeline: You cannot provide cluster settings when using serverless compute.

This happens when the configuration sets `serverless = false` together with a `clusters` block on a `databricks_pipeline` that has an `ingestion_definition` (e.g. a query-based MySQL/Postgres connector).

## Root cause

The Databricks Go SDK marshals the `Serverless bool` field with `omitempty`, so `serverless = false` is dropped from the request unless `"Serverless"` is present in the request's `ForceSendFields`.

Server-side, an ingestion pipeline with `serverless` **unset** is defaulted to `serverless = true` during canonicalization. The platform then validates the canonicalized spec and rejects the `clusters` block with the error above. So the dropped `serverless = false` is the entire cause of the failure.

## Fix

Append `"Serverless"` to `ForceSendFields` in `Create` and `Update`, but only when the user **explicitly set** `serverless` in their configuration. This is determined by reading the raw config (`GetRawConfigAt`) rather than `GetOkExists`, which also reads prior state. Reading raw config ensures we don't force-send for existing classic-compute pipelines whose state merely contains `serverless = false` — the over-firing that caused the [earlier attempt at this fix](https://github.com/databricks/terraform-provider-databricks/commit/8287eee) to be reverted.

### Why not clear `edition` / `channel`?

An earlier version of this PR cleared `edition` and `channel` for ingestion pipelines. That approach was based on an inaccurate diagnosis and has been dropped:

- The platform validates the **canonicalized** spec (after defaults are applied), so a cleared `edition` is simply re-defaulted to `Advanced`; there is no validation that rejects `edition`/`channel` for ingestion pipelines independent of serverless.
- Clearing `edition` would actually *introduce* a failure in workspaces pinned to the `Pro` edition via per-workspace overrides, where the serverless "must use the Advanced edition" check would then fail.

## Tests

`TestForceSendServerless` covers the three cases (`serverless = false` and `serverless = true` are force-sent; unset is not). The standard `qa.ResourceFixture` harness cannot exercise this because it builds config with `terraform.NewResourceConfigRaw`, which does not populate the cty `RawConfig` that `GetRawConfigAt` reads — so the test builds a `ResourceData` with `RawConfig` set directly, matching the real gRPC plan path.


mihaimitrea-db (approved):
Looks good

PR #5786, opened 2026-06-05
Add CI classifier to block PRs with breaking schema changes

Divyansh-db:
## Summary

- Adds a Go classifier under `scripts/schema-classifier/` that walks two `terraform providers schema -json` dumps and labels each delta as breaking or non-breaking against a frozen rule taxonomy (remove attribute, optional→required, type change, nesting-mode change, etc.).
- Adds a new `.github/workflows/schema.yml` that runs the classifier on every PR and **fails the check** if any breaking change is detected.
- Bypass: add `ALLOW_SCHEMA_BREAKING_CHANGE=true` on its own line in the PR description. The workflow uses the `pull_request.edited` event so the bypass takes effect on the next description edit — no new commit required. Mirrors the `NO_CHANGELOG=true` UX from `.github/workflows/next-changelog.yml`.

### Workflow design

One workflow, two jobs:

- `classify` runs on `ubuntu-latest` (where the public Go proxy is reachable). Builds the provider on base and head, runs the classifier, uploads the markdown report + outcome marker as a workflow artifact. Always exits 0 — purely informational.
- `status` runs on the protected runner (IP-allowlisted for the GitHub API). Downloads the artifact, reads the bypass directive from the PR body, posts/updates/deletes a sticky comment, and exits non-zero if breaking changes are present without an active bypass. **`status` is the required check** to mark in branch protection.

### Rule taxonomy (frozen, also encoded in `classify_test.go`)

Breaking: remove resource / data source / attribute; add required attribute; optional → required; remove computed from a settable attribute; settable → computed-only; remove sensitive; any attribute `type` change; nesting-mode change; min_items increase; max_items decrease; remove nested block; add nested block with `min_items > 0`.

Non-breaking: add resource / data source; add optional or computed-only attribute; required → optional; add computed to a still-settable attribute; add sensitive; max_items relaxed or removed; min_items decreased; add `deprecated`; description / `description_kind` change.

### Scope

The classifier inspects only what Terraform Core sees in `terraform providers schema -json`. Provider-side concepts like `ForceNew`, `ConflictsWith`, `ValidateFunc`, and `DiffSuppressFunc` are not present in that dump and are intentionally out of scope.

### Fork PRs

The check itself runs and gates correctly on fork PRs — `pull_request` events fire normally, the classifier runs, the workflow's pass/fail reflects the breaking-vs-not decision. What is **missing** on fork PRs is the in-PR sticky comment: `GITHUB_TOKEN` is intentionally read-only on `pull_request` events from forks, so `gh pr comment` no-ops (the step is `continue-on-error`). Fork contributors still see the failed check status and can read the classifier report in the Action log. If sticky comments on fork PRs become important, that would need a two-workflow setup using `pull_request_target` (similar to `next-changelog.yml`); we chose not to pay that complexity given the dominant same-repo workflow.

### Behavior matrix

| Scenario | Job status |
|---|---|
| No schema change | pass |
| Only non-breaking deltas | pass |
| Breaking deltas, no bypass | **fail** |
| Breaking deltas + `ALLOW_SCHEMA_BREAKING_CHANGE=true` on its own line in PR description | pass |

### Local use

`make diff-schema` now prints the classifier report after the existing `jd` view and exits non-zero on breaking changes. Use `ALLOW_BREAKING=1 make diff-schema` to see the report without the failure.

## Test plan

- [x] `go test ./scripts/schema-classifier/...` — 27 table-driven tests, one per rule taxonomy row + CLI behavior, all passing.
- [x] End-to-end CLI: hand-crafted base/head JSON pairs verified — breaking deltas exit 1, `--allow-breaking` exits 0, both text and markdown formats render correctly.
- [x] CI run on this PR — both `classify` and `status` jobs pass (no schema changes on this branch).
- [x] Demo PR (#5787) exercises a deliberate Optional→Required break — `status` fails when bypass directive is absent, passes when it's added on its own line in the PR description, and flips back to failing within minutes when the directive is removed.
- [ ] After review, mark `status` as a required check in repo Settings → Branches → branch protection rules for `main`. Manual repo-settings step, not done from a PR.

NO_CHANGELOG=true


hectorcast-db (commented):
TAG includes Private Preview elements, which are allowed to have breaking changes. As it is this will break the update workflow for TAG every time a team introduces a breaking change.

Divyansh-db:
> TAG includes Private Preview elements, which are allowed to have breaking changes. As it is this will break the update workflow for TAG every time a team introduces a breaking change.

This should only happen for update TF PRs. We can include the override line in the PR description for these PRs and that should be enough

scripts/schema-classifier/classify.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+func classifyAttributeChanged(path, name string, base, head Attribute) []Change {
+	var out []Change
+
+	if !jsonEqual(base.Type, head.Type) || !jsonEqual(base.NestedType, head.NestedType) {
+		out = append(out, Change{Path: path, Kind: "TypeChanged", Severity: Breaking,
+			Message: fmt.Sprintf("Attribute %q type changed (any type delta is breaking)", name)})
+	}
```
rauchy: For the nested case, this should only be `Breaking` if the nested type's schema is `Breaking` itself. Otherwise, adding an optional attribute to a nested type (which is a non-breaking change) would result in the `jsonEqual(base.NestedType, head.NestedType)` check failing. This would incorrectly classify the change as `Breaking`.

rauchy (changes requested):
Are we catching Computed (non-optional) -> Required anywhere?

rauchy (approved):

hectorcast-db (approved):
Approving. 
We will add an exception for TAG. Not ideal, but this will covers most common mistakes (handwritten changes).

PR #5795, opened 2026-06-10
[Fix] `databricks_mws_ncc_private_endpoint_rule`: mark `account_id` Computed to prevent spurious drift

### LINKED ISSUE #5347: [ISSUE] Issue with `databricks_mws_ncc_private_endpoint_rule` resource
### Configuration
Whenever we add additional S3 bucket in to the endpoint_rule configuration, instead of editing the resource it recreates it, which result in connectivity issue as we are using the S3 endpoint created with the NCC private rule to setup our S3 bucket security configuration.
It is also strange that it always recreates the private rule as it says the databricks account in the configured provider is different even though there is no such change, but we solved this by using lifecycle rule as you can see in the configuration below.

resource "databricks_mws_ncc_private_endpoint_rule" "ncc_private_storage_rule" {
  count                          = contains(["dev", "intg", "prep", "prod"], local.environment) ? 1 : 0
  provider                       = databricks.mws
  network_connectivity_config_id = databricks_mws_network_connectivity_config.databricks_mws_ncc[0].network_connectivity_config_id
  endpoint_service               = "com.amazonaws.eu-west-1.s3"
  enabled                        = true

  resource_names = 
    [
      "bucketA",
      "bucketB",
   ->adding "bucketC"
  ]
  lifecycle {
    ignore_changes = [account_id]
  }
}

### Expected Behavior
Whenever we add new bucket to the policy the resource to be updated not recreated

### Actual Behavior
Whenever we add new bucket to the policy the resource is being recreated

### Steps to Reproduce
Add new bucket to the policy of existing databricks_mws_ncc_private_endpoint_rule

terraform {
  required_version = ">=1.10.4"
  required_providers {
    databricks = {
      version = "1.85.0"
    }
    aws = {
      source  = "hashicorp/aws"
      version = "~> 5.81.0"
    }
  }
}

### Is it a regression?
First time we observed this issue was by using current (above mentioned) provider version

Divyansh-db:
## Changes

The `databricks_mws_ncc_private_endpoint_rule` resource has been intermittently planning a spurious `account_id` change that fails on apply.

**Symptom.** After a successful create, a subsequent plan shows:

```
~ account_id = "<some-account-id>" -> null
```

…and the follow-up apply fails with:

```
Error: cannot update mws ncc private endpoint rule: Update mask must be specified.
```

**Root cause.** The schema for this resource is auto-generated from the SDK response struct `settings.NccPrivateEndpointRule`, whose `AccountId` field is tagged `json:"account_id,omitempty"` (the field has `PUBLIC_UNDOCUMENTED` visibility on the Create/Update input schemas in the API spec — it's an output-only field). The provider's reflection layer at `common/reflect_resource.go:287` treats every `omitempty` JSON tag as `Optional: true`, so `account_id` ended up `Optional` in the schema. The resource's `SetComputed()` list at `mws/resource_mws_ncc_private_endpoint_rule.go:22` already covers other output-only fields (`rule_id`, `endpoint_name`, `connection_state`, etc.) but missed `account_id`.

When the backend's GET response populates `account_id` into state (e.g. after `terraform import`, or in any other path that bypasses `common.StructToData`'s "skip non-Computed, non-bool fields the user didn't configure" check) and HCL doesn't set `account_id`, the diff engine sees state value vs. null config and plans `account_id = "..." -> null`. The Update CRUD body only includes `enabled`, `domain_names`, and `resource_names`, so the spurious drift produces an empty `update_mask`, which the backend rejects.

**Fix.** Add `account_id` to the `SetComputed()` list, matching the sibling `databricks_mws_network_connectivity_config` resource. With `Optional + Computed`, the diff engine preserves the server-provided value across refreshes and never plans the spurious in-place update.

## Tests

- [x] `make test` run locally — new unit tests `TestResourceNccPrivateEndpointRule_AccountIdIsComputed` (schema invariant) and `TestResourceNccPrivateEndpointRule_NoDriftWhenBackendReturnsAccountId` (diff-engine behavioral) fail before the fix and pass after. Full mws unit suite still green.
- [x] covered with integration tests in `internal/acceptance` — new `TestMwsAccNccPrivateEndpointRule_NoSpuriousAccountIdDrift` (Azure / AWS branches, GCP explicitly skipped per the resource docs). Verified end-to-end against a real Azure account:
  - Before fix: Step 2 (apply with `account_id` removed) fails with the customer-reported `cannot update mws ncc private endpoint rule: Update mask must be specified.` from the backend.
  - After fix: all 3 steps pass; the no-op Update is skipped, plan stays empty.
- [ ] relevant change in `docs/` folder — `account_id` was never documented in the Argument Reference, so no doc change is needed.
- [ ] using Go SDK — n/a, schema customization only
- [ ] using TF Plugin Framework — n/a, this resource is SDKv2

Closes #5347.


mws/resource_mws_ncc_private_endpoint_rule.go
```diff
@@ -19,7 +19,7 @@ func ResourceMwsNccPrivateEndpointRule() common.Resource {
 			common.CustomizeSchemaPath(m, p).SetForceNew()
 		}
 
-		for _, p := range []string{"rule_id", "endpoint_name", "connection_state", "creation_time", "updated_time", "vpc_endpoint_id"} {
+		for _, p := range []string{"account_id", "rule_id", "endpoint_name", "connection_state", "creation_time", "updated_time", "vpc_endpoint_id"} {
```
tanmay-db, line 22: We finalised going ahead with computed + optional for this? Not sure if there were further discussions.

mws/resource_mws_ncc_private_endpoint_rule_acc_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+// Reproduces ES-1957213 / #5347.
```
tanmay-db: Internal info, we should remove

mws/resource_mws_ncc_private_endpoint_rule_test.go
```diff
@@ -294,3 +295,49 @@ func TestResourceNccPrivateEndpointRuleDelete_Error(t *testing.T) {
...
+// same reason; the omission here is the root cause of ES-1957213 / #5347.
+func TestResourceNccPrivateEndpointRule_AccountIdIsComputed(t *testing.T) {
+	s := ResourceMwsNccPrivateEndpointRule().Schema
+	accountID, ok := s["account_id"]
+	assert.True(t, ok, "account_id should be present in the schema")
+	assert.True(t, accountID.Computed,
+		"account_id must be Computed; otherwise a backend-populated value triggers spurious drift and an empty-update-mask error")
+}
+
+// Reproduces the customer-facing symptom of ES-1957213 / #5347: when a prior
```
tanmay-db: ditto

mws/resource_mws_ncc_private_endpoint_rule_acc_test.go
```diff
@@ -0,0 +1,94 @@
...
+	case acceptance.IsAzure(t):
+		ncc = `
+		resource "databricks_mws_network_connectivity_config" "this" {
+			name = "tf-{var.STICKY_RANDOM}"
+			region = "eastus2"
+		}`
+		ruleBase = `
+		resource "databricks_mws_ncc_private_endpoint_rule" "this" {
+			network_connectivity_config_id = databricks_mws_network_connectivity_config.this.network_connectivity_config_id
+			resource_id                    = "/subscriptions/2a5a4578-9ca9-47e2-ba46-f6ee6cc731f2/resourceGroups/deco-prod-azure-eastus2-rg/providers/Microsoft.Storage/storageAccounts/decotestprodunity"
```
tanmay-db: is this something which should be part of env variable?

mws/resource_mws_ncc_private_endpoint_rule_acc_test.go
```diff
@@ -0,0 +1,94 @@
...
+				if r.Primary.Attributes["account_id"] == "" {
+					return fmt.Errorf("account_id not populated in state after Step 1; bug precondition not met. attrs: %v", r.Primary.Attributes)
+				}
+				return nil
+			},
+		},
+		acceptance.Step{
+			Template: withoutAccountID,
+		},
+		acceptance.Step{
```
tanmay-db, line 91: what is the purpose of this step after the above step? is it to make sure there is no diff?

mws/resource_mws_ncc_private_endpoint_rule.go
Divyansh-db: Yes, we don't want to introduce any breaking changes

mws/resource_mws_ncc_private_endpoint_rule_acc_test.go
Divyansh-db: removed now

mws/resource_mws_ncc_private_endpoint_rule_test.go
Divyansh-db: removed now

mws/resource_mws_ncc_private_endpoint_rule_acc_test.go
Divyansh-db: removed now

mws/resource_mws_ncc_private_endpoint_rule_acc_test.go
Divyansh-db: yes

tanmay-db (approved):

PR #5797, opened 2026-06-11
Deprecate SDKv2 fallback for migrated PF resources

Divyansh-db:
## Summary

- Three resources (`databricks_library`, `databricks_quality_monitor`, `databricks_share`) and three data sources (`databricks_share`, `databricks_shares`, `databricks_volumes`) have been served by the Plugin Framework by default since their migration. The SDKv2 implementations remain only as opt-in fallbacks via the `USE_SDK_V2_RESOURCES` / `USE_SDK_V2_DATA_SOURCES` environment variables.
- This PR is **Phase 1 of a two-phase deprecation**: announce now, delete in the next major release. Documentation pages get a `~> **Deprecation**` banner, `NEXT_CHANGELOG.md` carries the announcement, and a runtime warning fires when a user is on the SDKv2 fallback path.
- The runtime warning is wired into the generic registration helpers (`getPluginFrameworkResourcesToRegister` / `getPluginFrameworkDataSourcesToRegister`), so **any future entry added to `migratedResources` / `migratedDataSources` inherits the deprecation signal automatically**. A doc comment above the two lists spells out the deprecation contract for whoever owns the next migration.
- Phase 2 (next major release) will delete the ~1,862 LOC of SDKv2 source plus the env-var fallback machinery and move the PF entries from `migratedResources` to `pluginFwOnlyResources`. `ConfigureAsSdkV2Compatible()` stays — existing user state files were written with block-shaped schemas and need that compatibility shim to be read.

## Runtime warning behavior

- **Who sees it:** only users who have set `USE_SDK_V2_RESOURCES` / `USE_SDK_V2_DATA_SOURCES` (or, in tests, used `WithSdkV2ResourceFallbacks` / `WithSdkV2DataSourceFallbacks`) for a name in `migratedResources` / `migratedDataSources`. Default users never trip the helper.
- **Frequency:** Terraform spawns a fresh provider process per CLI command. A `sync.Map` dedupe inside the helper means at most one warning per name per process, so a single `terraform plan` / `apply` prints one line per affected name — and every subsequent command re-emits, by design, until the user removes the override.
- **Visibility caveat:** the warning is a `log.Printf("[WARN] ...")`, which is the same channel SDKv2 schema deprecations use. Terraform's CLI only forwards `[WARN]` provider logs to the user when `TF_LOG` is set to `WARN` or louder; with `TF_LOG` unset the line is silently dropped. The docs banner + `NEXT_CHANGELOG` entry are therefore the loudest signals; the runtime warning catches users who run with logging on. If we want an unconditional in-band signal, we'd switch to `resp.Diagnostics.AddWarning` from the PF provider's `Configure`, at the cost of duplicating the iteration over the fallback list — happy to do that in a follow-up if reviewers prefer.

## Test plan

- [x] New unit tests in `internal/providers/pluginfw/pluginfw_test.go` cover: once-per-name dedup, env-var path, fallback-option path, and no-warning-when-clean.
- [x] `go test ./internal/providers/pluginfw/... ./internal/providers/sdkv2/... ./internal/providers/...` — all green.
- [x] `gofmt -l`, `go vet`, `staticcheck ./internal/providers/pluginfw/`, `./tools/validate_whitespace.py` — all clean.
- [ ] Manual smoke test on a workspace: with `USE_SDK_V2_RESOURCES="databricks_library"` set and `TF_LOG=WARN`, run `terraform plan` against a config that uses `databricks_library` and confirm the deprecation warning appears once and the resource still works through SDKv2.
- [ ] Manual smoke test without the env var: confirm no warning is emitted and the Plugin Framework continues to serve the resource.

### DIFF
--- a/internal/providers/pluginfw/pluginfw_rollout_utils.go
+++ b/internal/providers/pluginfw/pluginfw_rollout_utils.go
@@ -8,9 +8,11 @@ package pluginfw
 
 import (
 	"context"
+	"log"
 	"os"
 	"slices"
 	"strings"
+	"sync"
 
 	"github.com/databricks/databricks-sdk-go/config"
 	"github.com/databricks/terraform-provider-databricks/internal/providers/pluginfw/products/app"
@@ -29,7 +31,24 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 )
 
-// List of resources that have been migrated from SDK V2 to plugin framework
+// migratedResources lists resources that were originally implemented on SDKv2
+// and have been re-implemented on the Plugin Framework. Entries here are served
+// by the Plugin Framework by default; users can opt back to SDKv2 by listing
+// the resource name in the USE_SDK_V2_RESOURCES environment variable.
+//
+// Deprecation contract for entries in this list:
+//  1. The SDKv2 implementation must stay present and functional for the rest of
+//     this major version so that USE_SDK_V2_RESOURCES still works.
+//  2. The resource's documentation must carry a `~> **Deprecation**` banner
+//     warning that the SDKv2 implementation will be removed in the next major.
+//  3. The PR that adds an entry must announce the deprecation in NEXT_CHANGELOG.
+//  4. In the next major release, the SDKv2 source file and test must be deleted
+//     and the entry moved from migratedResources to pluginFwOnlyResources.
+//
+// getPluginFrameworkResourcesToRegister emits a one-time runtime warning whenever
+// a user steers a name in this list back to SDKv2, so entries inherit the
+// deprecation signal automatically.
+//
 // Keep this list sorted.
 var migratedResources = []func() resource.Resource{
 	library.ResourceLibrary,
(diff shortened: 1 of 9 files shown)


internal/providers/pluginfw/pluginfw_rollout_utils.go
```diff
@@ -176,9 +226,11 @@ func getPluginFrameworkDataSourcesToRegister(dataSourceFallbacks []string) []fun
 	// Loop through the map and add data sources if they're not specifically marked to use the SDK V2
 	for _, dataSourceFunc := range migratedDataSources {
 		name := getDataSourceName(dataSourceFunc)
-		if !shouldUseSdkV2DataSource(name) && !slices.Contains(dataSourceFallbacks, name) {
-			dataSources = append(dataSources, dataSourceFunc)
+		if shouldUseSdkV2DataSource(name) || slices.Contains(dataSourceFallbacks, name) {
```
chrisst: rewriting the conditional logic feels unnecessarily risky when you could have implemented this with an else

internal/providers/pluginfw/pluginfw_rollout_utils.go
```diff
@@ -29,22 +31,68 @@ import (
...
+// migratedResources / migratedDataSources is steered back to the SDKv2
+// implementation, either via the USE_SDK_V2_RESOURCES / USE_SDK_V2_DATA_SOURCES
+// environment variables or via the test-only WithSdkV2ResourceFallbacks /
+// WithSdkV2DataSourceFallbacks options. kind is "resource" or "data source".
+//
+// The warning is sent through log.Printf with a [WARN] prefix; the Terraform
+// CLI surfaces this to users running with TF_LOG=WARN or higher. This is the
+// same channel SDKv2 schema deprecations use.
+func emitSdkV2FallbackWarning(name, kind string) {
+	if _, loaded := warnedFallbackNames.LoadOrStore(name, struct{}{}); loaded {
```
chrisst: TBH this feels overkill for logging a warning. I'm not even sure I understand when or if this warning will be fired and I'm not sure I need to.

internal/providers/pluginfw/pluginfw_rollout_utils.go
Divyansh-db: Good call — fixed in b4c38791. The condition is back to the original `if !shouldUseSdkV2DataSource(name) && !slices.Contains(...)` and I moved the warning to an `else` branch. Diff is now strictly additive.

internal/providers/pluginfw/pluginfw_rollout_utils.go
Divyansh-db: Simplified in b4c38791 — dropped the helper, the `sync.Map` dedup, and the long doc comment. The log line is now inline in the `else` branch of both registration loops.

On *when* it fires: only when the user has set `USE_SDK_V2_RESOURCES` / `USE_SDK_V2_DATA_SOURCES` to one of `databricks_library`, `databricks_quality_monitor`, `databricks_share`, `databricks_shares`, or `databricks_volumes` (or, in tests, used `WithSdkV2ResourceFallbacks` / `WithSdkV2DataSourceFallbacks`). Default users never see it. Within a single `terraform plan` it prints at most once per affected name, since `Resources()` / `DataSources()` is called once per provider process.

Caveat worth flagging: provider `log.Printf("[WARN] ...")` only reaches the user's terminal when `TF_LOG` is set to `WARN` or higher; with `TF_LOG` unset (the common case) Terraform's plugin protocol filters it out. So today, only users running with logging on see the warning. If we want it unconditionally visible to anyone on the fallback path, the right tool is `resp.Diagnostics.AddWarning` from inside the PF provider's `Configure()` — happy to switch if you'd prefer that.

Context on *why* a warning at all: this PR is Phase 1 of a 2-phase deprecation. Phase 2 (next major) deletes the ~1,862 LOC of SDKv2 source plus the env-var fallback machinery itself, so anyone currently on the fallback needs a heads-up.

rauchy (approved):
The change LGTM, but I'm feeling like this thing might go unnoticed due to the ` log.Printf("[WARN] ...")` politeness. I'd be happy to see a louder shout.

Divyansh-db:
> The change LGTM, but I'm feeling like this thing might go unnoticed due to the  log.Printf("[WARN] ...") politeness. I'd be happy to see a louder shout.

The fallback decision is made inside getPluginFrameworkResourcesToRegister / getPluginFrameworkDataSourcesToRegister, which are called from Resources(ctx) / DataSources(ctx). Both of those return a []func() resource.Resource — they don't receive a *Response with a diagnostics bag, so there's no place to attach a Diagnostics.AddWarning at that call site. To use diagnostics, the emission has to move to Configure(ctx, req, *resp), which has resp.Diagnostics, and this would require replication of the list of migrated resources in Configure() method

PR #5802, opened 2026-06-12
Fix `databricks_instance_pool` ignoring `enable_elastic_disk = false`

### LINKED ISSUE #5462: [Bug] `databricks_instance_pool` ignores `enable_elastic_disk = false` due to `omitempty` JSON tag

Divyansh-db:
## Summary

`enable_elastic_disk = false` on `databricks_instance_pool` is silently ignored, causing an infinite plan/replace cycle. This PR removes the `omitempty` JSON tag on the `EnableElasticDisk` field so `false` is explicitly serialized to the API.

## Credit

This re-submits @Bafff's fix from #5454. CI tests can't run on PRs from forks, which is why the original got stalled — I'm re-opening from an upstream branch so the test suite can run. The diagnosis, code change, and unit test are all theirs. Thanks @Bafff!

## Root cause

The `EnableElasticDisk` field in both `InstancePool` and `InstancePoolAndStats` was tagged `json:"enable_elastic_disk,omitempty"`. Because `false` is the zero value for `bool` in Go, `omitempty` omits it from the serialized JSON. The API never receives `enable_elastic_disk: false` and defaults to `true`.

Combined with `tf:"force_new"`, this creates an infinite cycle:

1. User sets `enable_elastic_disk = false`.
2. Terraform creates the pool — but `omitempty` drops the field from the request.
3. API defaults to `true`.
4. Next `terraform plan` sees drift (`true → false`), wants to recreate.
5. Repeat forever.

Direct REST API calls confirm Databricks **does** accept and persist `enable_elastic_disk: false`:

```bash
# Create with false
curl -X POST ".../api/2.0/instance-pools/create" \
  -d '{"instance_pool_name":"test","node_type_id":"Standard_F4","enable_elastic_disk":false,...}'

# Read back — confirmed false
curl ".../api/2.0/instance-pools/get?instance_pool_id=..."
# Returns: "enable_elastic_disk": false
```

So the bug is purely client-side serialization.

## Fix

- Remove `omitempty` from the JSON tag on `EnableElasticDisk` in both `InstancePool` and `InstancePoolAndStats` so `false` is sent explicitly.
- Add `tf:"optional"` on `InstancePool.EnableElasticDisk`. The schema generator's `isOptional()` (in `common/reflect_resource.go`) checks for either `omitempty` on the JSON tag *or* `optional` in the `tf` tag — without the new tag, removing `omitempty` would flip the schema to Required, which would itself be a breaking change. The pre-existing `s["enable_elastic_disk"].Default = true` is preserved, so users who omit the attribute see no behavior change.

## Test plan

- [x] `go test ./pools/...` passes.
- [x] New regression test `TestResourceInstancePoolCreate_ElasticDiskDisabled` asserts the outgoing Create request body contains `"enable_elastic_disk": false`. It uses a `map[string]any` `ExpectedRequest` so the assertion does not depend on the struct's own JSON tags.
- [x] Existing tests are unaffected (they already populated `EnableElasticDisk: true` in their `ExpectedRequest`, so removing `omitempty` is a no-op for them).
- [x] `make fmt lint ws` clean.

Fixes #5462

### DIFF
--- a/pools/resource_instance_pool.go
+++ b/pools/resource_instance_pool.go
@@ -89,7 +89,7 @@ type InstancePool struct {
 	GcpAttributes                      *InstancePoolGcpAttributes      `json:"gcp_attributes,omitempty" tf:"force_new,suppress_diff"`
 	NodeTypeID                         string                          `json:"node_type_id,omitempty" tf:"suppress_diff,force_new,conflicts:instance_pool_fleet_attributes"`
 	CustomTags                         map[string]string               `json:"custom_tags" tf:"optional"`
-	EnableElasticDisk                  bool                            `json:"enable_elastic_disk,omitempty" tf:"force_new,suppress_diff"`
+	EnableElasticDisk                  bool                            `json:"enable_elastic_disk" tf:"optional,force_new,suppress_diff"`
 	DiskSpec                           *InstancePoolDiskSpec           `json:"disk_spec,omitempty" tf:"force_new"`
 	PreloadedSparkVersions             []string                        `json:"preloaded_spark_versions,omitempty" tf:"force_new"`
 	PreloadedDockerImages              []clusters.DockerImage          `json:"preloaded_docker_images,omitempty" tf:"force_new,slice_set,alias:preloaded_docker_image"`
@@ -118,7 +118,7 @@ type InstancePoolAndStats struct {
 	DefaultTags                        map[string]string                `json:"default_tags,omitempty" tf:"computed"`
 	CustomTags                         map[string]string                `json:"custom_tags,omitempty"`
 	IdleInstanceAutoTerminationMinutes int32                            `json:"idle_instance_autotermination_minutes"`
-	EnableElasticDisk                  bool                             `json:"enable_elastic_disk,omitempty"`
+	EnableElasticDisk                  bool                             `json:"enable_elastic_disk"`
 	DiskSpec                           *InstancePoolDiskSpec            `json:"disk_spec,omitempty"`
 	PreloadedSparkVersions             []string                         `json:"preloaded_spark_versions,omitempty"`
 	State                              string                           `json:"state,omitempty"`
--- a/pools/resource_instance_pool.go
+++ b/pools/resource_instance_pool.go
@@ -118,7 +118,7 @@ type InstancePoolAndStats struct {
 	DefaultTags                        map[string]string                `json:"default_tags,omitempty" tf:"computed"`
 	CustomTags                         map[string]string                `json:"custom_tags,omitempty"`
 	IdleInstanceAutoTerminationMinutes int32                            `json:"idle_instance_autotermination_minutes"`
-	EnableElasticDisk                  bool                             `json:"enable_elastic_disk"`
+	EnableElasticDisk                  bool                             `json:"enable_elastic_disk" tf:"optional"`
 	DiskSpec                           *InstancePoolDiskSpec            `json:"disk_spec,omitempty"`
 	PreloadedSparkVersions             []string                         `json:"preloaded_spark_versions,omitempty"`
 	State                              string                           `json:"state,omitempty"`
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -10,7 +10,7 @@
 
 
 ### Bug Fixes
-* Fixed `databricks_instance_pool` ignoring `enable_elastic_disk = false` due to `omitempty` JSON tag, which caused an infinite plan/replace cycle ([#NNNN](https://github.com/databricks/terraform-provider-databricks/pull/NNNN)).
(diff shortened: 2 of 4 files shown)


pools/resource_instance_pool.go
```diff
@@ -89,7 +89,7 @@ type InstancePool struct {
 	GcpAttributes                      *InstancePoolGcpAttributes      `json:"gcp_attributes,omitempty" tf:"force_new,suppress_diff"`
 	NodeTypeID                         string                          `json:"node_type_id,omitempty" tf:"suppress_diff,force_new,conflicts:instance_pool_fleet_attributes"`
 	CustomTags                         map[string]string               `json:"custom_tags" tf:"optional"`
-	EnableElasticDisk                  bool                            `json:"enable_elastic_disk,omitempty" tf:"force_new,suppress_diff"`
```
tanmay-db, line 92: Why does this change fix the issue? Do we need/have to remove omitempty?

tanmay-db (commented):
Could you add an integration test which would have failed before (diff) and now would pass (no diff)

pools/resource_instance_pool.go
```diff
@@ -118,7 +118,7 @@ type InstancePoolAndStats struct {
 	DefaultTags                        map[string]string                `json:"default_tags,omitempty" tf:"computed"`
 	CustomTags                         map[string]string                `json:"custom_tags,omitempty"`
 	IdleInstanceAutoTerminationMinutes int32                            `json:"idle_instance_autotermination_minutes"`
-	EnableElasticDisk                  bool                             `json:"enable_elastic_disk,omitempty"`
```
tanmay-db, line 121: We don't need force_new suppress_diff here, as it's needed above?

pools/resource_instance_pool.go
Divyansh-db: With `bool` + `omitempty`, Go's `encoding/json` can't distinguish "user set false" from "unset" — both serialize as omitted. So `enable_elastic_disk: false` never reaches the API, which then falls back to its server-side default `true`, producing the infinite plan/replace loop. Removing `omitempty` forces explicit serialization.

The structurally cleaner alternative is `*bool` (which the SDK v2 codegen already uses upstream), but that's a larger type-change that would ripple through the schema generator — this is the minimum-viable fix that keeps the existing field type.

pools/resource_instance_pool.go
Divyansh-db: Correct — `InstancePoolAndStats` is the schema source for the **data source** `databricks_instance_pool` (via `poolDetails` in `data_instance_pool.go:29`), not the resource. `force_new` and `suppress_diff` are resource-only directives; they're meaningless on a data source attribute. Only `tf:"optional"` is needed here, to keep the data source attribute Optional after losing the `omitempty` signal that `isOptional()` reads.

Divyansh-db:
@tanmay-db Added in 44f3b35d — `TestAccInstancePool_EnableElasticDiskFalse` in `pools/resource_instance_pool_acc_test.go`. It creates a pool with `enable_elastic_disk = false` and relies on the implicit post-apply refresh-plan to fail on any drift. Before this PR's fix, the create request dropped the field, the API stored `true`, and the refresh-plan would show a `true → false` recreate diff. After the fix, the second plan is empty and the test passes.

tanmay-db (approved):

PR #5803, opened 2026-06-12
Run fork PR unit tests offline from a pre-warmed Go module cache

chrisst:
## Changes

Adds a "Warm Go Cache" workflow and a fork-aware path to the `tests` job so that PRs from forks can pass the required build/unit-test check without JFrog credentials.

### Problem
Fork PRs currently fail the required `tests` check unconditionally: the build authenticates to the JFrog Go module proxy via GitHub OIDC, and GitHub does not issue OIDC tokens to fork PR workflow runs, so the very first setup step fails (`Unable to get ACTIONS_ID_TOKEN_REQUEST_URL`) before any code is compiled. This blocks all external contributions and any Databricks employee contributing from a personal fork.

### Solution
This ports the cache-warming SOP 

- **`warm-go-cache.yml` (new)** — the sole writer of the `go-modules-*` Actions cache. Runs on the hardened runner group with JFrog OIDC, resolves modules with the exact command CI uses (`jf go mod vendor`), verifies the cache resolves offline, and saves it under a timestamped key. Triggers: daily cron (beats 7-day cache eviction), pushes to `main` touching `go.mod`/`go.sum`, and manual dispatch.
- **`tests` job fork routing** — fork PRs skip JFrog entirely: restore the warmed cache (restore-only), resolve modules offline through Go's `file://` proxy, run `make test` as usual. The check context stays `tests`. Same-repo runs are unchanged.

### Maintainer flow for fork PRs that change dependencies
If a fork PR modifies `go.mod`/`go.sum`, its `tests` check fails on cache miss. After reviewing the dependency change, dispatch **Actions → Warm Go Cache → Run workflow → pr_number = N**: the warmer fetches *only* `go.mod`/`go.sum` from the fork (never source code; the PR number and head ref are validated before use), rebuilds the cache, and the contributor re-runs the failed check. The fork-PR auto-comment and CONTRIBUTING.md document this.

### Security properties
- No fork-controlled code ever executes in a privileged context; the warmer overlays dependency manifests only.
- Only the warmer (main branch or maintainer-dispatched) writes the cache; PR workflows are restore-only, and `workflow_dispatch` inputs / PR head fields are validated against injection.
- Fork runs hold no credentials at all; integration tests keep their existing manual trigger flow.

## Tests
- `actionlint` and the repo's conftest policies pass on all changed workflow/action files.
- The offline-resolution step in the warmer self-verifies the cache: a broken cache fails the warmer, not contributor PRs.
- Note for reviewers: the fork path only works after the warmer's first run on `main` — please dispatch "Warm Go Cache" once right after merge.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -13,6 +13,9 @@ jobs:
       labels: linux-ubuntu-latest
 
     permissions:
+      # Required for JFrog OIDC auth on same-repo runs. GitHub never issues
+      # OIDC tokens to fork PR runs, so the permission is inert there; GitHub
+      # offers no way to condition job-level permissions on the event.
       id-token: write
       contents: read
 
@@ -20,15 +23,39 @@ jobs:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
+      # Fork PRs cannot authenticate to JFrog (GitHub does not issue OIDC
+      # tokens to fork runs); they resolve modules offline from the cache
+      # pre-warmed by the "Warm Go Cache" workflow instead.
+      - name: Detect fork PR
+        id: fork-check
+        shell: bash
+        run: |
+          if [ "${{ github.event.pull_request.head.repo.full_name }}" != "" ] && \
+             [ "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ]; then
+            echo "is_fork=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_fork=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Setup Go build environment
         uses: ./.github/actions/setup-go-build-environment
         with:
           go-version-file: go.mod
+          is-fork: ${{ steps.fork-check.outputs.is_fork }}
 
       - name: Pull external libraries
         shell: bash
+        env:
+          IS_FORK: ${{ steps.fork-check.outputs.is_fork }}
         run: |
-          jf go mod vendor
+          if [ "$IS_FORK" = "true" ]; then
+            # Resolve modules offline from the pre-warmed cache. If this fails
+            # after a go.mod/go.sum change, a maintainer must run the
+            # "Warm Go Cache" workflow with this PR's number first.
+            go mod vendor
+          else
+            jf go mod vendor
+          fi
 
           # Point native go commands at the local module cache instead of
           # the network. go run pkg@version needs module lookups even when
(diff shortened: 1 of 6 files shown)


.github/workflows/push.yml
```diff
@@ -13,22 +13,49 @@ jobs:
...
 
+      # Fork PRs cannot authenticate to JFrog (GitHub does not issue OIDC
+      # tokens to fork runs); they resolve modules offline from the cache
+      # pre-warmed by the "Warm Go Cache" workflow instead.
+      - name: Detect fork PR
+        id: fork-check
+        shell: bash
+        run: |
+          if [ "${{ github.event.pull_request.head.repo.full_name }}" != "" ] && \
+             [ "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ]; then
```
hectorcast-db: FYI, I vaguely remember that this was not "good enough" for a 100% foolproof solution. E.g., some entity may not be using a fork and still not have access to secrets/proxy.

But it should cover most of the cases.

.github/workflows/warm-go-cache.yml
```diff
@@ -0,0 +1,113 @@
...
+on:
+  push:
+    branches: [main]
+    paths:
+      - go.mod
+      - go.sum
+      - .github/workflows/warm-go-cache.yml
+  schedule:
+    - cron: "0 6 * * *" # Daily; GitHub evicts caches not accessed for 7 days
+  workflow_dispatch:
```
hectorcast-db, line 29: Maybe I am being over cautious, but workflow executions are public. Someone could open a benign PR, monitor the workflow execution and, when this workflow is triggered, quickly push a malicious update to the PR. Should we use a commit hash instead of PR number?

I think this may be ok. The reason is that malicious versions should not be in the proxy to begin with, so they would not be able to be downloaded into the cache even with such "switcharoo". We can use the hash if we want, but I am fine with using the PR number too.

.github/workflows/push.yml
hectorcast-db: I remember: some bots (e.g., Dependabot ) don't use a fork, but uses a scoped token which doesn't have access to secrets.
So such PRs will not work with this approach. Maybe we can do a similar approach that we did for tests. Try to read the token: if it's not there for any reason, default to the cache.

https://github.com/databricks/terraform-provider-databricks/blob/main/.github/workflows/integration-tests.yml#L31-L51

.github/workflows/warm-go-cache.yml
chrisst: I'm ok with pr for now too

.github/workflows/push.yml
chrisst: SG, I'll take a stab at this.

chrisst: updated.

hectorcast-db (approved):

PR #5812, opened 2026-06-16
Add `clear_cloud_attributes_on_remove` to `databricks_cluster`

renaudhartert-db:
## Summary

Removing a cloud-attributes block (`aws_attributes`, `azure_attributes`, `gcp_attributes`) from a `databricks_cluster` was silently suppressed: `terraform plan` reported no changes and the value (e.g. `instance_profile_arn`) stayed on the cluster, with no way to clear it through Terraform. The cloud-attribute blocks are diff-suppressed to hide the perpetual diff the platform causes by echoing back default cloud attributes the configuration never set. (ES-1927346)

This adds an opt-in, per-cluster `clear_cloud_attributes_on_remove` flag (default `false`, so existing behavior is unchanged). When it is `true` and a cloud-attributes block is removed from configuration, the removal produces a real diff and `Edit` that clears it. Keeping the block, even partially specified, preserves the historical suppression, so a partially-set block does not perpetually diff against platform-returned defaults; only removing the whole block clears.

## Notes for reviewers

`common/customizable_schema.go` and `common/reflect_resource.go` are generated (`DO NOT EDIT`) files. The resource depends on a new `SetConditionalSuppressDiff` helper added there. That helper is behavior-preserving for all existing callers: the existing `SetSuppressDiff`/`diffSuppressor` now route through it with an always-true predicate. The change is being landed at its generator source separately and will reconcile on the next regeneration; it is included here so the resource change builds and can be reviewed as a unit.

## Tests

Table-driven unit tests in `clusters/resource_cluster_clear_on_remove_test.go` cover the behavior at three altitudes: the diff (both flag values, plus explicit-empty and empty-block edges), the read (server defaults for an unconfigured block are dropped, never stored), and the update (whether the diff produces a clearing `Edit`).


clusters/resource_cluster_clear_on_remove_test.go
```diff
@@ -0,0 +1,363 @@
...
+	"github.com/databricks/databricks-sdk-go/service/compute"
+	"github.com/databricks/terraform-provider-databricks/common"
+
+	ctyjson "github.com/hashicorp/go-cty/cty/json"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/stretchr/testify/mock"
+)
+
+// These tests cover the ES-1927346 behavior and the clear_cloud_attributes_on_remove
```
Divyansh-db: We should not list the internal information here

Divyansh-db (commented):
Discussed offline, one test case with the flag is missing.

Divyansh-db (approved):

PR #5817, opened 2026-06-17
docs(budget): example for creating budgets for Genie usage

zcking:
## Changes

Update docs with an example illustrating how to create and manage [Budgets for Genie](https://docs.databricks.com/aws/en/genie/budgets) now that Genie is going PAYGO next month. 

UI changes make this feel like a totally different set of fields ("resource tags") but it's actually the same old familiar `filter { ... tags { ... } }` configuration. The example makes this more obvious for users looking to manage this specific cost control.

## Tests

- [X] `make fmt-docs` run
- [X] renderes successfully in [Terraform Registry Doc Preview Tool](https://registry.terraform.io/tools/doc-preview)
- [X] tested the example HCL configuration in internal account


alexott (approved):
lgtm, only either add changelog item, or add a corresponding tag into the PR description

PR #5819, opened 2026-06-18
Migrate databricks_mws_ncc_private_endpoint_rule to the Plugin Framework (opt-in)

renaudhartert-db:
## Summary

This adds a Plugin Framework (PF) implementation of `databricks_mws_ncc_private_endpoint_rule`. The SDKv2 implementation stays the default; the PF implementation is opt-in, served only when its name is listed in the `DATABRICKS_TF_ENABLED_PF_RESOURCES` environment variable. Opting in requires no HCL or state changes: the PF schema is a one-to-one parity port of the SDKv2 schema (verified by dumping `terraform providers schema -json` under both modes and diffing the resource; the only difference is that the `gcp_endpoint` block enforces "at most one" with a validator rather than a schema-level `max_items`, which is functionally identical).

The PF Create waits for the rule to finish provisioning before returning, so problems surface at apply time instead of on the next plan. It polls `connection_state` until the rule leaves `CREATING`: `PENDING` and `ESTABLISHED` are treated as success; `CREATE_FAILED`, `REJECTED`, `DISCONNECTED`, and `EXPIRED` surface as apply-time errors; and any unknown future state is treated as success so a provider built against an older SDK cannot fail an apply on a value it does not recognize. The shipped SDKv2 resource does no polling and returns immediately, so this synchronous behavior is new for opt-in users.

## Breaking change

Server-populated attributes on the SDKv2 resource (`rule_id`, `account_id`, `endpoint_name`, `vpc_endpoint_id`, `connection_state`, `creation_time`, `updated_time`, `deactivated`, `deactivated_at`, `error_message`, and `gcp_endpoint.psc_endpoint_uri`) are now read-only. Previous releases accepted writes to these attributes silently while the API ignored or overwrote them, producing perpetual drift. Configurations that explicitly set any of these must remove the assignment; the value is still available in state after apply.

## How registration stays safe

A resource is served by exactly one provider. The SDKv2 provider drops an opt-in resource from its map precisely when the PF provider registers it, both driven by the same env var, so there is no double registration. An invariant test asserts every opt-in resource also has an SDKv2 implementation.

## Tests

Unit tests cover the Create polling state machine (success, `CREATE_FAILED`, the rejected/disconnected/expired error path, the unknown-state forward-compatibility path, and persisting state before polling so destroy still works), CRUD, import, ID round-trip, schema invariants (`account_id` computed-only, the group_id/domain_names/resource_names mutual exclusion), and the empty-list-to-null mapping. `make fmt lint ws` and the affected unit tests pass.

ALLOW_SCHEMA_BREAKING_CHANGE=true


internal/providers/pluginfw/products/networks/privateendpointrule/resource_model.go
```diff
@@ -0,0 +1,201 @@
...
+// fromAPI copies server-returned fields into m. The caller separately sets
+// NetworkConnectivityConfigId from the URL path; we don't trust it from the
+// response so the value stays consistent with how the resource was addressed.
+func (m *model) fromAPI(ctx context.Context, rule *settings.NccPrivateEndpointRule) diag.Diagnostics {
+	var diags diag.Diagnostics
+	m.RuleId = types.StringValue(rule.RuleId)
+	m.AccountId = types.StringValue(rule.AccountId)
+	m.EndpointService = types.StringValue(rule.EndpointService)
+	m.GroupId = types.StringValue(rule.GroupId)
+	m.ResourceId = types.StringValue(rule.ResourceId)
```
Divyansh-db: `endpoint_service`, `group_id`, and `resource_id` are `Optional`/non-`Computed`, and each is unset on one cloud or the other (AWS leaves `group_id`/`resource_id`; Azure leaves `endpoint_service`). The server returns the unset ones as `""`, so `types.StringValue(...)` here puts a *known* `""` in state while the plan has `null` — the first apply fails with `inconsistent result after apply: .<attr>: was null, but now cty.StringVal("")`. The list fields already map empty→null via `stringsToList`; the scalars need the same:

```go
func stringOrNull(s string) types.String {
    if s == "" {
        return types.StringNull()
    }
    return types.StringValue(s)
}
```

```go
m.EndpointService = stringOrNull(rule.EndpointService)
m.GroupId         = stringOrNull(rule.GroupId)
m.ResourceId      = stringOrNull(rule.ResourceId)
```

internal/providers/pluginfw/products/networks/privateendpointrule/resource.go
```diff
@@ -0,0 +1,268 @@
...
+		return
+	default:
+		// PENDING and ESTABLISHED are the expected success states (the rule is
+		// usable or awaiting customer approval). The loop already excluded
+		// CREATING, and a timed-out poll returned an error above, so anything
+		// reaching here is one of those two or an enum value the vendored SDK
+		// predates. Treat unknown future states as success rather than failing
+		// the apply on a value we cannot classify; a later read surfaces it.
+		resp.Diagnostics.Append(plan.fromAPI(ctx, rule)...)
+		resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
Divyansh-db, line 189: `enabled` is `Optional + Computed`, so `enabled = false` in config is a *known* planned value. `CreatePrivateEndpointRule` has no `Enabled` field, so the server creates with its default and `fromAPI` overwrites `plan.Enabled` with it; if they differ, the apply fails the consistency check. (Omitting `enabled` plans as unknown and is fine.) Preserve the planned value and let the next Update reconcile:

```go
configuredEnabled := plan.Enabled
resp.Diagnostics.Append(plan.fromAPI(ctx, rule)...)
if !configuredEnabled.IsNull() && !configuredEnabled.IsUnknown() {
    plan.Enabled = configuredEnabled
}
resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```

internal/providers/pluginfw/products/networks/privateendpointrule/resource_schema.go
```diff
@@ -0,0 +1,172 @@
...
+				},
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"psc_endpoint_uri": schema.StringAttribute{
+							Computed: true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"service_attachment": schema.StringAttribute{Optional: true},
```
Divyansh-db: `gcp_endpoint.service_attachment` is `Optional` with no `RequiresReplace`, but `toUpdateRequest` never adds `gcp_endpoint` to the mask — so editing it plans an in-place update that's dropped (empty mask → Update skips the call), then state is written with the new value while the server keeps the old → perpetual diff. Match its create-only siblings:

```go
"service_attachment": schema.StringAttribute{
    Optional: true,
    PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
},
```

internal/providers/pluginfw/products/networks/privateendpointrule/resource.go
```diff
@@ -0,0 +1,268 @@
...
+	CreatePrivateEndpointRule(ctx context.Context, req settings.CreatePrivateEndpointRuleRequest) (*settings.NccPrivateEndpointRule, error)
+	GetPrivateEndpointRule(ctx context.Context, req settings.GetPrivateEndpointRuleRequest) (*settings.NccPrivateEndpointRule, error)
+	UpdatePrivateEndpointRule(ctx context.Context, req settings.UpdateNccPrivateEndpointRuleRequest) (*settings.NccPrivateEndpointRule, error)
+	DeletePrivateEndpointRule(ctx context.Context, req settings.DeletePrivateEndpointRuleRequest) (*settings.NccPrivateEndpointRule, error)
+}
+
+type resourcePrivateEndpointRule struct {
+	api apiClient
+
+	// backoff controls the Create polling loop. It is wxposed as a field
```
Divyansh-db: Nit: typo — `wxposed` → `exposed`.

Divyansh-db (approved):

PR #5821, opened 2026-06-20
Try to fix integration test for data notification destination and entitlements

alexott:
Integration tests are failing in the last few days, so this is a fix for them

NO_CHANGELOG=true

## Tests

- [ ] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/internal/providers/pluginfw/products/notificationdestinations/data_notification_destinations_acc_test.go
+++ b/internal/providers/pluginfw/products/notificationdestinations/data_notification_destinations_acc_test.go
@@ -1,13 +1,20 @@
 package notificationdestinations_test
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"testing"
+	"time"
 
-	"github.com/databricks/terraform-provider-databricks/internal/acceptance"
+	"github.com/databricks/databricks-sdk-go/service/settings"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


scim/entitlement_provider_config_test.go
```diff
@@ -56,6 +56,7 @@ func TestMwsAccEntitlements_AccountProvider_ProviderConfig(t *testing.T) {
 
 		resource "databricks_entitlements" "this" {
 			service_principal_id  = databricks_service_principal.this.id
+			workspace_access      = true
```
tanmay-db, line 59: Why are we adding this? Is this some recent API change?

internal/providers/pluginfw/products/notificationdestinations/data_notification_destinations_acc_test.go
```diff
@@ -38,33 +45,76 @@ func CheckDataSourceNotificationsPopulated(t *testing.T) func(s *terraform.State
...
 
-func TestAccNotificationsCreation(t *testing.T) {
-	acceptance.WorkspaceLevel(t, acceptance.Step{
-		Template: `
-			resource "databricks_notification_destination" "email_notification" {
-				display_name = "email notification destination"
-				config {
-					email {
-					addresses = ["abc@gmail.com"]
+func waitForNotificationsInList(t *testing.T) func(s *terraform.State) error {
```
tanmay-db, line 53: Does notification creation take time? I am wondering if this should be raised with API team (LRO) instead of retry/wait on the client side

tanmay-db (commented):
Hi @alexott, thanks for the PR, could you add more context in the PR description regarding the change / if there is a recent API behaviour change?  \+ @rauchy for integration tests failures

PR #5824, opened 2026-06-22
Fix privilege swap in databricks_grant by splitting PATCH into

cereddy:
Changes

  Fix databricks_grant and databricks_grants failing when changing a principal's privileges
  from one set to another (e.g. SELECT → MODIFY) in a single terraform apply.

  The Databricks API rejects a PATCH request that both adds and removes privileges for the  
  same principal in one call. The fix splits the update into two sequential API calls:      
  removes first, then adds. This mirrors the previous workaround of deleting the grant and  
  recreating it, but happens transparently within a single apply.

  Tests

  - make test run locally
  - relevant change in docs/ folder
  - covered with integration tests in internal/acceptance
  - using Go SDK
  - using TF Plugin Framework
  - has entry in NEXT_CHANGELOG.md file

  ---
  A few notes on the unchecked boxes:
  - docs/: no doc change needed — the resource behaviour is the same from the user's        
  perspective, it just stops erroring.
  - integration tests: ideally you'd add one in internal/acceptance, but that requires a    
  live Databricks workspace. Fine to leave unchecked and let the maintainers decide if they 
  want to add one.
  - Go SDK / TF Plugin Framework: not applicable here, the change is in the SDKv2
  permissions layer.


tanmay-db (commented):
Hi @cereddy, thanks for reaching out, could you please tell if this is something that never worked or was working and recently started having issues? 

> 	// The Databricks API rejects requests that both add and remove privileges for the
	// same principal in one call. Split into removes-first then adds to avoid this.

cereddy:
Hi @tanmay-db, I've had this issue multiple times in the last 2/3 years, so I think it has never worked.
Thank you

cereddy:
Hi @tanmay-db, I've pushed an additional commit with tests:
  - Direct unit tests for UpdatePermissions in catalog/permissions/permissions_test.go 
  - Fixed TestGrantCreate, TestGrantCreateMetastoreId, and TestWaitUntilReady in resource_grants_test.go which were still expecting the old single combined PATCH
  - Added TestUcAccGrantSwapPrivilege acceptance test for the end-to-end scenario

PR #5829, opened 2026-06-24
[Fix] Send enabled=false on databricks_ip_access_list create and update

tejaskochar-db:
## Changes

`databricks_ip_access_list` silently ignored `enabled = false`. The attribute was dropped from both the create (`POST`) and update (`PATCH`) request bodies, so a list configured as disabled was created — and left — enabled. Because the API defaults `enabled` to `true`, this produced a permanent plan diff (`true -> false`) that `apply` never resolved: the corrective `PATCH` also omitted `enabled`.

There are two distinct causes:

- **Update:** `settings.UpdateIpAccessList.Enabled` is a `bool` tagged `omitempty`. The SDK marshaler omits a zero-valued field (`false`) unless its name is listed in `ForceSendFields`, and the resource never force-sent it — so `enabled: false` never reached the wire. (`enabled: true` was unaffected: a non-zero value isn't subject to `omitempty`.)
- **Create:** `settings.CreateIpAccessList` has no `Enabled` field at all — the create API always creates a list enabled, so disabling requires a follow-up update.

### Fix

- **Update:** force-send `enabled`, so the configured value is always written.
- **Create:** after creating, if the list is configured `enabled = false`, issue a follow-up update to disable it. This mirrors `databricks_external_location`, which sets `owner`/`isolation_mode` through a follow-up update because the create call cannot carry them. If the follow-up fails, the error is wrapped to make clear the list was created but not disabled, and the resource ID is retained so a later apply reconciles it (rather than orphaning the created list).

### Why force-sending `enabled` is safe

`enabled` has a schema default of `true`, so the value handed to the SDK is always the effective desired state — never an accidental zero value:

| User config | Resolved value | Sent to API |
|---|---|---|
| (omitted) | `true` (schema default) | `true` |
| `enabled = true` | `true` | `true` |
| `enabled = false` | `false` | `false` |

For `true` (explicit or via the default) the request body is unchanged from before this PR — a non-zero bool was always serialized. The only behavioral change is that `false` is now sent instead of being dropped. The create-side follow-up update likewise fires only when the resolved value is `false`, so no extra request is made in the common (enabled) case.

## Tests

- [x] `make test` run locally (`access` package; new unit tests below)
- [ ] relevant change in `docs/` folder (not needed — the docs already document `enabled` defaulting to `true`; this restores that documented behavior)
- [x] using Go SDK

New unit tests assert the exact request bodies via `assert.JSONEq`:

- `TestIPACLUpdate_Disabled` — the `PATCH` body now contains `"enabled": false` (this test fails before the fix).
- `TestIPACLCreate_Disabled` — create issues the `POST`, then a follow-up `PATCH` carrying `"enabled": false`.
- `TestIPACLCreate_Disabled_UpdateError` — when the follow-up update fails, the wrapped error surfaces and the resource ID is retained.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bug Fixes
 
 * Fix permanent permissions drift when `user_name` casing in `databricks_permissions` `access_control` blocks differs from the API response ([#5757](hattps://github.com/databricks/terraform-provider-databricks/issues/5757)).
+* Fix `databricks_ip_access_list` ignoring `enabled = false` on create and update ([#5829](https://github.com/databricks/terraform-provider-databricks/pull/5829))
 
 ### Documentation
 
--- a/access/resource_ip_access_list.go
+++ b/access/resource_ip_access_list.go
@@ -2,7 +2,9 @@ package access
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/databricks/databricks-sdk-go"
 	"github.com/databricks/databricks-sdk-go/service/settings"
 	"github.com/databricks/terraform-provider-databricks/common"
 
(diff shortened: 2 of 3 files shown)


hectorcast-db (changes requested):
This won't work. The underlying issue with with Terraform itself. The "Get" method used to read the value cannot differentiate between `false` and `not set by the user`.
https://github.com/hashicorp/terraform-plugin-sdk/issues/817

The only solution is to migrate to Plugin Framework.

PR #5831, opened 2026-06-25
Make notification destination acceptance tests robust to list eventual consistency

tejaskochar-db:
## Changes

`TestAccNotificationsCreation` created a notification destination and then asserted it appeared in the `databricks_notification_destinations` data source within the same apply. The list API is eventually consistent, so the list can lag the create: the just-created destination is not returned yet and the assertion fails, which has made these acceptance tests fail consistently.

This splits the test so neither path depends on a freshly created destination being listed:

- `TestAccNotificationDestinationCreate` verifies creation through the get-by-id path, which is read-after-write consistent, asserting directly on resource state.
- `TestAccNotificationDestinationsDataSource` validates the data source against a destination already present on the workspace (`TEST_NOTIFICATION_DESTINATION_ID`); it skips when that fixture is not configured.

## Tests

`TestAccNotificationDestinationCreate` passes against a live workspace. `TestAccNotificationDestinationsDataSource` runs once the fixture destination is present and skips otherwise.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -15,3 +15,5 @@
 ### Exporter
 
 ### Internal Changes
+
+* Make notification destination acceptance tests robust to the eventual consistency of the notification destinations list API.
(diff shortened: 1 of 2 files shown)


alexott:
https://github.com/databricks/terraform-provider-databricks/pull/5821 ?

hectorcast-db (approved):

PR #5833, opened 2026-06-25
Share the inert-path classifier so fork/manual integration runs can skip docs-only PRs

chrisst:
## What

Extracts the inert-path allowlist + matcher into one shared script, `.github/scripts/classify_inert.mjs` (+ tests), and points the existing `integration-tests.yml` jobs (`detect-changes`, `carry-forward`) at it instead of each keeping its own copy.

## Why

Skip tests was working for maintainer prs but not for contributor prs via the manual integration test workflow. This allows me to extend this to the contributor prs as well.

## Notes

- `detect-changes` classifies the PR's changed files (paginated `listFiles`, >300 files ⇒ non-inert); `carry-forward` imports the same `isInert`. All other behavior is unchanged.
- Fail-open: any error classifies as non-inert, so a bug can never mint a false "passed" check.
- Tests 13/13; actionlint + conftest pass.

NO_CHANGELOG=true

### DIFF
--- a/.github/scripts/classify_inert.mjs
+++ b/.github/scripts/classify_inert.mjs
@@ -1,22 +1,11 @@
-// Shared inert-path classifier — the single source of truth for which changed
-// files are "inert" (documentation and repository metadata that can never
-// affect an integration-test outcome). It is consumed in three places:
-//
-//   1. integration-tests.yml `detect-changes` (this repo, via github-script)
-//   2. integration-tests.yml `carry-forward`  (this repo, via github-script)
-//   3. The internal fork-PR / manual-dispatch integration-test workflow, which
-//      sparse-checks this exact file out of `main` and imports it so those runs
-//      share the same allowlist.
-//
-// Keeping the allowlist and the matching logic here — and nowhere else — stops
-// the call sites from drifting apart. The file is intentionally dependency-free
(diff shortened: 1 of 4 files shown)


simonfaltum (commented):
Code review finding:

High: `.github/workflows/integration-tests.yml` checks out `.github/scripts/classify_inert.mjs` from the PR merge commit in both `detect-changes` and `carry-forward`, then trusts that file to decide whether to skip or carry forward the required `Integration Tests` check. For `pull_request`, `actions/checkout` defaults to the PR merge branch, so a future PR can modify `classify_inert.mjs` to return `true` for code changes and have this workflow use that unmerged policy to mint/carry a green integration check.

Please check out the classifier from a trusted ref, e.g. `github.event.pull_request.base.sha` or `origin/main`, in both jobs. If the file is missing during this bootstrap PR, fail open to `inert_only=false` / `carried=false`.

Validation I ran locally:
- `node --test .github/scripts/classify_inert.test.mjs` passes, 13/13.
- `node --check` on both `.mjs` files passes.
- `git diff --check origin/main...HEAD` passes.

`actionlint` is not installed locally, so I did not reproduce that check.

simonfaltum (approved):
Reviewed; no blocking findings.

PR #5834, opened 2026-06-25
Allow setting `user_api_scopes = []` on `databricks_app` to disable OBO authorization

Divyansh-db:
## Changes

`databricks_app` could not set `user_api_scopes = []` (to disable On-Behalf-Of user authorization). The apply failed with `Provider produced inconsistent result after apply ... user_api_scopes: was [], but now null`, and the configured empty list otherwise produced a perpetual diff.

**Root cause:** the Apps API does not echo `user_api_scopes` back in its response when OBO authorization is inactive, so the response deserializes to a null list. Terraform treats a known empty list and `null` as different values, and `user_api_scopes` is an `Optional` (non-`Computed`) attribute, so the post-apply state (`null`) does not match the configured value (`[]`), which Core rejects.

This change reconciles the configured value into state when the API omits the field, mirroring the `SyncFields` reconciliation generated for `databricks_app_space`. The guard is intentionally **narrow** — it only restores a known, non-null, **empty** list (an empty list is indistinguishable from "unset" on the wire, so we never invent one) — and it is applied in Create, Read, and Update.

## Tests

- Unit test `TestReconcileEmptyUserApiScopes` covering all branches (configured empty / non-empty / null / unknown × API null / empty / values).
- Acceptance test `TestAccAppResource_EmptyUserApiScopes` creating a `databricks_app` with `user_api_scopes = []` and asserting a stable, empty plan (no inconsistent-result error, no perpetual diff).

_This pull request and its description were written by Isaac, a Claude Code AI agent._

### DIFF
--- a/internal/providers/pluginfw/products/app/resource_app_test.go
+++ b/internal/providers/pluginfw/products/app/resource_app_test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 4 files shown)


internal/providers/pluginfw/products/app/resource_app_test.go
```diff
@@ -63,6 +65,39 @@ func TestResourceApp_SchemaPreserved(t *testing.T) {
...
+		// API reports values (e.g. OBO active / out-of-band change) -> trust the API.
+		{"configured empty, api has values -> keep api", empty, sql, sql},
+		{"configured empty, api empty -> keep api empty", empty, empty, empty},
+		// Narrow guard: a non-empty configured value is never restored from null.
+		{"configured non-empty, api null -> keep api null", sql, null, null},
+		// Unset stays unset; we must not invent an empty list.
+		{"configured null, api null -> keep api null", null, null, null},
+		{"configured null, api has values -> keep api", null, sql, sql},
+		// Unknown (e.g. interpolated) is not treated as a known empty list.
+		{"configured unknown, api null -> keep api null", unknown, null, null},
```
mihaimitrea-db: [nit] I find the test easier to read if they are one argument per line like this:

```
{
   "configured empty, api null -> restore empty",
    empty,
    null,
    empty
}
```

mihaimitrea-db (approved):
LGTM, please fix styling of the tests before merging.

PR #5837, opened 2026-06-27
[Fix] Retry on potential eventual consistency case in permissions

### LINKED ISSUE #5827: [ISSUE] `databricks_permissions` still produces "inconsistent final plan" on the access_control set (SQL warehouse, v1.118.0, after #5091)
### Configuration

A `databricks_permissions` on a SQL warehouse whose `access_control` mixes a static group block with dynamic service-principal blocks:

```hcl
resource "databricks_permissions" "sql_endpoint_usage" {
  sql_endpoint_id = databricks_sql_endpoint.this.id

  access_control {
    group_name       = "users"
    permission_level = "CAN_USE"
  }

  dynamic "access_control" {
    for_each = var.service_principal_application_ids # set of SP application IDs
    content {
      service_principal_name = access_control.value
      permission_level       = "CAN_USE"
    }
  }
}
```

### Expected Behavior

Apply succeeds; the `access_control` set is consistent between plan and apply.

### Actual Behavior

Intermittently, apply fails with:

```
Error: Provider produced inconsistent final plan

When expanding the plan for module….databricks_permissions.sql_endpoint_usage to include
new values learned so far during apply, provider produced an invalid new value for
.access_control: planned set element cty.ObjectVal(map[string]cty.Value{
  "group_name":cty.StringVal("users"), "permission_level":cty.StringVal("CAN_USE"),
  "service_principal_name":cty.NullVal(cty.String), "user_name":cty.NullVal(cty.String)})
does not correlate with any element in actual.
```

### Steps to Reproduce

1. `terraform apply` the config above as part of a larger plan.
2. Intermittently it fails on the permissions resource with the above error.
3. Re-run `terraform apply` — it usually succeeds.

- Terraform `>= 1.12`
- `databricks/databricks` provider `1.118.0`

### Is it a regression?

The same "inconsistent final plan in the Permissions resource" was addressed in #5091 (merged 2025-10-22), which is included in v1.118.0 (released 2026-06-17) — yet it still occurs here. #5091's description notes it was an *attempt* to fix an *intermittent* issue, so this looks like an incomplete fix for the SQL-warehouse / `users`-group case. Older related closed issues: #2608, #2159.

### Debug Output

The block above is what the apply surfaces; happy to attach a `TF_LOG=DEBUG` gist on request.

### Important Factoids

- Intermittent — some applies fail, a re-run passes.
- `access_control` is a set mixing a static `group_name = "users"` element and dynamic `service_principal_name` elements; the element that fails to correlate is the static `users`/`CAN_USE` one.

Happy to help validate/test; this looks like a provider-internal set-consistency issue (likely the same root cause #5091 targeted), so deferring to maintainers on the approach.

alexott:
## Changes

@tanmay-db it looks like another case of eventual consistency in #5827

Resolves #5827

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug Fixes
 
+* Fix intermittent inconsistent final plan errors in `databricks_permissions` for SQL warehouse access control reads after permission updates ([#5837](https://github.com/databricks/terraform-provider-databricks/issues/5837)).
 * Fix permanent permissions drift when `user_name` casing in `databricks_permissions` `access_control` blocks differs from the API response ([#5757](hattps://github.com/databricks/terraform-provider-databricks/issues/5757)).
 
 ### Documentation
--- a/permissions/resource_permissions.go
+++ b/permissions/resource_permissions.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/service/iam"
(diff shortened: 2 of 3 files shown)


hectorcast-db (commented):
These are a bit of a cat and mouse problem, where we keep trying to patch API inconsistencies. We should focus on fixing the API, which should either wait to return or explicitely use async annotations. 

For Permissions in particular, this seems more problematic to me than the rest. Isn't this a Singleton? If so, a race condition by some other component updating the permissions would cause TF to just hand forever waiting for a sync which will never happen.

PR #5845, opened 2026-06-30
Add git_credential_id attribute to databricks_repo resource

jamiejackson:
The Databricks platform now requires explicit credential selection when creating repos via the API (GIT_CLI_CANNOT_CHOOSE_CREDENTIAL error). The Repos API accepts an undocumented git_credential_id parameter that resolves this, but the databricks_repo resource does not expose it.

This adds an optional git_credential_id attribute (ForceNew) that is passed through to POST /api/2.0/repos on resource creation. When unset, behavior is unchanged (omitempty ensures the field is not serialized).

## Changes

- Added `GitCredentialID int64` to `reposCreateRequest` struct with `json:"git_credential_id,omitempty"`
- Added `git_credential_id` as an optional, ForceNew `schema.TypeInt` field in the resource schema
- Wired the field into the Create function
- Added `TestResourceRepoCreateWithGitCredentialID` unit test
- Added `TestAccRepo_WithGitCredentialID` acceptance test
- Added docs for the new attribute in `docs/resources/repo.md`
- Added changelog entry in `NEXT_CHANGELOG.md`

**Future consideration:** A `databricks_git_credentials` data source would allow looking up the credential ID dynamically rather than hardcoding it. Not included here to keep scope tight, but would be a natural follow-up.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK

Unit test added (`TestResourceRepoCreateWithGitCredentialID`). Acceptance test added (`TestAccRepo_WithGitCredentialID`) -- creates a `databricks_git_credential` and references its ID in `databricks_repo`. End-to-end tested against a production workspace exhibiting the GIT_CLI_CANNOT_CHOOSE_CREDENTIAL error -- repo creation succeeds with the credential ID specified.

Note: This uses the legacy SDKv2 resource pattern, matching the existing `databricks_repo` implementation.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### New Features and Improvements
 
+* Add optional `git_credential_id` attribute to `databricks_repo` resource to allow explicit credential selection when automatic selection fails (`GIT_CLI_CANNOT_CHOOSE_CREDENTIAL`).
 * Deprecate the SDKv2 fallback implementations of `databricks_library`, `databricks_quality_monitor`, and `databricks_share` resources, and the `databricks_share`, `databricks_shares`, and `databricks_volumes` data sources. These resources have been served by the Plugin Framework by default since their migration; the SDKv2 implementations remain only as opt-in fallbacks via the `USE_SDK_V2_RESOURCES` / `USE_SDK_V2_DATA_SOURCES` environment variables. Setting either environment variable for any of these names now emits a runtime warning (visible with `TF_LOG=WARN` or higher), and the SDKv2 implementations will be removed in the next major release of the provider.
 
 ### Bug Fixes
--- a/docs/resources/repo.md
+++ b/docs/resources/repo.md
@@ -30,6 +30,7 @@ The following arguments are supported:
 
 * `url` -  (Required) The URL of the Git Repository to clone from. If the value changes, Git folder is re-created.
 * `git_provider` - (Optional, if it's possible to detect Git provider by host name) case insensitive name of the Git provider.  Following values are supported right now (could be a subject for a change, consult [Repos API documentation](https://docs.databricks.com/dev-tools/api/latest/repos.html)): `gitHub`, `gitHubEnterprise`, `bitbucketCloud`, `bitbucketServer`, `azureDevOpsServices`, `gitLab`, `gitLabEnterpriseEdition`, `awsCodeCommit`.
+* `git_credential_id` - (Optional) The ID of the Git credential (`databricks_git_credential`) to use for this repo.  Required when the workspace has multiple credentials or when automatic credential selection fails (`GIT_CLI_CANNOT_CHOOSE_CREDENTIAL`).  If the value changes, the Git folder is re-created.
 * `path` - (Optional) path to put the checked out Git folder. If not specified, , then the Git folder will be created in the default location.  If the value changes, Git folder is re-created.
 * `branch` - (Optional) name of the branch for initial checkout. If not specified, the default branch of the repository will be used.  Conflicts with `tag`.  If `branch` is removed, and `tag` isn't specified, then the repository will stay at the previously checked out state.
 * `tag` - (Optional) name of the tag for initial checkout.  Conflicts with `branch`.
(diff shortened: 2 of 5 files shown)


tanmay-db:
> now requires explicit credential selection when creating repos via the API

Hi @jamiejackson , thanks for reaching out. I am trying to understand the root cause of the problem. Could you confirm if the issue is a regression on terraform provider or from the API? i.e. was the terraform apply working before and now not working on the same version?

jamiejackson:
The HCL/provider hadn't changed, but starting recently, it could no longer be used to create new `databricks_repo` resources, as noted.

Between the last successful run and the failure, there had been a Databricks upgrade.

Therefore, it seems that the API changed with the upgrade, or the upgrade caused conditions that prevented the repo from creating without the explicit credential ID.

GrantIsEaton:
Hi @jamiejackson @tanmay-db thanks for your attention to this issue. 

**TLDR**
This PR should not be needed, the issue that caused users who have opted into the Git CLI preview in their admin settings preview to get error `GIT_CLI_CANNOT_CHOOSE_CREDENTIAL` unexpectedly has been resolved. 

~~**Why we dont need this PR**~~
~~1. the underlying issue is fixed and rolled out as of end of last week~~
~~2. UpdateRepo and CreateRepo APIs (plus terraform) should be getting access & documentation to the git_credential_id optional parameter. Current ETA is EOD tomorrow for public APIs and wednesday EOD for terraform. These properties are optional and really are not needed given 1.~~

Please feel free to tag me if any future issues arise.

edit: actually we do need this PR so thank you. I thought this code would be autogenerated on our side but apparently it is manual for now. let's proceed with the review.

docs/resources/repo.md
```diff
@@ -30,6 +30,7 @@ The following arguments are supported:
 
 * `url` -  (Required) The URL of the Git Repository to clone from. If the value changes, Git folder is re-created.
 * `git_provider` - (Optional, if it's possible to detect Git provider by host name) case insensitive name of the Git provider.  Following values are supported right now (could be a subject for a change, consult [Repos API documentation](https://docs.databricks.com/dev-tools/api/latest/repos.html)): `gitHub`, `gitHubEnterprise`, `bitbucketCloud`, `bitbucketServer`, `azureDevOpsServices`, `gitLab`, `gitLabEnterpriseEdition`, `awsCodeCommit`.
+* `git_credential_id` - (Optional) The ID of the Git credential (`databricks_git_credential`) to use for this repo.  Required when the workspace has multiple credentials or when automatic credential selection fails (`GIT_CLI_CANNOT_CHOOSE_CREDENTIAL`).  If the value changes, the Git folder is re-created.
```
GrantIsEaton, line 33:
```suggestion
* `git_credential_id` - (Optional) The ID of the Git credential (`databricks_git_credential`) to use for authentication with the Git provider remote.
```

let's just keep it plain and simple. we dont actually need to specify this param if you have multiple git credentials, you can still rely on git credential defaults (this was a bug before that we fixed, it was not working)

NEXT_CHANGELOG.md
```diff
@@ -5,8 +5,8 @@
 ### Breaking Changes
 
 ### New Features and Improvements
-* Add resource and data source for `databricks_postgres_data_api`.
 
+* Add optional `git_credential_id` attribute to `databricks_repo` resource to allow explicit credential selection when automatic selection fails (`GIT_CLI_CANNOT_CHOOSE_CREDENTIAL`).
```
GrantIsEaton, line 9:
```suggestion
* Add optional `git_credential_id` attribute to `databricks_repo` resource to allow explicit credential selection
```

repos/resource_repo.go
```diff
@@ -177,6 +178,11 @@ func ResourceRepo() common.Resource {
 			ConflictsWith: []string{"branch"},
 			ValidateFunc:  validation.StringIsNotWhiteSpace,
 		}
+		s["git_credential_id"] = &schema.Schema{
+			Type:     schema.TypeInt,
+			Optional: true,
+			ForceNew: true,
```
GrantIsEaton, line 184: i dont think we need this. reason being if you specify credential id 1 or credential id 2 the resulting repo / git folder is the same.

I would say it makes more sense that if you need a way to pick a credential for a certain repository url to wait until we offer the association endpoint (or just always pass the right credential id)

GrantIsEaton:
opened https://github.com/databricks/terraform-provider-databricks/pull/5877 to address feedback

PR #5855, opened 2026-07-07
[Fix] Fix updating a column comment on a VIEW in databricks_sql_table

### LINKED ISSUE #4250: [ISSUE] Issue with `databricks_sql_table` resource
### Configuration

Create any view using the `databricks_sql_table` resource which includes a column comment, and then change this comment and try to re-apply terraform.

```hcl
resource "databricks_sql_table" "my_view" {
  cluster_id = local.cluster_id

  name = "view"
  catalog_name = databricks_catalog.this.name
  schema_name = databricks_schema.this.name
  table_type = "VIEW"

  view_definition = format("SELECT name FROM %s WHERE id == 1", databricks_sql_table.table.id)

  comment = "A view managed by terraform"

  column {
     name = "name"
     comment = "This comment is changed after the view is created"
  }
}
```

### Expected Behavior

When changing a column comment on a view and running `terraform apply`, the view would be successfully updated with the new comment.

### Actual Behavior

When applying the change to a column comment, terraform tries to modify the resource and fails with an error like:

```
Error: cannot update sql table: cannot execute ALTER VIEW `my_catalog`.`my_schema`.`my_view` ALTER COLUMN `name` COMMENT 'An updated column comment.': [PARSE_SYNTAX_ERROR] Syntax error at or near 'ALTER'.(line 1, pos 72)
```

### Steps to Reproduce

```
$ tf version
Terraform v1.7.5
on windows_amd64
+ provider registry.terraform.io/azure/azapi v1.15.0
+ provider registry.terraform.io/databricks/databricks v1.58.0
+ provider registry.terraform.io/hashicorp/azurerm v3.109.0
+ provider registry.terraform.io/hashicorp/null v3.2.3
```

### Is it a regression?
No 

### Important Factoids

This is partly a databricks issue, because you can't run `ALTER VIEW` at all - the same issue actually occurs if you try and edit column comments directly in the databricks UI. In the absence of databricks fixing this, would it be possible for the provider to be updated so that changes to column comments force replacement of views? There's no harm in replacing views, since they don't contain data so this would be safe to do, and would make it much easier to manage views with column comments using terraform. Currently we have to manually `taint` or delete views before applying whenever changing comments, which is quite frustrating

graph226:
## Changes

Changing a column comment on a `databricks_sql_table` with `table_type = "VIEW"` produced a perpetual, un-appliable diff. On update the provider generated:

```sql
ALTER VIEW `cat`.`sch`.`v` ALTER COLUMN `c` COMMENT '...'
```

which is **not valid Databricks SQL** — `ALTER VIEW` has no `ALTER COLUMN` clause — and fails at apply time with:

```
[PARSE_SYNTAX_ERROR] Syntax error at or near 'ALTER'. SQLSTATE: 42601
```

So the comment change could never be applied: every `terraform plan` showed the same diff and every `apply` errored.

Databricks *does* support setting a column comment on a view — just via a different statement, `COMMENT ON COLUMN` (Unity Catalog allows `relation_name` to be a table, streaming table, view, or materialized view). This PR emits `COMMENT ON COLUMN` for views while keeping `ALTER TABLE ... ALTER COLUMN ... COMMENT` for tables:

```sql
COMMENT ON COLUMN `cat`.`sch`.`v`.`c` IS '...'
```

The change is applied **in place** — the view is not recreated, so grants and dependent objects are preserved. This matches the "Expected Behavior" described in #4250 (the view is updated with the new comment) rather than forcing a replacement.

Fixes #4250.

## Verified against a live workspace

**1. Raw SQL** — the statement the released provider emits vs. the one this PR emits, run against a real Unity Catalog SQL warehouse (temporary view, created and dropped):

| Statement | Result |
|---|---|
| `ALTER VIEW … ALTER COLUMN id COMMENT '…'` (released provider) | ❌ `PARSE_SYNTAX_ERROR` SQLSTATE 42601 |
| `` COMMENT ON COLUMN `cat`.`sch`.`v`.`id` IS '…' `` (this PR) | ✅ succeeds; comment updated in place |

**2. End-to-end with a locally built provider** — built this branch and pointed Terraform at it with a `dev_overrides` block, then ran a real plan/apply cycle against a live workspace:

```hcl
resource "databricks_sql_table" "probe" {
  name            = "…"
  catalog_name    = "…"
  schema_name     = "…"
  table_type      = "VIEW"
  warehouse_id    = "…"
  view_definition = "SELECT nullif(1, 2) AS id"
  column {
    name    = "id"
    comment = var.col_comment # changed between applies
  }
}
```

- `apply` (create) → view created.
- `apply` (change only the column comment) → `Plan: 0 to add, 1 to change, 0 to destroy` (in place, **not** a replacement); the provider logged `Executing Sql: COMMENT ON COLUMN … IS '…'`; `Apply complete! 0 added, 1 changed, 0 destroyed`. This is exactly the operation that fails on the released provider.
- `apply` again → `No changes. Your infrastructure matches the configuration.` (no perpetual diff; clean convergence).
- Confirmed via the Tables API that the column comment was updated.

## Tests

Added `TestResourceSqlTableUpdateView_ColumnComment`, which drives an update that changes a view's column comment and asserts the provider issues `COMMENT ON COLUMN ... IS 'new comment'` and never `ALTER ... ALTER COLUMN`. Existing view/table update tests continue to pass (table column comments still use `ALTER TABLE ... ALTER COLUMN`).

- [x] `make test` run locally (`go test ./catalog/`, `go vet ./catalog/`, `gofmt` — all clean; full `catalog` suite passes)
- [x] using Go SDK


Divyansh-db:
Hi @graph226 , thanks for the contribution. I have added the acceptance tests for the changes for validation and regresstion guard on top of the changes. https://github.com/databricks/terraform-provider-databricks/pull/5881.

Divyansh-db (approved):

PR #5856, opened 2026-07-07
[Fix] Fix column nullability handling on a VIEW in databricks_sql_table

graph226:
## Changes

A view column's nullability is **derived from the view query** and cannot be set independently — but `databricks_sql_table` treated it like a table column, which broke views in three ways:

1. **`CREATE` emitted `NOT NULL` for view columns.** `serializeColumnInfo` appended `NOT NULL` whenever `nullable = false`, so the provider generated:

   ```sql
   CREATE VIEW `cat`.`sch`.`v` (`id`  NOT NULL COMMENT '...') AS ...
   ```

   Views don't support `NOT NULL` column constraints, so the create failed.

2. **`UPDATE` emitted an invalid `ALTER VIEW ... ALTER COLUMN`.** When the configured nullability differed from the server's, `alterExistingColumnStatements` generated:

   ```sql
   ALTER VIEW `cat`.`sch`.`v` ALTER COLUMN `id` DROP NOT NULL
   ```

   `ALTER VIEW` has no `ALTER COLUMN` clause, so this fails with `[PARSE_SYNTAX_ERROR] SQLSTATE: 42601` — the same class of bug as #4250, but on the nullability path.

3. **Permanent plan diff.** A view column's server-reported nullability often differs from the configured value (e.g. `SELECT 1 AS id` yields a `NOT NULL` column, while the schema default is `nullable = true`), so every plan showed a nullability diff that could never be applied cleanly.

### Fix

- `serializeColumnInfo`: don't append `NOT NULL` for views.
- `alterExistingColumnStatements`: don't emit the nullability `ALTER` for views.
- Add a diff suppression on `column.nullable` for views, so the server-derived value doesn't produce a permanent diff.

Nullability handling for **tables** is unchanged (`NOT NULL` in `CREATE`, `ALTER TABLE ... ALTER COLUMN ... SET/DROP NOT NULL` on update).

> Note: this is a sibling of #5855, which fixes the column **comment** path on views (`COMMENT ON COLUMN`). The two are independent — this PR only touches nullability — but together they make view columns behave correctly.

## Verified against a live workspace

Built this branch and pointed Terraform at it with a `dev_overrides` block, then ran real plan/apply cycles against a live workspace with a view whose column is `NOT NULL`-inferred (`SELECT 1 AS id`) and `nullable = false` in config — the exact shape that failed before:

- **Create** → `CREATE VIEW … (`id`  COMMENT '…')` (no `NOT NULL`); `Apply complete! 1 added`. *(Previously: create failed.)*
- **Re-apply (no config change)** → `No changes. Your infrastructure matches the configuration.` *(Previously: a permanent `nullable` diff.)*
- **Change `view_definition`** → only `ALTER VIEW … AS …` is emitted, never `ALTER VIEW … ALTER COLUMN … NOT NULL`; `Apply complete! 1 changed`. *(Previously: the nullability `ALTER` failed with `PARSE_SYNTAX_ERROR`.)*

## Tests

Added `TestResourceSqlTableView_NullableColumnHandling`, asserting that `CREATE VIEW` for a view omits `NOT NULL` and that a nullability change on a view column produces no `ALTER ... ALTER COLUMN` statement. Updated `TestResourceSqlTableCreateStatement_ViewWithComments`, which previously asserted the (invalid) `NOT NULL` output for a view.

- [x] `make test` run locally (`go test ./catalog/`, `go vet ./catalog/`, `gofmt` — all clean; full `catalog` suite passes)
- [x] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -21,6 +21,7 @@
 ### Bug Fixes
 
 * Fix import for jobs with >100 tasks ([#5417](https://github.com/databricks/terraform-provider-databricks/pull/5417)).
+* Fix handling of column nullability on a `VIEW` in `databricks_sql_table` (#PLACEHOLDER). A view column's nullability is derived from the view query and can't be set independently. The provider previously emitted `NOT NULL` in `CREATE VIEW` (which Databricks rejects) and `ALTER VIEW ... ALTER COLUMN ... SET/DROP NOT NULL` on update (which fails with `PARSE_SYNTAX_ERROR`), and it reported a permanent diff when the server-derived nullability differed from the configured value. The `NOT NULL` constraint and the `ALTER` are no longer emitted for views, and nullability diffs are suppressed for view columns.
 
 ### Documentation
 
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -21,7 +21,7 @@
 ### Bug Fixes
 
 * Fix import for jobs with >100 tasks ([#5417](https://github.com/databricks/terraform-provider-databricks/pull/5417)).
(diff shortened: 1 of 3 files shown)


graph226:
Closing in favour of #5958.

While looking at the nullability path I found that it shares a root cause with the column *membership* and *comment* paths: a view's columns are derived from its query, and the provider treats them like table columns throughout. Fixing only nullability here would leave `ALTER VIEW ... ADD/DROP/RENAME COLUMN` still being emitted (and still rejected), and would need rebasing onto that fix afterwards.

#5958 covers everything in this PR plus those cases, and folds in the `VIEW` special case added by #5855 so the whole view-column path lives in one function. Sorry for the churn.

PR #5860, opened 2026-07-07
Allow configuring cascade deletion for pipelines 

ronaldz-db:
## Changes

For pipeline deletion, we offer a [cascade](https://docs.databricks.com/api/workspace/pipelines/delete) parameter that controls whether the pipeline deletion cascades to datasets as well. We are extending support to Terraform, which is a key customer ask.  

Based on DABs core [feedback](https://github.com/databricks/cli/pull/5846#discussion_r3551290057), we will name this new Terraform field for pipelines as `cascade_on_destroy`

We are making a corresponding change in DABs as well: https://github.com/databricks/cli/pull/5846

## Tests

- [x] `make test` run locally
- [x] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### New Features and Improvements
 
+* Added `cascade` attribute to `databricks_pipeline` to control whether deleting a pipeline also deletes its datasets (materialized views, streaming tables, and views). Defaults to `true`; set to `false` to preserve the datasets on delete ([#5860](https://github.com/databricks/terraform-provider-databricks/pull/5860)).
+
 ### Bug Fixes
 
 ### Documentation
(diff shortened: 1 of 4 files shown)


alexott:
I have another PR: https://github.com/databricks/terraform-provider-databricks/pull/5600 that doesn't expose `cascade` directly but instead explicitly preserves tables.

ronaldz-db:
@alexott oh great! There was a bug in the Go SDK that was recently fixed: https://github.com/databricks/databricks-sdk-go/commit/a00222de4b5130d02b0c24dcf038291df7047491

We can now merge either PR, but imo we should keep the `cascade` naming for consistency. Any thoughts?

alexott:
I'm not sure about `cascade`. We were bitten many times by using default values when default is changed on the backend...

ronaldz-db:
Synced offline with Alex - unfortunately we do need to provide a default value here and cannot solely rely on the server default. That said, we do need to be careful here and in DABs for changes to defaults

Issue #4822, opened 2025-06-27
[FEATURE] delta share recipients data source


EMayneKT:
### Use-cases
Get a list of existing delta share receipients using the `list share recipients` endpoint ([documented here](https://docs.databricks.com/api/workspace/recipients/list)) 

### Attempted Solutions
There seems to be no way to get a list of existing recipients.

### Proposal
Implement as a data source under `sharing/data_recipients.go`

chrisst:
Added in https://github.com/databricks/terraform-provider-databricks/pull/5868?timeline_per_page=5

PR #5868, opened 2026-07-09
Add `databricks_recipients` data source

chrisst:
Adds a `databricks_recipients` data source that lists Delta Sharing recipient names in the metastore. Fills the gap called out in #4822 — there's a `databricks_recipient` resource but no way to enumerate existing recipients.

It's a thin Plugin Framework data source modeled on `databricks_shares`: calls `Recipients.ListAll` and returns the names. Includes docs and an acceptance test (`TestUcAccDataSourceRecipients`), which I ran end to end against a UC workspace.


internal/providers/pluginfw/products/sharing/data_recipients.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	w, clientDiags := d.Client.GetWorkspaceClientForUnifiedProviderWithDiagnostics(ctx, workspaceID)
+
+	resp.Diagnostics.Append(clientDiags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	recipients, err := w.Recipients.ListAll(ctx, sharing.ListRecipientsRequest{})
```
Divyansh-db, line 100: The Go SDK states that ordering is not guaranteed by the API. I think we should sort and store to avoid unnecessary diffs or use sets

chrisst: good catch, I'll move to sets. I think this was based on the shares datasource which it appears was originally a set in sdkv2 and regressed when moved to the plugin framework. But that's a battle for another day.

Divyansh-db (approved):

PR #5869, opened 2026-07-10
Add trace_location to databricks_mlflow_experiment

james-fletcher-db:
## What did you change, and why?

Adds a `trace_location` block to `databricks_mlflow_experiment`, letting an experiment's traces be stored in a Unity Catalog schema (`uc_trace_location { catalog, schema, table_prefix }`).

The resource's schema is auto-derived from `ml.Experiment` via `common.StructToSchema`, so the block appears once the SDK carries the field. This PR adds the schema customization the field's contract requires:

- **`trace_location`, `uc_trace_location`, `catalog`, `schema` → `ForceNew`.** The location is immutable server-side (the `UpdateExperiment` RPC only renames), so changing it must replace the experiment.
- **`catalog`, `schema` → `Required`**, with case-insensitive diff suppression (they are UC identifiers, matching how other UC names are handled in the provider).
- **`table_prefix` → plain `Optional` (create-only).** The server echoes it on read only when the user set it, so a plain `Optional` round-trips without drift. When omitted, the server generates a default and the field stays empty on read.
- **`effective_table_prefix` → `Computed`.** Read-only; carries the prefix actually in effect (the user value, or the server default when omitted) — the AIP-129 effective-value split, so the resolved value lives here instead of in the user-owned `table_prefix`.

### Example

```hcl
resource "databricks_mlflow_experiment" "with_uc_traces" {
  name = "${data.databricks_current_user.me.home}/uc-traces-experiment"

  trace_location {
    uc_trace_location {
      catalog      = "my_catalog"
      schema       = "my_schema"
      table_prefix = "my_experiment" # optional; server generates one if omitted
    }
  }
}
```

### Notes for reviewers

- The data source `databricks_mlflow_experiment` also gains `trace_location` as a read attribute (it embeds `ml.Experiment`) — additive, non-breaking.
- **Availability:** the field is server-gated by the `enableMlflowTracingV6PublicApi` SAFE flag, so it only functions on enabled workspaces.

## How do you know it works?

- `go build ./mlflow/...` — green.
- `go test ./mlflow/` — `TestExperimentCreateWithTraceLocation` (create + round-trip of `table_prefix` and `effective_table_prefix`) and `TestExperimentTraceLocationOmittedNoForceNew` (server-set location + config omits the block → no forced replacement) both pass.
- `make fmt` (no reformatting), `make ws` (0 failed), `make lint`/staticcheck (clean).
- `make diff-schema` — **non-breaking**: 1 additive change (the computed `effective_table_prefix`); `trace_location` itself is surfaced by the SDK already on `main`.

### DIFF
--- a/mlflow/resource_mlflow_experiment.go
+++ b/mlflow/resource_mlflow_experiment.go
@@ -33,6 +33,21 @@ func ResourceMlflowExperiment() common.Resource {
 			}
 			common.CustomizeSchemaPath(m, "artifact_location").SetForceNew().SetSuppressDiff()
 			common.CustomizeSchemaPath(m, "name").SetRequired().SetCustomSuppressDiff(experimentNameSuppressDiff)
+			// Immutable server-side, so a change must replace. Computed so an
+			// experiment adopted without the block keeps its server value instead of
+			// diffing to zero and forcing replacement.
+			common.CustomizeSchemaPath(m, "trace_location").SetForceNew().SetComputed()
+			common.CustomizeSchemaPath(m, "trace_location", "uc_trace_location").SetForceNew().SetComputed()
+			// Case-insensitive identifiers: suppress case-only diffs so a normalized
+			// server echo does not force replacement.
+			common.CustomizeSchemaPath(m, "trace_location", "uc_trace_location", "catalog").
+				SetForceNew().SetRequired().SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
+			common.CustomizeSchemaPath(m, "trace_location", "uc_trace_location", "schema").
+				SetForceNew().SetRequired().SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
+			// Echoed on read only when the user set it, so plain Optional round-trips
+			// without drift; the resolved value is read from effective_table_prefix.
+			common.CustomizeSchemaPath(m, "trace_location", "uc_trace_location", "table_prefix").SetForceNew().SetOptional()
+			common.CustomizeSchemaPath(m, "trace_location", "uc_trace_location", "effective_table_prefix").SetComputed()
 			// the API never accepts description, but we need to keep this for backwards compatibility
 			m["description"] = &schema.Schema{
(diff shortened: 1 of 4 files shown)


mlflow/resource_mlflow_experiment.go
```diff
@@ -33,6 +33,21 @@ func ResourceMlflowExperiment() common.Resource {
 			}
 			common.CustomizeSchemaPath(m, "artifact_location").SetForceNew().SetSuppressDiff()
 			common.CustomizeSchemaPath(m, "name").SetRequired().SetCustomSuppressDiff(experimentNameSuppressDiff)
+			// Immutable server-side, so a change must replace. Computed so an
+			// experiment adopted without the block keeps its server value instead of
+			// diffing to zero and forcing replacement.
+			common.CustomizeSchemaPath(m, "trace_location").SetForceNew().SetComputed()
+			common.CustomizeSchemaPath(m, "trace_location", "uc_trace_location").SetForceNew().SetComputed()
```
james-fletcher-db: @hectorcast-db The `trace_location` / `uc_trace_location` blocks are currently set `Computed` so an imported/adopted experiment retains its server-set location in state. If you think it would be better to drop the `Computed`, the only cost is import not persisting that location to Terraform state, though it still exists on the experiment. It should not be destructive or cause a spurious `ForceNew`.

hectorcast-db: As far as I understand, whether it is computed or not, `ForceNew` will cause TF to destroy and recreate the resource if the server sends any value which is not in the user template. 

The main difference is that without compute, it will happen immediately, and without it, it will be delayed to the next update.

james-fletcher-db: Sounds like it would be best to drop the Computed tag, just updated it. 

I also added a test to check the behavior when the server sends a value not in the user template - I wasn't able to show it forcing a destroy and recreate in my testing, but I could be misunderstanding how it works.

hectorcast-db (approved):

PR #5877, opened 2026-07-13
Add git_credential_id attribute to databricks_repo resource

GrantIsEaton:
## Changes

Adds an optional `git_credential_id` attribute to the `databricks_repo` resource. This lets you always specify, by ID, which Git credential (`databricks_git_credential`) is used to authenticate with the Git provider remote, if desired. When it's not set, behavior is unchanged (the field is omitted from requests via `omitempty`).

The attribute is wired into both the create and update paths, so it can be set when a repo is created and changed in place afterwards.

Note that `git_credential_id` is a **per-operation** parameter rather than persisted repo state — the API uses it to authenticate the specific clone/update request and does not return it from `GetRepo`. So it is sent on every create and on every update that has one configured (not only when the value changes), otherwise a branch-only update would authenticate with the caller's default credential instead. Because the API never returns it, Terraform cannot detect drift on this attribute.

- Added `GitCredentialID` to `reposCreateRequest` with `json:"git_credential_id,omitempty"`
- Added `git_credential_id` as an optional `schema.TypeInt` field in the resource schema
- Wired the field into the Create path, the post-clone checkout request, and the Update path
- Added `TestResourceRepoCreateWithGitCredentialID`, `TestResourceRepoUpdateGitCredentialID`, and `TestResourceRepoUpdateBranchSendsUnchangedGitCredentialID` unit tests
- Added docs for the new attribute in `docs/resources/repo.md`
- Added changelog entry in `NEXT_CHANGELOG.md`

No acceptance test: acceptance tests run in parallel against a shared workspace, and `TestAccGitCredentials` asserts the workspace holds exactly one Git credential, so a test that creates a second one races it. The unit tests assert the exact create/update request bodies instead.

## Tests

- [x] `make test` run locally
- [ ] using Go SDK


repos/resource_repo.go
```diff
@@ -46,10 +46,11 @@ func (r ReposInformation) RepoID() string {
 }
 
 type reposCreateRequest struct {
```
alexott, line 48: ideally we should just switch to Go SDK structs here

tanmay-db: Hi @alexott , I agree but given the migration to Go SDK might become a rabbit hole, I think it's fine to just add a field here for now.

alexott: Agree. let's address it separately

GrantIsEaton: Thanks @alexott — I went ahead and built out the full migration to the Go SDK structs so we could see what it actually takes, in a stacked demonstration PR: #5883.

Short version: it's mechanically doable but not free, and I'd lean toward keeping this PR small and doing the migration as a follow-up (which also matches @tanmay-db's read above).

The main frictions I hit:

- **The schema can't move to the SDK struct.** The schema is generated from `ReposInformation` via `StructToSchema` and relies on TF aliases (`git_provider` ← `provider`, `commit_hash` ← `head_commit_id`) that `workspace.RepoInfo` doesn't carry. Driving the schema off the SDK struct would silently rename user-facing attributes. So the SDK structs can only take over the request/response path; the schema struct stays, and `Read` hand-maps between the two.
- **Test churn.** The SDK appends a `?` to GET/DELETE URLs, so every HTTP fixture (in `repos` *and* `exporter`) had to change, and the provider-local `List`/`ListAll` + `ReposListResponse` had to go (listing already goes through the SDK in the exporter). Net ~+123/−242 across 5 files, plus the unified-host acceptance test's whole premise (raw-HTTP vs SDK) had to be rewritten.
- **Risk profile.** This PR is a tiny additive field; the migration rewrites the entire CRUD path for *all* `databricks_repo` users, so it deserves its own review and real acceptance-test runs rather than riding along here.

Also worth noting: the migration only compiles because `main` now pins an SDK version where `GitCredentialId` exists on the repo structs — on the SDK version this PR was originally cut against, it wasn't there at all.

Happy to land #5883 separately if we think it's worth it. I've also rebased this PR on `main` to clear the merge conflict.

alexott: we have a way of customizing aliases for Go SDK structs as well. I.e., https://github.com/databricks/terraform-provider-databricks/blob/main/pipelines/resource_pipeline.go#L131 - but this may require wrapping the Go SDK struct into TF-bound struct

alexott: but let's do Go SDK migration as a separate PR

GrantIsEaton: @alexott  Sorry this was all claude in ^; i did not mean for it to respond to you just yet :) I think it knows better than me here though, hope this was useful.  I agree that we should not do this as a part of this change (get the TF out to our customers) but we can consider it as a follow-up

GrantIsEaton: @alexott I think @tanmay-db will review this PR tomorrow but LMK if it looks ok for you as a short-term adjustment prior to https://github.com/databricks/terraform-provider-databricks/pull/5883

re that PR / feedback: 
Thanks for the feedback @alexott ! Claude updated the PR based on your thoughts, which did seem to simplify things. I don't have time this second to dive into that PR but here is what claude had to say as a response:

> Thanks — the Aliases() pointer was exactly what I was missing. I updated the demo PR (#5883) to generate the schema straight from workspace.RepoInfo via a thin wrapper + Aliases(), so there's no parallel hand-maintained struct anymore. Confirmed it round-trips correctly through schema-gen, StructToData, and DataToStruct, so git_provider/commit_hash stay stable for users.
Your caveat holds though: it still needs the wrapper type plus re-declaring the force-new/computed/validator bits in CustomizeSchema() (since those no longer come from tf: tags), and Read has to map GetRepoResponse → RepoInfo. So it's cleaner and clearly the idiomatic direction, but it's a full CRUD rewrite touching all databricks_repo users rather than a drop-in. I'd suggest landing it as its own follow-up rather than bundling it into #5877.

repos/resource_repo.go
tanmay-db, line 101: Just to confirm, this field is not part of response right?

tanmay-db:
Triggered integration tests

repos/resource_repo.go
GrantIsEaton: its a different API's response (list vs update/ create)

GrantIsEaton:
Hi @tanmay-db can i get some help reviewing?

repos/resource_repo.go
tanmay-db, line 226: @GrantIsEaton, could you please confirm no change is required here? This won't have the git credential set above on 213.

I am asking because below in update we are setting this:

> if d.HasChange("git_credential_id") {
> 				req["git_credential_id"] = int64(d.Get("git_credential_id").(int))
> 			}
> 			return reposAPI.Update(d.Id(), req)

GrantIsEaton: Good catch!

tanmay-db (approved):
@GrantIsEaton , I am not sure why the integration test was removed. I think we should add it back since the test itself was passing

PR #5881, opened 2026-07-14
Add acceptance tests for databricks_sql_table column comment updates

Divyansh-db:
## Changes

Adds acceptance-test coverage for column-comment updates on `databricks_sql_table`, covering both the `VIEW` and table code paths.

> **This PR is stacked on #5855.** Its base branch mirrors #5855's fix commits, so the diff here is **only the new test file** and the branch is fully runnable (the view test passes against a live workspace because the fix is present in the base). Please review/merge #5855 first; this PR can then be rebased onto `main` and merged. If #5855's branch is updated, I'll refresh the base here to match.

#5855 fixes the invalid `ALTER VIEW ... ALTER COLUMN ... COMMENT` SQL by emitting `COMMENT ON COLUMN` for views. That PR includes a unit test asserting the generated SQL; these acceptance tests exercise the full plan/apply lifecycle against a live Unity Catalog workspace, which is what actually reproduces the reported symptom (perpetual, un-appliable diff) and proves the update is applied in place.

### Tests added

- `TestUcAccSqlTableViewColumnCommentUpdate` — creates a `VIEW` with a column comment, then changes only the column comment, and asserts:
  - the change is an **in-place update, not a replacement** (`ExpectResourceAction(..., Update)`), so grants and dependent objects are preserved;
  - the new comment is persisted (read back into state);
  - re-applying the same config is a **no-op** (`ExpectResourceAction(..., Noop)`) — i.e. no perpetual diff.
- `TestUcAccSqlTableColumnCommentUpdate` — the table control: same flow with `table_type = "MANAGED"`, ensuring the fix does not regress in-place column-comment updates on tables (which continue to use `ALTER TABLE ... ALTER COLUMN ... COMMENT`).

### Verified against a live workspace

Both tests were run end-to-end against a live Unity Catalog workspace and pass. The emitted SQL confirms the intended behavior:

- view update → `COMMENT ON COLUMN `cat`.`sch`.`v`.`id` IS 'updated comment'` (in place; no `DROP`/`CREATE VIEW`);
- table update → `ALTER TABLE `cat`.`sch`.`t` ALTER COLUMN `id` COMMENT 'updated comment'`.

- [x] `make fmt lint ws` run locally
- [x] covered with integration tests in `internal/acceptance` (this PR)
- [x] using Go SDK

---

NO_CHANGELOG=true


catalog/resource_sql_table_acc_test.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// TestUcAccSqlTableViewColumnCommentUpdate is the regression test for the bug
+// where updating a column comment on a VIEW emitted the invalid
+// `ALTER VIEW ... ALTER COLUMN ... COMMENT`, which Databricks rejects with a
+// PARSE_SYNTAX_ERROR, leaving the change stuck as a perpetual, un-appliable diff.
+//
+// It asserts the full behaviour the fix promises against a live workspace:
+//   - Step 2 changes ONLY the column comment. This apply is exactly the
+//     operation that errored on the released provider; here it must succeed,
```
tanmay-db, line 56: I see the integration tests passed without the external PR changes, the comment says it errored on the provider (main), is this expected?

Divyansh-db: These tests are on top of the changes. The base of this PR is the clone of the changes and the tests run on top of it

tanmay-db (approved):

PR #5914, opened 2026-07-31
Fix `databricks_share` inconsistent result when `comment` is set outside Terraform

Divyansh-db:
## Changes

Managing a `databricks_share` without a `comment` in configuration would break once a comment was set on the share out-of-band (for example, in the UI): the next plan/apply failed with

```
produced an unexpected new value: .comment: was null, but now cty.StringVal(...)
```

and, in the case where the comment was subsequently cleared in the UI, `.comment: was null, but now cty.StringVal("")`.

**Root cause:** in the Plugin Framework implementation, `comment` was `Optional` but not `Computed`. The shares Update API has no field mask — presence is the mask, and sending `""` is the only way to clear a comment — so once a comment has ever been set the field is permanently `""`-or-a-value and a null configuration can never round-trip to a null server value. Because the attribute was not `Computed`, Terraform Core rejected the server value as inconsistent with the planned (null) value. This is a regression from the SDKv2 implementation, which suppressed the diff on `comment`.

**Fix:**
- Make `comment` `Optional`+`Computed` with `UseStateForUnknown`, so a value set outside Terraform is adopted into state instead of erroring, and unconfigured `comment` does not churn to "known after apply" on unrelated updates.
- In `Update`, only send `comment` on the wire when the plan carries a concrete, known change to it (guarding against the unknown plan value that `Optional`+`Computed` allows), and set `ForceSendFields` so an explicit `comment = ""` still clears the description instead of being dropped by the SDK's `omitempty`.

Resulting semantics:
- Omitted `comment`: preserve/adopt the server value.
- `comment = "text"`: manage and update it.
- `comment = ""`: explicitly clear the description.

Note this makes `databricks_share` treat an omitted `comment` as "not managed", which matches what the SDKv2 `databricks_share` did via `SetSuppressDiff`. It differs from `databricks_catalog`/`schema`/`volume`/`registered_model`, which instead clear a previously-managed comment when it is removed from config. Happy to align with that family in a follow-up if preferred — it needs the raw config to distinguish "absent" from "resolved-to-state" once the attribute is `Computed`.

## Tests

- `TestShareCommentConfigToState` — table test specifying what the user writes versus what ends up in state, given what the server holds: an explicit value (including `""`) is authoritative and sent; an omitted `comment` is not sent and state adopts the server value (null or not). Also asserts the planned value equals the post-apply state in every case, which is the invariant whose violation caused this bug.
- `TestShareCommentChanged` — the `Update` send/skip decision across null / unknown / empty / value plan and state combinations.
- Schema assertion that `comment` is `Optional`+`Computed` with the expected plan modifier.
- `TestUcAccShareCommentSetOutsideTerraform` — acceptance test: create a share with no `comment`, set one out-of-band, then re-apply the comment-less config and assert an empty plan, that state holds exactly the out-of-band value, and that the value still exists on the server. Verified against a UC workspace.


tanmay-db (commented):
I am not sure if comment should be optional+computed. Is comment/similar field handled like this for other resources on pluginfw?

Is this only specific to share because of:
> The Delta Sharing backend stores an empty string rather than omitting the field once a comment has ever been set

> only send comment on the wire when the plan carries a concrete, known change to it (guarding against the unknown plan value that Optional+Computed allows),

what happens if we always send?

internal/providers/pluginfw/products/sharing/resource_share.go
```diff
@@ -157,6 +169,16 @@ func (r *ShareResource) Schema(ctx context.Context, req resource.SchemaRequest,
...
 		c.AddPlanModifier(stringplanmodifier.UseStateForUnknown(), "created_by")
 
+		// comment is Optional+Computed so that a comment set out-of-band (e.g. in the
+		// UI) is adopted into state instead of failing the apply with "produced an
+		// unexpected new value". The Delta Sharing backend stores an empty string
+		// rather than omitting the field once a comment has ever been set, so a null
+		// config can never round-trip to a null server value. UseStateForUnknown keeps
+		// the refreshed value pinned on updates that change other attributes (the
+		// framework otherwise marks null-config computed attributes unknown).
+		c.SetComputed("comment")
```
tanmay-db, line 170: I am not sure if comment should be optional+computed. Is comment/similar field handled like this for other resources on pluginfw?

Is this only specific to share because of:
> The Delta Sharing backend stores an empty string rather than omitting the field once a comment has ever been set

> only send comment on the wire when the plan carries a concrete, known change to it (guarding against the unknown plan value that Optional+Computed allows),

what happens if we always send?

Divyansh-db: Good questions — answering each:

**Is `comment` Optional+Computed elsewhere in PF?** No. Every other PF resource with a `comment` (`policy_info`, `secret_uc`, `recipient_federation_policy`, the three `ai_gateway_*` services) is `Optional` only — i.e. exactly what `databricks_share` was before this PR. They're latent candidates for the same failure if their backends ever return a value for an unset `comment`; share is just where it was actually hit. Worth noting the opposite precedent too: the SDKv2 `databricks_catalog`/`schema`/`volume`/`registered_model` keep `comment` plain `Optional` and instead *clear* it (`ForceSendFields` on `Comment`) when a previously-managed comment is removed from config. So "adopt the server value" vs "clear it" is a real fork; this PR takes the first, which is also what the SDKv2 `databricks_share` did via `SetSuppressDiff` on `comment` before the PF migration.

**Is it share-specific?** Yes, in the sense that the shares Update API makes a null config unsatisfiable. There's no field mask on the request — presence *is* the mask: omit `comment` and the existing value is preserved, and sending `""` is the only way to clear. So once a comment has ever been set, the field is permanently `""`-or-a-value and "absent" is unreachable. A null config can therefore never round-trip, which is precisely what an Optional-not-Computed attribute requires.

**What if we always send?** It's worse — it breaks the first apply of every share that has no `comment` in config. With `comment` Optional-not-Computed, Core plans `null`; always sending `""` makes the server store and echo `""`, the converter turns that into a known `""` (the SDK marks present basic-type fields in `ForceSendFields` on unmarshal), and planned `null` != actual `""` fails the consistency check on create. I verified this. The variant that does work is a schema default of `""`, but that still requires `Computed` and would propose wiping existing descriptions on upgrade.

Happy to switch to the "clear when previously managed" semantics if you'd prefer consistency with the catalog/schema/volume family — it's a bigger change (needs the raw config to tell "absent" from "resolved-to-state" once the attribute is Computed), so I'd rather not fold it into this fix.

tanmay-db (approved):

Issue #5174, opened 2025-10-31
[ISSUE] Issue with `databricks_group` resource.  When adding an external group (i.e. Entra group) to the workspace level it does not create the group at the account level


haveyoumetcp:
### Configuration

```hcl

provider "databricks" {
  alias = "workspace"
  host = azurerm_databricks_workspace.databricks_workspace.workspace_url
}

data "azuread_group" "test_cdp_user_group" {
  for_each       = toset(var.test_databricks_workspace_user_groups)
  display_name   = each.key
  security_enabled = true
}

resource "databricks_group" "test_cdp_workspace_user_databricks_groups" {
  for_each     = data.azuread_group.test_cdp_user_group
  display_name = each.value.display_name
  external_id  = each.value.object_id
  provider     = databricks.workspace
  force        = true
}
```

### Expected Behavior
The external group is added to the account level. 

When adding a group via the UI in the workspace it automatically adds it the account.

### Actual Behavior
The external group is not added to the account level.

### Steps to Reproduce

1. `terraform apply

### Terraform and provider versions

v1.96.0

### Is it a regression?
Not sure it ever worked, but mostly an issue now with the Automated Identity Federation and not importing groups via SCIM.

### Important Factoids
I can add the group when using the account level connection with this provider:
``` hcl
provider "databricks" {
  alias      = "account"
  host       = "https://accounts.azuredatabricks.net"
  account_id = var.databricks_account_id
}
```

However, there's a huge issue with this.  If a group is deleted (i.e. in a development process), then the group is deleted at the account level and will impact production processes.

### Would you like to implement a fix?

No

alexott:
Please collect debug logs with both creation and deletion of the group

alexott:
in workspace context we're using workspace-level API, so deletion shouldn't affect account. If it happens, then it's a SCIM backend problem that should be reported via support ticket.

haveyoumetcp:
This issue isn't about deletion.  This issue is specifically about the posted Expected and Actual behavior:

Expected Behavior
The external group is added to the account level.

When adding a group via the UI in the workspace it automatically adds it the account.

Actual Behavior
The external group is not added to the account level.

alexott:
There is no public API yet for adding an account-level group in the workspace context. Otherwise, it would have been done a long time ago

haveyoumetcp:
> There is no public API yet for adding an account-level group in the workspace context. Otherwise, it would have been done a long time ago

[...]

PR #5932, opened 2026-08-07
Add resolve-by-external-id data sources for users, groups, and service principals

### LINKED ISSUE #5414: [ISSUE] Issue with `databricks_permission_assignment` failing with "Principal not found in account" for Entra ID-only identities (Automatic Identity Management enabled)
### Configuration

```hcl
resource "databricks_permission_assignment" "user_test" {
  provider = databricks.account
  provider_config {
    workspace_id = data.terraform_remote_state.workspace.outputs.workspace_unique_id
  }
  user_name   = "user@company.com" # Exists in Entra ID, not yet provisioned in Databricks account
  permissions = ["USER"]
}
```

This was also tested with a workspace-level provider.
```hcl
resource "databricks_permission_assignment" "user_test" {
  user_name   = "user@company.com" # Exists in Entra ID, not yet provisioned in Databricks account
  permissions = ["USER"]
}
```

Tested for users, groups and service principals.

### Expected Behavior

With Automatic Identity Management enabled, I expect that referencing an Entra ID-only identity (not yet provisioned in the account), like a user, group, or service principal by `user_name`, `group_name`, or `service_principal_name`, would automatically resolve and add/link the identity from Entra ID to the Databricks workspace, as happens in the Databricks UI. After that, eventually, the identity would be automatically provisioned in the account, following AIM behaviour.

In the Databricks UI, it is possible to add an identity that exists only in Entra ID directly to a workspace. The UI shows a badge indicating the identity is only in Entra ID -> `in Microsoft Entra ID`, and the assignment works as expected.
Image

### Actual Behavior

When applying, I get:
```
Error: cannot create permission assignment: RESOURCE_DOES_NOT_EXIST: Principal not found in account.
```
This happens for any identity that exists in Entra ID but is not yet provisioned in the Databricks account. **For identities already provisioned in the account, the resource works as expected and adds/links them to the workspace.**

### Additional Context
- Automatic Identity Management is enabled on the Databricks account
- I upgraded from provider v1.90.0 to v1.107.0 to use this feature

### Steps to Reproduce
1. Ensure Automatic Identity Management is enabled
2. Reference an Entra ID-only identity in `databricks_permission_assignment`, using identity name, ex: `group_name`
3. Run `terraform apply`

### What should happen?
The provider should resolve and add the Entra ID-only identity to the workspace, matching the UI behavior.

### What actually happens?
```
Error: cannot create permission assignment: RESOURCE_DOES_NOT_EXIST: Principal not found in account.
```
It is a fact that the identity will not exist in the account yet, but it will be visible as AIM is enabled, and will be provisioned later. The Terraform resource should be able to identify the identity, which is searchable on Entra ID, matching this info panel located in the account, for example. Image

`databricks` v1.107.0
`terraform` v1.7.0

### Is it a regression?

I don't think so, but checking the docs https://registry.terraform.io/providers/databricks/databricks/1.107.0/docs/resources/permission_assignment and https://github.com/databricks/terraform-provider-databricks/issues/5018#issuecomment-3370287427, my interpretation is that this should be working. Also, using the UI, it is possible to add an Entra ID-only identity to the workspace, and it would be nice if this is matched when working with Terraform.

Please let me know if you need more logs or details! Apologies if this is not clear enough, but I can try to clarify this further.
Thank you in advance :)

arman1371:
## Changes
- Adds `databricks_user_by_external_id`, `databricks_group_by_external_id`, and `databricks_service_principal_by_external_id` data sources, wrapping the Databricks `iamv2` `resolve-by-external-id` APIs to resolve a Databricks principal from the external ID assigned to it by the customer's IdP.
- Each data source works with both the account-level and workspace-level provider, inferring the API level from the provider host with an optional `api = "account" | "workspace"` override (same pattern as `databricks_users`).
- **Note (get-or-create semantics):** these APIs create the principal if one with the given `external_id` does not already exist yet, and require the account to be onboarded to Automatic Identity Management (AIM). This is called out in each data source's documentation.

## Tests
- [x] `make test` run locally
- [x] using Go SDK
- [x] `go test ./internal/providers/pluginfw/products/iamv2/...` — unit tests covering `resolveApiLevel` dispatch (account/workspace/unified-host/invalid values) and schema generation for all three data sources
- [x] `go test ./internal/providers/pluginfw/...` — full plugin framework suite passes
- [x] `make build`, `make lint`, `make ws`, `go vet ./...`
- [x] `make fmt-docs`

Fixes #5414 , #5174


kristeey:
Would it make sense to expose this as a managed resource instead of a data source to have complete control over the lifecycle of the group/identity? Problem with the data source as I see it is that even though it imports/resolves the group just fine but if the data source is removed, the group will still exist in databricks. Also, from a terraform perspective the data source has "create" capabilities since the api it uses will "create"/add the external group to databricks if it exist in the external IDP but not in databricks just yet. 

```hcl
resource "databricks_external_group" "this" {
  external_id = "<idp-group-id>"
}
```
Suggested lifecycle:
- Create: look up the group by external_id; if it exists, fail with import instructions. Otherwise call ResolveGroup to add it to databricks.
- Read: fetch the group using its stored Databricks group ID.
- Delete: call the IAM v2 account-group delete endpoint.
- Import: adopt an existing external group using its Databricks group ID.

This would make creation and deletion visible in Terraform plans and avoid mutating Databricks while reading a data source. Removing the resource would explicitly remove the group from Databricks, while importing would represent deliberate ownership of an existing group.
Note: There will be a small lookup/resolve race because ResolveGroup does not expose create-only semantics, but I think it is acceptable.

arman1371:
I think this has been overtaken by 1205ee74, which auto-generated `databricks_account_iam_external_group_v2` / `_external_user_v2` / `_external_service_principal_v2` as data sources plus `databricks_account_iam_group_v2` and others as full-CRUD resources. All on `main` now. The generated data source keeps the side effect and just documents it: *"Reading this data source has a side effect: it resolves the group against the IdP and provisions the group in the account if it does not already exist."*

Worth stating the strongest form of your point, though: because `external_id` is known at plan time, Terraform reads the data source during **plan**, not apply. So `terraform plan` alone,  including an unattended CI run on a PR,  provisions the group in the account, with no diff and no approval. That's a real problem.

Given the overlap, should I close this as superseded or should I still convert it to managed resource?

PR #5395, opened 2026-02-15
[Fix] [Security Impact] Mark `personal_access_token` as sensitive in `databricks_git_credential`

jlieow:
## Security Impact

The `personal_access_token` field is a credential that grants access to Git providers (GitHub, GitLab, Azure DevOps). Without `Sensitive: true`, the PAT is exposed in plain text in CLI output, which can leak into CI/CD logs, terminal history, screen recordings, and any system that captures Terraform output giving unauthorised users access.

## Changes

The `personal_access_token` field on `databricks_git_credential` is not marked `Sensitive: true` in the Terraform schema. This causes the PAT value to be displayed in plain text in `terraform plan` and `terraform apply` output instead of being displayed as `(sensitive value)`.

For the Terraform script:
```
resource "databricks_git_credential" "gitlab" {
  git_username          = "myuser"
  git_provider          = "gitLab"
  personal_access_token = "this_should_not_be_exposed"
}
```

Terraform plan outputs:
```
Terraform will perform the following actions:

  # databricks_git_credential.github will be created
  + resource "databricks_git_credential" "github" {
      + git_provider          = "gitHub"
      + git_username          = "username"
      + id                    = (known after apply)
      + personal_access_token = "this_should_not_be_exposed"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

**Root cause:** `s["personal_access_token"]` is not set as sensitive.

**Fix:** Set `s["personal_access_token"].Sensitive = true` in the resource's schema customization callback.

---

### Mark `personal_access_token` as sensitive (`repos/resource_git_credential.go`, line 33)

**Before:**
```go
s["personal_access_token"].DefaultFunc = schema.MultiEnvDefaultFunc([]string{
    "GITHUB_TOKEN",               // https://registry.terraform.io/providers/integrations/github/latest/docs
    "GITLAB_TOKEN",               // https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs
    "AZDO_PERSONAL_ACCESS_TOKEN", // https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs
}, nil)
```

**After:**
```go
s["personal_access_token"].Sensitive = true
s["personal_access_token"].DefaultFunc = schema.MultiEnvDefaultFunc([]string{
    "GITHUB_TOKEN",               // https://registry.terraform.io/providers/integrations/github/latest/docs
    "GITLAB_TOKEN",               // https://registry.terraform.io/providers/gitlabhq/gitlab/latest/docs
[...]

PR #5950, opened 2026-08-17
[Fix] [Security Impact] Mark plaintext credential fields as sensitive in `databricks_secret_uc`

jlieow:
## Security Impact

`value` and `effective_value` on `databricks_secret_uc` hold Unity Catalog secret material — passwords, tokens, keys. Without `Sensitive: true`, these values are displayed in plain text in CLI output, which can leak into CI/CD logs, terminal history, screen recordings, and any system that captures Terraform output, giving unauthorised users access.

Affected fields:
- `databricks_secret_uc.value`
- `databricks_secret_uc.effective_value`
- `data.databricks_secret_uc.value` / `.effective_value`
- `data.databricks_secret_ucs.secrets[*].value` / `.effective_value`

The plural data source is the widest exposure: it renders every secret in a schema as unredacted list elements in a single plan.

## Changes

The secret value fields are not marked `Sensitive: true` in the Terraform schema. This causes the secret to be displayed in plain text in `terraform plan` and `terraform apply` output instead of being displayed as `(sensitive value)`.

For the Terraform script:
```hcl
resource "databricks_secret_uc" "this" {
  catalog_name = "my_catalog"
  schema_name  = "my_schema"
  name         = "my_secret"
  value        = "SUPER_SECRET_SENTINEL_12345"
  comment      = "Sample secret"
}
```

Terraform plan outputs:
```

  # databricks_secret_uc.this will be created
  + resource "databricks_secret_uc" "this" {
      + catalog_name    = "my_catalog"
      + comment         = "Sample secret"
      + effective_value = (known after apply)
      + name            = "my_secret"
      + schema_name     = "my_schema"
      + value           = "SUPER_SECRET_SENTINEL_12345"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

**Root cause:** these fields are not set as sensitive. `resource_secret_uc.go:276` calls `SetRequired()` on `value` but never `SetSensitive()`, and the resource passes `nil` as the schema-customization callback, so there was no hand-editable place to set it.

**Fix:** replace the `nil` customization argument with a callback calling `.SetSensitive()` on both fields, in the resource and both data sources.

After the fix:
```
      + effective_value = (sensitive value)
      + value           = (sensitive value)
```

Note: this is the same class of issue as #5395 (`databricks_git_credential.personal_access_token`) and #5409 (`databricks_model_serving` plaintext credentials). Unlike those, `databricks_secret_uc` is an auto-generated Plugin Framework resource — see Caveat below.

---

### Mark secret value fields as sensitive (`internal/providers/pluginfw/products/secret_uc/resource_secret_uc.go:291`)

**Before:**
```go
attrs, blocks := tfschema.ResourceStructToSchemaMap(ctx, Secret{}, nil)
```

**After:**
```go
attrs, blocks := tfschema.ResourceStructToSchemaMap(ctx, Secret{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
	// The secret payload must never be rendered in plan/apply output. `value` is the
	// write path and `effective_value` the read path, so both carry secret material.
	c.SetSensitive("value")
	c.SetSensitive("effective_value")

	return c
})
```

**Why:** Ensures the secret is masked as `(sensitive value)` in plan/apply output. The customization callback runs after the generated `ApplySchemaCustomizations` (`tfschema/struct_to_schema.go:158` then `:181`), so it composes with the generated `SetRequired`/`SetComputed` instead of conflicting with them.

The same change is applied to `data_secret_uc.go:243`. For the plural data source, `data_secret_ucs.go:75` uses the nested path so every listed secret is masked:

```go
c.SetSensitive("secrets", "value")
c.SetSensitive("secrets", "effective_value")
```

**Why `effective_value` too:** it is currently always null, because `GetSecretRequest` has no `include_value` field to request the value. But it is declared as the read path for the secret, so it will carry real secret material if that is ever wired up. Marking it now avoids reintroducing the leak later.

---

## Caveat: auto-generated files

All three files are auto-generated (`// Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.`, `linguist-generated=true` in `.gitattributes:306-308`), so this fix can be overwritten by the next `Update SDK API to <sha>` regeneration. The three schema tests are the safety net — they fail loudly if the customization callback is lost, turning a silent security regression into a red build.

The durable fix is generator-side. Worth noting the gap is provider-wide, not specific to this resource: `grep -rn "SetSensitive" internal/providers/pluginfw/products/` returns zero hits across all 155 generated files, so any generated resource with a credential field has the same exposure. Happy to split that into a separate issue.

---

### Unit Tests

- **TestResourceSecret_SecretValuesAreSensitive**: Builds the `databricks_secret_uc` resource schema and asserts `value` and `effective_value` are `Sensitive`. This test would have caught the original bug.
- **TestDataSourceSecret_SecretValuesAreSensitive**: Same assertion for the `databricks_secret_uc` data source schema.
- **TestDataSourceSecrets_SecretValuesAreSensitive**: Asserts the nested `secrets[*].value` and `secrets[*].effective_value` in the `databricks_secret_ucs` data source are `Sensitive`.

Verified these fail without the fix: reverting the three source files produces 6 failing assertions across all 3 tests.

### Validation

Whole-provider schema comparison (binaries built from `af211e28` vs this branch, `terraform providers schema -json` diffed key by key) shows **exactly 6 differences, all additive `sensitive: true` flags** — no attribute added, removed, retyped, or changed in required/optional/computed status. No breaking changes.

Sentinel occurrences in `terraform plan -no-color` output: **1 before the fix, 0 after.**

- [x] `make test` run locally
- [ ] relevant change in `docs/` folder — N/A, `docs/resources/secret_uc.md` is generated and does not document sensitivity
- [x] using Go SDK

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -11,6 +11,7 @@
 * Add support for `model_service`, `mcp_service`, and `model_provider_service` securables in `databricks_grant` and `databricks_grants` ([#5941](https://github.com/databricks/terraform-provider-databricks/pull/5941)).
 
 ### Bug Fixes
+* Mark `value` and `effective_value` as sensitive in `databricks_secret_uc` and its data sources. These fields hold Unity Catalog secret material but were rendered in cleartext in `terraform plan` and `terraform apply` output, which could leak secrets into CI/CD logs and terminal history.
 * Fixed `databricks_share` failing with `produced an unexpected new value: .comment` when a share's comment is set or removed outside of Terraform (e.g. in the UI). The `comment` attribute is now `Optional`+`Computed`, so an out-of-band value is adopted into state instead of erroring, while `comment = ""` still explicitly clears the description.
 
 ### Documentation
--- a/internal/providers/pluginfw/products/secret_uc/data_secret_uc.go
+++ b/internal/providers/pluginfw/products/secret_uc/data_secret_uc.go
@@ -240,7 +240,14 @@ func (r *SecretDataSource) Metadata(ctx context.Context, req datasource.Metadata
 }
 
 func (r *SecretDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
-	attrs, blocks := tfschema.DataSourceStructToSchemaMap(ctx, SecretData{}, nil)
+	attrs, blocks := tfschema.DataSourceStructToSchemaMap(ctx, SecretData{}, func(c tfschema.CustomizableSchema) tfschema.CustomizableSchema {
+		// The secret payload must never be rendered in plan/apply output. `value` is the
+		// write path and `effective_value` the read path, so both carry secret material.
+		c.SetSensitive("value")
+		c.SetSensitive("effective_value")
+
+		return c
+	})
 	resp.Schema = schema.Schema{
 		Description: "Terraform schema for Databricks Secret",
 		Attributes:  attrs,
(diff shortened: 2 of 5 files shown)


alexott:
This is autogenerated resource so no manual changes are allowed. Raise issue with the corresponding team to fix this at generation time

PR #5977, opened 2026-08-28
fix(databricks_app): manage input_only git_source, source_code_path, caller_credential_id

atreyadbrx:
## Changes

Makes the `databricks_app` resource able to manage the app's `INPUT_ONLY` git-deployment fields — `git_source`, `source_code_path`, and `git_repository.caller_credential_id` (and, transitively, `git_repository.auto_deploy`). Today, setting `git_source` or `source_code_path` fails with **"Provider produced inconsistent result after apply"** and then a perpetual diff.

Two root causes, both fixed here:

1. **Not echoed on read.** These fields are `INPUT_ONLY` — the Apps API accepts them on write but never returns them (the read-back is the separate `OUTPUT_ONLY` `default_git_source`/`default_source_code_path`). The hand-written resource overwrote state from the empty response because it never ran the reconciliation the generated model already provides. Fix: call the generated `App.SyncFieldsDuringCreateOrUpdate` / `SyncFieldsDuringRead` at the state-set sites — the convention every other plugin-framework resource follows (e.g. `database_instance`'s `stopped`/`effective_stopped`). This carries the configured value across reads for all input-only App fields, guarded to known/non-null.
   - `SyncFieldsDuringRead` does not reconcile `user_api_scopes` (only `effective_user_api_scopes`), so the existing `reconcileEmptyUserApiScopes` restore is kept on the read path to avoid regressing #5834.

2. **Computed nested fields.** `git_source`'s nested `git_repository`/`resolved_commit` are generated as `Computed` (because `GitSource` is a shared message reused in output contexts). For an input-only `git_source` they can never be populated, so a configured value was left unknown after apply. Fix: strip `Computed` from those descendants in `ApplySchemaCustomizations` (same nested-override pattern the resource already uses for `provider_config.workspace_id`), so a configured `git_source` is fully known. This is the only genuinely new logic in the PR.

This unblocks `git_repository.auto_deploy`, whose prerequisite (a settable `git_source` branch) now round-trips.

Scope: this is the standalone `databricks_app` resource. It does not touch the DABs bundle→Terraform path (databricks/cli#6399), which deliberately drops these fields; that change is in a different repo and does not conflict.

## Tests

- `TestResourceApp_GitSourceInputOnlySchema` (unit) — asserts `git_source` is settable and its nested `git_repository`/`resolved_commit` are not `Computed`.
- `TestReconcileEmptyUserApiScopes` (unit, existing) — the `user_api_scopes` empty-list restore still exercised on the read path.
- `TestAccAppResource_GitSource` (acceptance) — a git-backed app applies, then a re-apply is a no-op (regression coverage for inconsistent-result/perpetual-diff, mirroring `TestAccAppResource_EmptyUserApiScopes`).

Verified end-to-end against a live workspace: applied a `databricks_app` with `git_repository{auto_deploy=true, caller_credential_id}` + `git_source{branch}`, second plan clean (no perpetual diff). Integration tests pass in CI.

## Schema breaking-change check

ALLOW_SCHEMA_BREAKING_CHANGE=true

The bypass is intentional. The classifier flags exactly one change — `git_source.git_repository.caller_credential_id` is no longer `Computed` (`ComputedRemoved`) — which is precisely this fix. That field is `INPUT_ONLY` (never returned by the API), so `Computed` was a shared-message codegen artifact, not a real server-populated value; leaving it `Computed` is what produced the "unknown after apply" failure. `git_source` could not be managed before this PR, so no existing state relies on the prior behavior, and the top-level `git_repository.caller_credential_id` is unchanged.


atreyadbrx:
Ready for review 🙏 — @rauchy @tanmay-db @Divyansh-db (cc @databricks/field-dev-terraform).

This generalizes the input-only preservation @Divyansh-db added in #5834 to `git_source` / `source_code_path` / `git_repository.caller_credential_id`, by adopting the generated `SyncFields` (the same convention as `database_instance`'s `stopped`/`effective_stopped`), plus a schema override so `git_source`'s nested descendants aren't `Computed`.

Heads-up @andrewnester: this intersects the premise in cli#6399 that these fields can't be managed — happy to align on the approach. Note the acceptance test (`TestAccAppResource_GitSource`) still needs `TF_ACC` + a workspace and hasn't been run yet.

rauchy (approved):

PR #6003, opened 2026-09-10
Add parent support to databricks_connection for schema-level connections

mohaned2014:
## Summary

Schema-level (L3) Unity Catalog connections live inside a schema (`catalog.schema.name`) and are GA, but the `databricks_connection` resource could only manage metastore-level (L1) connections. This adds an optional `parent` argument so schema-level connections can be managed as code.

## What changed

- **`parent` argument** (`schemas/{catalog}.{schema}`), optional and `ForceNew`. When set, the connection is created inside that schema; when omitted, the connection is metastore-level, exactly as before.
- **Identity:** the resource id becomes `metastore_id|full_name`. For L1 connections `full_name == name`, so this is byte-identical to the previous `metastore_id|name` id — **no state migration**, and existing L1 configurations/state are unaffected.
- **Read round-trip:** the API returns only `full_name`, so `parent` is reconstructed from it on read (empty for L1), keeping the plan clean for the `parent`/identity fields.

## Testing

- Unit: `TestSchemaParentFromFullName`, `TestConnectionsCreate_SchemaLevel`, and all existing connection tests pass; `gofmt`/`go vet` clean.
- Acceptance: `TestUcAccConnectionsSchemaLevelResourceFullLifecycle` — validated live against a staging Unity Catalog workspace (create → read → destroy).

## Known pre-existing diff (not addressed here)

The acceptance test sets `ExpectNonEmptyPlan: true`. An HTTP connection re-plans dirty on `options` because the server omits the write-only `bearer_token` on read. This is **pre-existing** and affects all HTTP connections (metastore- and schema-level alike) — reproduced on a plain L1 HTTP connection with no `parent`. It is an in-place no-op update (no recreation or data loss) and does not block managing the connection; it will be handled separately.


catalog/resource_connection.go
```diff
@@ -81,7 +96,9 @@ func ResourceConnection() common.Resource {
 				}
 			}
 			d.Set("metastore_id", conn.MetastoreId)
-			pi.Pack(d)
+			// Address connections by full_name: for L3 it is catalog.schema.name, and for L1 it
+			// equals the name, so this preserves the existing metastore_id|name id for L1 (no migration).
+			d.SetId(conn.MetastoreId + "|" + conn.FullName)
```
eddiebkheet, line 129: **Correctness (P1).** The post-create owner update just above (line 92, `updateConnectionRequest.Name = createConnectionRequest.Name`) still addresses the connection by its **short** name, but a schema-level connection must be addressed by its full name — as this `SetId` now does.

Scenario: a `databricks_connection` with `parent` **and** `owner` set. Create succeeds, then the owner `PATCH` is sent to `/connections/{short_name}`, which doesn't resolve the L3 connection — it 404s, or if a metastore-level connection with the same short name exists it silently updates the **wrong** connection's owner. L1 is unaffected (`createConnectionRequest.Name == FullName`).

This path is uncovered — the L3 unit and acceptance tests don't set `owner`.

catalog/resource_connection.go
```diff
@@ -40,15 +55,15 @@ func ResourceConnection() common.Resource {
 			for _, v := range []string{"owner", "read_only"} {
 				common.CustomizeSchemaPath(m, v).SetComputed()
 			}
-			for _, v := range []string{"read_only", "properties", "comment", "connection_type"} {
+			for _, v := range []string{"read_only", "properties", "comment", "connection_type", "parent"} {
 				common.CustomizeSchemaPath(m, v).SetForceNew()
 			}
 			common.CustomizeSchemaPath(m, "options").SetSensitive().SetCustomSuppressDiff(suppressComputedFields)
 			common.CustomizeSchemaPath(m, "name").SetCustomSuppressDiff(common.EqualFoldDiffSuppress)
```
eddiebkheet, line 84: `parent` is added to the ForceNew list (line 58) but, unlike `name` here, has neither an `EqualFoldDiffSuppress` nor a `ValidateFunc`:

1. **Recreation risk.** Read unconditionally rewrites `parent` from the server's `full_name` (line 141). If the configured `parent` differs only in identifier case/format from what the API returns — the exact reason `name` carries `EqualFoldDiffSuppress` — the next plan diffs a ForceNew field and destroys/recreates the connection.
2. **No plan-time validation.** A malformed `parent` (`main.default` without the `schemas/` prefix, or `schemas/main` with no `.`) is accepted at plan and only fails at apply.

catalog/connection_test.go
```diff
@@ -67,3 +67,41 @@ func TestUcAccConnectionsWithoutOwnerResourceFullLifecycle(t *testing.T) {
...
+}
+
+func TestUcAccConnectionsSchemaLevelResourceFullLifecycle(t *testing.T) {
+	acceptance.UnityWorkspaceLevel(t, acceptance.Step{
+		Template: schemaLevelConnectionTemplate(),
+		// A schema-level HTTP connection re-plans dirty on `options` because the server omits the
+		// write-only bearer_token on read. That is a pre-existing round-trip issue affecting all
+		// HTTP connections (metastore- and schema-level alike), not specific to schema connections,
+		// and it does not block create/read/update/delete. Tracked separately.
+		ExpectNonEmptyPlan: true,
```
eddiebkheet: `ExpectNonEmptyPlan: true` passes on *any* non-empty plan, which also masks a `parent`/identity ForceNew regression — the thing this feature most needs to prove doesn't drift.

The stated cause (server omits write-only `bearer_token`) doesn't obviously hold: `bearer_token` is already in `sensitiveOptions` and re-injected from config on read specifically to prevent that drift, so the dirty plan may have a different cause.

docs/resources/connection.md
```diff
@@ -125,11 +125,29 @@ resource "databricks_connection" "pbi" {
...
+  }
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
 - `name` - (Required) Name of the connection.
+- `parent` - (Optional) Parent schema of a schema-level connection, in the format `schemas/{catalog}.{schema}`. When set, the connection is created inside that schema and its `full_name` becomes `{catalog}.{schema}.{name}`; when omitted, the connection is metastore-level. Change forces creation of a new resource.
```
eddiebkheet, line 150: Two doc spots outside this diff aren't updated for L3:

- The `## Import` section (line 181) still shows only `<metastore_id>|<name>` and doesn't cover the schema-level form `<metastore_id>|<catalog>.<schema>.<name>`.
- The `id` attribute reference (line 167, "in form of `<metastore_id>|<name>`") is now inaccurate for L3.

alex-khakhlyuk (approved):

catalog/resource_connection.go
mohaned2014: Updated and test with unit test as Schema connections doesn't support owner change.

catalog/resource_connection.go
mohaned2014: Updated.

catalog/connection_test.go
mohaned2014: You're right on both counts.

Cause: it's not omitted bearer_token (that's re-injected from config on read). The real cause is the HTTP-bearer backend adding a server-managed auth_scheme option that's never in config. I've corrected the misleading comment.

Masking: the re-apply and Day-2 steps now assert plancheck.ExpectResourceAction(..., ResourceActionUpdate), so ExpectNonEmptyPlan can no longer hide a parent/identity ForceNew regression — that would plan a replace and fail the check (it caught exactly that on an earlier ForceNew slip). I also added a live host-update step proving an in-place L3 update round-trips.

The remaining non-empty plan is only that pre-existing auth_scheme option diff (all HTTP-bearer connections, L1 and L3 alike)

docs/resources/connection.md
mohaned2014: Updated

eddiebkheet (approved):

catalog/resource_connection.go
```diff
@@ -116,7 +150,17 @@ func ResourceConnection() common.Resource {
 					conn.Options[key] = element
 				}
 			}
-			return common.StructToData(conn, s, d)
+			// Drop an empty environment_settings object so it does not materialize as a block
+			// that diffs against a config which omits it.
+			if es := conn.EnvironmentSettings; es != nil && es.EnvironmentVersion == "" && len(es.JavaDependencies) == 0 {
```
Divyansh-db, line 155: Is it a new behaviour introduced by L3 connections? Why do we need it now?

mohaned2014: @Divyansh-db Yeah, it's specific to schema-level connections. Metastore-level ones don't return environment_settings at all when it's unset, but schema-level ones send back an empty object — and without dropping it, Terraform sees that as a block it needs to remove on every plan. It's a no-op for existing connections. The tidier fix is really on the backend but it will take couple of weeks to be rolled out.

Divyansh-db: @mohaned2014 , we do not indent to handle server side issues in the provider code. It is very difficult to remove such hacky code in future without breaking some existing customer. I would recommend you to ask the concerned service team for fixing this as the fix for it dosen't belong in the provider

mohaned2014: @Divyansh-db make sense, removed the logic and pushed a backend fix.

PR #6015, opened 2026-09-18
Remove Workspace test for unusual case

hectorcast-db:
## Changes
Remove test for unusual case. The server changed the behaviour and started normalising the value, which caused test.

There is not much value in this test, so we are removing it.
## Tests

- [X] `make test` run locally
- [] relevant change in `docs/` folder
- [] using Go SDK
- [] using TF Plugin Framework
- [] has entry in `NEXT_CHANGELOG.md` file

NO_CHANGELOG=true


tejaskochar-db (approved):
wow

